PR #1, opened 2020-04-06
Add initial fabric-ios and fabric-macos sources

markavitale:
## Summary
This repository will be a combination of [our existing fabric-ios repository](www.github.com/Office/ui-fabric-ios) and a previously internal only fabric-macos. Bring in the commit histories of those repos and make them sibling directories.

This will be our starting source to move forward with the combined fluentui-apple repository.

## Verification
Locally launch and test the iOS and macOS sources, ensuring all unit tests succeed and the test apps are usable as expected.


ios/.ado/fabric-ios-publish-nuget.yml
```diff
@@ -0,0 +1,89 @@
+# This file defines the Apple publish build steps
```
anandrajeswaran, line 1: and we're good with the .ado folder living on GitHub? Or should that be entirely private since it's specific to ado?

harrieshin: the CI for nuget package will live in ADO still right? although, there isn't a huge secret here so i don't mind ...

markavitale: I think that for some of our builds, specifically our internal NuGet publishing builds, we'll still want those to be ADO pipelines built from this source. This pattern exists similarly in https://github.com/microsoft/fluentui-react-native. I prefer to keep the build definitions under source control along with the source.

PR #2, opened 2020-04-07
[CI] Add initial CI for both platforms

markavitale:
Add a GitHub action for CI.
Build and Test
- macOS Debug and Release configurations of the test app
- iOS Debug simulator configuration of the unit test target

Build
- iOS release simulator configuration of the unit test target
- iOS Debug and Release simulator configurations of the test app target
- iOS Debug and Release device configurations of the test app target

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,98 +9,24 @@ on:
       - master
 
 jobs:
-  macos_debug_build_test:
+  xcodebuild:
     runs-on: macos-10.15
-    
-    steps:
-    - uses: actions/checkout@v2
-    - name: Switch to Xcode 11.2.1
-      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
-    - name: macOS Debug Build and Test
-      run: |
-        source scripts/xcodebuild_wrapper.sh
-        macos_build OfficeUIFabricTestApp Debug build test
-  
-  macos_release_build_test:
-    runs-on: macos-10.15
-    
-    steps:
-    - uses: actions/checkout@v2
-    - name: Switch to Xcode 11.2.1
-      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
-    - name: macOS Release Build and Test
-      run: |
-        source scripts/xcodebuild_wrapper.sh
-        macos_build OfficeUIFabricTestApp Release build test
+    strategy:
+      matrix:
+        build_command: [
+          'macos_build OfficeUIFabricTestApp Debug build test',
+          'macos_build OfficeUIFabricTestApp Release build test',
+          'ios_simulator_build OfficeUIFabric Debug build test',
+          'ios_simulator_build OfficeUIFabric Release build',
+          'ios_simulator_build Demo.Development Debug build',
+          'ios_simulator_build Demo.Development Release build',
+          'ios_device_build Demo.Development Debug build',
+          'ios_device_build Demo.Development Release build',
+        ]
 
-  ios_simulator_debug_build_test:
-    runs-on: macos-10.15
-    
     steps:
     - uses: actions/checkout@v2
     - name: Switch to Xcode 11.2.1
       run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
-    - name: iOS Simulator Debug Build and Test
-      run: |
-        source scripts/xcodebuild_wrapper.sh
-        ios_simulator_build OfficeUIFabric Debug build test
-  
-  ios_simulator_release_build:
-    runs-on: macos-10.15
-    
-    steps:
-    - uses: actions/checkout@v2
-    - name: Switch to Xcode 11.2.1
-      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
-    - name: iOS Simulator Release Build
-      run: |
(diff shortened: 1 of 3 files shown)


scripts/ci.sh
```diff
@@ -0,0 +1,73 @@
...
+# \param $6+ build commands
+function invoke_xcodebuild()
+{
+    /usr/bin/xcodebuild \
+        -"$1" "$2" \
+        -scheme "$3" \
+        -configuration "$4" \
+        -sdk "$5" \
+        "${@:6}" \
+        CODE_SIGNING_ALLOWED=NO
```
anandrajeswaran, line 19: Is this copied from elsewhere? This line might be worth a comment

scripts/ci.sh
```diff
@@ -0,0 +1,73 @@
...
+{
+    invoke_xcodebuild project "macos/xcode/OfficeUIFabric.xcodeproj" OfficeUIFabricTestApp "$1" macosx build test
+}
+
+# Run an iOS simulator xcodebuild invocation with the specified scheme, configuration, and build commands
+#
+# \param $1 scheme
+# \param $2 configuration
+# \param $3+ build commands
+function ios_simulator()
```
anandrajeswaran, line 37: should this be ios_simulator_build or ios_build_simulator?

scripts/ci.sh
```diff
@@ -0,0 +1,73 @@
...
+{
+    invoke_xcodebuild workspace "ios/OfficeUIFabric.xcworkspace" "$1" "$2" iphonesimulator "${@:3}" -destination "platform=iOS Simulator,name=iPhone 8"
+}
+
+# Run an iOS device xcodebuild invocation with the specified scheme and configuration
+#
+# \param $1 scheme
+# \param $2 configuration
+function ios_build_device()
```
anandrajeswaran, line 46: very minor, but above in macos_build_test we establish the pattern of {platform}_{actions}. This one feels like {platform}_{actions}_{platformPart2}. Not sure what the proper answer is, but should be consistent

scripts/ci.sh
markavitale: specifically the code signing bit? I basically tried to mimic the xcodebuild invocation from ADO's xcode build task

scripts/ci.sh
markavitale: Yep I needed a PR to actually have the action run and be registered, but this all needs to be cleaned up and parallelized into multiple jobs. This iteration was more of a proof of concept.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+    steps:
+    - uses: actions/checkout@v2
+    - name: Switch to Xcode 11.2.1
+      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
+    - name: iOS Simulator Release Build
+      run: |
+        source scripts/xcodebuild_wrapper.sh
+        ios_simulator_build OfficeUIFabric Release build
+
+  ios_testapp_simulator_debug_build:
```
anandrajeswaran, line 60: testapp before ios? Per another comment it feels like it is in between two "platform" tokens

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+        ios_simulator_build OfficeUIFabric Debug build test
+  
+  ios_simulator_release_build:
+    runs-on: macos-10.15
+    
+    steps:
+    - uses: actions/checkout@v2
+    - name: Switch to Xcode 11.2.1
+      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
+    - name: iOS Simulator Release Build
```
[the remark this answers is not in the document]
markavitale: I think we should. There are apps and build systems that will consume this that rely on release simulator builds succeeding, so we should confirm that. This runs in parallel so it doesn't add to the total CI time, although it does use up an available build agent. We can revisit this parallelization if we're seeing issues with agent availability.

markavitale:
@tom-un are you aware of any way we could extract the steps we have duplicated 8 times into a sort of template? Just internally, I don't think we want to publish an action to the marketplace or anything. I'm new to GitHub Actions.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,106 @@
...
+        source scripts/xcodebuild_wrapper.sh
+        macos_build OfficeUIFabricTestApp Debug build test
+  
+  macos_release_build_test:
+    runs-on: macos-10.15
+    
+    steps:
+    - uses: actions/checkout@v2
+    - name: Switch to Xcode 11.2.1
+      run: sudo xcode-select --switch /Applications/Xcode_11.2.1.app/Contents/Developer
```
HeyImChris, line 30: Minor but could be nice to pull this out too. Then when we're on our new xcode version we only need to update it in one place and not risk forgetting about it for a flavor

markavitale: Totally agree. I'd like to pull out all of these steps and just take in the one interesting line at the bottom, but I'm not quite sure how to do that in GitHub Actions. I'll try to read up on it.

markavitale: I believe the latest iteration addresses this concern by switching to a strategy matrix and only defining these steps once.

.github/workflows/ci.yml
markavitale: Switched the naming to the specific build command we're running to make it easier for developers to match the behavior on their local box.

PR #7, opened 2020-04-10
[Carthage] Workaround scheme name Carthage bug

markavitale:
## Summary
It seems Carthage has a bug where if two schemes have the same name, even in different xcode projects, it will trip up and build one of those schemes twice and won't build the other scheme.

Work around this by renaming the schemes in our projects to be unique. Note there's plenty more renaming to do for rebranding this repo from OfficeUIFabric to FluentUI, but for now let's scope this change to just get things building for Carthage users.

## Verification
- Make a local test app that uses Carthage and depends on both iOS and macOS FluentUI frameworks, ensure both build exactly once
- Sanity build and run each test app
- Rely on CI for verification that scheme name changes didn't break anything else


HeyImChris:
Worth opening a bug on Carthage with a test app reproing this?

HeyImChris:
What were the xcode projs and schemes that had duplicated names?

markavitale:
> Worth opening a bug on Carthage with a test app reproing this?
I've got a test app I'm working on turning into a bug to submit to that repo.

> What were the xcode projs and schemes that had duplicated names?
ios/OfficeUIFabric.xcworkspace has an OfficeUIFabric scheme, and macos/xcode/OfficeUIFabric.xcodeproj also had an OfficeUIFabric scheme.

PR #9, opened 2020-04-13
[Rebrand] Rename fabric-macos to FluentUI

markavitale:
Rename all of the files, frameworks, modules, and more from OfficeUIFabric branding to FluentUI.

There are still a few remnants around some packaging files, but this is the bulk of the renaming work.

Verification:
Ensure that assets and strings are still being loaded as expected.


macos/FluentUITestApp/TestAvatarViewController.swift
```diff
@@ -4,7 +4,7 @@
 //
 
 import AppKit
-import OfficeUIFabric
+import FluentUI
```
anandrajeswaran, line 7: we've probably already decided this 8 times (sorry), but we're ok with the macOS and iOS modules having the same name? All code consuming them *should* only be searching in the place for the appropriate platform, but likely most code consuming this doesn't have their search paths as strictly managed as Office does

macos/xcode/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -14,16 +14,17 @@
 		8061BF7E22EF936800F2D245 /* DatePickerView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8061BF7C22EF936700F2D245 /* DatePickerView.swift */; };
 		8061BF7F22EF936800F2D245 /* DatePickerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8061BF7D22EF936700F2D245 /* DatePickerController.swift */; };
 		8061BF8322EF957200F2D245 /* TestDatePickerController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8061BF8222EF957200F2D245 /* TestDatePickerController.swift */; };
-		80FEB30F230DE91D00F9FF3E /* OfficeUIFabric.strings in Resources */ = {isa = PBXBuildFile; fileRef = 80FEB2F8230DE53F00F9FF3E /* OfficeUIFabric.strings */; };
-		8F16282123CCD17900F81956 /* OfficeUIFabric.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 8F16282023CCD17900F81956 /* OfficeUIFabric.xcassets */; };
+		80FEB30F230DE91D00F9FF3E /* FluentUI.strings in Resources */ = {isa = PBXBuildFile; fileRef = 80FEB2F8230DE53F00F9FF3E /* FluentUI.strings */; };
+		8F16282123CCD17900F81956 /* FluentUI.xcassets in Resources */ = {isa = PBXBuildFile; fileRef = 8F16282023CCD17900F81956 /* FluentUI.xcassets */; };
```
anandrajeswaran, line 18: I posed another version of this question elsewhere, but do we want to push for things (mostly source files) in the FluentUI project to have _ios and _macos suffixes for easy CMD+Shift+O? Or is that a lot of change for very little benefit

macos/FluentUITestApp/TestAvatarViewController.swift
harrieshin: most of the platform that using cocoapod to consume this library via pod file would be something like 
target 'FooiOS' do
    platform :ios, '11.0'
 ...
end

target 'FoomacOS' do
    platform :osx, '10.15'
    ...
end

which should generate the right xcconfigt file for those search path correctly ...

markavitale: Carthage should also make this very simple (and does right now with both frameworks being named "OfficeUIFabric" by detecting the platform and putting it in an appropriate folder in the build output, either `Carthage/Build/iOS` or `Carthage/Build/macOS`. So outside of those two supported package managers supporting it, I'd say we can trust clients rolling their own dependency management to handle this case.

macos/xcode/FluentUI.xcodeproj/project.pbxproj
markavitale: So this to me is a bit of a different question. These are internal implementation details to the framework and as such we can change them at-will later without client impact. Since we're currently still in an "xcodeproj per platform" configuration, I'd say leave these as-is for now and consider changing if we find issues later.

PR #11, opened 2020-04-14
[FluentUI-macOS] Use preconditionFailure() instead of fatalError()

markavitale:
fatalError() has a default parameter which takes in the #file and on failure will log the full source path of the file. This also encodes our full source path into the binary even in release configurations which is a security concern.

preconditionFailure works in essentially the same way but doesn't encode the full source path in release configurations.

### DIFF
--- a/macos/FluentUI/AvatarView.swift
+++ b/macos/FluentUI/AvatarView.swift
@@ -60,7 +60,7 @@ open class AvatarView : NSView {
 	}
 	
 	@available(*, unavailable) required public init?(coder decoder: NSCoder) {
-		fatalError("init(coder:) has not been implemented")
+		preconditionFailure("init(coder:) has not been implemented")
 	}
 
 	override open func updateLayer() {
--- a/macos/FluentUI/AvatarView.swift
+++ b/macos/FluentUI/AvatarView.swift
@@ -59,8 +59,9 @@ open class AvatarView : NSView {
 		updateAvatarViewContents()
 	}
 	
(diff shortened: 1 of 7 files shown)


macos/FluentUI/DatePicker/DatePickerView.swift
```diff
@@ -104,7 +104,7 @@ class DatePickerView: NSView {
 	}
 	
 	required init?(coder decoder: NSCoder) {
```
anandrajeswaran, line 106: would recommend updating these to also have @available(*, unavailable) to truly make this a compile-time break

macos/FluentUI/AvatarView.swift
```diff
@@ -60,7 +60,7 @@ open class AvatarView : NSView {
 	}
 	
 	@available(*, unavailable) required public init?(coder decoder: NSCoder) {
-		fatalError("init(coder:) has not been implemented")
+		preconditionFailure("init(coder:) has not been implemented")
```
anandrajeswaran, line 63: For all of this, imo the "init(coder:) has not been implemented" can be removed. As written, it makes it sound like we want to implement it but haven't. And having @available(*, unavailable) makes the intention quite clear so I don't think the text adds much

markavitale: We may want to revisit this in the future to unblock clients who use storyboards/xibs for our public components, but short term making the APIs unavailable should get the message across.

PR #12, opened 2020-04-14
Remove alpha value from MSButton colors and use the property primary …

harrieshin:
…theme color.
Before | After
------------ | -------------
before_light  | after_light
before_dark | after_dark

### DIFF
--- a/ios/OfficeUIFabric/Controls/MSButton.swift
+++ b/ios/OfficeUIFabric/Controls/MSButton.swift
@@ -131,7 +131,7 @@ open class MSButton: UIButton {
 
     public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
-        if #available(iOS 12, *) {
+        if #available(iOS 13, *) {
             if previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle {
                 updateBorderColor()
             }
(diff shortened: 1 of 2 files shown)


ios/OfficeUIFabric/Controls/MSButton.swift
```diff
@@ -129,6 +129,15 @@ open class MSButton: UIButton {
         proposedTitleLabelWidth = bounds.width - (contentEdgeInsets.left + contentEdgeInsets.right)
     }
 
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+        if #available(iOS 13, *) {
+            if previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle {
+                updateBorderColor()
```
markavitale, line 136: We should be able to handle this by just setting the color in updateLayer(), which is the Apple recommendation I believe.

markavitale: We can probably get rid of all the methods updating colors and put them directly in updateLayer. Anywhere we were explicitly changing colors, just call setNeedsUpdateLayer.

harrieshin: huh... isn't updateLayer a macOS thing?

markavitale: Doh. Ignore my dopey mixing up of NSView and UIView calls. According to https://developer.apple.com/documentation/xcode/supporting_dark_mode_in_your_interface you should be fine with the traitCollectionDidChange location.

PR #13, opened 2020-04-15
update podspec and rename iOS Fabric folder names

harrieshin:
Verified with " pod lib lint FluentUI.podspec "


ios/FluentUI/Info.plist
```diff
@@ -15,7 +15,7 @@
 	<key>CFBundlePackageType</key>
 	<string>FMWK</string>
 	<key>CFBundleShortVersionString</key>
-	<string>0.0.1</string>
+	<string>$(MARKETING_VERSION)</string>
```
markavitale, line 18: @harrieshin is this intentional?

harrieshin: fixed! thanks!

PR #16, opened 2020-04-15
rename pod name from FluentUI to MicrosoftFluentUI

harrieshin:


harrieshin:
> Can we also update the version numbers in the macOS info plist to match? I think it's currently at 0.1.0, but since we've not published, moving backwards should be fine.

fixed!

PR #18, opened 2020-04-15
[CocoaPods] Make a separate resource bundle for macOS

markavitale:
For CocoaPods support, we should have a separate resource bundle that's copied into the framework and referenced by name instead of packing the resources directly into our framework. This allows clients to choose whether they want to link dynamically or statically.


MicrosoftFluentUI.podspec
```diff
@@ -15,5 +15,5 @@ Pod::Spec.new do |s|
   
   s.osx.deployment_target = "10.13"
```
harrieshin, line 16: you need to change the podspec version so that when you update the git tag it include your changes.

PR #27, opened 2020-04-23
adjust the corner radius for square avatar view

harrieshin:
reflect new design for group(square) avatar view in large, xlarge, xxlarge size.

### DIFF
--- a/ios/FluentUI/People Picker/MSAvatarView.swift
+++ b/ios/FluentUI/People Picker/MSAvatarView.swift
@@ -7,6 +7,7 @@ import UIKit
 
 // MARK: MSAvatarSize
 
+/// `MSAvatarSize` detemines the size, font, and corner radius for `MSAvatarView`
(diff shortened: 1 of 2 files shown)


ios/FluentUI/People Picker/MSAvatarView.swift
```diff
@@ -48,6 +48,17 @@ import UIKit
             return CGSize(width: 70, height: 70)
         }
     }
+
+    var squareCornerRadius: CGFloat {
```
markavitale, line 52: The other vars in here are public - just wanted to make sure this is internal by design.

markavitale: nit: Would also prefer this new var was documented, even if it is internal.

harrieshin: yeah i wasn't really inclined to make this public. but i can document it.

PR #32, opened 2020-04-25
[Localization] Add workflow to localize strings in the repository

markavitale:
## Summary
Add a localization workflow that calls into the internal Microsoft Touchdown Build Service to get up-to-date localizations for all of the strings in our repository

### scripts/GetLocalizedFiles.sh
A bash script that relatively simply wraps the Touchdown Build REST API. This is mostly based on a bash script written by the Touchdown build team (distributed by internal CocoaPod), modified for being called externally and simplified for our purposes.

### scripts/localize.sh
A simple shell script that grabs the relevant API input (Team ID, Application ID, Application Secret) and invokes GetLocalizedFiles.sh for each platform where we need strings localized. It's built so that it checks the environment for certain variables and if they aren't available, it prompts the user interactively. In the workflow we'll expect these env variables to be set so the script doesn't get stuck at the interactive prompt.

### .github/workflows/localize.yml
Add a workflow that's triggered on pushes to master and also triggered on a schedule (daily at midnight). This dual triggering allows any push to immediately register new strings with the service, while letting a future scheduled run to update the strings once they have been localized in the service.

This checks out the source, runs the localize.sh script (which modifies the loc files in-place) and then determines if there have been any changes and creates a PR if there have been changes.

## Verification
Allow this to run on my PR branch temporarily (not just on master) and allow the service calls to be made. See the automatic PR opened in my fork (note that since this call used the ?localizationTarget=false parameter, the strings are returned in English for all locales). https://github.com/markavitale/fluentui-apple/pull/4


scripts/localize.sh
```diff
@@ -0,0 +1,26 @@
...
+    read TDBUILD_AAD_APPLICATION_CLIENT_SECRET
+    stty echo
+    printf "\n"
+fi
+
+# Localize iOS resources
+scripts/GetLocalizedFiles.sh -t $TDBUILD_TEAM_ID -u -a $TDBUILD_AAD_APPLICATION_CLIENT_ID -p $TDBUILD_AAD_APPLICATION_CLIENT_SECRET -f ios/FluentUI/Resources/Localization/en.lproj -r ios/Localization/en.lproj -o ios/FluentUI/Resources/Localization
+
+# Localize macOS resources
+scripts/GetLocalizedFiles.sh -t $TDBUILD_TEAM_ID -u -a $TDBUILD_AAD_APPLICATION_CLIENT_ID -p $TDBUILD_AAD_APPLICATION_CLIENT_SECRET -f macos/FluentUIResources-macos/Strings/en.lproj -r macos/Localization/en.lproj -o macos/FluentUIResources-macos/Strings
```
markavitale, line 26: @anandrajeswaran @harrieshin, the `-r macos/Localization/en.lproj` is the relative path that these strings should be stored under on the tdbuild localization server. It doesn't actually need to correspond to any paths in our repo. This allows us to be referencing the same localization files on the service but still be able to move/refactor our code in the repo so a moved file doesn't trigger relocalization of all the strings. 

Do you all like the <platform>/Localization/en.lproj pattern?

.github/workflows/localize.yml
```diff
@@ -0,0 +1,34 @@
...
+    branches:
+      - master
+  # Also run every day at midnight to pick up new localizations
+  schedule:
+    # * is a special character in YAML so you have to quote this string
+    - cron:  '0 0 * * *'
+
+jobs:
+  Localize:
+    runs-on: macos-10.15
```
anandrajeswaran, line 14: is there a way we can have this specified in only one place (I think other scripts use this?)

scripts/GetLocalizedFiles.sh
```diff
@@ -0,0 +1,156 @@
...
+echo "using oauth."
+else
+echo "using NTLM."
+fi
+
+# parse json and return value of the key
+function jsonValue() {
+KEY=$1
+num=$2
+jsonParseCmd=`awk -F"[,:}]" '{for(i=1;i<=NF;i++){if($i~/'$KEY'\042/){print $(i+1)}}}' | tr -d '"' | sed -n ${num}p`
```
anandrajeswaran, line 63: yay, more awk

markavitale: Yeah this comes from the original script from the localization team, not interested in rewriting/clarifying it.

.github/workflows/localize.yml
markavitale: I've spent some time looking through this and don't think we can abstract the runs-on variable. We could look into moving the localization step into the ci.yml file, and have it conditionally run for pushes to master with an [if statement](https://help.github.com/en/actions/reference/context-and-expression-syntax-for-github-actions), but for now I want to keep them separate for our own sanity standing this up and not wanting to break our CI.

As of right now, it's duplicated between this file and `.gitub/workflows/ci.yml`. While the same string shows up in the `.ado` folder, the fact that ADO Pipelines and GitHub Actions use the same VM image name isn't guaranteed to be true going forward.

scripts/localize.sh
markavitale: I've looked into it a bit more, and based on other examples, it probably makes the most sense to get rid of everything other than just `platform_name`. Other examples don't provide the en.lproj, and the Localization relative directory just kind of feels implied.

scripts/localize.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# Check if environment variables have been defined for our team ID, AAD App ID, and AAD App Secret for automated workflows before requesting them interactively
+if [ -z $TDBUILD_TEAM_ID ]; then
```
Saadnajmi, line 4: If I understand this correctly, the secrets will exist on github / the build server, and developers of FluentUI-Apple won't need them for setup on local machines?

markavitale: If you are manually running the localization workflow on your machine and you don't have the env variables, this will prompt you interactively to provide each of the variables in the command line. I guess you _could_ set the env variables locally, and it would work exactly as it does in CI.

markavitale: But maybe to more correctly answer your question, I don't anticipate developers running the localization pipeline locally unless they are actively changing how the pipeline works. The whole localization story should be abstracted from developers by this pipeline.

.github/workflows/localize.yml
```diff
@@ -0,0 +1,34 @@
+name: Localize
+
+on: 
+  push:
+    branches:
+      - master
```
harrieshin, line 6: could we make this to be only on changes for xib or .string files later?

markavitale: Not that I'm aware of. See https://help.github.com/en/actions/reference/events-that-trigger-workflows for a list of ways you can invoke. Since we won't know what files changed until we've actually checked the commit, and we can only check the contents of the commit when we're in the workflow, I don't think we can avoid running the workflow.

Luckily, this is in line with how the API is designed to be called, and no PR will be made if no changes are detected (which will be true for the majority of runs of this pipeline).

PR #35, opened 2020-04-28
[Localization] Run Workflow on Linux and scope to strings file changes

markavitale:
# Summary
macOS agents on GitHub are more limited and thus more costly to run on (in terms of parallelism). We shouldn't use macOS agents for workflows where we can instead use Linux. Since our localization workflow is just bash, it runs fine on linux agents.

# Changes
- Switch runs-on to the latest Ubuntu image
- Only run this workflow on push when there are changes to strings or stringsdict files, as those are the only files we localize
- Note, the scheduled runs will still happen regardless.

### DIFF
--- a/.github/workflows/localize.yml
+++ b/.github/workflows/localize.yml
@@ -4,6 +4,9 @@ on:
   push:
     branches:
       - master
+    paths:
+      - '**.strings'
+      - '**.stringsdict'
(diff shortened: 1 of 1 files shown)


markavitale:
Not sure why the CI builds aren't running. I'm going to try closing and re-opening the PR to kick it.

PR #40, opened 2020-04-30
rename all the iOS classes to have MSF prefix for objc and no prefix for swift

harrieshin:
For swift, the naming convention is Foo
For objc, the naming convention is MSFFoo


ios/FluentUI/Controls/Button.swift
```diff
@@ -5,9 +5,13 @@
...
 import UIKit
 
-// MARK: MSButtonStyle
+// MARK: ButtonStyle
 
-@objc public enum MSButtonStyle: Int, CaseIterable {
+@available(*, deprecated, renamed: "ButtonStyle")
+public typealias MSButtonStyle = ButtonStyle
+
+@objc(MSFButtonStyle)
```
markavitale, line 13: I looked through the code but didn't quite find it. Is there a spot where we're doing the Obj-C typealiases for MSFButton -> MSButton?

harrieshin: @markavitale  i haven' decided to drop it for Objc typealias. the problem is that most of the Objc directly import "FluentUI-Swift.h" rather than "FluentUI.h" Also, Office devmain apps are only clients that are heavily Objc and I rather just fixed them all in devmain. do you have strong preference to support objc type alias?

markavitale: I was hoping this change wouldn't break existing clients. We definitely have existing obj-c clients that bring in the framework via #import <FluentUI/FluentUI.h>. The only folks bringing in FluentUI-Swift.h directly are probably clients that have explicitly turned off clang modules, which is really probably only Office since it's a default setting for new projects.

markavitale: In the end, this probably covers 99% of our bases and works for Swift clients. The fixes for Obj-C clients is really just adding a single letter that Xcode will almost surely make the recommendation ("did you mean MSFButton?"). I think let's move forward with this, and we can revisit if this ends up being a difficult migration for objective-c clients.

PR #43, opened 2020-05-01
[Accessibility] Enable full macOS date picker accessibility

lcapkovic:
## Summary
We limited date picker accessibility as a workaround for not having localized accessibility strings. Now that localization works, we can get rid of that workaround.

## The Change
Remove all references to `accessibilityTemporarilyRestricted` and its definition.

## Verification
Accessibility inspector in the mac test app to check we have the correct view hierarchy and strings. VoiceOver sanity check.

### DIFF
--- a/macos/FluentUI/DatePicker/CalendarView.swift
+++ b/macos/FluentUI/DatePicker/CalendarView.swift
@@ -58,7 +58,7 @@ class CalendarView: NSView {
 		}
 		
 		// Accessibility
-		setAccessibilityElement(!DatePickerView.accessibilityTemporarilyRestricted)
+		setAccessibilityElement(true)
 		setAccessibilityLabel(NSLocalizedString(
 			"DATEPICKER_ACCESSIBILITY_CALENDAR_VIEW_LABEL",
 			tableName: "FluentUI",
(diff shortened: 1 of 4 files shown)


macos/FluentUI/DatePicker/CalendarView.swift
```diff
@@ -58,7 +58,7 @@ class CalendarView: NSView {
 		}
 		
 		// Accessibility
-		setAccessibilityElement(!DatePickerView.accessibilityTemporarilyRestricted)
+		setAccessibilityElement(true)
```
markavitale, line 61: I'm assuming the default value is `false` for everywhere we're explicitly setting to `true`?

harrieshin: > By default, NSView and its subclasses set this value to NO; however, if your NSView subclass adopts one of the accessibility protocols, the system changes the default value to YES.

markavitale: Yeah from the snippets in the review it wasn't immediately obvious if we were subclassing NSControls or just pure NSViews.

PR #46, opened 2020-05-02
add new properties to customize popupmenucontroller and drawercontroller colors

harrieshin:
Instead of forcing clients to override the sdk static color variable to change the colors of popupmenucontroller or drawer controller, add new properties for clients to change the colors of the instance of those classes.

Before | After
------------ | -------------
Screen Shot 2020-05-01 at 4 40 47 PM|Simulator Screen Shot - iPhone Xʀ - 2020-05-06 at 17 09 58

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
@@ -140,7 +140,7 @@ class PopupMenuDemoController: DemoController {
             PopupMenuItem(image: UIImage(named: "3-day-view-24x24"), title: "3-Day", isEnabled: false, isSelected: calendarLayout == .threeDay, onSelected: { self.calendarLayout = .threeDay })
         ]
 
-        let menuBackgroundColor = UIColor(red: 0.2, green: 0.2, blue: 0.2, alpha: 1.0)
+        let menuBackgroundColor: UIColor = .darkGray
 
         for item in items {
             item.titleColor = .white
@@ -154,7 +154,7 @@ class PopupMenuDemoController: DemoController {
 
         controller.backgroundColor = menuBackgroundColor
         controller.resizingHandleViewBackgroundColor = menuBackgroundColor
-        controller.separatorColor = UIColor(red: 0.3333333333, green: 0.3333333333, blue: 0.3333333333, alpha: 1.0)
+        controller.separatorColor = .lightGray
 
         present(controller, animated: true)
     }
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -120,6 +120,12 @@ open class DrawerController: UIViewController {
         case popover
     }
 
+    @objc open var backgroundColor: UIColor = Colors.Drawer.background {
```
markavitale, line 123: Each of these public properties should have `///` documentation.

ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
```diff
@@ -129,4 +130,32 @@ class PopupMenuDemoController: DemoController {
...
+            item.imageSelectedColor = .white
+            item.accessoryCheckmarkColor = .white
+            item.backgroundColor = menuBackgroundColor
+        }
+
+        controller.addItems(items)
+
+        controller.backgroundColor = menuBackgroundColor
+        controller.resizingHandleViewBackgroundColor = menuBackgroundColor
+        controller.separatorColor = UIColor(red: 0.3333333333, green: 0.3333333333, blue: 0.3333333333, alpha: 1.0)
```
markavitale, line 157: Any reason not to use [opaqueSeparatorColor](https://developer.apple.com/documentation/uikit/uicolor/3173133-opaqueseparatorcolor?language=objc)?

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -66,6 +66,17 @@ open class PopupMenuController: DrawerController {
...
         return height
     }
+
+    open override var backgroundColor: UIColor {
+        get {
+            return super.backgroundColor
+        }
+        set {
+            super.backgroundColor = newValue
+            tableView.backgroundColor = newValue
```
markavitale, line 76: Any reason not to handle that with a didSet instead of overriding the getter and setter?

ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
markavitale: (If Fluent wants a different color, that's fine, but this is in the test code and is hardcoded)

harrieshin: i don't care either way. i was just copying from PDFkit's codebase.

harrieshin: opaqueSeparatorColor is only iOS 13 so i probably use just lightGray or something else

PR #51, opened 2020-05-06
add @objc

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
add @objc for some of the public initializers
fix crash on launch in demo app in Catalina development because it couldn't initialize the storyboard with "TableViewCell" properly.

### Verification

Xcode 11.4 Catalina 
12.4 iPhone XR simulator
13.4.1 iPhone 11 Pro Max Simulator
### Pull request checklist

This PR has considered:


anandrajeswaran:
Technically we could and probably should move the demo app fix to a separate PR. Probably ok in this case because I can't imagine we'd ever want to revert or pull either change without the other in practice

PR #54, opened 2020-05-07
add alternative primary colors in demo app

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
- Some clients can change their app primary color within the same app session. Update Colors.swift to not cache Colors.primary values for control's static property.
- Demo app changes:
-- add alternative primary color palette(green) in the demo app
-- use UITableView to show the colors rather than UIStackView.
-- add accessory button in the table header view to switch app primary color

### Verification

|Simulator Screen Shot - iPhone 11 Pro Max - 2020-05-06 at 23 20 07| Simulator Screen Shot - iPhone Xʀ - 2020-05-06 at 23 12 31 Simulator Screen Shot - iPhone Xʀ - 2020-05-06 at 23 14 27|

### Pull request checklist

This PR has considered:


ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/Colors/PrimaryColor.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
```
anandrajeswaran, line 1: For disambiguity, should we call this DemoPrimaryColor.colorset?

ios/FluentUI/Core/Colors.swift
```diff
@@ -288,14 +287,14 @@ public final class Colors: NSObject {
     }
 
     public struct Badge {
-        public static var background: UIColor = primaryTint40
```
anandrajeswaran, line 291: I like this change a lot as is - making these computed should be a huge help as we figure out how to make colors truly dynamic for the case of multiple fluent components with different themes active at the same time (multi-window).

Getting greedy, is this also a chance to start pushing some of these out of Colors.swift into the files they actually affect? Would help a lot with clarity of code and understanding impact of changes, but also could help with bundle size if we go down a path where a client can handpick the controls they want and not get all of them

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -7,84 +7,157 @@ import FluentUI
...
+            DemoColorItem(text: Colors.Palette.warningTint40.name, color: Colors.Palette.warningTint40.color),
+            DemoColorItem(text: Colors.Palette.warningTint10.name, color: Colors.Palette.warningTint10.color),
+            DemoColorItem(text: Colors.Palette.warningShade40.name, color: Colors.Palette.warningShade40.color),
+            DemoColorItem(text: Colors.Palette.warningShade30.name, color: Colors.Palette.warningShade30.color)
+        ])
+    ]
+    private var tableView: UITableView!
+    private var usingDefaultPrimaryColor: Bool = false
+
+    override func viewDidLoad() {
```
anandrajeswaran, line 73: This looks like it's copied code, but easy to fix: this looks like a loadView rather than viewDidLoad. viewDidLoad is used for additional configuration after loading from a xib. loadView is where programmatic hierarchy is created and usually culminates in a "view = " call. See hopefully https://github.com/anandrajeswaran/apple-ux-guide/blob/master/ViewControllers.md later today :)

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -7,84 +7,157 @@ import FluentUI
...
+        super.viewDidLoad()
+
+        tableView = UITableView(frame: view.bounds, style: .grouped)
+        tableView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```
anandrajeswaran, line 77: Should we just use constraints here to set a good example? Then we can avoid passing in the dummy size (view.bounds) above

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -7,84 +7,157 @@ import FluentUI
...
+        tableView.register(TableViewCell.self, forCellReuseIdentifier: TableViewCell.identifier)
+        tableView.register(TableViewHeaderFooterView.self, forHeaderFooterViewReuseIdentifier: TableViewHeaderFooterView.identifier)
+        tableView.dataSource = self
+        tableView.delegate = self
+        tableView.separatorStyle = .none
+        tableView.backgroundColor = Colors.Table.backgroundGrouped
```
anandrajeswaran, line 83: Remove

ios/FluentUI/Core/Colors.swift
markavitale: That feels like a separate change to me logically, even if it's something we choose to do next.

ios/FluentUI/Core/Colors.swift
```diff
@@ -311,21 +310,21 @@ public final class Colors: NSObject {
     }
 
     public struct BarButtonItem {
-        public static var primary: UIColor = foreground8
+        public static var primary: UIColor { return UIColor(light: Colors.primary, dark: .white) }
         public static var secondary: UIColor = foreground2d
```
markavitale, line 314: Out of curiosity, so in this change we're making the primary color settable by the clients live, but not the secondary? Is that something we'd want to do in the future or not something we expect to be a client pain point?

NOT SAYING TO HANDLE DOCUMENTATION OF THIS CLASS IN THIS CHANGE: this is another case where "simple" or "trivial" documentation of APIs would allow us to adapt it to indicate that some of these properties are readonly-computed and others are just value assignments. A client could before this change call 
```Swift
Colors.BarButtonItem.primary = foo
Colors.BarButtonItem.secondary = bar
```

But now the first line will give an error while the second line would not. Maybe this is a desired outcome because it forces consistency with the primary colors and gives freedom to clients with secondary colors, but it's hard to understand just by looking at the code without any documentation.

ios/FluentUI/Core/Colors.swift
harrieshin: i understand the desire and completely agree with your direction. but i don't think it is something i should do in _this_ PR

anandrajeswaran: Yup, makes sense

ios/FluentUI/Core/Colors.swift
harrieshin: Yeah it seems hard to find the write documentation right now. what if I write, this comment in the mean time 
`    /// Variation of App brand colors. If an application is a hub of different apps, `primary` color could change within the same foreground session.
    /// It is recomended not to cache `primary` color because it could change.
 
@objc public static var primary: UIColor = communicationBlue`

markavitale: Sounds good!

PR #66, opened 2020-05-19
Link control for macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

We are adding a very simple Link Control, which is implemented as a simple stylized subclass of NSButton. 

The class contains two properties, "content" and "url", which match the interface of the [FluentUI React Native Link](https://developer.microsoft.com/en-us/fluentui#/controls/crossplatform/link). This is intentional, so that it's easier to wrap this native control in react native in the future.

In the process, I also replaced the "Hover Button" from the test app, since that's a relic from when we crated the repo

### Verification

| Test App with Hover Button View Controller Screen Shot 2020-05-19 at 4 37 43 PM | Test App with Link View Controller Screen Shot 2020-05-19 at 4 38 59 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/66)


macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+/// A lightweight hyperlink control
+@objc(MSFLink)
+open class Link : NSButton {
```
harrieshin, line 10: should it be linkButton?

Saadnajmi: I wanted to clearly distinguish it from being a button, since [FluentUI React Native](https://developer.microsoft.com/en-us/fluentui#/controls/crossplatform) has them as separate controls. Also reserve the right to reimplement this with NSTextView or something else if we find it's better off that way 🙂

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,48 @@
...
+	@objc public init(content: String, url: NSURL) {
+		self.content = content
+		self.url = url
+		super.init(frame: .zero)
+		alignment = .natural
+		isBordered = false
+		target = self
+		action = #selector(linkClicked(_:))
+		
+		let attributedTitleAttributes = [
```
markavitale, line 25: This name feels a bit redundant.

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,48 @@
...
+
+/// A lightweight hyperlink control
+@objc(MSFLink)
+open class Link : NSButton {
+	
+	/// Initialies a link with a content string that is displayed, and an underlying
+	/// - Parameters:
+	///   - content: The visible text of the link that the user sees.
+	///   - url: The URL that is opened when the link is clicked
+	@objc public init(content: String, url: NSURL) {
```
markavitale, line 16: Not sure how I feel about the "content" name for the string. Buttons, which this is, have titles, not "content". https://developer.apple.com/documentation/appkit/nsbutton/1524430-title

markavitale, line 16: I'm guessing this might be the control spec, but do we like having this as a URL as opposed to a closure that we call when this is invoked?

Saadnajmi: I can look into what it would take to keep it as the property "title", and then have React Native map the "content" prop to the native "title" property

Saadnajmi: Not sure I understand what you mean?

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,48 @@
...
+		self.url = url
+		super.init(frame: .zero)
+		alignment = .natural
+		isBordered = false
+		target = self
+		action = #selector(linkClicked(_:))
+		
+		let attributedTitleAttributes = [
+			NSAttributedString.Key.foregroundColor : NSColor.linkColor,
+			NSAttributedString.Key.font : NSFont.labelFont(ofSize: NSFont.labelFontSize)
```
Saadnajmi, line 27: I went with labelFontSize bc it seemed like the right thing to use here, but "systemFontSize" is also completely fair. I'm not sure if they actually differ

macos/FluentUI/Link.swift
markavitale: As it works right now, this is a button that looks like a link. We could scope the control to only work as a link that opens a URL, or we could make it more generic and let clients handle what happens when this button is pressed, even if it's something other than opening a URL.

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,48 @@
...
+	/// - Parameters:
+	///   - content: The visible text of the link that the user sees.
+	///   - url: The URL that is opened when the link is clicked
+	@objc public init(content: String, url: NSURL) {
+		self.content = content
+		self.url = url
+		super.init(frame: .zero)
+		alignment = .natural
+		isBordered = false
+		target = self
```
markavitale, line 22: These are standard button APIs and the fact that we rely on clients not changing them might be an issue. Should we at the least override the setter for target and action with a preconditionFailure(), make them unavailable, or some other way of saying "don't do this, our control relies on it"?

macos/FluentUI/Link.swift
Saadnajmi: just `titleAttributes`?

macos/FluentUI/Link.swift
Saadnajmi: Ah, I specifically wanted to make it only open URL's and nothing else. I figured if clients wanted to do more, they could use the (nonexistent) button control

macos/FluentUI/Link.swift
Saadnajmi: Oh! That's a great idea! I've been wondering how to stop clients from changing properties I need constant 😃

macos/FluentUITestApp/TestControlsViewController.swift
```diff
@@ -17,8 +17,8 @@ fileprivate struct Constants {
 class TestControlsViewController: NSViewController, NSTableViewDelegate, NSTableViewDataSource {
 	private let controls: [(title: String, type: NSViewController.Type)] = [
 							("FluentUI-macOS (placeholder)", TestPlaceholderViewController.self),
-							("Hover Button", TestHoverButtonViewController.self),
 							("Avatar View", TestAvatarViewController.self),
+							("Link", TestLinkViewController.self),
```
anandrajeswaran, line 21: very minor: can we move down a line to keep these alphabetically ordered

macos/FluentUI/Link.swift
Saadnajmi: I removed the "content" property altogether, and just called the function parameter "title"

macos/FluentUI/Link.swift
Saadnajmi: Reverted this because I like systemFontSize better  ¯\_(ツ)_/¯

macos/FluentUI/Link.swift
Saadnajmi: Few problems with this:
If I just blindly put `preconditionFailure` calls on the didSet of `target`, `action`, etc..., then this would crash trying to set those in init. I can add a private flag `doneInitializing` that we have a `guard` on inside `didSet`, but that feels less clean.

Also, `NSButton` (along with any other alternative class I use) have lots of properties.. should I set them all to crash if you access them? or just the ones that I override, like we mentioned? This problem happens whether I subclass or return a stylized object, so I'm inclined to just clearly document "Please don't mess with X properties" and add `preconditionFailure` calls with the message "Improper use of Link Control, please do not override this property, or subclass/make your own custom object if you wish to use this"

anandrajeswaran: Upon thinking this through, I think differently - do we need to restrict clients? If they need to change the alignment or target, we should just let them? As long as the default they get from using fluentui APIs gives them the appearance they want, I don't think we should spend effort trying to protect them and instead should assume they are well aware of AppKit and can hopefully find other properties to perfectly adapt this to their needs?

anandrajeswaran: Related - overriding target seems like a very necessary thing if we're providing a default behavior of opening a URL. I could easily see someone asking to have it open their help viewer or settings or something like that and that functionality should already be baked into the control by AppKit

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,41 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+/// A lightweight hyperlink control
+@objc(MSFLink)
+open class Link : NSButton {
```
anandrajeswaran, line 10: Are we just making all controls "open" because someone will eventually want to be able to use an override for standard NSButton override points? Probably ok but just curious what our guiding principle is on that

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,41 @@
...
+	
+	/// Initialies a link with a content string that is displayed, and an underlying
```
anandrajeswaran, line 12: and an underlying what? Also update "content" here

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,41 @@
...
+	/// - Parameters:
+	///   - content: The visible text of the link that the user sees.
```
anandrajeswaran, line 14: name of parameter is now title

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,41 @@
...
+	///   - url: The URL that is opened when the link is clicked
+	@objc public init(title: String, url: NSURL) {
+		self.url = url
+		super.init(frame: .zero)
+		alignment = .natural
+		isBordered = false
+		target = self
+		action = #selector(linkClicked(_:))
```
anandrajeswaran, line 22: can we simplify by not taking in a sender parameter in linkClicked?

macos/FluentUI/Link.swift
anandrajeswaran, line 14: Interesting thing to address - what if someone wants to change the title later? What do they do? Maybe saying in the comments that it is not currently supported is best path?

macos/FluentUITestApp/TestLinkViewController.swift
```diff
@@ -4,12 +4,12 @@
 //
 
 import AppKit
+import FluentUI
 
-class TestHoverButtonViewController: NSViewController {
+class TestLinkViewController: NSViewController {
 	override func loadView() {		
 		let containerView = NSStackView(frame: .zero)
```
anandrajeswaran, line 11: as a very minor point to help make the UI declarative (I know I started this bad pattern) - more standard would be to do something like
`let link = Link(title: "Bing", url: NSURL(string: "https://www.bing.com")!)
let containerView = NSStackView(views: [link])

Also are there any other types of links we want to add here for illustrative purposes? Maybe one where the client customizes the target (and shows an alert) if we agree that's a thing they should be able to do?
`

macos/FluentUITestApp/TestControlsViewController.swift
```diff
@@ -17,9 +17,9 @@ fileprivate struct Constants {
 class TestControlsViewController: NSViewController, NSTableViewDelegate, NSTableViewDataSource {
 	private let controls: [(title: String, type: NSViewController.Type)] = [
 							("FluentUI-macOS (placeholder)", TestPlaceholderViewController.self),
```
markavitale, line 19: Unrelated to this change, should we remove this placeholder view controller?

macos/FluentUI/Link.swift
markavitale: Our default has been "there's no reason to prevent users from subclassing these controls" and there are certainly valid reasons for subclassing most AppKit control types.

macos/FluentUI/Link.swift
markavitale: I guess we could override didSet of title and route it through attributedTitle with our link attributes? If we wanted to enforce that titles on this button have the link style.

macos/FluentUI/Link.swift
HeyImChris, line 12: Initializes? spelling

macos/FluentUITestApp/TestControlsViewController.swift
Saadnajmi: I vote we keep it but replace the logo with the FluentUI Logo, and maybe have a Link (hehe) to the website

macos/FluentUITestApp/TestLinkViewController.swift
Saadnajmi: I added a link with an optional hover effect, and I'll add one more with an overriden target/action. Maybe also some links inside text to show how they don't wrap when they reach the edge of a paragraph (this is intended)

macos/FluentUI/Link.swift
Saadnajmi: Yep! Ended up overriding the title property. I could make setting the alternateTitle/attributedTitle a `preconditionFailure`, but I'll go with the earlier suggestion of "give clients a nice default, and let them do what they like afterwards"

Saadnajmi:
I added an optional property `showsUnderlineWhileMouseInside` because there is at least one example of a link in Office which does so (in case we want to replace it with the FluentUi-Apple link). I chose the name of the property by basing it off of the existing NSButton property `showsBorderWhileMouseInside`

macos/docs/Controls/Link.md
```diff
@@ -0,0 +1,54 @@
...
+
+## Best Practices
+### Do
+- Use to navigate to another page, window, or help topic; display a definition; initiate a command; or choose an option.
+- Make Links discoverable by visual inspection alone. Users shouldn't have to interact with your program to find links.
+- Use Links that give specific descriptive information about the result of clicking on the link, using as much text as necessary. Users should be able to - accurately predict the result of a link from its link text and optional Tooltip.
+- Use text that suggests clicking, such as a command starting with an imperative verb like "Show", "Print", "Copy", or "Delete".
+
+### Don't
+- Use if the action is destructive or irreversible. Because users associate links with navigation (and the ability to back out), Links aren't appropriate for commands with significant consequences.
```
Saadnajmi, line 18: All of this is lifted from the [FluentUI Web Link](https://developer.microsoft.com/fluentui#/controls/web/link) and [FluentUI React Native Link](https://developer.microsoft.com/fluentui#/controls/crossplatform/link) documentation (which happens to be the same). All that was removed was a reference to a tooltip.

macos/FluentUITestApp/TestLinkViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+import FluentUI
+
+class TestLinkViewController: NSViewController {
+	override func loadView() {
+		let linkWithNoHover = Link(title: "Link to Bing", url: NSURL(string: "https://wwww.bing.com")!)
```
Saadnajmi, line 11: Here's a fun test... what's an `NSURL` that _isn't_ a webpage that we could put here (like a file or image for instance)

harrieshin: nit: for iOS we use any example of nsurl to point to this GitHub website lol

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,127 @@
...
+		self.title = ""
+		initialize()
+	}
+	
+	@available(*, unavailable)
+	required public init?(coder decoder: NSCoder) {
+		preconditionFailure()
+	}
+	
+	open func initialize() {
```
anandrajeswaran, line 43: why open? Just entirely private?

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,127 @@
...
+		super.updateTrackingAreas()
+
+		// Remove existing trackingArea
+		if let trackingArea = trackingArea {
+			removeTrackingArea(trackingArea)
+			self.trackingArea = nil
+		}
+
+		// Create a new trackingArea
+		let opts: NSTrackingArea.Options = [.mouseEnteredAndExited, .activeAlways]
```
anandrajeswaran, line 85: "options" to avoid abbreivations?

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,127 @@
...
+	
+	@objc private func linkClicked() {
+		if let url = url {
+			NSWorkspace.shared.open(url as URL)
+		}
+	}
+}
+
+fileprivate let linkAttributes: [NSAttributedString.Key : Any] = [
+	.foregroundColor : NSColor.linkColor
```
anandrajeswaran, line 121: very minor. I thought convention in Swift was no space before semi-colon but space after. Not part of this review (and can ignore this comment in this review), but time for a swift format step_

macos/FluentUITestApp/TestLinkViewController.swift
anandrajeswaran: Curious why an NSURL that doesn´t point to a website is a case. I think this code treats URLs opaquely?

macos/FluentUITestApp/TestLinkViewController.swift
```diff
@@ -0,0 +1,37 @@
...
+		let containerView = NSStackView(views: [linkWithNoHover, linkWithHover, linkWithHoverAndNoURL, linkWithOverridenTargetAction])
+		containerView.edgeInsets = NSEdgeInsets(top: 100, left: 100, bottom: 100, right: 100)
+		containerView.orientation = .vertical
+		containerView.distribution = .gravityAreas
+		view = containerView
+	}
+	
+	@objc private func displayAlert() {
+		let alert = NSAlert()
+        alert.messageText = "Alert"
```
anandrajeswaran, line 33: spacing may be off?

macos/docs/Controls/Link.md
```diff
@@ -0,0 +1,54 @@
...
+
+
+## Best Practices
+### Do
+- Use to navigate to another page, window, or help topic; display a definition; initiate a command; or choose an option.
+- Make Links discoverable by visual inspection alone. Users shouldn't have to interact with your program to find links.
+- Use Links that give specific descriptive information about the result of clicking on the link, using as much text as necessary. Users should be able to - accurately predict the result of a link from its link text and optional Tooltip.
+- Use text that suggests clicking, such as a command starting with an imperative verb like "Show", "Print", "Copy", or "Delete".
+
+### Don't
```
anandrajeswaran, line 17: Probably also want a "Don't" for use link for navigation internal to your app? That conflicts with the valid help case, but there's gotta be some wording to explain that these should be used for navigating to external content and buttons should be used instead in most cases

macos/docs/Controls/Link.md
markavitale: The overview doesn't feel like it applies very well here. We don't support "visited" or "unvisited" link colors, it's really only used for "opening a given URL in an external browser", not "navigating to another page, window, or Help topic..." etc.

macos/FluentUI/Link.swift
Saadnajmi: Swift Format is definitely the next thing on my list of "Stuff-I-have-already-worked-on-that-belongs-public" now

macos/FluentUITestApp/TestLinkViewController.swift
Saadnajmi: I'm pretty sure the NSURL passed in can just as easily be a path to a file as it can be a path to a webpage, distinguished by setting the sheme to `file` or `https`. I feel that there is not a strong reason to restrict this control to only `NSURL`'s with the `https` scheme, especially given a client can easily override the `target/action`

anandrajeswaran: Agreed, but do we actually need to test that case? Is that testing the button? Or is that testing AppKit's openURL?

macos/FluentUI/Link.swift
```diff
@@ -0,0 +1,127 @@
...
+	/// Initializes a hyperlink with a title and no URL, useful if you plan to override the Target/Action
+	/// - Parameters:
+	///   - title: The visible text of the link that the user sees.
+	@objc public init(title: String) {
+		super.init(frame: .zero)
+		self.title = title
+		initialize()
+	}
+	
+	public override init(frame frameRect: NSRect) {
```
Saadnajmi, line 32: Add `@objc`

macos/FluentUITestApp/TestLinkViewController.swift
Saadnajmi: Fair. I guess we don't need to test it. Just the realization that I _could_ do that was kinda cool

PR #67, opened 2020-05-20
Button control for macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

We are adding a Stylized "Button" control for macOS. The styling is roughly based off of the FluentUI Native iOS and Android controls. There are 3 styles (Primary Filled, Primary Outline, and Borderless), which are set at initialization.

There's a number of things that still need to be addressed before check in:
- Colors (I used a lot of placeholder colors to approximate what we want)
- Icon Support (This will be needed for FluentUI React Native, and currently does not work will with this implementation
- Documentation (There is internal documentation, but we also need markdown pages for the website. This will be the last thing to do)

### Verification

Tested Light/Dark and Voiceover on macOS 10.15, along with rest/hover/pressed mouse states

| Placeholder NSButton Placeholder NSButton | Fluent Styled Buttons Fluent Styled Buttons |

### Pull request checklist

This PR has considered:

Buttons in test app

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/67)


macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+			break
+		case .borderless:
+			textColor = primaryColor
+			break
+		}
+		
+		if (isEnabled) {
+			if (mouseClicked) {
+				if #available(macOS 10.14, *) {
+					layer?.backgroundColor = fillColor.withSystemEffect(.pressed).cgColor
```
Saadnajmi, line 141: Side note: I didn't know these "withSystemEffect" API's existed till today and I'm glad they do 😁

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,48 @@
...
+	override func loadView() {
+		
+		let enabledButtons: [NSButton] = [
+			NSButton(title: "NSButton", target: nil, action: nil),
+			Button(title: "Primary Filled", style: .primaryFilled),
+			Button(title: "Primary Outline", style: .primaryOutline),
+			Button(title: "Borderless", style: .borderless),
+		]
+
+		let disabledButtons: [NSButton] = [
```
Saadnajmi, line 19: There's avoidable duplication here... but maps hurt my head

anandrajeswaran: particularly because we follow up with another loop we probably should use a map

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,48 @@
...
+			let spacing = stackView.spacing
+			stackView.edgeInsets = NSEdgeInsets(top: spacing, left: 0, bottom: spacing, right: 0)
+			return stackView
+		}
+		
+		// set our view to a horizontal stack view containing the vertical stack views
+		let containerView = NSStackView(views: stackViews)
+		containerView.alignment = .top
+		let spacing = containerView.spacing
+		containerView.edgeInsets = NSEdgeInsets(top: 0, left: spacing, bottom: 0, right: spacing)
```
anandrajeswaran, line 42: probably want directional edge insets here and elsewhere?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
+//
+// Copyright Microsoft Corporation
+//
```
anandrajeswaran, line 3: Is this the copyright appropriate for our license?

macos/FluentUITestApp/TestButtonViewController.swift
harrieshin: what's the reason why adding NSButton?

Saadnajmi: @harrieshin The NSButton on the top row was just for visual / size comparison..
I'll see if I can redo this with map

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+
+import AppKit
+
+fileprivate struct Constants {
+	
+	static let borderWidth: CGFloat = 1
+	
+	static let cornerRadius: CGFloat = 3
+	
+	static let minimumWidth: CGFloat = 120
```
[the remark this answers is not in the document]
Saadnajmi: Interesting. Just tested this, and it seems NSButton does have a minimum width of 28.
I added the minimum width originally because it was in a design for another button I made. 

Screen Shot 2020-05-20 at 2 40 30 PM

Here's what FluentButton looks like with a minimum width of 30.. I feel I prefer this.

Screen Shot 2020-05-20 at 2 48 47 PM

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+	static let horizontalPadding:  CGFloat = 2
+	
+	private init() {}
+}
+
+@objc public enum ButtonStyle: Int, CaseIterable {
+    case primaryFilled
+    case primaryOutline
+    case borderless
+}
```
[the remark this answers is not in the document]
Saadnajmi: Noticed this is not publicly documented in FluentUI-iOS either, which has 5 styles
```
public enum ButtonStyle: Int, CaseIterable {
    case primaryFilled
    case primaryOutline
    case secondaryOutline
    case tertiaryOutline
    case borderless
```

Followup change may be to add those in as well, since I'm not actually sure when to use `secondaryOutline` and `tertiaryOutline`. @harrieshin Thoughts?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+	}
+	
+	open override var intrinsicContentSize: NSSize {
+		var intrinsicContentSize = super.intrinsicContentSize;
+
+		if (intrinsicContentSize.width > Constants.minimumWidth) {
+			intrinsicContentSize.width += (Constants.horizontalPadding * 2);
+		} else {
+			intrinsicContentSize.width = Constants.minimumWidth;
+		}
```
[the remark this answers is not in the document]
Saadnajmi: Not sure I understand what you are proposing?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+open class Button: NSButton {
+	/// The primary color of the button, AKA, the fill color in the primaryFilled style, and the outline in the primaryOutline style
+	@objc open var primaryColor: NSColor = NSColor.clear
+	
+	private var trackingArea: NSTrackingArea?
+	
+	private var mouseEntered: Bool = false
+	private var mouseClicked: Bool = false
+	
+	override open var wantsUpdateLayer: Bool {
```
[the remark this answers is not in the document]
Saadnajmi: Realized that xcode auto fills these in as open when I wrote this code. I'm changing the unnecessary "opens" to "public". Probably don't ever want a client to change this in a subclass, so I'm changing it and others to public

macos/FluentUI/Button.swift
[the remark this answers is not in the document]
anandrajeswaran: intrinsicContentSize.width = min(intrinsicContentSize.width + Constants.horizontalPadding * 2, Constants.minimumWidth)

macos/FluentUI/Button.swift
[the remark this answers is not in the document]
Saadnajmi: Actually, it looks like NSButton doesn't have a minimum width, just nicely enforced horizontal padding. I removed the minimum width, and increased the horizontal padding to compensate

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+}
+
+// MARK: - Button
+
+/// A fluent styled button, with hover effects and a corner radius.
+@IBDesignable
+@objc(MSFButton)
+open class Button: NSButton {
+	/// The primary color of the button, AKA, the fill color in the primaryFilled style, and the outline in the primaryOutline style
+	@objc open var primaryColor: NSColor = NSColor.clear
```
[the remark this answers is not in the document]
Saadnajmi: I was following the precedent of AvatarView being `open`.

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+	
+	private var trackingArea: NSTrackingArea?
```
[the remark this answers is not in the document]
Saadnajmi: Updated to match our swift guide

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+    public override init(frame: CGRect) {
+		preconditionFailure()
+    }
+
+    public required init?(coder aDecoder: NSCoder) {
+		preconditionFailure()
+    }
+	
+	open func initialize() {
+		// Do common initialization work
```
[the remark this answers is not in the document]
Saadnajmi: Now there's more than one again, so a need for a helper returns.

macos/FluentUI/Button.swift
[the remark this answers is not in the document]
Saadnajmi: Removed the minimum width and the need for this code.

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+		mouseClicked = false
+		needsDisplay = true
+	}
+	
+	open override func updateLayer() {
+		layer?.borderWidth = Constants.borderWidth
+		layer?.masksToBounds = true
+		layer?.cornerRadius = Constants.cornerRadius
+				
+		var fillColor : NSColor = NSColor.clear
```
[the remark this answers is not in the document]
Saadnajmi: Went with  private computed properties over closures, which should(?) give the same level of immutability benefits

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+				} else {
+					layer?.backgroundColor = fillColor.cgColor
+				}
+			}
+		} else {
+			if #available(macOS 10.14, *) {
+				layer?.backgroundColor = fillColor.withSystemEffect(.disabled).cgColor
+			} else {
+				layer?.backgroundColor = NSColor.systemGray.withAlphaComponent(0.25).cgColor
+			}
```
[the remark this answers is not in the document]
Saadnajmi: Between the `if let layer = layer`, splitting all the color picking out to separate private computed properties, and reducing things to one call, I feel this is now much cleaner. Thank you for the suggestions =)

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,188 @@
...
+	
+	static let borderWidth: CGFloat = 1
+	
+	static let cornerRadius: CGFloat = 3
+	
+	static let minimumWidth: CGFloat = 120
+	
+	static let verticalPadding: CGFloat = 2
+	
+	static let horizontalPadding:  CGFloat = 2
```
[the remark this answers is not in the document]
Saadnajmi: Would it be better if I added comments to each constant to specify exactly what it does?
Horizontal padding is extra space between the title and the left/right edge. Likewise for vertical padding

Saadnajmi:
`TestButtonViewController` is still very much a work in progress, but I have an idea for a nice grid of buttons with different combos of parameters, and a control to change the title/color of all the fluentButtons to test how that behaves at runtime

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+/// Indicates what style our button is drawn as
+@objc public enum ButtonStyle: Int, CaseIterable {
+    case primaryFilled	// Solid fill color
+    case primaryOutline	// Clear fill color, solid outline
+    case borderless		// clear fill color, clear outline
```
harrieshin, line 12: nit: capitalization of "clear" to be consistent with rest

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+
+	@available(*, unavailable)
+	required public init?(coder decoder: NSCoder) {
+		preconditionFailure()
+	}
+	
+	open func initialize() {
+		// Do common initialization work
+		isBordered = false
+		wantsLayer = true
```
harrieshin, line 42: why are we setting wants layer to true when we override the property below?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+
+		return intrinsicContentSize;
+	}
+	
+	override public var wantsUpdateLayer: Bool {
+		return true
+	}
+	
+	/// The primary color of the button, AKA, the fill color in the primaryFilled style, and the outline in the primaryOutline style
+	@objc public var primaryColor: NSColor = .clear {
```
harrieshin, line 111: if the default style is .primaryFilled wouldn't it make more sense the default color is accent color/system blue?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+	
+	private var mouseEntered = false
+	private var mouseDown = false
+	
+	private var fillColor: NSColor {
+		return style == ButtonStyle.primaryFilled ? primaryColor : .clear
+	}
+	
+	private var outlineColor: NSColor {
+		return style == ButtonStyle.primaryOutline ? primaryColor.withAlphaComponent(0.4) : .clear
```
harrieshin, line 149: not related to this PR but when is MacOS defining primary color shades?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+	
+	private var textColor: NSColor {
+		return style == ButtonStyle.primaryFilled ? .white : primaryColor
+	}
+	
+	private var restBackgroundColor: NSColor {
+		return fillColor
+	}
+
+	private var hoverBackgroundColor: NSColor {
```
harrieshin, line 160: for iOS, we also define disabled button color. does Mac have that?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+
+	private var pressedBackgroundColor: NSColor {
+		if #available(macOS 10.14, *) {
+			return fillColor.withSystemEffect(.pressed)
+		} else {
+			return fillColor.withAlphaComponent(0.25)
+		}
+	}
+	
+	private var disabledBackgroundColor: NSColor {
```
harrieshin, line 176: personal opinion: the outline button doesn't look disabled unless the outline color matches the title color.

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+/// A fluent styled button, with hover effects and a corner radius.
+@IBDesignable
+@objc(MSFButton)
+open class Button: NSButton {
+
+	@objc public init(title: String, style: ButtonStyle = .primaryFilled) {
+		super.init(frame: .zero)
+		self.title = title
+		self.style = style
+        initialize()
```
anandrajeswaran, line 26: minor: inconsistent indentation (hi Swift-format)

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+        super.init(frame: frameRect)
+        initialize()
+	}
+
+	@available(*, unavailable)
+	required public init?(coder decoder: NSCoder) {
+		preconditionFailure()
+	}
+	
+	open func initialize() {
```
anandrajeswaran, line 39: does this need to be open?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+		// Do common initialization work
+		isBordered = false
+		wantsLayer = true
+
+		if #available(macOS 10.14, *) {
+			primaryColor = .controlAccentColor
+		} else {
+			primaryColor = .systemBlue
+		}
+		update()
```
anandrajeswaran, line 49: Can we come up with a more specific name? As code evolves we tend to want to split these up for clarity and perf gains so starting with a more explicit title can do that by design

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+		return true
+	}
+	
+	/// The primary color of the button, AKA, the fill color in the primaryFilled style, and the outline in the primaryOutline style
+	@objc public var primaryColor: NSColor = .clear {
+		didSet {
+            if primaryColor != oldValue {
+                update()
+            }
+		}
```
anandrajeswaran, line 116: odd indentation here and in the style var

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,207 @@
...
+	}
+	
+	private func update() {
+		if let layer = layer {
+			layer.borderWidth = borderWidth
+			layer.masksToBounds = true
+			layer.cornerRadius = cornerRadius
+			layer.backgroundColor = layerBackgroundColor.cgColor
+			layer.borderColor = outlineColor.cgColor
+		}
```
anandrajeswaran, line 135: these parts can just go in updateLayer? That would be needed to ensure we properly update from light to dark mode where we use cgColors and not NSColors opaquely

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+		containerView.orientation = .horizontal
+		
+		view = containerView
+	}
+	
+	private func testNSButtons() -> [NSButton] {
```
anandrajeswaran, line 38: minor: name sounds like an xctest :). Since this is a private fun in a test file, can just be something like "makeNSButtons/makeFluentButtons"

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,67 @@
...
+		return [
+			NSButton(title: "NSButton", target: nil, action: nil),
+			NSButton(title: "+", target: nil, action: nil)
```
anandrajeswaran, line 41: Harrie commented on this earlier, but do we really want AppKit controls in our demo app in any scenario?

macos/FluentUI/Button.swift
Saadnajmi: This is (very) confusing, and I recommend reading the [Layer](https://developer.apple.com/documentation/appkit/nsview/1483298-layer), [wantsLayer](https://developer.apple.com/documentation/appkit/nsview/1483695-wantslayer), and [wantsUpdateLayer](https://developer.apple.com/documentation/appkit/nsview/1483461-wantsupdatelayer) documentation for more info, but TL;DR:

Setting `wantsLayer` determines if the view is layer-backed, 
Overriding `wantsUpdateLayer` tells the view whether we use `draw(_:)` or `updateLayer()` to update the contents of the layer.

macos/FluentUI/Button.swift
Saadnajmi: I think we're figuring that out right now, and that is one reason why this is a draft PR and not a real one

macos/FluentUI/Button.swift
Saadnajmi: That seems fair, and the iOS / macOS buttons should align. I think that would be blocked until we get primary color shades for macOS.

macos/FluentUITestApp/TestButtonViewController.swift
Saadnajmi: Fair enough, I took out all the NSButtons.

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,214 @@
...
+	@objc public init(title: String, style: ButtonStyle = .primaryFilled) {
+		super.init(frame: .zero)
+		self.title = title
+		self.style = style
+		initialize()
+    }
+	
+	public override init(frame frameRect: NSRect) {
+        super.init(frame: frameRect)
+		initialize()
```
anandrajeswaran, line 31: inconsistent spacing

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,214 @@
...
+		}
+	}
+
+	/// The primary style of the button
+	@objc public var style: ButtonStyle = .primaryFilled {
+		didSet {
+            if style != oldValue {
+                update()
+            }
+		}
```
anandrajeswaran, line 125: inconsistent spacing

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,214 @@
...
+	private var mouseDown = false
+	
+	private var fillColor: NSColor {
+		return style == ButtonStyle.primaryFilled ? primaryColor : .clear
+	}
+	
+	private var outlineColor: NSColor {
+		let baseOutlineColor = style == ButtonStyle.primaryOutline ? primaryColor.withAlphaComponent(0.4) : .clear
+		if isEnabled {
+			return style == ButtonStyle.primaryOutline ? primaryColor.withAlphaComponent(0.4) : .clear
```
anandrajeswaran, line 161: just return baseOutlineColor? Or am I missing something

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+			button.image = NSImage(named: NSImage.addTemplateName)
+			button.imagePosition = .imageLeading
+			return button
+		}
+		let disabledButtons = buttons().map { button -> NSButton in
+			button.isEnabled = false
+			return button
+		}
+		
+		let gridView = NSGridView(frame: .zero)
```
anandrajeswaran, line 43: love it! NSGridView is great

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+		gridView.rowSpacing = 20
+		gridView.columnSpacing = 20
```
anandrajeswaran, line 45: pull out to constant? Good to treat demo code like production code

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+	
+		// Insert the Row Titles as the last step
+		let rowLabels: [NSView] = [
+			emptyCell,
+			NSTextField(labelWithString: "Title"),
+			NSTextField(labelWithString: "Image"),
+			NSTextField(labelWithString: "Title and Image"),
+			NSTextField(labelWithString: "Disabled")
+		]
+		gridView.insertRow(at: 0, with: rowLabels)
```
anandrajeswaran, line 66: should we just do this as the first step so we can addRow instead of inserting at 0? Probably not, but just hoping to learn why we needed to do it this way

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+		gridView.addRow(with: [])	//Padding
+
+		let containerView = NSStackView()
+		containerView.orientation = .vertical
+		containerView.translatesAutoresizingMaskIntoConstraints = false
+		containerView.edgeInsets = NSEdgeInsets(top: 20, left: 20, bottom: 20, right: 20)
```
anandrajeswaran, line 72: pull that 20 into a constant per other comment

macos/FluentUITestApp/TestButtonViewController.swift
Saadnajmi: gridView is weird. If I add the row first, then add the columns, it adds the new column not under my label row, but to the right (Trailing) of the label row. This was the simplest way I found to ensure the row was added at the top, and that everything else was just an "addColumn"

macos/FluentUITestApp/TestButtonViewController.swift
Saadnajmi: Yes... very great.. so easy to use (/sarcasm)

macos/FluentUI/Button.swift
Saadnajmi: This got mixed up and should just be baseOutlineColor, good catch.

Saadnajmi:
Documentation needs to updated before this goes in

macos/FluentUI/Button.swift
Saadnajmi: I meant this to be the most generic form of update.. `updateColorsAndStyle()`? or `updateButtonColorsAndStyle`

macos/FluentUI/Button.swift
Saadnajmi: Spoke with our designer Kaan earlier, and all the color work / figuring out primary color shades is still very much an open question for how they play with semantic colors. I will consider it not a blocker for this review

macos/FluentUI/Button.swift
Saadnajmi: For some more context: Kaan agrees here that the outline should probably be gray, but moreso we need to align what we do with disabled colors with what Apple does. In this case, the disabled outline button has the outline color with the "disabled" system effect applied on top of it. The text in the control has some.. other filter being applied on top of it from disabling the control that I haven't quite figured out where it's coming from yet (I'm not manually greying out the text). I'll consider this not a blocker for this review

macos/docs/Controls/Button.md
```diff
@@ -0,0 +1,39 @@
...
+let primaryFilledButton = Button(title: "FluentUI Button", style: .primaryFilled),
+let primaryOutlineButton = Button(title: "FluentUI Button", style: .primaryOutline),
+let BorderlessButton = Button(title: "FluentUI Button", style: .borderless),
+```
+
+```Swift
+// You can set an image to ony display an image
+let buttonWithImage = Button()
+buttonWithImage.imagePosition = .imageOnly
+buttonWithImage.image = NSImage(named: NSImage.addTemplateName)
```
Saadnajmi, line 29: Benefit of writing documentation: You realize things like: "This totally could have just been one line with it's own initializer

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+		self.title = title
+		self.style = style
+		initialize()
+    }
+	
+	@objc public init(image: NSImage, style: ButtonStyle = .primaryFilled) {
+		super.init(frame: .zero)
+		self.style = style
+		initialize()
+		// Image must be set after `wantsLayer = true` so that the image is on the layer and not the underlying NSButtonCell
```
anandrajeswaran, line 32: Can we guard/fail this in code?

macos/FluentUI/Button.swift
anandrajeswaran: I guess that fits in to the general goal of swiftUI of making UI declarative by pushing all updatable properties together. This seems fair in general, but we should just verify we're not calling this multiple times on loading

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+		isBordered = false
+		wantsLayer = true
+		update()
+	}
+	
+	override public var image: NSImage? {
+		didSet {
+			if imagePosition != .imageOnly {
+				preconditionFailure("We currently do not support adding both an image and a title. Please set the imagePosition to .imageOnly before setting an image")
+			}
```
anandrajeswaran, line 58: Probably more of a TODO than a preconditionFailure? This definitely seems like something we will support so at the very least we should remove the instructions for how to workaround it and replace with anything useful that could help someone contribute a fix if they run into this

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+			rect: bounds,
+			options: opts,
+			owner: self,
+			userInfo: nil
+		)
+		addTrackingArea(trackingArea)
+		self.trackingArea = trackingArea
+	}
+	
+	override public func mouseEntered(with event: NSEvent) {
```
anandrajeswaran, line 83: open override? Does it actually need to be public?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+	
+	override public func mouseUp(with event: NSEvent) {
+		mouseDown = false
+		needsDisplay = true
+	}
+	
+	open override func updateLayer() {
+		if let layer = layer {
+			layer.borderWidth = borderWidth
+			layer.masksToBounds = true
```
anandrajeswaran, line 106: Do we need to be doing this? Isn't this the default?

Regardless we can't be doing this. From https://developer.apple.com/library/archive/releasenotes/AppKit/RN-AppKit/:
macOS 10.13 is more consistent around updates to properties of views' layers (that is, layers that are the .layer of some NSView, however created). As before, if an application modifies such a layer by changing any of the following properties, the behavior of the application is undefined: bounds, position, zPosition, anchorPoint, anchorPointZ, transform, affineTransform, frame, hidden, geometryFlipped, masksToBounds, opaque, compositingFilter, filters, shadowColor, shadowOpacity, shadowOffset, shadowRadius, shadowPath, layoutManager.

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+			}
+		}
+	}
+
+	private func update() {
+		self.attributedTitle = NSAttributedString(string: title, attributes: [.foregroundColor : textColor])
+		needsDisplay = true
+	}
+	
+	private func disabledColor(color: NSColor) -> NSColor {
```
anandrajeswaran, line 149: more swift-y might be
`
private func disabledColor(**for** color: NSColor) -> NSColor
`

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,230 @@
...
+		if #available(macOS 10.14, *) {
+			return color.withSystemEffect(.disabled)
+		} else {
+			return color.withAlphaComponent(0.25)
+		}
+	}
+	
+	private static var defaultPrimaryColor: NSColor {
```
anandrajeswaran, line 157: since this is static, might move it down to be more a constant and not in the middle of a bunch of instance configuration?

macos/FluentUITestApp/TestButtonViewController.swift
```diff
@@ -0,0 +1,89 @@
...
+		gridView.insertRow(at: 0, with: rowLabels)
+		gridView.addRow(with: [])	//Padding
+
+		let containerView = NSStackView()
+		containerView.orientation = .vertical
+		containerView.translatesAutoresizingMaskIntoConstraints = false
+		containerView.edgeInsets = NSEdgeInsets(top: containerViewEdgeInsets,
+												left: containerViewEdgeInsets,
+												bottom: containerViewEdgeInsets,
+												right: containerViewEdgeInsets)
```
anandrajeswaran, line 75: indentation looks off?

macos/docs/Controls/Button.md
```diff
@@ -0,0 +1,36 @@
...
+Buttons are one of the core controls that make an app feel native to the platform it's on. It’s important to respect the platform's paradigms in order to help the user feel at home on macOS and keep the experience quality high.
+
+![ButtonViews.png](.attachments/ButtonViews.png)
+
+## Best Practices
+### Do
+- Make sure the label conveys a clear purpose of the button to the user.
+- Describe the action the button will perform, ideally with a verb. Use concise, specific, self-explanatory labels, usually a single word.
+
+### Don't
```
anandrajeswaran, line 13: Should clients not override our cell? Or can/should they?

macos/docs/Controls/Button.md
```diff
@@ -0,0 +1,36 @@
...
+- Place the default focus on a button that performs a destructive actionn. Instead, place the default focus on the button that performs the "safe action" (e.g "Cancel").
+
+## Usage
+```Swift
+// There are 3 styles to choose from, the default being primary filled
+let primaryFilledButton = Button(title: "FluentUI Button", style: .primaryFilled),
```
anandrajeswaran, line 19: Do we describe when each should be used?

macos/FluentUI/Button.swift
Saadnajmi: No, but I thought we defaulted to public unless otherwise needed?

macos/FluentUI/Button.swift
Saadnajmi: @markavitale mentioned we could reference a github issue here

macos/FluentUI/Button.swift
anandrajeswaran: I would default the opposite - only expose things for a reason. I don't know if it's verifiable fact that not making things public reduces size, but that makes sense in my head - I'd assume it takes extra instructions/code to expose something publicly

macos/FluentUI/Button.swift
Saadnajmi: What do you mean? Clients should never be able to set an image before `wantsLayer = true`

anandrajeswaran: Well if the order is notable enough that it's worth a comment in code, we might as well add verification so someone refactoring it will know if they screwed it up?

macos/FluentUI/Button.swift
Saadnajmi: I rewrote this to remove the need for update() -> Logic is now done in `didSet` for `title`, and usages of `update()` have been replaced with `needsDisplay = true`

macos/FluentUI/Button.swift
Saadnajmi: Added guards to `willSet` of the `title` and `image` property. 
Question: Is there something better we can do than `exit(-1)` if the guard fails? Not sure if we havor throwing an exception, or preconditionFailure, or that's arbitrary

macos/FluentUI/Button.swift
Saadnajmi: Removed the need of this `preconditionFailure` by making sure the imagePosition is imageOnly in willSet. Agree this is something we definitely want to add later on

macos/FluentUI/Button.swift
Saadnajmi: Removed the call

macos/docs/Controls/Button.md
Saadnajmi: ... That's effectively undefined behavior. Is it worth mentioninig?

macos/docs/Controls/Button.md
Saadnajmi: That would be great :D, if I knew when each was actually used.
Context: I chose these 3 styles by looking at what we do on other platforms, and the documentation for each platform didn't specify when to use each one. I'm not sure that is standardized.

macos/FluentUITestApp/TestButtonViewController.swift
Saadnajmi: Github defaulting to 8 spaces for a tab really helps catch weirdness...
changed it so there shouldn't be weird indentation

macos/docs/Controls/Button.md
anandrajeswaran: Sounds like we can file an issue on PM/Design for missing documentation?

macos/docs/Controls/Button.md
anandrajeswaran: Per another comment in this review, I find it nice when frameworks callout what behavior would be undefined :). Particularly when it's something like tampering with a cell which is something anyone who has ever played with an NSButton has done

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,241 @@
...
+	override public var image: NSImage? {
+		willSet {
+			guard wantsLayer == true else {
+				exit(-1)
+			}
+			imagePosition = .imageOnly
+		}
+	}
+	
+	override public var title: String {
```
anandrajeswaran, line 68: This looks non-optional but I don't see how we ensure it has a non-nil value. Does the compiler not complain about this?

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,241 @@
...
+		willSet {
+			guard wantsLayer == true else {
+				exit(-1)
```
anandrajeswaran, line 71: Would prefer a preconditionFailure than an exit(-1) since this seems like exactly what that would be used for

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,241 @@
...
+			self.image = image
+		}
+	}
+	
+	override public var image: NSImage? {
+		willSet {
+			guard wantsLayer == true else {
+				exit(-1)
+			}
+			imagePosition = .imageOnly
```
anandrajeswaran, line 64: Alternatively, can we not do this and simply rely on the client to set the imagePosition they want? Eventually we do want to support this property so may be better to leave behavior undefined than explicitly add code we want to be able to delete soon

macos/FluentUI/Button.swift
Saadnajmi: This is non-optional. By default, a newly created NSButton has the title "Button" if none is set

anandrajeswaran: Oh yeah, totally forgot it's an override. Carry on

macos/FluentUI/Button.swift
Saadnajmi: Went for that approach

macos/docs/Controls/Button.md
Saadnajmi: Added a line under "Usage" in the documentation, since that's the most "technical" part of the documentation

macos/docs/Controls/Button.md
Saadnajmi: Perhaps. Considering I just "engineering best guessed" the design and then validated it with design in post, I feel once we have clients use the button, we'll get more design tokens / redlines / specifc requests, and we can update documention to match those new things. Regardless, I'll try and bring up this area is lacking

macos/FluentUI/Button.swift
```diff
@@ -0,0 +1,241 @@
...
+	
+	private var mouseEntered = false
+	private var mouseDown = false
+	
+	private var fillColor: NSColor {
+		return style == ButtonStyle.primaryFilled ? primaryColor : .clear
+	}
+	
+	private var outlineColor: NSColor {
+		let baseOutlineColor = style == ButtonStyle.primaryOutline ? primaryColor.withAlphaComponent(0.4) : .clear
```
markavitale, line 178: nit: should some of these alpha values be constants?

macos/docs/Controls/Button.md
```diff
@@ -0,0 +1,37 @@
+# Button
+
+## Overview
+Buttons are one of the core controls that make an app feel native to the platform it's on. It’s important to respect the platform's paradigms in order to help the user feel at home on macOS and keep the experience quality high.
```
markavitale, line 4: I don't have a great idea of how to improve this, but I feel like the overview should at least in some way touch on what the control actually is/does. Maybe we can take some inspiration from the Web button overview? Note some things are different here, especially their focus on navigation, because of web paradigms. 

https://developer.microsoft.com/en-us/fluentui#/controls/web/button

Or maybe the NSButton docs are a better starting point: https://developer.apple.com/documentation/appkit/nsbutton

Maybe something like: 

> Buttons are a standard control that users can interact with to take actions within your application.

would be sufficient. 

https://developer.microsoft.com/en-us/fluentui#/controls/web/button

macos/docs/Controls/Button.md
```diff
@@ -0,0 +1,37 @@
...
+let primaryOutlineButton = Button(title: "FluentUI Button", style: .primaryOutline),
+let borderlessButton = Button(title: "FluentUI Button", style: .borderless)
+```
+
+```Swift
+// You can display an image instead of a title
+let buttonWithImageAndStyle = Button(image: NSImage(named: NSImage.addTemplateName)!, style: .primaryFilled)
+```
+
+Overriding the underlying NSButtonCell will result in udefined behavior.
```
markavitale, line 29: I almost prefer something a little less formal like

> The FluentUI Button achieves its styling in part by customizing the underlying NSButtonCell. Any direct modifications to the cell will result in undefined behavior.

markavitale, line 29: nit: spelling (udefined -> undefined).

I use a VS Code spell checker extension for writing my documentation that helps a decent amount. It knows how to handle camel case and other common coding concepts that standard spell checkers can't. https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker

Saadnajmi: Except this is not true. The customization is achieved by ignoring the bezel, and modifying the layer. I do nothing to the underlying NSButtonCell. Theoretically, you can do whatever you want and things will probably be fine: I just haven't tested that thoroughly hence undefined behavior.

Also worth noting that for future iterations, I will definitely mess with the NSButtonCell so this guidance will change

macos/docs/Controls/Button.md
Saadnajmi: This snippet of text was what I got from the iOS and Android. I'm not particularly happy or beholden to it. I'll add some extra info on how we subclass NSButton but also add hover effects in comformance to macOS (since I literrally use the system effect for [hover](https://developer.apple.com/documentation/appkit/nscolor/systemeffect/rollover) to color that state)

macos/FluentUI/Button.swift
Saadnajmi: By having these colors all on the bottom of my swift file, I was effectively treating them as constants. Furthermore, this control will have to implement the color providing protocol @anandrajeswaran added, so all of this will change regardless. Given there's only 4, it should be easy enough for me to pull them out as real constants

macos/docs/Controls/Button.md
Saadnajmi: Side note: I don't like how this sentence lies alone. How would you feel about adding a subheader under usage with something like:
```markdown
### Not recommended
- Overriding NSButtonCell is currently undefined behavior
- Displaying both a title and an image will currently not render properly 
```

PR #74, opened 2020-05-23
Modify MSFAvatarView to support custom border image

lugonzal:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Add frameImage public property for MSFAvatarView

customBorderImage can be set so that the Avatar has a 3 pixel frame (~broder) based on this image. This image will be used to create a color, which eventually is set as the background color of the avatar frame view.

- Why not expose a color property instead of an image? Because the avatar view size is variable and not public. If a client creates a color with an image, the image is fixed in size and the color patter won't be updated when avatar size changes occur.
- Why not expose a frame/border view property instead of an image? Because the Avatar View uses really creates 2 avatar views (big and small) for contract/extract. One view can't be shared by 2 hierarchies, and can't be deep copied either.
- Why not expose a gradient and leave the convert-to-image details to Fluent? Because this way we provide the flexibility to have a predefined image or a runtime created image as a color.

### Verification

Create Demo for both. Played with using the avatar frame in navigation view controller (expanded, contracted).

image image

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/74)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarViewDemoController.swift
@@ -12,20 +12,25 @@ class AvatarViewDemoController: DemoController {
 
         createSection(withTitle: "Circle style for person", name: "Kat Larrson", image: UIImage(named: "avatar_kat_larsson")!, style: .circle)
         createSection(withTitle: "Square style for group", name: "NorthWind Traders", image: UIImage(named: "site")!, style: .square)
+        createSection(withTitle: "With image based frame", name: "Kat Larrson", image: UIImage(named: "avatar_kat_larsson")!, style: .circle, withFrame: true)
     }
 
-    private func createSection(withTitle title: String, name: String, image: UIImage, style: AvatarStyle) {
+    private func createSection(withTitle title: String, name: String, image: UIImage, style: AvatarStyle, withFrame: Bool = false) {
         addTitle(text: title)
         for size in AvatarSize.allCases.reversed() {
-            let imageAvatar = createAvatarView(size: size, name: name, image: image, style: style)
-            let initialsAvatar = createAvatarView(size: size, name: name, style: style)
+            let imageAvatar = createAvatarView(size: size, name: name, image: image, style: style, withFrame: withFrame)
+            let initialsAvatar = createAvatarView(size: size, name: name, style: style, withFrame: withFrame)
             addRow(text: size.description, items: [imageAvatar, initialsAvatar], textStyle: .footnote, textWidth: 100)
         }
         container.addArrangedSubview(UIView())
     }
 
-    private func createAvatarView(size: AvatarSize, name: String, image: UIImage? = nil, style: AvatarStyle) -> UIView {
+    private func createAvatarView(size: AvatarSize, name: String, image: UIImage? = nil, style: AvatarStyle, withFrame: Bool = false) -> UIView {
         let avatarView = AvatarView(avatarSize: size, withBorder: true, style: style)
+        if withFrame, let frameImage = colorfulImageForFrame() {
+            avatarView.frameImage = frameImage
+        }
+
         avatarView.setup(primaryText: name, secondaryText: "", image: image)
 
         let avatarContainer = UIView()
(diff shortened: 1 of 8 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -151,6 +151,14 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
         }
     }
 
+    /// If set, the accessory button title color will be based on this image, overiding the acessory style's title color.
+    /// The image will be adjusted to the button size, so the entire image based color is evenly distributed.
+     @objc open var imageForAccessoryButtonColor: UIImage? {
```
harrieshin, line 156: I am not sure this is the direction we want to go for the continuous support of this control. shouldn't this be something like the TableViewCell which we expose the custom accessory button and client can do whatever they want with it? I think it will make it easier for like zero query search scenarios of client want to show ">"  button etc .... this image for accessory button color seems too specific

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -276,6 +310,32 @@ open class AvatarView: UIView {
...
+        if frameView.frame == expectedFrame {
+            return
+        }
+
+        frameView.frame = expectedFrame
+        frameView.layer.cornerRadius = cornerRadius(for: expectedFrame.width)
+
+        var image = frameImage
+        if frameImage.size != expectedFrame.size {
+            UIGraphicsBeginImageContextWithOptions(expectedFrame.size, false /* opaque */, 0.0 /* scale */)
```
harrieshin, line 328: we should be using UIGraphicsImageRenderer

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -128,6 +140,24 @@ open class AvatarView: UIView {
         }
     }
 
+    private var frameView: UIView? {
```
harrieshin, line 143: customBorderView? customBackgroundView?

ios/FluentUI/People Picker/Avatar.swift
```diff
@@ -28,6 +29,9 @@ open class AvatarData: NSObject, Avatar {
     public var secondaryText: String
     public var image: UIImage?
 
+    /// An image that can be used as a frame (outer wide broder) for the avatar view
```
harrieshin, line 32: nit spellcheck: border?

ios/FluentUI/People Picker/Avatar.swift
```diff
@@ -28,6 +29,9 @@ open class AvatarData: NSObject, Avatar {
...
+    @objc public var frameImage: UIImage?
```
harrieshin, line 33: customBorderImage?

ios/FluentUI/People Picker/Avatar.swift
harrieshin: isn't there a caveat? the avatar view itself should also have has border to be true?

ios/FluentUI/People Picker/AvatarView.swift
harrieshin: could we just user borderView and combine the logic?

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -94,6 +94,7 @@ open class AvatarView: UIView {
         static let borderWidth: CGFloat = 2
```
harrieshin, line 94: should there be more documentation on borderWidth vs fameborderwidth difference?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarViewDemoController.swift
```diff
@@ -12,20 +12,25 @@ class AvatarViewDemoController: DemoController {
...
-            let initialsAvatar = createAvatarView(size: size, name: name, style: style)
+            let imageAvatar = createAvatarView(size: size, name: name, image: image, style: style, withFrame: withFrame)
+            let initialsAvatar = createAvatarView(size: size, name: name, style: style, withFrame: withFrame)
             addRow(text: size.description, items: [imageAvatar, initialsAvatar], textStyle: .footnote, textWidth: 100)
         }
         container.addArrangedSubview(UIView())
     }
 
-    private func createAvatarView(size: AvatarSize, name: String, image: UIImage? = nil, style: AvatarStyle) -> UIView {
+    private func createAvatarView(size: AvatarSize, name: String, image: UIImage? = nil, style: AvatarStyle, withFrame: Bool = false) -> UIView {
```
harrieshin, line 28: this param name doesn't seem clear. frame make it sounds like user is setting the frame of the avatar or not using auto layout.

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
lugonzal: it's a good point. It'll take me a a bit extra work and refactoring, so I'll revert this and try to make it a separate tentative change, since it's not that important for the Pride update.

ios/FluentUI/People Picker/AvatarView.swift
lugonzal: agree name is not great. I just wanted it to be clear that this is a thicker border, almost like a picture frame. But the work frame in code means something different... I think I'll go with customBorderView.

ios/FluentUI/People Picker/Avatar.swift
lugonzal: I was thinking it was a good assumption that if a customBorderImage is set, the client intends the view to have a border (regardless of the hasBorder value). Also, hasBorder refers to the usual border, not the customBorder, I can make that clearer in documentation

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewHeaderFooterViewDemoController.swift
```diff
@@ -52,7 +52,6 @@ class TableViewHeaderFooterViewDemoController: DemoController {
 }
 
 // MARK: - TableViewHeaderFooterViewDemoController: UITableViewDataSource
-
```
lugonzal, line 55: discard

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -276,6 +314,30 @@ open class AvatarView: UIView {
...
+
+        var image = customBorderImage
+        if customBorderImage.size != expectedFrame.size {
+            let renderer = UIGraphicsImageRenderer(size: expectedFrame.size)
+            image = renderer.image { context in
+                customBorderImage.draw(in: CGRect(origin: .zero, size: CGSize(width: expectedFrame.size.width, height: expectedFrame.size.height)))
+            }.withRenderingMode(.alwaysOriginal)
+        }
+
+        customBorderView.backgroundColor = UIColor.init(patternImage: image)
```
harrieshin, line 338: nit: UIColor(patternImage: image)

ios/FluentUI/People Picker/Avatar.swift
```diff
@@ -28,6 +29,9 @@ open class AvatarData: NSObject, Avatar {
     public var secondaryText: String
     public var image: UIImage?
 
+    /// An image that can be used as a frame (outer wide border) for the avatar view
+    @objc public var customBorderImage: UIImage?
```
markavitale, line 33: nit: can we move the documentation up to the declaration of this in the protocol?

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -120,6 +123,19 @@ open class AvatarView: UIView {
         }
     }
 
+    /// Image to be used as border around the avatar. It will be scaled to fit the avatar size.
+    /// If set, the hasBorder initializer value will be ignored, since it's assumed that the client inteds
```
markavitale, line 127: >inteds [](start = 100, length = 6)

nit spellcheck: intends

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -120,6 +123,19 @@ open class AvatarView: UIView {
...
+    /// to have a custom border.
+    @objc open var customBorderImage: UIImage? {
```
markavitale, line 129: Do we want to specify that this image will be treated as a pattern image?

PR #75, opened 2020-05-25
Switch to on-demand customizable colors

anandrajeswaran:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
- Introduce a `ColorProviding` protocol to allow FluentUI to ask for customizable colors on demand instead of requiring them all to be pushed before fluent ui is used. This will allow improved perf by adhering more to a pay-for-play model and will allow host apps to have simultaneous experiences showing different themes
- Deprecate `Colors.primary` and friends in favor of these APIs, but use those as a fallback in case a color provider is not set
- Modify all controls using these colors to only ask for a color when they have been added to a window, often relying on `viewDidAppear:` and `didMoveToWindow` override points
- `initializeAppearance` in `FluentUIframework` force loads from the bundle before it needs to and prevents simultaneous experiences with different themes by using the global UIAppearance for customization of iOS controls. Deprecate that variant and add a version that takes in a primary color and using the scoped versions UIAppearance APIs so appearance is only impacted if a named type is in the hierarchy. There are still some colors/images force loaded from the bundle in this path, but that can be figured out as a follow-up change.
- Add multi-window support to the demo apps and simplify the master/detail pattern since the built-in complexity from Xcode's default is not really needed. Now in the Colors demo, we can swap out the current UIWindow with a subclass that conforms to `ColorProviding` to have the colors dynamically update using the above pull model.
- **Breaking Change** - `navigationBarColor` in UINavigationItem extension renamed to `customNavigationBarColor`

### Verification
Click through all controls in all demos for the giant matrix of
- On iPad Pro (4th gen), iPhone 11 (notched), and iPhone 7 (not notched)
- iOS 12.4 and iOS 13.4
- Default, Green, and No theme (simulate client who hasn't adopted yet)
- Light and Dark Mode

| 
image
 | image
 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/75)


harrieshin:
can we add a note that MSFNavigationController api has changed for customNavigationBarColor? I think that is one of the "breaking change"?

harrieshin:
in your screen shot, there back button differs from the original. 1) there shouldn't be "back" text. (2) it should aways be gray unless it is "done" or checkmark button from barbuttonitems.swift ....

ios/FluentUI/People Picker/PeoplePicker.swift
```diff
@@ -328,8 +328,8 @@ open class PeoplePicker: BadgeField {
         guard let draggedBadge = draggedBadge,
             let personaBadgeDataSource = draggedBadge.dataSource as? PersonaBadgeViewDataSource,
             let destinationPeoplePicker = destinationBadgeField as? PeoplePicker else {
-            assertionFailure("Badge dataSource is not of type PersonaBadgeViewDataSource or destination badge field is not of type MSPeoplePicker")
-            return
+                assertionFailure("Badge dataSource is not of type PersonaBadgeViewDataSource or destination badge field is not of type MSPeoplePicker")
```
harrieshin, line 331: nit: while you are here do you mind changing it to MSFPeoplePicker?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -118,3 +130,5 @@ class TabBarItemView: UIView {
         imageView.highlightedImage = item.selectedImage(isInPortraitMode: isInPortraitMode)
     }
 }
+
+private let unselectedColor = Colors.foreground2c
```
harrieshin, line 134: is it out of scope to move the entire constant struct to the bottom?

ios/FluentUI/Table View/ActionsCell.swift
```diff
@@ -191,6 +194,10 @@ open class ActionsCell: UITableViewCell {
             )
         )
     }
+    open override func didMoveToWindow() {
```
harrieshin, line 197: super nit: new line

ios/FluentUI.Demo/FluentUI.Demo/Info.plist
```diff
@@ -31,8 +31,6 @@
 	<true/>
 	<key>UILaunchStoryboardName</key>
 	<string>LaunchScreen</string>
-	<key>UIMainStoryboardFile</key>
-	<string>Main</string>
```
harrieshin, line 35: should we also delete the storyboard file which declares a split view but we no longer implement split view delegate with your change?

ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemeWindows.swift
```diff
@@ -0,0 +1,74 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class DemoColorThemeDefaultWindow: UIWindow, ColorProviding {
```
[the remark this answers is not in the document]
anandrajeswaran: Yup, currently this happens by calling Colors.primary. As clients adopt and we fully deprecate that, the code in Colors.swift will call communicationBlue directly

PR #85, opened 2020-06-01
Add new property in DrawerController

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

DrawerController thinks it should present as UIPopoverPresentationController when the horizontal size is compact. But for larger phones like iPhone 11 Pro Max even in portrait orientation, horizontal size is regular. 
(1) make sure in iPhone the vertical drawers are presented in slider mode. 
(2) add a new property that clients can force the slider to be drawn in preferred content width in landscape in compact horizontal size.

### Verification

|Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-01 at 08 40 56  | Simulator Screen Shot - iPhone 11 Pro Max - 2020-06-01 at 08 28 28 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/85)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -318,7 +318,13 @@ class DrawerPresentationController: UIPresentationController {
         var contentFrame = bounds.inset(by: marginsForContentView())
 
         var contentSize = presentedViewController.preferredContentSize
-        let landscapeMode = contentFrame.width > contentFrame.height
+
+        let landscapeMode: Bool
+        if let windowSize = sourceViewController.view.window?.frame.size {
+            landscapeMode = windowSize.width > windowSize.height
+        } else {
+            landscapeMode = false
+        }
 
         if presentationDirection.isVertical {
             if contentSize.width == 0 ||
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -16,21 +16,21 @@ class DrawerPresentationController: UIPresentationController {
 
     let presentationDirection: DrawerPresentationDirection
 
-    private let shouldUseWindowFullWidth: Bool
+    private let shouldUseWindowFullWidthInLandscape: Bool
     private let sourceViewController: UIViewController
     private let sourceObject: Any?
     private let presentationOrigin: CGFloat?
     private let presentationOffset: CGFloat
     private let presentationBackground: DrawerPresentationBackground
 
-    init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?, source: UIViewController, sourceObject: Any?, presentationOrigin: CGFloat?, presentationDirection: DrawerPresentationDirection, presentationOffset: CGFloat, presentationBackground: DrawerPresentationBackground, adjustHeightForKeyboard: Bool, shouldUseWindowFullWidth: Bool) {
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -315,8 +318,12 @@ class DrawerPresentationController: UIPresentationController {
         var contentFrame = bounds.inset(by: marginsForContentView())
 
         var contentSize = presentedViewController.preferredContentSize
+        let landscapeMode = contentFrame.width > contentFrame.height
+
         if presentationDirection.isVertical {
-            if contentSize.width == 0 || traitCollection.horizontalSizeClass == .compact {
+            if contentSize.width == 0 ||
+                (traitCollection.userInterfaceIdiom == .phone && landscapeMode && shouldUseWindowFullWidthInLandscape) ||
```
lugonzal, line 325: what about iPhones in landscape mode that are considered compact, shouldn't they be presented as sliders regardless of shouldUseWindowFullWidthInLandscape?

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -315,8 +318,12 @@ class DrawerPresentationController: UIPresentationController {
...
+                (traitCollection.horizontalSizeClass == .compact && !landscapeMode) {
```
lugonzal, line 326: what about iPad in compact view, while the iPad is in landscape mode?

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -315,8 +318,12 @@ class DrawerPresentationController: UIPresentationController {
         var contentFrame = bounds.inset(by: marginsForContentView())
 
         var contentSize = presentedViewController.preferredContentSize
+        let landscapeMode = contentFrame.width > contentFrame.height
```
lugonzal, line 321: is this a good assumption? Can't the drawer have variable sizes, such that will make it longer or wider regardless of the device window state? Should we based this on the screen size instead?

harrieshin: ah yes. thanks for pointing out that mistake. I will use the sourceViewcontroller's view.windows size.

ios/FluentUI/Drawer/DrawerPresentationController.swift
harrieshin: It is shown as a slider in landscape. the request from List team was only on landscape it should show the "preferred" width. here is a screenshot of iPhoneSE
Simulator Screen Shot - iPhone SE (1st generation) - 2020-06-02 at 19 45 26

ios/FluentUI/Drawer/DrawerPresentationController.swift
harrieshin: even in iPad compact mode in device landscape orientation, it would look more like iPhone portrait. wouldn't it be?

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-06-02 at 20 12 24

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-06-02 at 20 12 28

Simulator Screen Shot - iPad Pro (12 9-inch) (3rd generation) - 2020-06-02 at 20 12 34

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -315,8 +318,18 @@ class DrawerPresentationController: UIPresentationController {
         var contentFrame = bounds.inset(by: marginsForContentView())
 
         var contentSize = presentedViewController.preferredContentSize
+
+        let landscapeMode: Bool
+        if let windowSize = sourceViewController.view.window?.frame.size {
```
lugonzal, line 323: will this work for iPad compact? Not sure if this window is the whole screen or just the narrow split view

harrieshin: The window size is the narrow splitview.

PR #87, opened 2020-06-02
change the tabbar background effect to be .systemChromeMaterial

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

iOS 13 and above, use systemChromeMaterial blur effect for TabBarView
Update the demo app so that example of NavigationController would show tab bar as well.

### Verification

tested iPad/iPhone : light and dark mode
[demo_tabbar.mov.zip](https://github.com/microsoft/fluentui-apple/files/4719997/demo_tabbar.mov.zip)

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/87)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -214,6 +214,16 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
     }
 
+    private let tabBarView: TabBarView = {
+        let tabBarView = TabBarView()
+        tabBarView.items = [
+            TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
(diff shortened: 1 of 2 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -247,7 +270,8 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
 
     func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
         let cell = tableView.dequeueReusableCell(withIdentifier: TableViewCell.identifier, for: indexPath) as! TableViewCell
-        cell.setup(title: "Cell #\(1 + indexPath.row)", accessoryType: .disclosureIndicator)
+        let imageView = UIImageView(image: UIImage(named: "excelIcon"))
```
lugonzal, line 273: nit: Does this icon make sense here? At least in the tableviewcell demo, the cell title is a spreadsheet name

harrieshin: it might not? the reason why i wanted some sort of color cell image is to show translucency of the tabbar background in dark mode. i do wonder if we should change this particular large title navigation controller demo to the next level and make it look more like office as we add more controls

PR #89, opened 2020-06-04
Use 24x24 TabBar icon size when label is shown

jonathanwangg:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes
`TabBarItemView.swift`

TabBar icon size will use 24x24 (portrait image size) when a label is present, otherwise it will use 28x28 (landscape image size). I reused the `landscapeImageSize` variable because it was the correct size that was required, but an alternative solution would be to create a new variable specifically for when a label is added (e.g. `portraitImageSizeWithLabel`)

### Verification
The change was manually verified in the Xcode simulator using an iPhone 11. The change in icon size was verified visually in both portrait and landscape orientations. 

| image | after_portrait |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/89)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
@@ -33,11 +33,21 @@ class TabBarViewDemoController: DemoController {
 
         let updatedTabBarView = TabBarView(showsItemTitles: showsItemTitles)
         updatedTabBarView.delegate = self
-        updatedTabBarView.items = [
-            TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
-            TabBarItem(title: "New", image: UIImage(named: "New_28")!, selectedImage: UIImage(named: "New_Selected_28")!, landscapeImage: UIImage(named: "New_24")!, landscapeSelectedImage: UIImage(named: "New_Selected_24")!),
-            TabBarItem(title: "Open", image: UIImage(named: "Open_28")!, selectedImage: UIImage(named: "Open_Selected_28")!, landscapeImage: UIImage(named: "Open_24")!, landscapeSelectedImage: UIImage(named: "Open_Selected_24")!)
-        ]
+        
+        if showsItemTitles {
+            updatedTabBarView.items = [
+                TabBarItem(title: "Home", image: UIImage(named: "Home_24")!, selectedImage: UIImage(named: "Home_Selected_24")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
+                TabBarItem(title: "New", image: UIImage(named: "New_24")!, selectedImage: UIImage(named: "New_Selected_24")!, landscapeImage: UIImage(named: "New_24")!, landscapeSelectedImage: UIImage(named: "New_Selected_24")!),
+                TabBarItem(title: "Open", image: UIImage(named: "Open_24")!, selectedImage: UIImage(named: "Open_Selected_24")!, landscapeImage: UIImage(named: "Open_24")!, landscapeSelectedImage: UIImage(named: "Open_Selected_24")!)
+            ]
+        } else {
+            updatedTabBarView.items = [
+                TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
+                TabBarItem(title: "New", image: UIImage(named: "New_28")!, selectedImage: UIImage(named: "New_Selected_28")!, landscapeImage: UIImage(named: "New_24")!, landscapeSelectedImage: UIImage(named: "New_Selected_24")!),
+                TabBarItem(title: "Open", image: UIImage(named: "Open_28")!, selectedImage: UIImage(named: "Open_Selected_28")!, landscapeImage: UIImage(named: "Open_24")!, landscapeSelectedImage: UIImage(named: "Open_Selected_24")!)
+            ]
+        }
+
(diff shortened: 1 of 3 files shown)


harrieshin:
should we change the function parameters of Tabbarview now it is different?

lugonzal:
@harrieshin Jonathan's code is not actually using the landscape image passed in init for MSFTabBarItem (is that what you meant?), he's just changing the size of the view that will host the portrait image. I wonder, should he be using the landscapeImage, if using the landscapeImage size? Or is it ok to use the portrait image in a smaller frame. If the latter is ok, what's the need for having a landscape image at all?

harrieshin:
@lugonzal yeah I meant for tabbaritem rather than tabbar. well.... I am hesitant to say we should reused the image of 28x28 of portrait to squeeze it in 21x21 frame because designers technically produced different images(pdf) and could have different line weight and spacing depending on the original image size. I think portrait and landscape naming might be fine? But, we need better documentation on those properties. And for sure we should update the demo app how it is calling the control.

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -110,8 +110,8 @@ class TabBarItemView: UIView {
         if isInPortraitMode {
             container.axis = .vertical
             container.spacing = spacingVertical
-            imageHeightConstraint?.constant = portraitImageSize
-            imageWidthConstraint?.constant = portraitImageSize
+            imageHeightConstraint?.constant = titleLabel.isHidden ? portraitImageSize : landscapeImageSize
+            imageWidthConstraint?.constant = titleLabel.isHidden ? portraitImageSize : landscapeImageSize
```
lugonzal, line 114: if we're using the landscapeImageSize, we should also use the item's landscapeImage. See below, when we're doing:

imageView.image = ...
imageView.highlightedImage = ...

This will involve perhaps modifying the unselected/selected Image methods to take another param, and also modifying the comments on the initializer of MSFTabBarItem, indicating that the landscapeImage will be used for portrait if the tab bar item shows a lablel

harrieshin:
The title of this PR shouldn't contain ADO bug number since it isn't related to this public github repo

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -110,8 +110,8 @@ class TabBarItemView: UIView {
         if isInPortraitMode {
             container.axis = .vertical
             container.spacing = spacingVertical
-            imageHeightConstraint?.constant = portraitImageSize
-            imageWidthConstraint?.constant = portraitImageSize
+            imageHeightConstraint?.constant = titleLabel.isHidden ? portraitImageSize : landscapeImageSize
```
anandrajeswaran, line 113: Since we're now using "landscapeImageSize" in portrait, the name "landscapeImageSize" is now misleading. Maybe we can rename it to something else to reflect how it is different rather than when it is used?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -33,11 +33,21 @@ class TabBarViewDemoController: DemoController {
...
         updatedTabBarView.delegate = self
-        updatedTabBarView.items = [
-            TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
-            TabBarItem(title: "New", image: UIImage(named: "New_28")!, selectedImage: UIImage(named: "New_Selected_28")!, landscapeImage: UIImage(named: "New_24")!, landscapeSelectedImage: UIImage(named: "New_Selected_24")!),
-            TabBarItem(title: "Open", image: UIImage(named: "Open_28")!, selectedImage: UIImage(named: "Open_Selected_28")!, landscapeImage: UIImage(named: "Open_24")!, landscapeSelectedImage: UIImage(named: "Open_Selected_24")!)
-        ]
+        
+        if showsItemTitles {
+            updatedTabBarView.items = [
+                TabBarItem(title: "Home", image: UIImage(named: "Home_24")!, selectedImage: UIImage(named: "Home_Selected_24")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
```
harrieshin, line 39: do we need to pass in landscape images in this case?

jonathanwangg: I was thinking we still required `landscapeImage` when an image is unselected in portrait mode since we return `landscapeImage` when label is shown.

harrieshin: both selected and landscape images are optional param for tabbaritem initialization, aren't they? the TabBarItem and tabbaritemview should handle them accordingly.

ios/FluentUI/Tab Bar/TabBarItem.swift
```diff
@@ -21,8 +21,8 @@ open class TabBarItem: NSObject {
     /// - Parameter title: Used for tabbar item view's label and for its accessibilityLabel.
     /// - Parameter image: Used for tabbar item view's imageView and for its accessibility largeContentImage unless `largeContentImage` is specified.
     /// - Parameter selectedImage: Used for imageView when tabbar item view is selected.  If it is nil, it will use `image`.
-    /// - Parameter landscapeImage: Used for imageView when tabbar item view in landscape. If it is nil, it will use `image`.
-    /// - Parameter landscapeSelectedImage: Used for imageView when tabbar item view is selected in landscape. If it is nil, it will use `selectedImage`.
+    /// - Parameter landscapeImage: Used for imageView when tabbar item view in landscape. If it is nil, it will use `image`. Will be used in portrait mode if the tab bar item shows a label.
+    /// - Parameter landscapeSelectedImage: Used for imageView when tabbar item view is selected in landscape. If it is nil, it will use `selectedImage`. Will be used in portrait mode if the tab bar item shows a label.
```
harrieshin, line 25: nit: let make sure this documentation is using full sentence.

PR #101, opened 2020-06-23
ContactView

jonathanwangg:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

`ios/FluentUI/People Picker/ContactView.swift`
This is a new View that will display a "Contact" which consists of an `AvatarView` and either 1 or 2 `UILabel`s. There are two methods to create a `ContactView`. Both methods allow the consumer to provide an `AvatarView`. The difference between the two methods is the text being passed in. The first method allows a first name and a last name which will create two `UILabel`s. The second method allows for a more generic single string which I expect to be either a phone number, email address, or a first name. Consideration for Right-To-Left locales was made and deemed unneccessary due to the fact that we are center-aligning the text when provided a first and last name.

`ios/FluentUI/Core/Colors.swift`
Added a struct to specify the colors being used for the labels within a `ContactView".

`ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactViewDemoController.swift`
This is a demo within the demo app that contains 5 possible `ContactView`s. It includes:
1) `ContactView` with image
2) `ContactView` without an image and using initials instead
3) `ContactView` without an image and using only a first name as an identifier
3) `ContactView` without an image and using an email address as an identifier
4) `ContactView` without an image and using a phone number as an identifier

### Things to address
1) Use a pawn / placeholder image instead of a "#" for emails or phone numbers.
2) Is it fine to use a grey colour background since the background of a `ContactView` is white in order to see the bounds?
3) `ContactView` is currently accepting an `AvatarView` as a parameter for it's constructor, but after some discussion with Mathieu, we would prefer to have the consumer pass in an image and have `ContactView` call the constructor for `AvatarView`.
4) Pressed state implementation has not been started.
5) The height of the labels are slightly different between a `ContactView` that uses 1 label and a `ContactView` that uses 2 labels. This is due to the intrinsic size of the text because of the font being used (i.e. subhead and footnote vs subhead). When a long email or phone number is used, the singular label will extend to a second line and both lines will have height of two subhead rather than a single subhead and single footnote. 
6) The `ContactView` for a phone number when using extra extra extra large setting for accessibility text looks awkward as the first line only shows "+1" and then a large white space.

### Verification

Testing was done via manual verification of height and other properties by using "Debug View Hierarchy" with the iOS simulator.

| Demo app with accessibility setting for text set to extra small | image |
| Demo app with accessibility setting for text set to large (default) | image |
| Demo app with accessibility setting for text set to extra extra extra large | image |
| Demo app with accessibility setting for text set to large (default) and dark mode | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/101)


ios/FluentUI/Core/Colors.swift
```diff
@@ -465,6 +465,11 @@ public final class Colors: NSObject {
         public static var background: UIColor = background1
     }
 
+    public struct Contact {
+        public static var firstName: UIColor = foreground1
```
lugonzal, line 469: >foreground1 [](start = 47, length = 11)

your fork might be out of date. These colors were renamed

jonathanwangg: Got it, I'll pull from upstream. Thanks for letting me know!

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+open class ContactView: UIView {
+    private struct Constants {
+        static let contactWidth: CGFloat = 70.0
+        static let avatarViewWidth: CGFloat = 70.0
+        static let avatarViewHeight: CGFloat = 70.0
+        static let firstNameLabelMinimumHeight: CGFloat = 12.0
+        static let spacingBetweenAvatarAndFirstLabel: CGFloat = 13.0
+    }
+
+    private var avatarView: AvatarView
```
lugonzal, line 17: >var [](start = 12, length = 3)

can this be let? we always assign it on init

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Initializes the contact view with an avatar view, first name, and last name
+    ///
+    /// - Parameters:
+    ///   - avatarView: The AvatarView that will be displayed within the ContactView
+    ///   - firstName: String that will be the text of the top label
+    ///   - lastName: String that will be the text of the bottom label
+    public init(avatarView: AvatarView, firstName: String, lastName: String) {
+        self.avatarView = avatarView
+        self.firstNameLabel = UILabel()
```
lugonzal, line 30: >self [](start = 8, length = 4)

no need for "self." unless a name conflict is here

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+    private var lastNameLabel: UILabel?
+    private var identifierLabel: UILabel?
+
+    /// Initializes the contact view with an avatar view, first name, and last name
+    ///
+    /// - Parameters:
+    ///   - avatarView: The AvatarView that will be displayed within the ContactView
+    ///   - firstName: String that will be the text of the top label
+    ///   - lastName: String that will be the text of the bottom label
+    public init(avatarView: AvatarView, firstName: String, lastName: String) {
```
MathieuKavalec, line 28: As we discussed, the AvatarView should be created internally instead of passed as a parameter to this ContactView control.

jonathanwangg: Yup, I noted this in the "Things to address" section point 3. I'll be getting to it after making the suggested changes.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+open class ContactView: UIView {
+    private struct Constants {
+        static let contactWidth: CGFloat = 70.0
```
MathieuKavalec, line 10: nit: Shouldn't this just be ViewWidth
The contact width and view width are the same.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Yes, that makes more sense

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Ah I see! Thanks for pointing it out

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+    ///   - avatarView: The AvatarView that will be displayed within the ContactView
+    ///   - firstName: String that will be the text of the top label
+    ///   - lastName: String that will be the text of the bottom label
+    public init(avatarView: AvatarView, firstName: String, lastName: String) {
+        self.avatarView = avatarView
+        self.firstNameLabel = UILabel()
+        self.lastNameLabel = UILabel()
+        super.init(frame: .zero)
+        self.backgroundColor = Colors.surfacePrimary
+        self.translatesAutoresizingMaskIntoConstraints = false
```
lugonzal, line 34: >self.translatesAutoresizingMaskIntoConstraints = false [](start = 8, length = 54)

why is this necessary? This is usually set before a view is added to a superview, so perhaps clients should take care of this

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    /// Initializes the contact view with an avatar view and an identifier
+    ///
+    /// - Parameters:
+    ///   - avatarView: The AvatarView that will be displayed within the ContactView
+    ///   - identifier: String that will be used to identify the contact (e.g. email, phone number, first name)
+    public init(avatarView: AvatarView, identifier: String) {
+        self.avatarView = avatarView
+        super.init(frame: .zero)
+        self.backgroundColor = Colors.surfacePrimary
```
lugonzal, line 48: can we make a private initializer that both public initializers call into? That way we can share some of this code, and future updates will be less prone to bugs

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+        setupLastNameLabel(using: lastName)
+        setupLayout()
+    }
+
+    /// Initializes the contact view with an avatar view and an identifier
+    ///
+    /// - Parameters:
+    ///   - avatarView: The AvatarView that will be displayed within the ContactView
+    ///   - identifier: String that will be used to identify the contact (e.g. email, phone number, first name)
+    public init(avatarView: AvatarView, identifier: String) {
```
lugonzal, line 45: > public init [](start = 3, length = 12)

add @objc so that public members are available to Objective C clients

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+        setupLayout()
+    }
+
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    private func setupLayout() {
+        var constraints = [NSLayoutConstraint]()
+        constraints.append(self.widthAnchor.constraint(equalToConstant: Constants.contactWidth))
```
lugonzal, line 60: >self [](start = 27, length = 4)

make sure you remove all the usages of "self."

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+            constraints.append(contentsOf: lastNameLayoutConstraints())
+        } else {
+            constraints.append(contentsOf: identifierLayoutConstraints())
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    private func avatarLayoutConstraints() -> [NSLayoutConstraint] {
+        addSubview(avatarView)
```
lugonzal, line 74: >addSubview [](start = 8, length = 10)

can we move this out of this method? The method name doesn't indicate that an action like this will take place

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+        constraints.append(avatarView.centerXAnchor.constraint(equalTo: self.centerXAnchor))
+        constraints.append(avatarView.topAnchor.constraint(equalTo: self.topAnchor))
+        constraints.append(avatarView.leadingAnchor.constraint(equalTo: self.leadingAnchor))
+        constraints.append(avatarView.trailingAnchor.constraint(equalTo: self.trailingAnchor))
+        return constraints
+    }
+
+    private func firstNameLayoutConstraints() -> [NSLayoutConstraint] {
+        var constraints = [NSLayoutConstraint]()
+        if let firstNameLabel = firstNameLabel {
```
lugonzal, line 89: >if let firstNameLabel = firstNameLabel { [](start = 8, length = 40)

guard let firstNameLabel = firstNameLabel else {
       return nil
}

Then we can make contraints a constant (let)

ios/FluentUI/People Picker/ContactView.swift
lugonzal: agree

---
In reply to: [444526820](https://github.com/microsoft/fluentui-apple/pull/101#discussion_r444526820) [](ancestors = 444526820)

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactViewDemoController.swift
```diff
@@ -0,0 +1,47 @@
...
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class ContactViewDemoController: DemoController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view.backgroundColor = UIColor.lightGray
```
lugonzal, line 12: >lightGray [](start = 39, length = 9)

I don't know if I like this, but I see how it shows that the card is a square. I wonder if we should set the background that one would use in a real life scenario (same color of the card), for better demonstration purpose

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+    private func setupIdentifierLabel(using identifier: String) {
+        identifierLabel = UILabel(frame: .zero)
+
+        if let identifierLabel = identifierLabel {
+            identifierLabel.adjustsFontForContentSizeCategory = true
+            identifierLabel.font = Fonts.subhead
+            identifierLabel.numberOfLines = 2
+            identifierLabel.text = identifier
+            identifierLabel.textAlignment = .natural
+            identifierLabel.textColor = Colors.Contact.firstName
```
lugonzal, line 169: this looks like same as FirstName. Let's share the code, and perhaps just use one label (titleLable) for both

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,172 @@
...
+        static let contactWidth: CGFloat = 70.0
+        static let avatarViewWidth: CGFloat = 70.0
+        static let avatarViewHeight: CGFloat = 70.0
+        static let firstNameLabelMinimumHeight: CGFloat = 12.0
+        static let spacingBetweenAvatarAndFirstLabel: CGFloat = 13.0
+    }
+
+    private var avatarView: AvatarView
+    private var firstNameLabel: UILabel?
+    private var lastNameLabel: UILabel?
```
lugonzal, line 19: >lastNameLabel [](start = 16, length = 13)

we can be a bit more flexible and name these more generically, like titleLable, subtitleLabel

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Ah, gotcha. Thanks for letting me know. I've removed it.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Will do!

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Agreed

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactViewDemoController.swift
jonathanwangg: Are you suggesting to keep the background the same colour as the other demo controllers? This is definitely possible, but I feel like it would be harder for the consumer to understand where the `ContactView` begins and where it ends since the `ContactView` colour is the same as the background from what I understand.

lugonzal: i'm on the fence. I think it looks unclean in the demo and it would be better to represent it as the real users will representing, which is, with a matching color. But up to you :)

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: I understand what you mean by using `guard` in order to return and exit out of the function early. I'm a bit puzzled regarding making `constraints` a `let`. Since if it's a `let`, we won't be able to append the constraints to array one by one. Are you suggesting to initialize the array with all of the constraints like `let constraints = [constraint_1, constraint_2, ...]`?

lugonzal: yes, that's what I'm suggesting

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,184 @@
...
+open class ContactView: UIView {
+    private struct Constants {
+        static let contactWidth: CGFloat = 70.0
+        static let avatarViewWidth: CGFloat = 70.0
+        static let avatarViewHeight: CGFloat = 70.0
+        static let titleLabelMinimumHeight: CGFloat = 12.0
+        static let spacingBetweenAvatarAndFirstLabel: CGFloat = 13.0
+    }
+
+    // TODO: Set it as the default image (pawn)
```
MathieuKavalec, line 17: AvatarView should take care of showing the pawn, not this control.
I'd remove that TODO comment as you would have no extra work to do in this class to support that.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,184 @@
...
+    @objc public var avatarImage: UIImage? {
+        didSet {
+            if let avatarImage = avatarImage {
```
MathieuKavalec, line 20: If the image being set is nil, we should honor that and set the avatar view's image to nil too.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,184 @@
...
+        setupSubtitleLabel(using: subtitle)
+        setupLayout()
+    }
+
+    /// Initializes the contact view by creating an avatar view with an identifier
+    ///
+    /// - Parameters:
+    ///   - identifier: String that will be used to identify the contact (e.g. email, phone number, first name)
+    @objc public init(identifier: String) {
+        avatarView = AvatarView(avatarSize: .extraExtraLarge, withBorder: false, style: .circle)
```
MathieuKavalec, line 51: You don't need titleLabel = UILabel() here?

jonathanwangg: Right, I made the correction when making changes for https://github.com/microsoft/fluentui-apple/pull/101#discussion_r444592675

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Got it. Setting the image to the pawn within `AvatarView` makes a lot more sense.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Good catch. This is definitely needed!

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Great suggestion! I took a look into `convenience` and `designated` initializers for this and will be pushing it soon.

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec, line 20: If the UIImage is nil, then you won't get into this condition. 
We would want to set the avatar view's image to nil if this property is set to nil.

jonathanwangg: Commit c2dad3000a5dce7e5bd6ba82d38c12cd8bc87ccd addresses this issue.

harrieshin:
please update your screenshots of your PR as well.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+
+        if let titleLabel = titleLabel, let subtitleLabel = subtitleLabel {
+            constraints.append(contentsOf: titleLabelLayoutConstraints())
+            constraints.append(contentsOf: subtitleLabelLayoutConstraints())
+            addSubview(titleLabel)
+            addSubview(subtitleLabel)
+        } else if let titleLabel = titleLabel {
+            constraints.append(contentsOf: identifierLayoutConstraints())
+            addSubview(titleLabel)
+        }
```
lugonzal, line 87: I think this might be simpler to write as:

if let titleLabel = titleLabel {
      if let subtitleLabel = subtitleLabel {
      } else {
      }
}

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+        static let titleLabelMinimumHeight: CGFloat = 12.0
+        static let spacingBetweenAvatarAndFirstLabel: CGFloat = 13.0
+    }
+
+    @objc public var avatarImage: UIImage? {
+        didSet {
+            setupAvatarView(with: titleLabel?.text, and: subtitleLabel?.text, or: nil)
+        }
+    }
+    private let avatarView: AvatarView
```
lugonzal, line 22: nit: new line

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+        static let contactWidth: CGFloat = 70.0
+        static let avatarViewWidth: CGFloat = 70.0
+        static let avatarViewHeight: CGFloat = 70.0
+        static let titleLabelMinimumHeight: CGFloat = 12.0
+        static let spacingBetweenAvatarAndFirstLabel: CGFloat = 13.0
+    }
+
+    @objc public var avatarImage: UIImage? {
+        didSet {
+            setupAvatarView(with: titleLabel?.text, and: subtitleLabel?.text, or: nil)
```
lugonzal, line 19: >nil [](start = 82, length = 3)

what if the view uses an identifier and a client changes the image?

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+
+        backgroundColor = Colors.surfacePrimary
+        setupLayout()
+    }
+
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    private func setupAvatarView(with firstName: String?, and lastName: String?, or identifier: String?) {
```
lugonzal, line 65: >setupAvatarView [](start = 17, length = 15)

I think we should just have 2 methods:
setupAvatarView(with firstName: String, and lastName: String)
setupAvatarView(with identifier: String)

there is no code shared between the cases where we have both title/subtitle and the case we have an identifier. Also, that way we don't need the params to be optional

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+            addSubview(titleLabel)
+        }
+
+        addSubview(avatarView)
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    private func avatarLayoutConstraints() -> [NSLayoutConstraint] {
+        avatarView.translatesAutoresizingMaskIntoConstraints = false
```
lugonzal, line 95: >avatarView.translatesAutoresizingMaskIntoConstraints = false [](start = 8, length = 60)

this belongs next to the line that adds it as a subview. This is only necessary before the constraints are activated. Same in the other constraint method

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
...
+            titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            titleLabel.bottomAnchor.constraint(equalTo: bottomAnchor)
+        ]
+    }
+
+    private func setupTitleLabel(using firstName: String) {
+        titleLabel = UILabel(frame: .zero)
+
+        if let titleLabel = titleLabel {
```
lugonzal, line 164: > if let titleLabel = titleLabel { [](start = 7, length = 33)

it's kind of weird to do this when we just initialized it in the line above. Instead you could do:
let label = UILabel(...)
// configure label
titleLabel = label

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,189 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+open class ContactView: UIView {
```
lugonzal, line 8: >ContactView [](start = 11, length = 11)

add this line above:
@objc(MSFContactView)

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: It looks much easier to read this way!

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Why does the line belong next to the line that adds it as a subview? From my code, `avatarLayoutConstraints()` is always called before the constraints are activated. Is it because we only want to set the `translatesAutoresizingMaskIntoConstraints` property a single time and not each time `avatarLayoutConstraints()` is called?

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: I see what you mean. I initialized it and then immediately checked whether or not it was nil which is redundant.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Agreed, the parameters look much cleaner done that way.

ios/FluentUI/People Picker/ContactView.swift
lugonzal: that is one of the reasons. But also because this method's name indicates it's a getter and doesn't indicate it will perform an action like this. Also, translatesAutoresizingMaskIntoConstraints is important before activating constraints so it makes sense that code is in the same place

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Addressed

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+
+        backgroundColor = Colors.surfacePrimary
+        setupLayout()
+    }
+
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    private func setupAvatarView(with firstName: String, and lastName: String) {
```
MathieuKavalec, line 75: Did we want lastName to be optional?

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+        avatarView.setup(primaryText: identifier, secondaryText: nil, image: avatarImage)
+    }
+
+    private func setupLayout() {
+        var constraints = [NSLayoutConstraint]()
+        constraints.append(widthAnchor.constraint(equalToConstant: Constants.contactWidth))
+        constraints.append(contentsOf: avatarLayoutConstraints())
+
+        avatarView.translatesAutoresizingMaskIntoConstraints = false
+        labelContainer.addSubview(avatarView)
```
MathieuKavalec, line 90: Not really a "labelContainer" if it contains the avatar view. Is this intentional? If so, we probably need a better name. Also, why is this container view needed if it contains all the views that our root view contains?

jonathanwangg: Oh this was not intentional. That was a mistake and should have only contained `titleLabel` and `subtitleLabel`. I'll fix that in a new commit.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+            avatarView.leadingAnchor.constraint(equalTo: leadingAnchor),
+            avatarView.trailingAnchor.constraint(equalTo: trailingAnchor)
+        ]
+    }
+
+    private func labelContainerLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            labelContainer.widthAnchor.constraint(equalToConstant: Constants.contactWidth),
+            labelContainer.centerXAnchor.constraint(equalTo: centerXAnchor),
+            labelContainer.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndFirstLabel),
```
MathieuKavalec, line 125: avatarView is a subview of labelContainer, this constraint doesn't really make much sense.
Is it possible that avatarView was supposed to be a subview of self instead?

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+            labelContainer.leadingAnchor.constraint(equalTo: leadingAnchor),
+            labelContainer.trailingAnchor.constraint(equalTo: trailingAnchor),
+            labelContainer.bottomAnchor.constraint(equalTo: bottomAnchor),
+            labelContainer.heightAnchor.constraint(greaterThanOrEqualToConstant: 2 * Constants.labelMinimumHeight),
```
MathieuKavalec, line 129: Use .0 for literal floating-point numbers.
In this case, 2.0 * Contants...

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Yes, you're correct. `labelContainer` should have only contained the `titleLabel` and `subtitleLabel`. `avatarView` is supposed to be a subview of `self`. I have that fixed now in a commit.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+            labelContainer.heightAnchor.constraint(lessThanOrEqualToConstant: Constants.titleLabelMaximumHeight + Constants.subtitleMaximumHeight)
```
MathieuKavalec, line 130: Why do we need a max height? 
Don't we risk vertically truncating text when the accessibility font size is set to a high value?

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+
+    private func titleLabelLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            titleLabel.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndFirstLabel),
+            titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            titleLabel.centerXAnchor.constraint(equalTo: centerXAnchor),
+            titleLabel.widthAnchor.constraint(equalTo: widthAnchor),
+            titleLabel.heightAnchor.constraint(greaterThanOrEqualToConstant: Constants.labelMinimumHeight),
+            titleLabel.heightAnchor.constraint(lessThanOrEqualToConstant: Constants.titleLabelMaximumHeight)
```
MathieuKavalec, line 142: Same comment as above

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+        ]
+    }
+
+    private func titleLabelLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            titleLabel.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndFirstLabel),
+            titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            titleLabel.centerXAnchor.constraint(equalTo: centerXAnchor),
+            titleLabel.widthAnchor.constraint(equalTo: widthAnchor),
```
MathieuKavalec, line 140: For correctness, the title label's leading, trailing, centerX and width anchors should be anchored to the labelContainer's anchors, not the self view's anchors.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+            labelContainer.trailingAnchor.constraint(equalTo: trailingAnchor),
+            labelContainer.bottomAnchor.constraint(equalTo: bottomAnchor),
+            labelContainer.heightAnchor.constraint(greaterThanOrEqualToConstant: 2 * Constants.labelMinimumHeight),
+            labelContainer.heightAnchor.constraint(lessThanOrEqualToConstant: Constants.titleLabelMaximumHeight + Constants.subtitleMaximumHeight)
+        ]
+    }
+
+    private func titleLabelLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            titleLabel.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndFirstLabel),
```
MathieuKavalec, line 136: equalTo: labelContainer.topAnchor instead.

A subview should always position itself relative to its superview or other views inside its superview.
Currently, this is a violation of view hierarchy encapsulation.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: I originally set a max height because of what I saw in the designs, but your comment is still valid because there are text sizes even larger than what is shown in the designs.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+
+        return [
+            subtitleLabel.topAnchor.constraint(equalTo: titleLabel.bottomAnchor),
+            subtitleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            subtitleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            subtitleLabel.centerXAnchor.constraint(equalTo: centerXAnchor),
+            subtitleLabel.widthAnchor.constraint(equalTo: widthAnchor),
+            subtitleLabel.bottomAnchor.constraint(equalTo: bottomAnchor),
+            subtitleLabel.heightAnchor.constraint(greaterThanOrEqualToConstant: Constants.labelMinimumHeight),
+            subtitleLabel.heightAnchor.constraint(lessThanOrEqualToConstant: Constants.subtitleMaximumHeight)
```
MathieuKavalec, line 159: Same comments as above.
- Why do we need a max height?
- Layout constraints must be relative to the superview (labelContainer).

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    private func identifierLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            titleLabel.widthAnchor.constraint(equalTo: widthAnchor),
+            titleLabel.centerXAnchor.constraint(equalTo: centerXAnchor),
+            titleLabel.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndFirstLabel),
+            titleLabel.leadingAnchor.constraint(equalTo: leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: trailingAnchor),
+            titleLabel.bottomAnchor.constraint(equalTo: bottomAnchor)
```
MathieuKavalec, line 170: Same comment as above: layout constraints must be relative to the superview (labelContainer) or other views inside the labelContainer.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,196 @@
...
+            }
+        }
+    }
+
+    private let avatarView: AvatarView
+    private var titleLabel: UILabel
+    private var subtitleLabel: UILabel?
+    private var labelContainer: UIView
+
+    /// Initializes the contact view by creating an avatar view with a first name and last name
```
lugonzal, line 35: >first name and last name [](start = 71, length = 24)

I would word this as something like "primary and secondary text"

jonathanwangg: Yes, of course. I must have missed this since the refactoring tool doesn't change the comments.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,196 @@
...
+            titleLabel.widthAnchor.constraint(equalTo: labelContainer.widthAnchor),
+            titleLabel.centerXAnchor.constraint(equalTo: labelContainer.centerXAnchor),
+            titleLabel.topAnchor.constraint(equalTo: avatarView.bottomAnchor, constant: Constants.spacingBetweenAvatarAndLabelContainer),
+            titleLabel.leadingAnchor.constraint(equalTo: labelContainer.leadingAnchor),
+            titleLabel.trailingAnchor.constraint(equalTo: labelContainer.trailingAnchor),
+            titleLabel.bottomAnchor.constraint(equalTo: labelContainer.bottomAnchor)
+        ]
+    }
+
+    private func setupTitleLabel(using firstName: String) {
```
lugonzal, line 171: >firstName [](start = 39, length = 9)

title

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,196 @@
...
+        let label = UILabel(frame: .zero)
+        label.adjustsFontForContentSizeCategory = true
+        label.font = Fonts.subhead
+        label.text = firstName
+        label.textAlignment = .center
+        label.textColor = Colors.Contact.title
+
+        if subtitleLabel == nil {
+            label.numberOfLines = 2
```
lugonzal, line 180: >2 [](start = 34, length = 1)

can you make it part of Constants

ios/FluentUI/People Picker/ContactView.swift
lugonzal: I think we should generalize it

lugonzal: The reason why I think this is very important is because I'm not sure clients will always be able to make the distinction between a name and last name, that structure and it's ordering I'm not sure if it's global and applies for every language.

jonathanwangg: Got it, thanks for explaining!

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -0,0 +1,196 @@
...
+        backgroundColor = Colors.surfacePrimary
+        setupLayout()
+    }
+
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    private func setupAvatarView(with firstName: String, and lastName: String) {
+        let fullName = firstName + " " + lastName
```
lugonzal, line 76: >fullName = firstName + " " + lastName [](start = 12, length = 37)

let's call this primaryText, and change the params to title and subtitle. We can't assume that clients passed a name as title and a last name as a subtitle.

I'm actually not sure if it would be correct to assume that we can concatenate both in this way

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Fixed this along with `setupSubtitleLabel()` accordingly.

ios/FluentUI/People Picker/ContactView.swift
lugonzal: Does the avatar view really need a primaryText set? If not, maybe we should avoid setting it since we're not sure

---
In reply to: [448534319](https://github.com/microsoft/fluentui-apple/pull/101#discussion_r448534319) [](ancestors = 448534319)

jonathanwangg: > let's call this primaryText, and change the params to title and subtitle. We can't assume that clients passed a name as title and a last name as a subtitle.
> 
> I'm actually not sure if it would be correct to assume that we can concatenate both in this way

I see what you mean regarding the concatenation. Do you have any suggestions on how to go about passing a first and last name as the `primaryText` argument to `avatarView.setup()`?

jonathanwangg: The `AvatarView` doesn't actually require a `primaryText` as it is an `Optional`. What do you mean by "avoid setting it since we're not sure if the text is correct"? I think we need to set pass in a `primaryText` argument when we call `avatarView.setup()` in order to display the one/two text labels.

lugonzal: This actually makes me think of an issue we might run into for the next Phase of development (the collection of Contact Cards). Clients will need to know the distinction between a name and a last name (hopefully People APIs know that distinction, it would be terrible if clients need to do text splits haha). To handle a person object, we use the Persona class in Fluent, which only has a fullName property. I imagine we'll use the same object to hold the data source for contact cards, we might need to modify that to accommodate for a first name and last name in that class.

lugonzal: > > Does the avatar view really need a primaryText set? If not, maybe we should avoid setting it since we're not sure if the text is correct
> > In reply to: [448534319](https://github.com/microsoft/fluentui-apple/pull/101#discussion_r448534319) [](ancestors = 448534319)
> 
> The `AvatarView` doesn't actually require a `primaryText` as it is an `Optional`. What do you mean by "avoid setting it since we're not sure if the text is correct"? I think we need to set pass in a `primaryText` argument when we call `avatarView.setup()` in order to display the one/two text labels.

Oh! Of course, avatarView needs that to setup the initials image, when there's no actual image... I just took a look at the method that creates the initials "initialLetters" and they just take the first letter from the first and second word on the fullname, regardless of whether they are actually the name and last name of a person. Fascinating, not the cleanest in my opinion.

So, I guess it is ok that we pass that concatenation as primaryText to avatarView

PR #103, opened 2020-06-25
Separator for macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

A new simple Separator control, modeled after the corresponding control in iOS.

### Verification

Tested Light/Dark and Voiceover (there is none) on macOS 10.15

Screen Shot 2020-06-26 at 11 31 52 AM

### Pull request checklist

This PR has considered:


macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+
+import AppKit
+
+@objc(MSFSeparatorStyle)
+public enum SeparatorStyle: Int {
+    case `default`
+    case shadow
+
+    fileprivate var color: NSColor {
+        switch self {
```
Saadnajmi, line 14: Inconsistent indentation. Really gotta get Swift Format in here

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+@objc(MSFSeparatorStyle)
+public enum SeparatorStyle: Int {
+    case `default`
+    case shadow
```
Saadnajmi, line 11: I initially made 2 styles to copy what we have on iOS. However, with the color choices I made (with input from design), there would be a significant different between 10.13 and 10.14. Either I find a better color than systemGray (I don't think I will), or I remove the "shadow" style altogether.

I'm more partial to removing the second style, considering that when I write the documentation, the first question will be "Why 2 styles, when is each used?" annnnnd I can't actually answer that 😅. For macOS we probably only need one style

macos/FluentUI/Separator.swift
```diff
@@ -77,17 +75,4 @@ open class Separator: NSView {
...
-	}
-
-	open override var intrinsicContentSize: CGSize {
-		switch orientation {
-		case .horizontal:
-			return CGSize(width: NSView.noIntrinsicMetric, height: frame.height)
-		case .vertical:
-			return CGSize(width: frame.width, height: NSView.noIntrinsicMetric)
-		}
-	}
```
Saadnajmi: We had this on the iOS control. I'm not sure the benefit of specifying `NSView.noIntrinsicMetric`, so I removed it. Would anyone have insight on what it is useful for?

Saadnajmi: I found this within the documentation for `AppKit.NSLayoutConstraint`

"Note that not all views have an intrinsicContentSize.  A horizontal slider has an intrinsic height, but no intrinsic width - the slider artwork has no intrinsic best width.  A horizontal NSSlider returns (NSViewNoIntrinsicMetric, <slider height>) for intrinsicContentSize.  An NSBox returns (NSViewNoIntrinsicMetric, NSViewNoIntrinsicMetric).  The _intrinsic_ content size is concerned only with data that is in the view itself, not in other views."

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+	open override var wantsUpdateLayer: Bool {
+		return true
+	}
+	
+	open override func updateLayer() {
+		if let layer = layer {
+			layer.backgroundColor = style.color.cgColor
+		}
+	}
+}
```
Saadnajmi, line 82: Added this comment on the commit I removed the ovverride, but ew-commenting here so it's not lost.
iOS contains the following override:
```
	open override var intrinsicContentSize: CGSize {
		switch orientation {
		case .horizontal:
			return CGSize(width: NSView.noIntrinsicMetric, height: frame.height)
		case .vertical:
			return CGSize(width: frame.width, height: NSView.noIntrinsicMetric)
		}
	}
```
 I'm not sure the benefit of specifying NSView.noIntrinsicMetric, so I removed it. Would anyone have insight on what it is useful for? 

I found this within the documentation for AppKit.NSLayoutConstraint

"Note that not all views have an intrinsicContentSize. A horizontal slider has an intrinsic height, but no intrinsic width - the slider artwork has no intrinsic best width. A horizontal NSSlider returns (NSViewNoIntrinsicMetric, ) for intrinsicContentSize. An NSBox returns (NSViewNoIntrinsicMetric, NSViewNoIntrinsicMetric). The intrinsic content size is concerned only with data that is in the view itself, not in other views."

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+    case horizontal
+    case vertical
+}
+
+
+@objc(MSFSeparator)
+open class Separator: NSView {
+	private var orientation: SeparatorOrientation = .horizontal
+	
+	private var style: SeparatorStyle = .default
```
anandrajeswaran, line 39: Since it's private this is not mutable? Or am I missing something? If it's not mutable, we can make it a let and we also don't need the updateLayer/wantsUpdateLayer. We can just set it once on init and overall do less work when the system asks us to display

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+@objc(MSFSeparatorOrientation)
+public enum SeparatorOrientation: Int {
+    case horizontal
+    case vertical
+}
+
+
+@objc(MSFSeparator)
+open class Separator: NSView {
+	private var orientation: SeparatorOrientation = .horizontal
```
anandrajeswaran, line 37: similar here - private implies not mutable, so make it a let?

macos/FluentUI/Separator.swift
anandrajeswaran: In this case we should use intrinsicContentSize instead of a constraint (can remove that code in initialize) - in general the principle of autolayout is that a view knows the size it wants to be if it's not a container of other views. NoIntrinsicMetric means the control is not particular about its size in a particular dimension, which makes sense for a separator. It's particular about its width when vertical or its height when horizontal,  but otherwise requires its container to constrain it to the size it wants

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+	
+	private var style: SeparatorStyle = .default
+
+	@objc public override init(frame frameRect: NSRect) {
+		super.init(frame: frameRect)
```
harrieshin, line 42: why let it available?

Saadnajmi: Can also remove the initWithFrame override altogether and force clients to specify the orientation. That makes a little more sense to me

Saadnajmi: Also good to point out I need to create the markdown documentation file as part of this review :D

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+    }
+}
+
+@objc(MSFSeparatorOrientation)
+public enum SeparatorOrientation: Int {
+    case horizontal
+    case vertical
+}
+
+
```
lugonzal, line 34: nit: extra line here. Swift lint might tell you about these

macos/FluentUI/Separator.swift
lugonzal, line 37: >private [](start = 1, length = 7)

in Fluent iOS we place public methods before private methods. I like that guideline for readability, what do you think?

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+	private var orientation: SeparatorOrientation = .horizontal
+	
+	private var style: SeparatorStyle = .default
+
+	@objc public override init(frame frameRect: NSRect) {
+		super.init(frame: frameRect)
+		initialize()
+	}
+	
+	@objc public init(orientation: SeparatorOrientation = .horizontal) {
```
lugonzal, line 46: >init [](start = 14, length = 4)

this initializer can be achieved with the one below in swift since we have a default value. Is the reason to have it here so that it's available to ObjC? If so, why not make one for the other case (init with just a style)?

I wonder how Swift would decide which initializer to use, since the sintax could point to any of them

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+    case shadow
+
+    fileprivate var color: NSColor {
+		switch self {
+		case .default:
+			if #available(OSX 10.14, *) {
+				return NSColor.separatorColor
+
+			} else {
+				return NSColor.systemGray
```
markavitale, line 20: [NSColor's quaternaryLabelColor](https://developer.apple.com/documentation/appkit/nscolor/1534635-quaternarylabelcolor) might be a better fallback? "Use this color for the least important text in your interface and for separators between text items."

Whereas [separatorColor](https://developer.apple.com/documentation/appkit/nscolor/2998831-separatorcolor) is "The color to use for separators between different sections of content. Do not use this color for split view dividers or window chrome dividers."

The distinction between those two types of separators is maybe interesting on macOS.

macos/FluentUI/Separator.swift
markavitale: Here's the relevant Microsoft swift style guide page: https://microsoft.github.io/swift-guide/FileContentOrdering.html

macos/FluentUI/Separator.swift
markavitale: I think right now these things are `var`s because they're populated in `initialize` which gets called _after_ `super.init(...)`. I think if we were to set them before, (which might result in minor code duplication) we could make them `let`s

macos/FluentUI/Separator.swift
markavitale: And also just code documentation. At the least everything public needs documentation, ideally just everything.

macos/FluentUI/Separator.swift
markavitale: Or if we can't get a solid read on what colors we want as a fallback, maybe just defining the fallback color in the asset catalog would work.

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,82 @@
...
+@objc(MSFSeparatorStyle)
+public enum SeparatorStyle: Int {
+    case `default`
+    case shadow
+
+    fileprivate var color: NSColor {
+		switch self {
+		case .default:
+			if #available(OSX 10.14, *) {
+				return NSColor.separatorColor
```
markavitale, line 17: Is `NSColor` necessary here? does just `return .separatorColor` work? If so we should use the shorthand syntax: https://microsoft.github.io/swift-guide/TypeInference.html

macos/FluentUI/Separator.swift
Saadnajmi: We don't have swift lint on the macOS project yet :(, but thank you for pointing this out!

macos/FluentUI/Separator.swift
Saadnajmi: Was just about to link that page :D 
Thanks, I wrote this before I was aware of that section. I'll update to reflect

macos/FluentUI/Separator.swift
Saadnajmi: Didn't know you could have a let without assigning it till the init. Thanks

macos/FluentUI/Separator.swift
Saadnajmi: Removed the var altogether

macos/FluentUI/Separator.swift
Saadnajmi: Code documentation added. Markdown still pending

macos/FluentUI/Separator.swift
Saadnajmi: Good point. I removed the default value, and we only have one initializer now

macos/FluentUI/Separator.swift
Saadnajmi: Forgot that .gridColor is probably the best fallback, and also what design requested. Switched to that.

macos/FluentUI/Separator.swift
Saadnajmi: Nope! Removed :D

macos/FluentUI/Separator.swift
Saadnajmi: @anandrajeswaran I'm fairly sure I still need `updateLayer/wantsUpdateLayer` so that the background color changes with light/dark mode. With testing, I could get away with not overriding `wantsUpdateLayer`, but I'm pretty sure that's just undefined behavior.

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,64 @@
...
+	let orientation: SeparatorOrientation
+	
+	/// Initializes a separator in the specified orientation
+	/// - Parameter orientation: The orientation of the separator, vertical or horizontal
+	@objc public init(orientation: SeparatorOrientation) {
+		self.orientation = orientation
+		super.init(frame: .zero)
+		wantsLayer = true
+		if let layer = layer {
+			layer.backgroundColor = separatorColor.cgColor
```
anandrajeswaran, line 25: oh wait you're right I'm wrong. We can't resolve the cgColor here because we're not in a known draw path so results may vary. Back to updateLayer it must go. Sorry for the bad advice

anandrajeswaran: Actually it's already in updateLayer. Just remove this and we're good

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,64 @@
...
+	open override var intrinsicContentSize: CGSize {
+		switch orientation {
+		case .horizontal:
+			return CGSize(width: NSView.noIntrinsicMetric, height: separatorThickness)
+		case .vertical:
+			return CGSize(width: separatorThickness, height: NSView.noIntrinsicMetric)
+		}
+	}
+	
+	private var separatorColor: NSColor {
```
anandrajeswaran, line 53: let instead of var possible? Or no

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,64 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+@objc(MSFSeparatorOrientation)
+public enum SeparatorOrientation: Int {
```
anandrajeswaran, line 9: These are obvious, but since public they probably need documentation?

macos/FluentUI/Separator.swift
Saadnajmi: `'let' declarations cannot be computed properties`

Nope :(

macos/FluentUI/Separator.swift
Saadnajmi: Changed to: 

```swift
public enum SeparatorOrientation: Int {
    case horizontal // Separator with no intrinic width
    case vertical // Separator with no intrinic height
}
```

macos/FluentUI/Separator.swift
lugonzal: private let separatorColor: NSColor = {
       // code
    }()

---
In reply to: [446368706](https://github.com/microsoft/fluentui-apple/pull/103#discussion_r446368706) [](ancestors = 446368706)

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,61 @@
...
+    case horizontal // Separator with no intrinic width
+    case vertical // Separator with no intrinic height
+}
+
+@objc(MSFSeparator)
+open class Separator: NSView {
+	let orientation: SeparatorOrientation
```
lugonzal, line 16: >let [](start = 1, length = 3)

can we be explicit about its private access

macos/FluentUI/Separator.swift
```diff
@@ -0,0 +1,61 @@
...
+			return .separatorColor
+		} else {
+			return.gridColor
+		}
+	}
+}
+
+// MARK: - Constants
+
+fileprivate let separatorThickness: CGFloat = 1.0
```
lugonzal, line 61: >separatorThickness [](start = 16, length = 18)

can we make this a private static let within Separator?

Saadnajmi: Sure. Also did so for separatorColor

macos/FluentUI/Separator.swift
Saadnajmi: Side note: I wrote the initializers as such because I copied it from the iOS initializers:
```
    @objc public override init(frame: CGRect) {
        super.init(frame: frame)
        initialize(style: .default, orientation: .horizontal)
    }

    @objc public init(style: SeparatorStyle = .default, orientation: SeparatorOrientation = .horizontal) {
        super.init(frame: .zero)
        initialize(style: style, orientation: orientation)
    }

    public required init?(coder aDecoder: NSCoder) {
    }
```

Might be worth going back and making sure we have matching initializers later on

PR #105, opened 2020-06-28
PillButton new style

isatoderici:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added an extra PillButtonStyle with new title/background colors and fonts for normal/selected states.

### Verification

1. Run FluentUI.Demo to open the app
2. Navigate to Pill Button Bar
3. See Transparent bar

Screen Shot 2020-06-27 at 9 21 20 PM
Screen Shot 2020-06-27 at 9 21 52 PM
Screen Shot 2020-06-29 at 3 30 06 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/105)

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -57,6 +57,24 @@ public enum PillButtonStyle: Int {
             return UIColor(light: Colors.primary(for: window), dark: Colors.primaryShade10(for: window))
         }
     }
+    
+    func font() -> UIFont {
+        switch self {
+        case .outline, .filled:
+            return Fonts.button4
+        case .transparent:
+            return Fonts.button1
+        }
+    }
+    
+    func selectedFont() -> UIFont {
+        switch self {
+        case .outline, .filled:
+            return Fonts.button4
+        case .transparent:
+            return Fonts.preferredFont(forTextStyle: .subheadline, weight: .bold)
+        }
+    }
 }
 
 // MARK: PillButton
@@ -70,7 +88,6 @@ open class PillButton: UIButton {
     private struct Constants {
         static let bottomInset: CGFloat = 6.0
         static let cornerRadius: CGFloat = 16.0
-        static let font: UIFont = Fonts.button4
         static let horizontalInset: CGFloat = 16.0
         static let topInset: CGFloat = 6.0
     }
@@ -105,7 +122,7 @@ open class PillButton: UIButton {
 
     private func setupView() {
         setTitle(pillBarItem.title, for: .normal)
-        titleLabel?.font = style == .transparent ? Fonts.button1 : Constants.font
+        titleLabel?.font = style.font()
         layer.cornerRadius = Constants.cornerRadius
         clipsToBounds = true
 
@@ -143,8 +160,6 @@ open class PillButton: UIButton {
     }
     
     private func updateFont() {
-        if style == .transparent {
-            titleLabel?.font = isSelected ? Fonts.preferredFont(forTextStyle: .subheadline, weight: .bold) : Fonts.button1
-        }
+        titleLabel?.font = isSelected ? style.selectedFont() : style.font()
     }
 }
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -14,6 +14,7 @@ public typealias MSPillButtonStyle = PillButtonStyle
 public enum PillButtonStyle: Int {
     case outline
     case filled
+    case transparent
 
     func backgroundColor(for window: UIWindow) -> UIColor {
         switch self {
@@ -21,6 +22,8 @@ public enum PillButtonStyle: Int {
             return Colors.PillButton.Outline.background
         case .filled:
             return UIColor(light: Colors.primaryShade10(for: window), dark: Colors.PillButton.Outline.background)
+        case .transparent:
+            return .clear
         }
     }
 
@@ -30,12 +33,14 @@ public enum PillButtonStyle: Int {
             return UIColor(light: Colors.primary(for: window), dark: Colors.surfaceQuaternary)
         case .filled:
             return UIColor(light: Colors.surfacePrimary, dark: Colors.surfaceQuaternary)
+        case .transparent:
+            return Colors.primaryTint40(for: window)
         }
     }
 
     var titleColor: UIColor {
         switch self {
-        case .outline:
+        case .outline, .transparent:
             return Colors.PillButton.Outline.title
         case .filled:
             return Colors.PillButton.Filled.title
@@ -48,6 +53,8 @@ public enum PillButtonStyle: Int {
             return  Colors.PillButton.Outline.titleSelected
         case .filled:
             return UIColor(light: Colors.primary(for: window), dark: Colors.PillButton.Outline.titleSelected)
+        case .transparent:
+            return UIColor(light: Colors.primary(for: window), dark: Colors.primaryShade10(for: window))
         }
     }
 }
@@ -76,6 +83,7 @@ open class PillButton: UIButton {
         didSet {
             updateAppearance()
             updateAccessibilityTraits()
+            updateFont()
         }
     }
 
@@ -97,7 +105,7 @@ open class PillButton: UIButton {
 
     private func setupView() {
         setTitle(pillBarItem.title, for: .normal)
-        titleLabel?.font = Constants.font
+        titleLabel?.font = style == .transparent ? Fonts.button1 : Constants.font
         layer.cornerRadius = Constants.cornerRadius
         clipsToBounds = true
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -14,13 +14,16 @@ public typealias MSPillButtonStyle = PillButtonStyle
...
+    case transparent
 
     func backgroundColor(for window: UIWindow) -> UIColor {
         switch self {
         case .outline:
             return Colors.PillButton.Outline.background
         case .filled:
             return UIColor(light: Colors.primaryShade10(for: window), dark: Colors.PillButton.Outline.background)
+        case .transparent:
+            return .clear
```
Saadnajmi, line 26: We'll probably want to add this to the Colors.Swift file, even if it is just `.clear`

lugonzal:
Since this is a public repo, I would remove any comment about Office features and usage.

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -30,12 +33,14 @@ public enum PillButtonStyle: Int {
             return UIColor(light: Colors.primary(for: window), dark: Colors.surfaceQuaternary)
         case .filled:
             return UIColor(light: Colors.surfacePrimary, dark: Colors.surfaceQuaternary)
+        case .transparent:
+            return Colors.primaryTint40(for: window)
```
Saadnajmi, line 37: Looking at the xcode declaration, I see the warning
```
'primaryTint40' is deprecated: renamed to 'setProvider(_:forWindow:)'
Use 'setProvider(_:forWindow:)' instead
```

Should we not be using this API anymore and replace it with the ColorProviding protocol @anandrajeswaran added?

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -97,7 +105,7 @@ open class PillButton: UIButton {
 
     private func setupView() {
         setTitle(pillBarItem.title, for: .normal)
-        titleLabel?.font = Constants.font
+        titleLabel?.font = style == .transparent ? Fonts.button1 : Constants.font
```
Saadnajmi, line 108: Either move `Fonts.button1` to the constants struct, or move `Constants.Font` out of the Constants struct and call the value directly?

ios/FluentUI/Pill Button Bar/PillButton.swift
lugonzal: this method is using the color providing protocol. Its implementation has a warning because we still need to support clients that do not use the protocol, so it calls deprecated var primaryTint40 in that case

---
In reply to: [447136201](https://github.com/microsoft/fluentui-apple/pull/105#discussion_r447136201) [](ancestors = 447136201)

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -133,4 +141,10 @@ open class PillButton: UIButton {
             }
         }
     }
+    
+    private func updateFont() {
+        if style == .transparent {
+            titleLabel?.font = isSelected ? Fonts.preferredFont(forTextStyle: .subheadline, weight: .bold) : Fonts.button1
```
lugonzal, line 147: >lected ? Fonts.pr [](start = 35, length = 17)

can we add style methods, "selectedFont" and "font", like we have "selectedTitleColor" and "titleColor"?

ios/FluentUI/Pill Button Bar/PillButton.swift
lugonzal, line 108: > titleLabel?.font [](start = 7, length = 17)

I think it'd be cleaner to just call setupFont here

lugonzal:
Can you also attach a screenshot of dark mode?

harrieshin:
is it really transparent style if we aren't making any alpha changes? (you can't see things behind the pill button) where does that name come from? design?

ios/FluentUI/Pill Button Bar/PillButton.swift
isatoderici: It is not a good practice to add variables provided by Apple in a global scope, because that will increase the project/framework size, as well as the layer complexity.

ios/FluentUI/Pill Button Bar/PillButton.swift
isatoderici: as this "font" and "selectedFont" are used only for .transparent case I didn't add separate style methods. Do you think we need them because they will increase the layer of complexity so instead of calling "Fonts.button1" we will call "font" that calls "Fonts.button1". Same for "selectedFont".

ios/FluentUI/Pill Button Bar/PillButton.swift
isatoderici: I think, in this case, a better approach is linked to @Saadnajmi comment but a bit different, to move Fonts.button1 to Constants as it is used in more than one place, and just because it is 1 line of code, do you think it needs an extra method?

isatoderici:
> is it really transparent style if we aren't making any alpha changes? (you can't see things behind the pill button) where does that name come from? design?

Does the design provide names as well? If not, can you come with a better proposal?

isatoderici:
> Since this is a public repo, I would remove any comment about Office features and usage in the PR description

Sounds good, makes sense! I added this description just to let you know why I need these changes.

ios/FluentUI/Pill Button Bar/PillButton.swift
lugonzal: I think the extra complexity is minor. It looks much more organized if we delegate all style specific code to the style enum.

ios/FluentUI/Pill Button Bar/PillButton.swift
lugonzal: I think it's better to not duplicate this code, even if it's 1 line. This will make sure that a future potential change in  this configuration is addressed in all cases, reducing the likelihood of bugs.

lugonzal:
> > is it really transparent style if we aren't making any alpha changes? (you can't see things behind the pill button) where does that name come from? design?
> 
> Transparent comes from button background color which for the normal state is .clear (= under the hood: white: 0.0, alpha: 0.0). Does the design provide names as well? If not, can you come with a better proposal?

Checkout the section "Segmented Controls" here: https://microsoft.sharepoint-df.com/:w:/t/APEXShared/EWY-LsUjfvpDnGd7TEIeuk0B4Ug6k4mXmouP7znfv8bMhw?e=5r91ke

Designers should add a new style here, and they will provide an adequate name. Can you double check with the designer of this new style what would be an appropriate name? I think "clear" might be better? Not sure. I'm signing off, but please double check this before I merge the PR

Saadnajmi:
> > > is it really transparent style if we aren't making any alpha changes? (you can't see things behind the pill button) where does that name come from? design?
> > 
> > 
> > Transparent comes from button background color which for the normal state is .clear (= under the hood: white: 0.0, alpha: 0.0). Does the design provide names as well? If not, can you come with a better proposal?
> 
> Checkout the section "Segmented Controls" here: https://microsoft.sharepoint-df.com/:w:/t/APEXShared/EWY-LsUjfvpDnGd7TEIeuk0B4Ug6k4mXmouP7znfv8bMhw?e=5r91ke
> 
> Designers should add a new style here, and they will provide an adequate name. Can you double check with the designer of this new style what would be an appropriate name? I think "clear" might be better? Not sure. I'm signing off, but please double check this before I merge the PR

Wrong link?

lugonzal:
> > > > is it really transparent style if we aren't making any alpha changes? (you can't see things behind the pill button) where does that name come from? design?
> > > 
> > > 
> > > Transparent comes from button background color which for the normal state is .clear (= under the hood: white: 0.0, alpha: 0.0). Does the design provide names as well? If not, can you come with a better proposal?
> > 
> > 
> > Checkout the section "Segmented Controls" here: https://www.figma.com/file/I5eZV4sJEzUasYcHu5oalcqb/Fluent-iOS?node-id=12824%3A18701
> > Designers should add a new style here, and they will provide an adequate name. Can you double check with the designer of this new style what would be an appropriate name? I think "clear" might be better? Not sure. I'm signing off, but please double check this before I merge the PR
> 
> Wrong link?

Yes :) Just updated it

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -48,6 +53,26 @@ public enum PillButtonStyle: Int {
...
+        }
+    }
+    
+    func selectedFont() -> UIFont {
+        switch self {
+        case .outline, .filled:
+            return Fonts.button4
+        case .transparent:
+            return Fonts.preferredFont(forTextStyle: .subheadline, weight: .bold)
         }
```
anandrajeswaran, line 76: Sorry if this has already been iterated away from based on previous comments, but thoughts on calling this func font(for selected: Bool) and combining the two cases? That could dedupe Fonts.btton4 for outline and filled styles  for a bit more clarity

PR #108, opened 2020-07-03
Side tab bar

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

SideBar is a vertical navigation control for iPad apps. 
This control can be configured to display an avatar view, and set of tabs at the top, and a set of buttons at the bottom.
SideBar is similar to TabBarView and reuses the TabBarItems for displaying tabs and the bottom buttons.
A maximum of 5 top tabs, and 5 bottom buttons can be set in the control.
A delegate protocols allows consumers of the control to handle tap events on the avatar view, top tab buttons and bottom buttons.

dark mode
light mode

### Verification

Tested iPad size transitions.
Tested iPad regular and compact.
Tested dark/light mode.
Tested large text accessibility and large content viewer.
Tested RTL
Tested VoiceOver and keyboard accessibility.
Make sure that the control renders correctly on iPhone, even though it is not intended to be used on that platform.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/108)


ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
```diff
@@ -18,6 +18,7 @@ let demos: [(title: String, controllerClass: UIViewController.Type)] = [
     ("HUD", HUDDemoController.self),
     ("Label", LabelDemoController.self),
     ("NavigationController", NavigationControllerDemoController.self),
+	("SideBar", SideBarDemoController.self),
```
anandrajeswaran, line 21: I think this is list is in alphabetical order

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
...
+
+class SideBarDemoController: DemoController {
+	private var sideBar: SideBar?
+	private var contentViewController: UIViewController?
+
+	override func viewDidLoad() {
+		super.viewDidLoad()
+
+		container.addArrangedSubview(createButton(title: "Show side bar", action: #selector(presentSideBar)))
+	}
```
anandrajeswaran, line 17: Should use loadView instead. loadView is used when constructing a hierarchy programmatically rather than using a xib, viewDidLoad is used for additional configuration after loading from a xib. Documentation pending at https://microsoft.github.io/apple-ux-guide/ViewControllers.html#construct-view-hierarchies-in-loadview

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class SideBarDemoController: DemoController {
+	private var sideBar: SideBar?
```
anandrajeswaran, line 10: A minor that this is persistent here, convention is to sort swift files from most visible symbols to least visible so it roughly reads in the order of importance. So private vars generally belong towards the bottom https://microsoft.github.io/swift-guide/FileContentOrdering.html

anandrajeswaran, line 10: here and elsewhere, could be lazy instead of optional?

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
...
+		super.viewDidLoad()
+
+		container.addArrangedSubview(createButton(title: "Show side bar", action: #selector(presentSideBar)))
+	}
+
+	@objc private func presentSideBar() {
+		let controller = UIViewController(nibName: nil, bundle: nil)
+		contentViewController = controller
+		controller.modalPresentationStyle = .fullScreen
+		controller.view.backgroundColor = Colors.background1
```
anandrajeswaran, line 23: minor: can we avoid unnecessary references to the monolithic colors file? The goal is eventually to break that up

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
...
+
+		sideBar = SideBar(frame: .zero)
+		sideBar!.insert(in: controller.view)
```
anandrajeswaran, line 26: per other comment, can avoid force unwrap by making this lazy

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
...
+	@objc private func dismissSideBar() {
+		dismiss(animated: false, completion: nil)
+	}
+
+	@objc private func toggleAvatarView(switchView: UISwitch) {
+		showAvatarView(switchView.isOn)
+	}
+
+	private func showAvatarView(_ show: Bool) {
+		var avatarView: AvatarView?
```
anandrajeswaran, line 95: Another minor thing: can we avoid a dependency on AvatarView here? There is a goal to make fluentui-apple "composable" by each client separately so they can only take the pieces they need and reduce overall bundle size impact. If we can force each component to be "independently" testable, that will help

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,122 @@
...
+		])
+
+		present(controller, animated: false)
+	}
+
+	private func createLabelAndSwitchRow(labelText: String, switchAction: Selector, isOn: Bool = false) -> UIView {
+		let stackView = UIStackView(frame: .zero)
+		stackView.axis = .horizontal
+		stackView.alignment = .center
+		stackView.spacing = 10.0
```
anandrajeswaran, line 72: Even though this is test code, should still split constants out to the bottom to set a good example for consumers who are using this as a template

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+		case bottom
+	}
+
+	private struct Constants {
+		static let maxTabCount: Int = 5
+		static let viewWidth: CGFloat = 62.0
+		static let avatarViewSize: CGFloat = 30.0
+		static let topItemSpacing: CGFloat = 38.0
+		static let bottomItemSpacing: CGFloat = 28.0
+	}
```
anandrajeswaran, line 30: Per other comment and swift-guide reference, move private stuff to the bottom

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+
+	@objc public weak var delegate: SideBarDelegate?
```
anandrajeswaran, line 32: public var so should get documentation - our swift guidelines match our objC guidelines there https://microsoft.github.io/objc-guide/Documentation/Patterns.html

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+
+	private var avatarViewGestureRecognizer: UITapGestureRecognizer?
+	@objc open var avatarView: AvatarView? {
```
anandrajeswaran, line 35: Per other comment - I'm initially leaning towards this just having a dependency on a view and not a specific AvatarView, but curious to hear your thoughts

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+		static let maxTabCount: Int = 5
+		static let viewWidth: CGFloat = 62.0
+		static let avatarViewSize: CGFloat = 30.0
+		static let topItemSpacing: CGFloat = 38.0
+		static let bottomItemSpacing: CGFloat = 28.0
+	}
+
+	@objc public weak var delegate: SideBarDelegate?
+
+	private var avatarViewGestureRecognizer: UITapGestureRecognizer?
```
anandrajeswaran, line 34: Can we just use a UIButton for more standard accessibility behavior?

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+		let allItems = items(in: section)
+		let numberOfItems = allItems.count
+		if numberOfItems > Constants.maxTabCount {
+			preconditionFailure("tab bar items can't be more than \(Constants.maxTabCount)")
+		}
+
+		let stackView = stackview(in: section)
+
+		for item in allItems {
+			let tabBarItemView = TabBarItemView(item: item, showsTitle: false)
```
anandrajeswaran, line 114: Does TabBarItemView really require us to install our own gesture recognizer? That might be a bug in TabBarItemView?

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+			}
+		}
+	}
+
+	@objc public override init(frame: CGRect) {
+		super.init(frame: frame)
+		initialize()
+	}
+
+	@objc public required init?(coder: NSCoder) {
```
anandrajeswaran, line 146: Can we just make this unavailable?

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+
+	private func setupLayout() {
+		if layoutConstraints.count > 0 {
+			NSLayoutConstraint.deactivate(layoutConstraints)
+			layoutConstraints.removeAll()
+		}
+
+		if let avatarView = self.avatarView {
+			layoutConstraints.append(contentsOf: [
+				avatarView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: 18.0),
```
anandrajeswaran, line 207: pull the constants (18.0 here, others below) out of line into constants to make them self-documenting

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+
+		NSLayoutConstraint.activate([widthAnchor.constraint(equalToConstant: Constants.viewWidth),
+									 borderLine.leadingAnchor.constraint(equalTo: trailingAnchor),
+									 borderLine.bottomAnchor.constraint(equalTo: bottomAnchor),
+									 borderLine.topAnchor.constraint(equalTo: topAnchor)])
+	}
+
+	private var layoutConstraints: [NSLayoutConstraint] = []
+
+	private func setupLayout() {
```
anandrajeswaran, line 199: setupConstraints may be a more accurate name?

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+	private func stackViewHeight(_ stackView: UIStackView) -> CGFloat {
+		let itemCount: CGFloat = CGFloat(stackView.arrangedSubviews.count)
+		var height: CGFloat = 24.0 * itemCount
+
+		if (itemCount > 0) {
+			height += stackView.spacing * (itemCount - 1)
+		}
+
+		return height
+	}
```
anandrajeswaran, line 237: Can we instead set the contentCompressionResistance/hugging priorities as appropriate on the items within the stack view so it automatically takes its height

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+	private func items(in section: Section) -> [TabBarItem] {
+		switch section {
+		case .top:
+			return topItems
+		case .bottom:
+			return bottomItems
+		}
+	}
+
+	private func stackview(in section: Section) -> UIStackView {
```
anandrajeswaran, line 248: minor, should be camelCase (stackView) to match UIKit

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,284 @@
...
+				return tabBarItemView
+			}
+		}
+
+		return nil
+	}
+
+	@objc private func handleAvatarViewTapped(_ recognizer: UITapGestureRecognizer) {
+		delegate?.sideBar?(self, didActivate: avatarView!)
+	}
```
anandrajeswaran, line 269: Per other comment, can we just let the client put a view in here themselve so they are free to handle it? On the other hand if this control is opinionated enough that the view needs to be an avatar view, then maybe it should be opinionated enough to handle that click ourselves? But we probably shouldn't do one and not the other.

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
MathieuKavalec: I originally designed this to take any view and then switched to avatar view.
The intention is to force a specific design for this control. The user MUST show an avatar view here. We don't want to give the flexibility of passing in any view here because we can't control how this will look.
The goal of this control library is to provide a unified set of controls and a unified design. We wouldn't apps to start placing in custom views here. We would want everyone to adopt avatar view and use that to display the user image. Therefore, we have a hard dependency on avatar view in this component.

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: A UIButton is a view. Avatar view is not a button. We would need to convert AvatarView to a button. Adding an invisible button above the avatar view would be hacky. This is out of scope for this change and I don't even think we want this. Tested accessibility and things work as expected. This is also how we handle the tap events on avatar view in other places in fluentUI (see LargeTitleView).

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: See my response to your other comment. I want this to explicitly be AvatarView so we can control the experience.

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: Yes, it does. This is how we handle touch events in TabBarView also.
Changing this is out of scope for this change.
Is there a process to file a bug and follow up for this?

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: As mentioned in my other responses, we want to force this view to be an avatar view that the client can provide. If no avatar view is provided, we don't display anything above the tab bar items.
As for touch events, we are making it easier on clients to intercept touch events by adding our own gesture recognizer. I personally like this approach because it simplifies the usage of this class. Otherwise, each client will need to add duplicated logic to handle touch events on the view. This also allows us to define the touch area for the avatar view. Right now, the touch area is just the view itself but, we could change that if we wanted to.

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: Unfortunately, the intrinsic size of TabBarItemView is zero height and zero width. TabBarItemView contains an image view and a label. Those take up space and are outside of the view's bounds. This definitely seems like a bug. I tried fixing TabBarItemView so that the root view contains all its subviews but this caused severe layout bugs in TabBarView. I think fixing TabBarItemView's behavior is out of scope for this change. What is the process to file a bug?
If TabBarItemView's view was sized correctly, we wouldn't need logic at all. Currently, our stack view thinks that TabBarItemView have a zero size.

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Delegate protocol to handle user events inside the side bar.
+@objc(MSFSideBarDelegate)
+public protocol SideBarDelegate {
+    /// Called after the view representing `TabBarItem` is selected.
+    @objc optional func sideBar(_ sideBar: SideBar, didSelect item: TabBarItem, fromTop: Bool)
```
lugonzal, line 12: >fromTop [](start = 80, length = 7)

please add documentation about this param

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+
+    /// Called after the avatar view is tapped in the side bar.
+    @objc optional func sideBar(_ sideBar: SideBar, didActivate avatarView: AvatarView)
+}
+
+/// View for a vertical side bar that can be used for app navigation.
+/// Optimized for horizontal regular + vertical regular size class configuration. Prefer using TabBarView for other size class configurations.
+@objc(MSFSideBar)
+open class SideBar: UIView {
```
lugonzal, line 21: >SideBar [](start = 11, length = 7)

if this uses tabBarItems mostly, why not call it SideTabBar? Just wonder if it'll be confused with the new iOS14 SideBar

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+    @objc open var topItems: [TabBarItem] = [] {
+        willSet {
+            willSetItems(in: .top)
+        }
+        didSet {
+            didSetItems(in: .top)
+        }
+    }
+
+    /// Tab bar iems to display in the bottom section of the side bar.
```
lugonzal, line 59: >bottom [](start = 39, length = 6)

we should add here that these items don't have a permanent selected state

lugonzal:
screenshots look awesome

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+    }
+
+    private var layoutConstraints: [NSLayoutConstraint] = []
+    private var avatarViewGestureRecognizer: UITapGestureRecognizer?
+    private let borderLine = Separator(style: .shadow, orientation: .vertical)
+
+    private let backgroundView: UIVisualEffectView = {
+        var style = UIBlurEffect.Style.regular
+        if #available(iOS 13, *) {
+            style = .systemMaterial
```
lugonzal, line 129: >systemMaterial [](start = 21, length = 14)

horizontal tab bar is systemChromeMaterial, do we want to keep that consistent?

MathieuKavalec: Yes, we do. I'll update this.

ios/FluentUI/Navigation/SideBar.swift
lugonzal, line 21: >SideBar [](start = 11, length = 7)

If this is not part of the Fluent design spec (figma, I can send link privately), could you please ask the designer to include it there?

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+        let allItems = items(in: section)
+        let numberOfItems = allItems.count
+        if numberOfItems > Constants.maxTabCount {
+            preconditionFailure("tab bar items can't be more than \(Constants.maxTabCount)")
+        }
+
+        let stackView = self.stackView(in: section)
+
+        for item in allItems {
+            let tabBarItemView = TabBarItemView(item: item, showsTitle: false)
```
lugonzal, line 226: >showsTitle [](start = 60, length = 10)

any plans to support this? I wonder if it would be as simple as setting this

MathieuKavalec: No plans, this isn't part of the figma designs. The designs show that, at some point, like the iOS 13 side bar, this control can be extended. When this happens, we will probably create a new control for the icon + text.

ios/FluentUI/Navigation/SideBar.swift
MathieuKavalec: Kaan did mention something about adding this to the mobile toolkit and he's working on it.
It is in the figma designs:
https://www.figma.com/file/0tO5Oliq4eyd48dL8gFAj1/WXP-iPad-L1-Fluent-Refresh?node-id=259%3A0

ios/FluentUI/Navigation/SideBar.swift
```diff
@@ -0,0 +1,296 @@
...
+                avatarViewGestureRecognizer = tapGesture;
+                view.addGestureRecognizer(tapGesture)
+            }
+
+            setupLayoutConstraints()
+        }
+    }
+
+    /// Tab bar iems to display in the top section of the side bar.
+    @objc open var topItems: [TabBarItem] = [] {
```
lugonzal: >TabBarItem [](start = 30, length = 10)

I think tabBarItemView is optimized for iPhone. If the "landscapeImage" property is set in the items (which in iPhone usually means a smaller image, because landscape has less vertical space), we'll use that image. On iPad landscape mode doesn't mean less vertical space. Perhaps we should comment that the lanscapeImage should not be set

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideBarDemoController.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    @objc private func presentSideBar() {
+        contentViewController.modalPresentationStyle = .fullScreen
+        contentViewController.view.backgroundColor = view.backgroundColor
+
+        sideBar.insert(in: contentViewController.view)
+        sideBar.delegate = self
+
+        sideBar.topItems = [
+            TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: UIImage(named: "Home_24")!, landscapeSelectedImage: UIImage(named: "Home_Selected_24")!),
```
lugonzal: >Home_24 [](start = 157, length = 7)

We should not set the landscape image here. Side bar should always use the "portrait" image which is bigger. The naming is not ideal for ipad since tabbaritem and tabBarItemView were optimized for iPhone

lugonzal:
Is it by design that the status bar is overlapping the sidebar?

MathieuKavalec:
> Is it by design that the status bar is overlapping the sidebar?

Yes, this is what the figma designs are showing.

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/TabBarView/Help_24.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "ic_help_24_outlined.pdf"
```
lugonzal, line 5: >ic_help_24_outlined [](start = 20, length = 19)

no need for these anymore

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideTabBarDemoController.swift
```diff
@@ -0,0 +1,128 @@
...
+
+    @objc private func presentSideTabBar() {
+        contentViewController.modalPresentationStyle = .fullScreen
+        contentViewController.view.backgroundColor = view.backgroundColor
+
+        sideTabBar.insert(in: contentViewController.view)
+        sideTabBar.delegate = self
+
+        sideTabBar.topItems = [
+            TabBarItem(title: "Home", image: UIImage(named: "Home_28")!, selectedImage: UIImage(named: "Home_Selected_28")!, landscapeImage: nil, landscapeSelectedImage: nil),
```
lugonzal, line 36: >landscapeImage: nil, landscapeSelectedImage: ni [](start = 125, length = 47)

these are optional params, no need to set nil

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,304 @@
...
+
+            avatarView?.removeFromSuperview()
+        }
+        didSet {
+            if let view = avatarView {
+                view.translatesAutoresizingMaskIntoConstraints = false
+                addSubview(view)
+
+                let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleAvatarViewTapped))
+                avatarViewGestureRecognizer = tapGesture;
```
lugonzal, line 46: the gesture is independent of the new avatartView, we perhaps don't have to create it every time and it can be a let

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,304 @@
...
+
+    private let bottomStackView: UIStackView = {
+        let bottomStackView = UIStackView(frame: .zero)
+        bottomStackView.axis = .vertical
+        bottomStackView.distribution = .fillEqually
+        bottomStackView.alignment = .fill
+        bottomStackView.translatesAutoresizingMaskIntoConstraints = false
+        bottomStackView.spacing = Constants.bottomItemSpacing
+
+        return bottomStackView
```
lugonzal, line 162: super nit, this code seems to be almost same as above, can we share it?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,304 @@
...
+    }
+
+    private func didSetItems(in section: Section) {
+        let allItems = items(in: section)
+        let numberOfItems = allItems.count
+        if numberOfItems > Constants.maxTabCount {
+            preconditionFailure("tab bar items can't be more than \(Constants.maxTabCount)")
+        }
+
+        let stackView = self.stackView(in: section)
```
lugonzal, line 231: >self [](start = 24, length = 4)

not needed

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,304 @@
...
+                                     borderLine.topAnchor.constraint(equalTo: topAnchor)])
+    }
+
+    private func setupLayoutConstraints() {
+        if layoutConstraints.count > 0 {
+            NSLayoutConstraint.deactivate(layoutConstraints)
+            layoutConstraints.removeAll()
+        }
+
+        if let avatarView = self.avatarView {
```
lugonzal, line 195: >self. [](start = 28, length = 5)

needed?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,304 @@
...
+}
+
+/// View for a vertical side tab bar that can be used for app navigation.
+/// Optimized for horizontal regular + vertical regular size class configuration. Prefer using TabBarView for other size class configurations.
+@objc(MSFSideTabBar)
+open class SideTabBar: UIView {
+    /// Delegate to handle user interactions in the side tab bar.
+    @objc public weak var delegate: SideTabBarDelegate?
+
+    /// The avatar view that displays above the top tab bar items.
```
lugonzal, line 30: >avatar view [](start = 12, length = 11)

since this is a fixed width view, we should suggest a size class for the avatar

ios/FluentUI/Tab Bar/SideTabBar.swift
MathieuKavalec: It is needed, otherwise the compiler complains because the variable name and method name are conflicting. I wanted to keep the var name "stackView" so decided to add the explicit self.

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,299 @@
...
+        didSet {
+            if let item = selectedTopItem {
+                itemView(with: item, in: .top)?.isSelected = true
+            }
+        }
+    }
+
+    @objc public override init(frame: CGRect) {
+        super.init(frame: frame)
+        initialize()
```
anandrajeswaran, line 88: Can just inline the contents of this if it's the only place it's being called?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,299 @@
...
+    /// Helper method to insert the view in a superview.
+    @objc public func insert(in superView: UIView) {
+        superView.addSubview(self)
+
+        NSLayoutConstraint.activate([
+            leadingAnchor.constraint(equalTo: superView.leadingAnchor),
+            topAnchor.constraint(equalTo: superView.topAnchor),
+            bottomAnchor.constraint(equalTo: superView.bottomAnchor)
+        ])
+    }
```
anandrajeswaran, line 105: This pattern still seems a bit odd to me. Why not make the public API a SideTabBarViewController so we aren't having this view take implicit ownership/knowledge of its superview?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,299 @@
...
+                avatarView.heightAnchor.constraint(equalToConstant: Constants.avatarViewSize),
+                topStackView.topAnchor.constraint(equalTo:avatarView.bottomAnchor , constant: Constants.avatarViewTopStackViewPadding)
+            ])
+        } else {
+            layoutConstraints.append(topStackView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: Constants.topStackViewTopPadding))
+        }
+
+        layoutConstraints.append(contentsOf: [
+            topStackView.widthAnchor.constraint(equalTo: widthAnchor),
+            topStackView.heightAnchor.constraint(equalToConstant: SideTabBar.stackViewHeight(topStackView)),
```
anandrajeswaran, line 193: Unable to find the other comment about this (thought I left one) but can't we achieve this using contentHuggingPriority and contentCompressionResistancePriority so we don't need to come back and update this if we change the factors we use for laying out the UIStackView?

ios/FluentUI/Tab Bar/SideTabBar.swift
MathieuKavalec: This is just a helper method to insert the view in its superview and position it along the leading edge.
I added a comment in the methods header to point that out.
The expectation is that, in most cases, this is how the view will get inserted in a view hierarchy so we can make it easier for consumers of this view.

ios/FluentUI/Tab Bar/SideTabBar.swift
MathieuKavalec: Explanation from my other comment:
The TabBarItemView's frame CGRectZero. This is definitely a bug. TabBarItemView contains an image view and a label that are laid out outside of the view's bounds. I tried quickly fixing TabBarItemView but this caused severe layout bugs in TabBarView. If TabBarItemView's root view was sized correctly, we could use contentHuggingPriority.
Fixing TabBarItemView is out of scope for this change and this current logic isn't adding much complexity.
Is there a way to file bugs for the FluentUI repo?

anandrajeswaran: We just use standard github issues to track issues. Can you file one and add a reference to it in a comment next to stackViewHeight so someone looking at this code later knows when it can go away?

ios/FluentUI/Tab Bar/SideTabBar.swift
anandrajeswaran: I guess a rephrasing is - is there any reason not to just expose a view controller instead of a view if the designed usage is as a decorator for a contentViewController?

MathieuKavalec: I was inspired by TabBarView which is just a view.
This control only contains view logic: it lays out subviews and intercepts touch events to pass them to a delegate. It could be a view controller, but I don't see any benefit. This class doesn't take care of any presentation logic, there is no controller logic pushing model data into views (MVC), there is no need for managing the lifecycle of the view, etc. 
I'd prefer to remove this helper method and make it clear that this is just a view that can be positioned anywhere.

ios/FluentUI/Tab Bar/SideTabBar.swift
MathieuKavalec: Filed a bug and added a comment.

ios/FluentUI/Tab Bar/SideTabBar.swift
anandrajeswaran: Yeah the point about being aligned with TabBarView makes sense but makes me nervous at the same time. Some thoughts:

- Fluent design does not really state that it can be positioned anywhere. It is part of fluent design that the TabBarView used to be restricted to the bottom of the screen and this control is being added because fluent design now says that a similar but different control can be used for the leading edge of the screen
- View controllers by nature don't take care of presentation logic of themselves, unless you're referring to the presentation of the content view controller. If that, I do kinda think it should take care of that presentation logic. If anyone is using this control, per fluent design the tabs should not execute actions but instead should swap out contentViewControllers.
- I guess the ultimate question - how is this control different than something like UIPageViewController or UITabBarController, both of which Apple chose UIViewControllers for? Default should be to follow Apple's patterns and idioms, so what is the distinction here that makes that undesirable?

MathieuKavalec: I see your point. The believe a good solution here would be to make SideTabBarView private and expose a view controller that manages the SideTabBarView and swaps view controllers.
It was my intention to build this view controller to start with because, depending on the window size class, we will need to alternate between SideTabBarView and TabBarView.
This view controller will involve a lot more work.

anandrajeswaran: So proposal is to check in as is and then add a view controller on top of it? That seems reasonable to me, but maybe we should start SideBarTabView as private if we intend it to be private later? Deprecating something is difficult and painful and since this is being done in the open since anyone can take a dependency on it

anandrajeswaran: At the least we should document this on SideBarTab View in the form of comments

MathieuKavalec: Yes, we will need to provide a view controller on top of it.
I spoke to our designer Today and there are still many gray areas when it comes to how this side bar will be used and where it will appear depending on screen size. 
The idea here is to have a master view controller that places all the subviews in the right place: header view, side tab bar view, tab bar view. The view controller would also take care of swapping UIViewControllers based on the selected tab.
For now, I'll try to make this SideBarTabView private. Hopefully that doesn't prevent us from using it in the demo app.

MathieuKavalec: Making SideBarTabView internal prevents us from using it in the test app. 
Is there a way to get around this limitation?
Do we really want to prevent SideBarTabView from being used outside of the intended view controller? I don't see any harm in giving more flexibility for how this view will be used.

anandrajeswaran: Yeah probably not much practical harm. The alternative is creating a trivial SideTabBarViewController wrapper around it for now, not trying to reconcile TabBarView yet and we can do that step later as you mentioned

MathieuKavalec: I will start working on the view controller in the near future (next week). 
By then, I'll be able to set SideBarTabView to internal and a temporary SideTabBarViewController won't be needed. 
I understand that this is open source and we need to be careful and intentional about what we expose but I don't think that this intermediate step is valuable, even short term.
We want to control how side tab bar and tab bar view are displayed and we can only do that with a full-fledge view controller that manages these views.

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,286 @@
...
+/// View for a vertical side tab bar that can be used for app navigation.
+/// Optimized for horizontal regular + vertical regular size class configuration. Prefer using TabBarView for other size class configurations.
+@objc(MSFSideTabBar)
+open class SideTabBar: UIView {
+    /// Delegate to handle user interactions in the side tab bar.
+    @objc public weak var delegate: SideTabBarDelegate?
+
+    /// The avatar view that displays above the top tab bar items.
+    /// The avatar view's size class should be AvatarSize.medium.
+    @objc open var avatarView: AvatarView? {
```
harrieshin, line 32: i am not sure we should expose avatarView if we are saying we going to restrict the size as "medium"

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,286 @@
...
+
+    private enum Section: Int {
+        case top
+        case bottom
+    }
+
+    private struct Constants {
+        static let maxTabCount: Int = 5
+        static let viewWidth: CGFloat = 62.0
+        static let avatarViewSize: CGFloat = 30.0
```
harrieshin, line 125: please use avater.size.meidum instead of defining its own constant

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -0,0 +1,286 @@
...
+        static let avatarViewTopPadding: CGFloat = 18.0
+        static let topStackViewTopPadding: CGFloat = 30.0
+        static let avatarViewTopStackViewPadding: CGFloat = 34.0
+        static let bottomStackViewBottomPadding: CGFloat = 10.0
+        static let topItemSpacing: CGFloat = 38.0
+        static let bottomItemSpacing: CGFloat = 28.0
+        static let itemHeight: CGFloat = 24.0
```
harrieshin, line 132: curious that you are passing in 28x28 images pdf but the item height is 24?

MathieuKavalec:
> side question; how does it work with iPad pointer?

Good question. I would need to download the new Xcode to test that.
This is definitely something we'll want to support, but if there's additional work, we'll look at that in a subsequent change.
No changes would need to be made in SideTabBar. The changes would have to go in TabBarItemView.

ios/FluentUI/Tab Bar/SideTabBar.swift
MathieuKavalec: This should be 28, my bad

MathieuKavalec: itemHeight will go away once issue#110 gets resolved. Currently, the TabBarItemViews don't have an intrinsic height based on their icon + label so we have to make this assumption about the internal size of the items.

ios/FluentUI/Tab Bar/SideTabBar.swift
harrieshin: I am also going to warn you that medium avatar size is going to change soon from 30 to 32.

MathieuKavalec: Fortunately, we can enforce that the avatar view's size is .medium since it's a readwrite property.
I like that we are giving the flexibility of passing in an AvatarView because there are many ways to configure the avatar view. Otherwise, we would need to create a mirror interface that pretty much matches the AvatarView public interface and then internally create an AvatarView based off of that. I think this would make the code more complex and less maintainable. With this approach, we are providing more flexibility, without giving too much flexibility since we are force the view to be an AvatarView and not any random UIView.

harrieshin:
now that you changed that tabbarview item sizing can you make sure you test those scenarios? landscape/portrait label/withoutlabel?

PR #109, opened 2020-07-06
Increase the range for the Calendar start/end year in Fluent. 

lugonzal:
Allow customizing the calendar's reference dates

### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Allow customizing the calendar's reference dates, so there's more flexibility for clients to decide what ranges are appropriate for their applications.

### Verification

Make sure date time pickers have a -100 + 100 range. And make sure all usages of these constants are run.

Before
Screen Shot 2020-07-06 at 12 03 45 PM
Screen Shot 2020-07-06 at 12 03 38 PM

After
Screen Shot 2020-07-06 at 12 02 33 PM
Screen Shot 2020-07-06 at 12 02 11 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/109)

### DIFF
--- a/ios/FluentUI/Core/CalendarConfiguration.swift
+++ b/ios/FluentUI/Core/CalendarConfiguration.swift
@@ -13,8 +13,8 @@ public typealias MSCalendarConfiguration = CalendarConfiguration
 @objc(MSFCalendarConfiguration)
 open class CalendarConfiguration: NSObject {
     private struct Constants {
-        static let startYearsAgo: Int = -100
-        static let endYearsInterval: Int = 200
+        static let startYearsAgo: Int = -3
+        static let endYearsInterval: Int = 10
     }
 
     // swiftlint:disable:next explicit_type_interface
@@ -22,8 +22,11 @@ open class CalendarConfiguration: NSObject {
 
     @objc open var firstWeekday: Int = Calendar.current.firstWeekday
 
-    let referenceStartDate: Date
-    let referenceEndDate: Date
+    /// By default, this is today minus 3 years. If overridden, make sure it's before (less than) the 'referenceEndDate'
+    @objc open var referenceStartDate: Date
+
+    /// By default, this is the default 'referenceStartDate' plus 10 years.  If overridden, make sure it's after (greater than) the 'referenceStartDate'
+    @objc open var referenceEndDate: Date
 
     @objc init(calendar: Calendar = .current) {
         // Compute a start date (January 1st on a year a default number of years ago)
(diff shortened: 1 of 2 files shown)


Saadnajmi:
Why isn't this just infinite and lazily loaded?

ios/FluentUI/Core/CalendarConfiguration.swift
```diff
@@ -13,10 +13,8 @@ public typealias MSCalendarConfiguration = CalendarConfiguration
 @objc(MSFCalendarConfiguration)
 open class CalendarConfiguration: NSObject {
     private struct Constants {
-        static let baseReferenceStartTimestamp: TimeInterval = 1420416000 // January 3 2015
-        static let baseReferenceEndTimestamp: TimeInterval = 1736035200 // January 5 2025
```
Saadnajmi, line 17: Were these constants just unused?

lugonzal:
> Why isn't this just infinite and lazily loaded?

I'm not sure why it was designed this way, I agree this should be infinite, but that'll take some big changes. This change unblocks a current client of the control in the short term.

ios/FluentUI/Core/CalendarConfiguration.swift
lugonzal: yes

lugonzal:
> > Why isn't this just infinite and lazily loaded?
> 
> I'm not sure why it was designed this way, I agree this should be infinite, but that'll take some big changes. This change unblocks a current client of the control in the short term.

changed the fix a bit, to be more flexible and prevent current clients to have a surprise new range by default.

ios/FluentUI/Core/CalendarConfiguration.swift
```diff
@@ -24,8 +22,11 @@ open class CalendarConfiguration: NSObject {
 
     @objc open var firstWeekday: Int = Calendar.current.firstWeekday
 
-    let referenceStartDate: Date
-    let referenceEndDate: Date
+    /// By default, this is today minus 3 years. If overridden, make sure it's before (less than) the 'referenceEndDate'
+    @objc open var referenceStartDate: Date
+
```
Saadnajmi, line 27: Could we add a guard in didset, or not because it had @objc ?

lugonzal: we could, but that would be assume clients will always set the start date first and then the end date... not sure if that's a correct assumption

PR #112, opened 2020-07-08
Rollback textPrimary and surfacePrimary to be variables

lugonzal:
Rollback textPrimary and surfacePrimary to be variables, to unblock clients that were overriding them, so they can upgrade to latest version.

### Platforms Impacted
- [X] iOS
- [ ] macOS

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/112)


lugonzal:
Rollback textPrimary and surfacePrimary to be variables, to unblock clients that were overriding them, so they can upgrade to latest version.

PR #113, opened 2020-07-09
Add arm64 macOS build arch for future Apple Silicon mac support

markavitale:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Add platform-specific versions of our publish_overrides.xcconfig file that include publish_overrides.xcconfig but then independently specify the extra archs necessary for each platform. On macOS we want to add arm64 (but not arm64e yet). On iOS we want to continue to build arm64e slices.

Update our Azure Pipeline yaml to respect the new override xcconfigs, and separately update our local script that mirrors the logic in the azure pipeline to match as well.

### Verification

Run `scripts/nuget_publish.sh` to simulate publishing the NuGet package, ensure it succeeds as expected. Ensure that the proper archs are built for each.

| image | image image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/113)

### DIFF
--- a/.ado/templates/fluentui-apple-publish-nuget-job.yml
+++ b/.ado/templates/fluentui-apple-publish-nuget-job.yml
@@ -18,7 +18,7 @@ jobs:
       xcode_actions: 'build'
       xcode_scheme: 'FluentUI-macOS'
       xcode_configuration: 'Debug'
-      xcode_extraArgs: '-xcconfig $(Build.Repository.LocalPath)/.ado/xcconfig/publish_overrides.xcconfig'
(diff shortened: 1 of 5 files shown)


markavitale:
Note: Since we're not building our NuGet publish with Xcode 12 yet, this shouldn't have any effect. I'll be making a followup change adding a separate beta publishing pipeline that uses the latest Xcode 12 beta to allow us to temporarily have side-by-side support for different xcode versions.

PR #117, opened 2020-07-10
Change navigationBar dark elevated color back to gray 900

lugonzal:
We need it to contrast with its non-elevated version to distinguish between a presented window and the window behind on iPhone. It sounds counterintuitive to select the same color, but iOS applies a dim to the window behind, which makes the current gray800 color look like gray900, which doesn't make a contrast.

### Platforms Impacted
- [X] iOS
- [ ] macOS

| Simulator Screen Shot - iPhone Xs Max - 2020-07-10 at 09 43 10 | Simulator Screen Shot - iPhone Xs Max - 2020-07-10 at 09 45 51 |

Issue that we're fixing:
Screen Shot 2020-07-08 at 3 31 52 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/117)

### DIFF
--- a/ios/FluentUI/Core/Colors.swift
+++ b/ios/FluentUI/Core/Colors.swift
@@ -473,11 +473,11 @@ public final class Colors: NSObject {
                 public static var textWeekend: UIColor = textOnAccent.withAlphaComponent(0.7)
             }
         }
-        public static var background: UIColor = NavigationBar.background
+        public static var background = UIColor(light: surfacePrimary, dark: surfaceTertiary)
     }
 
     public struct DateTimePicker {
-        public static var background: UIColor = NavigationBar.background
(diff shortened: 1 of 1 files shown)


harrieshin:
is this what Victor wanted? or we should just change the navigationbar background color?

lugonzal:
> is this what Victor wanted? or we should just change the navigationbar background color?

Ah yes! Early morning brain :)

harrieshin:
everything should be fine except i worry that calendar and datetimepicker background color would be incorrect now because its background was based on naivgationbar.background. we probably have to separate it?

PR #118, opened 2020-07-10
update version to 0.1.5

lugonzal:
### Platforms Impacted
- [X] iOS
- [X] macOS

### Description of changes

What's new:
* Change navigationBar dark elevated color back to gray 900 and fix pillButtonBar demo background color
* Add an Xcode 12 beta NuGet publish pipeline
* Add arm64 macOS build arch for future Apple Silicon mac support
* Allow overrides of textPrimary and surfacePrimary colors

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/118)

### DIFF
--- a/MicrosoftFluentUI.podspec
+++ b/MicrosoftFluentUI.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'MicrosoftFluentUI'
-  s.version          = '0.1.4'
+  s.version          = '0.1.5'
   s.summary          = 'Fluent UI is a set of reusable UI controls and tools'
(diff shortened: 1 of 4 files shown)


markavitale:
I'll note the "arm64 arch" stuff only applies to nuget builds used internally by Microsoft and doesn't apply directly to CocoaPods/Carthage consumers. There's some complexity there where we're still waiting on GitHub Actions/ADO to roll out support for building macOS Universal apps, but the infra should be ready for when they do roll that out.

PR #121, opened 2020-07-14
Fixing bug: when customEdgeInsets are set for the MSFButton and the b…

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Fixing bug: when customEdgeInsets are set for the MSFButton and the button is tapped, the update method resets the edge insets to the default value set for the style

### Verification

Changed the FluentUI Demo app to set the contentEdgeInsets of the button to zero when the style is borderless, tapped the button to make sure the insets would not be changed.

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/121)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
@@ -18,9 +18,9 @@ class ButtonDemoController: DemoController {
             let button = Button(style: style)
             button.setTitle("Button", for: .normal)
 
-			if style == .borderless {
-				button.contentEdgeInsets = .zero
-			}
+            if style == .borderless {
+                button.contentEdgeInsets = .zero
+            }
 
             let disabledButton = Button(style: style)
             disabledButton.isEnabled = false
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
@@ -18,10 +18,6 @@ class ButtonDemoController: DemoController {
             let button = Button(style: style)
             button.setTitle("Button", for: .normal)
 
-            if style == .borderless {
-                button.contentEdgeInsets = .zero
-            }
-
             let disabledButton = Button(style: style)
             disabledButton.isEnabled = false
             disabledButton.setTitle("Button", for: .normal)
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Controls/Button.swift
```diff
@@ -187,7 +195,9 @@ open class Button: UIButton {
 
         layer.borderWidth = style.hasBorders ? Constants.borderWidth : 0
 
-        contentEdgeInsets = style.contentEdgeInsets
+		if !isUsingCustomContentEdgeInsets {
```
harrieshin: wouldn't this be always false because it is initialization?

ios/FluentUI/Controls/Button.swift
```diff
@@ -121,6 +127,8 @@ open class Button: UIButton {
         }
     }
 
+	private var isUsingCustomContentEdgeInsets: Bool = false
```
harrieshin: your padding is off. please make sure you are using spaces instead of tab and also if you haven't install https://github.com/microsoft/fluentui-apple#swift-lint

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -18,6 +18,10 @@ class ButtonDemoController: DemoController {
             let button = Button(style: style)
             button.setTitle("Button", for: .normal)
 
+            if style == .borderless {
+                button.contentEdgeInsets = .zero
```
harrieshin, line 22: 2 question: has this been already talked with design to make sure zero inset is intentional? also, shouldn't this be handle in the enum ButtonStyle's contentEdgeInsets variable return value?

ios/FluentUI/Controls/Button.swift
rdeassis: It should all be fixed on my third commit to the branch....

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
rdeassis: So, the idea is that if the calling code is setting the button's contentEdgeInsets, it should not be overriden by the style...
This is an example of how the "Account Details" button was before and after I tapped it (while setting the contentEdgeInsets to zero):

Before:
before_tap

After:
after_tap

ios/FluentUI/Controls/Button.swift
rdeassis: It will unless the calling code sets the contentEdgeInsets property:

    open override var contentEdgeInsets: UIEdgeInsets {
        didSet {
            isUsingCustomContentEdgeInsets = true
        }
    }

harrieshin: but you need to initialize the button first before you can set the contentedgeinset of that instance?

PR #122, opened 2020-07-15
Add a SwiftUI test app for macOS

anandrajeswaran:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Light up a SwiftUI macOS demo app next to the existing macOS demo app.

**Why**
It's good to have some SwiftUI in the codebase so we can more easily reproduce errors if clients are including our controls in SwiftUI, but also SwiftUI make the concept of building a sidebar/content view setup easy and that's exactly what our test app is.

**Changes**
- Move test view controllers into a static lib so both AppKit and SwiftUI test apps can consume test view controllers without any duplication or extra steps needed by someone creating a control or adding tests
- Minor updates to the AppKit demo app to 1) avoid the need for a placeholder view controller (for now the window starts small and expands when one is selected) and 2) stop using the term "master" because we don't need to
- Fix a minor warning by not copying FluentUIUnitTest's Info.plist into its bundle
- Add a SwiftUI test app target in addition to the AppKit default one. The deployment target is 10.16 (Probably can move to 11.0 when beta 3 drops) so we can use SwiftUI Apps.
- Code actually needed for this is 1) a simple SwiftUI View to populate the sidebar and provide the selected content and 2) a trivial wrapper to allow SwiftUI to host an NSViewController

### Verification
- Sanity on 10.15 and Big Sur that the default test app still launches and switches tabs
- Delete NSUserDefaults to confirm first launch
- Not adding to CI yet since this target doesn't build with Xcode 11.

AppKit Test App
Screen Shot 2020-07-14 at 9 11 34 PM
SwiftUI Test App
Screen Shot 2020-07-14 at 9 09 55 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/122)


macos/FluentUITestViewControllers/TestViewControllers.swift
```diff
@@ -0,0 +1,22 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+public struct TestViewController: Identifiable {
+	public let id: String
```
harrieshin, line 9: title?

Saadnajmi:
I had a [branch](https://github.com/microsoft/fluentui-apple/compare/master...Saadnajmi:updated-test-app) where I was attempting to rewrite our original test app with NSSplitViewController so we could take advantage of the some of the Big Sur visual improvements. It might be worth merging some of those changes but also totally fine to leave that separate. 
Regardless, this is very cool and I can't wait to review it later today =)

macos/FluentUISwiftUITestApp/ContentView.swift
```diff
@@ -0,0 +1,29 @@
...
+			List {
+				ForEach(testViewControllers) { testViewController in
+					NavigationLink(destination: TestViewControllerWrappingView(type: testViewController.type)) {
+						Text(testViewController.id)
+					}
+				}
+			}
+			.listStyle(SidebarListStyle())
+		}
+		.frame(minWidth: 800, maxWidth: .infinity, minHeight: 500, maxHeight: .infinity)
```
Saadnajmi, line 21: Move to constants?

macos/FluentUISwiftUITestApp/TestViewControllerWrappingView.swift
```diff
@@ -0,0 +1,20 @@
...
+struct TestViewControllerWrappingView: NSViewControllerRepresentable {
+	
+	var type: NSViewController.Type
+	
+	func makeNSViewController(context: Context) -> some NSViewController {
+		return type.init(nibName: nil, bundle: nil)
+	}
+	
+	func updateNSViewController(_ nsViewController: NSViewControllerType, context: Context) {
+	}
```
Saadnajmi, line 19: I'm guessing this is an override and that's why it has to exist? If so, why isn't there the override keyword?

macos/FluentUITestViewControllers/TestViewControllers.swift
```diff
@@ -0,0 +1,22 @@
...
+public let testViewControllers = [TestViewController(id: "Avatar View",
+													 type: TestAvatarViewController.self),
+								  TestViewController(id: "Button",
+													 type: TestButtonViewController.self),
+								  TestViewController(id: "Date Picker",
+													 type: TestDatePickerController.self),
+								  TestViewController(id: "Link",
+													 type: TestLinkViewController.self),
+								  TestViewController(id: "Separator",
+													 type: TestSeparatorViewController.self)]
```
Saadnajmi, line 22: spacing looks off?

macos/xcode/FluentUI_swiftui_testapp.xcconfig
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+#include "FluentUI_common.xcconfig"
+
+MACOSX_DEPLOYMENT_TARGET = 10.16
```
Saadnajmi, line 8: Can the test app with SwiftUI run on 10.15?

macos/FluentUITestViewControllers/TestViewControllers.swift
anandrajeswaran: id is needed to conform to Identifiable, which is needed to be in a ForEach struct in SwiftUI. I was kinda being lazy by just purposing the name thing to id, but the more technically correct thing (maybe a more more clunky but more obvious) is probably to have a separate id property that we enforce as unique. Will do that in the next commit. Let me know what you think.

macos/FluentUITestViewControllers/TestViewControllers.swift
anandrajeswaran: github is bad with spaces/tabs, id (now 'title') is left-anchored to the same spot as 'type'. I might pick your brain about swift-format soon...

macos/xcode/FluentUI_swiftui_testapp.xcconfig
anandrajeswaran: Nope, because I chose to use the SwiftUI App construct. That's 10.16 (11.0) only, but it's what makes this so simple

anandrajeswaran:
> I had a [branch](https://github.com/microsoft/fluentui-apple/compare/master...Saadnajmi:updated-test-app) where I was attempting to rewrite our original test app with NSSplitViewController so we could take advantage of the some of the Big Sur visual improvements. It might be worth merging some of those changes but also totally fine to leave that separate.
> Regardless, this is very cool and I can't wait to review it later today =)

Nice, just took a quick look. I'd vote for leaving separate since the current PR only orthogonally affects the existing test app, but yeah definitely a good point that we should ultimately keep the behavior aligned and submitting these changes for split view in Big Sur in the SwiftUI test app temporarily violates that

Saadnajmi:
@anandrajeswaran Temporary violation probably fine. I would still want these changes in sooner since that branch has some problems with layout constraints. Also, the SwiftUII test app is prettier =)

PR #123, opened 2020-07-15
Add observation to UINavigationItem's accessoryView

harrieshin:
Add observation to UINavigationItem's accessoryView property so that shyheadercontroller should adjust its subview

### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Sometime client apps wants to change the accessoryView of the large title navigation controller, but the shyheaderView isn't updating its subview accordingly. Make sure that in the initialization process of shy header we add a listener to property changes.

### Verification
iPad/ iPhone large title : test collapsible searchbar scenarios.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/123)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -44,6 +44,8 @@ class ShyHeaderController: UIViewController {
         return maxHeight
     }
 
+    private var accessoryViewObservation: NSKeyValueObservation?
+
     private var navigationBarCenterObservation: NSKeyValueObservation?
     private var navigationBarHeightObservation: NSKeyValueObservation?
     private var navigationBarColorObservation: NSKeyValueObservation?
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -88,6 +90,10 @@ class ShyHeaderController: UIViewController {
         defer {
             contentScrollView = contentViewController.navigationItem.contentScrollView
         }
+
+        accessoryViewObservation = contentViewController.navigationItem.observe(\UINavigationItem.accessoryView) { [unowned self] item, _ in
```
anandrajeswaran, line 94: Curious - why "unowned" vs. "weak"?

harrieshin: i was just following the other pattern in this file. but my guess is that self doesn't have to be optional and it will always have value and not have check if it nil ....

PR #124, opened 2020-07-15
make sure demo app doesn't show "back" title next to backbarbuttonitem

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Minor fallout from commit 01c5c8d when we were deleting storyboard we forgot to programmatically set the "<" title to empty string.

### Verification

| Screen Shot 2020-07-14 at 10 52 24 PM|Screen Shot 2020-07-14 at 10 52 41 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/124)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -43,6 +43,7 @@ class DemoListViewController: UITableViewController {
             subtitle: FluentUIFramework.bundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
         )
         navigationItem.titleView = titleView
+        navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
 
         tableView.backgroundColor = Colors.Table.background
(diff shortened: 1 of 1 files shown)


ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -43,6 +43,7 @@ class DemoListViewController: UITableViewController {
             subtitle: FluentUIFramework.bundle.object(forInfoDictionaryKey: "CFBundleShortVersionString") as? String
         )
         navigationItem.titleView = titleView
+        navigationItem.backBarButtonItem = UIBarButtonItem(title: "", style: .plain, target: nil, action: nil)
```
anandrajeswaran, line 46: Thanks for fixing! Possibly worth adding a comment since it probably isn't obvious to most why we'd be adding an item with no title and why that turns out ok for VoiceOver?

PR #129, opened 2020-07-22
move color definition of the components out of Colors.swift file

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Move all the component color definition to the swift file that component is implemented rather than keeping in Colors.swift so that when we start breaking down the library in to submodules  it is easier to compile.
Defining each color as public extension of Colors does increase the binary size a bit for now (15K bytes on device build) but this reduces the breakage of existing clients of fluent library.

### Verification

launch demo iOS app

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/129)


ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -45,6 +45,16 @@ public protocol BadgeFieldDelegate: AnyObject {
     @objc optional func badgeFieldShouldReturn(_ badgeField: BadgeField) -> Bool
 }
 
+// MARK: - BadgeField Colors
+
+public extension Colors {
```
Saadnajmi, line 50: Do these need to be public?

harrieshin: it was public before so yes it has to be public for client apps to access these colors

PR #138, opened 2020-07-27
icon support for fluent buttons

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added support for icons in the fluentUI buttons.
The icons are only supported for the primaryFilled, primaryOutline and secondaryOutline.
The icons must be template images since the Button object will set the tint color for the icons. The tint color matches the color applied for the button's label.
The icons should ideally be 24x24 for primary and 20x20 for secondary but this is not enforced. The comment for the icon's property indicates this.  

### Verification

Tested in RTL, light/dark mode, large text sizes.
Added test code that toggles the image on a button to make sure that the layout updates correctly.
Tested with custom content edge insets to make sure that this property is respected even when an icon is set.

Before:
Screen Shot 2020-07-27 at 4 16 27 PM

After:
Screen Shot 2020-07-29 at 11 25 15 AM
Screen Shot 2020-07-29 at 11 25 27 AM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/138)


ios/FluentUI/Controls/Button.swift
```diff
@@ -101,6 +112,17 @@ open class Button: UIButton {
...
+    /// The button's image.
+    /// The image must be a template image.
+    /// For ButtonStyle.primaryFilled and ButtonStyle.primaryOutline, the image must be 24x24.
+    /// For ButtonStyle.secondaryOutline, the image must be 20x20.
+    /// For other styles, the image is not displayed.
+    @objc open var image: UIImage? {
+        didSet {
+            update()
+        }
+    }
```
Saadnajmi, line 124: Can we add a guard to make sure it's a template image?

ios/FluentUI/Controls/Button.swift
```diff
@@ -212,6 +281,47 @@ open class Button: UIButton {
...
+    private func disabledTitleAndImageColor(for window: UIWindow) -> UIColor {
+        return style == .primaryFilled ? normalTitleAndImageColor(for: window) : Colors.primaryTint20(for: window)
+    }
+
+    private var normalImageTintColor: UIColor?
+    private var highlightedImageTintColor: UIColor?
+    private var disabledImageTintColor: UIColor?
+
+    private var isUsingCustomContentEdgeInsets: Bool = false
+    private var isAdjustingCustomContentEdgeInsetsForImage: Bool = false
```
Saadnajmi, line 301: Why do we need this bool? I might be dumb but it almost feels like a semaphore which I'm not sure why we need?

Saadnajmi:
Totally going to steal some of the logic for macOS btw =)

ios/FluentUI/Controls/Button.swift
MathieuKavalec: Actually, this works even if the image is not a template image. There isn't really a way to know either as far as I can tell. I'll remove the comment. It should be expected that the image's tint color will be modified so the right kind of image must be provided.

ios/FluentUI/Controls/Button.swift
MathieuKavalec: Without this, we end up in infinite recursion.
The contentEdgeInsets can be set to a custom value. We implement the didSet property observer of contentEdgeInsets. In the didSet, we need to update the value of contentEdgeInsets if an image is already set so that we can add some padding between the image and the text. The contentEdgeInsets value will be reflected in the UI because we use titleEdgeInsets to offset the text and make it look like the original contentEdgeInsets values are respected. If we modify contentEdgeInsets in its own didSet callback, we will end up calling again into the didSet callback and it will never end.

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/Button/Placeholder_24.imageset/Contents.json
```diff
@@ -0,0 +1,16 @@
+{
+  "images" : [
+    {
+      "idiom" : "universal",
+      "filename" : "placeholder_24.pdf"
```
harrieshin, line 5: if this is coming from fluent icon repo, do you mind keeping the same name so we can easily update it later?

ios/FluentUI/Controls/Button.swift
```diff
@@ -120,26 +141,35 @@ open class Button: UIButton {
     open override var contentEdgeInsets: UIEdgeInsets {
         didSet {
             isUsingCustomContentEdgeInsets = true
+
+            if !isAdjustingCustomContentEdgeInsetsForImage && image(for: .normal) != nil {
```
harrieshin, line 145: should we not do image inset when user is directly setting content insets of the image and title ? isUsingCustomContentEdgeInsets

ios/FluentUI/Controls/Button.swift
```diff
@@ -120,26 +141,35 @@ open class Button: UIButton {
...
-        size.height = ceil(max(size.height, style.minTitleLabelHeight) + contentEdgeInsets.top + contentEdgeInsets.bottom)
+        var size = super.intrinsicContentSize
+        let labelHeight = titleLabel?.intrinsicContentSize.height ?? 0.0
+        size.height = ceil(max(labelHeight, style.minTitleLabelHeight)) + contentEdgeInsets.top + contentEdgeInsets.bottom
+
         return size
     }
 
-    /// if value is 0.0, CGFloat.greatestFiniteMagnitude is used to calculate the width of the `titleLabel` in `intrinsicContentSize`
-    private var proposedTitleLabelWidth: CGFloat = 0.0 {
```
harrieshin, line 134: if you are deleting this, please make sure you aren't regressing the fix commit 965e4cf in large text mode

ios/FluentUI/Controls/Button.swift
```diff
@@ -101,6 +112,17 @@ open class Button: UIButton {
         }
     }
 
+    /// The button's image.
+    /// The image must be a template image.
+    /// For ButtonStyle.primaryFilled and ButtonStyle.primaryOutline, the image must be 24x24.
+    /// For ButtonStyle.secondaryOutline, the image must be 20x20.
+    /// For other styles, the image is not displayed.
+    @objc open var image: UIImage? {
```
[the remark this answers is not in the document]
MathieuKavalec: I was heavily debating that with myself. The problem with using image(for: UIControl.State) is that we will ignore the state param. We generate the images for each state from this new image property. We also control the position of the image and add a few more restrictions like the image only being supported for some styles. Considering this, I felt that it made sense to have a custom image property.

ios/FluentUI/Controls/Button.swift
```diff
@@ -212,6 +281,47 @@ open class Button: UIButton {
...
+
+    private func highlightedTitleAndImageColor(for window: UIWindow) -> UIColor {
+        return style == .primaryFilled ? normalTitleAndImageColor(for: window) : Colors.primaryTint20(for: window)
+    }
+
+    private func disabledTitleAndImageColor(for window: UIWindow) -> UIColor {
+        return style == .primaryFilled ? normalTitleAndImageColor(for: window) : Colors.primaryTint20(for: window)
+    }
+
+    private var normalImageTintColor: UIColor?
```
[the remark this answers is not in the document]
MathieuKavalec: Are you talking about the disabledTitleAndImageColor or normalImageTintColor?
**disabledTitleAndImageColor** can't be cached because there are different variants based on the window.
**normalImageTintColor** is cached as a significant optimization to avoid recreating the tinted image each time the control updates. We only recreate the tinted image when the tint changes. We don't cache the tinted images because that would take up a lot of memory and the tint color depends on the window so it could change.

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/Button/Placeholder_24.imageset/Contents.json
MathieuKavalec: That is the name from the fluent icon repo.

ios/FluentUI/Controls/Button.swift
MathieuKavalec: Custom content edge insets are still supported with this logic. We also don't have a choice to use that. It's impossible to add padding between the image and label without modifying the content edge insets (image edge insets and label edge insets alone cannot solve this).

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/Button/Placeholder_24.imageset/Contents.json
MathieuKavalec: This is the name that's being used in figma. Is that the right name? I'm not sure what's the fluent icon repo.

MathieuKavalec: Resolving assuming that Fluent icons == what's on Figma

ios/FluentUI/Controls/Button.swift
MathieuKavalec: Good catch. I thought this was working but it only worked on large enough screens. On iPhone, the button was not sizing correctly. I updated the sizing logic to fix this issues and also make it work when an icon is set.

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/Button/Placeholder_24.imageset/Contents.json
MathieuKavalec: Ok, I copied over the resources directly from the fluent icon repo with the same name. Thanks.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -23,13 +23,42 @@ class ButtonDemoController: DemoController {
...
+                iconButton.image = image
+
+                let disabledIconButton = Button(style: style)
+                disabledIconButton.isEnabled = false
+                disabledIconButton.setTitle("Button", for: .normal)
+                disabledIconButton.image = image
+
+                addRow(items: [iconButton, disabledIconButton], itemSpacing: 20)
+
+                let iconOnlyButton = Button(style: style)
```
rdeassis, line 39: I don't see iconOnlyButton and disabledIconOnlyButton on the screenshots. Were they added after you took the screenshots?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -50,4 +79,15 @@ extension ButtonStyle {
             return "Borderless"
         }
     }
+
+    var image: UIImage? {
+        switch self {
+        case .primaryFilled, .primaryOutline:
+            return UIImage(named: "ic_fluent_circle_24_regular")!
```
harrieshin, line 86: totally find if the imageset name is more user "friendly" i only cared about the actualy pdf file name

ios/FluentUI/Controls/Button.swift
```diff
@@ -120,26 +141,47 @@ open class Button: UIButton {
...
-            if proposedTitleLabelWidth != oldValue {
-                invalidateIntrinsicContentSize()
+    open override func imageRect(forContentRect contentRect: CGRect) -> CGRect {
+        var rect = super.imageRect(forContentRect: contentRect)
+
+        if let image = image {
+            let imageHeight = image.size.height
+
+            if imageHeight > rect.size.height {
+                rect.origin.y -= round((imageHeight - rect.size.height) / 2.0)
```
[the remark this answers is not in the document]
MathieuKavalec: Not sure I understand your question. The image's height and origin should not be zero.
I will add a comment in code to clarify this logic:
// If the entire image doesn't fit in the default rect, increase the rect's height
// to fit the entire image and reposition the origin to keep the image centered.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
MathieuKavalec: I updated the screenshots.

PR #150, opened 2020-07-30
Button title and image support

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

We support setting an image, and a title inside our button, but not at the same time. Let's change that.

- Remove the intrinsicContentSize override
- Create a new FluentButtonCell class that is the default for our Button
- Override `drawingRect(forBounds:)`, `titleRect(forBounds:)`, and `image(forBounds:)`
- Update our test app to show the buttons in a few styles

Screen Shot 2020-08-28 at 7 54 10 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/150)


macos/FluentUI/Button.swift
```diff
@@ -53,6 +66,7 @@ open class Button: NSButton {
 		// Do common initialization work
 		isBordered = false
 		wantsLayer = true
+		layer?.contentsScale = window?.backingScaleFactor ?? 1.0
```
Saadnajmi, line 69: Not sure if I * need * to do this (it worked fine before), but we do it in Avatar View and DatePicker

macos/FluentUI/Button.swift
```diff
@@ -237,6 +261,109 @@ open class Button: NSButton {
...
+			case .imageTrailing:
+				x = isRTL ? Button.horizontalEdgePadding + imageSize.width + Button.titleAndImageRectInterspacing : Button.horizontalEdgePadding
+				y = Button.verticalEdgePadding
+				width = titleSize.width
+				height = titleSize.height
+				break
+			default:
+				// We explicitly do not support the other variants of imagePosition
+				break
+			}
```
Saadnajmi, line 348: Could also put a preconditionFailure here or something. As this flow goes, it'll follow whatever super returns, which'll probably always look bad

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -31,16 +31,33 @@ class TestButtonViewController: NSViewController {
...
 		}
 
 		let disabledButtonsWithImage = buttonsWithImage().map { button -> NSButton in
 			button.isEnabled = false
 			return button
 		}
+		
+		let isRTL = NSApp.userInterfaceLayoutDirection == .rightToLeft
+		let leadingArrowImage = isRTL ? NSImage(named: NSImage.goRightTemplateName)! : NSImage(named: NSImage.goLeftTemplateName)!
+		let trailingArrowImage = isRTL ? NSImage(named: NSImage.goLeftTemplateName)! : NSImage(named: NSImage.goRightTemplateName)!
```
Saadnajmi, line 47: We should replace this with the fluent arrow icon whenever we get that into our repo

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -52,6 +69,8 @@ class TestButtonViewController: NSViewController {
 		gridView.addColumn(with: disabledButtonsWithTitle)
 		gridView.addColumn(with: buttonsWithImage())
 		gridView.addColumn(with: disabledButtonsWithImage)
+		gridView.addColumn(with: buttonsWithTitleAndImage())
+		gridView.addColumn(with: disabledButtonsWithTitleAndImage)
```
Saadnajmi, line 73: This is starting to get crowded. I could rewrite this so there's only 1 column of disabled buttons, instead of 3.

macos/FluentUI/Button.swift
```diff
@@ -36,6 +36,19 @@ open class Button: NSButton {
 		imagePosition = .imageOnly
 		initialize(title: nil, image: image, style: style)
     }
+	
+	
+	/// Initializes a Fluent UI Button with a title,  image,  imagePosition, and style
+	/// - Parameters:
+	///   - title: String displayed in the button
+	///   - image: The NSImage to diplay in the button
+	///   - imagePosition: The position of the image, relative to the title
```
Saadnajmi, line 45: Should probably specify here we only explicitly support 4 of the enum values

macos/FluentUI/Button.swift
harrieshin: do we need noimage and imageonly enum? isn't that kind of obvious looking at if there is an image or if there is a title?

macos/FluentUI/Button.swift
harrieshin: i prefer not having default so ppl dont make mistakes

macos/FluentUI/Button.swift
Saadnajmi: User could set an empty image, and also the imagePosition to .imageOnly. I figured we had to support that, as that is how images in buttons work on macOS in my experience

macos/FluentUI/Button.swift
Saadnajmi: That would mean implementing the imagePositions like "imageAbove, below, overlaps"... which feel weird. Though I guess I've already done the math and can..

macos/FluentUITestViewControllers/TestButtonViewController.swift
Saadnajmi: Thanks!

macos/FluentUI/Button.swift
```diff
@@ -285,20 +286,49 @@ open class Button: NSButton {
...
+				width = imageSize.width
+				height = imageSize.height
+				break
+			case .imageAbove:
+				x = (rect.width - imageSize.width) / 2
 				y = Button.verticalEdgePadding
-				width = rect.width - Button.titleAndImageRectInterspacing - titleSize.width
+				width = imageSize.width
+				height = imageSize.height
+				break
```
Saadnajmi: We now can handle ALL the enum values.. for whoever ever wants to do this 
Screen Shot 2020-07-30 at 3 33 16 PM

macos/FluentUI/Button.swift
Saadnajmi: ... I went for it. This also adds an "@unknown default" case bc apparently swift wants that

macos/FluentUI/Button.swift
Saadnajmi: JK. We do all of them 😭

harrieshin:
even though mac doesn't have official toolkit design, i would expect it would be similar and image color will be the same as text color if the image is passed in as render-as-template?

Saadnajmi:
> even though mac doesn't have official toolkit design, i would expect it would be similar and image color will be the same as text color if the image is passed in as render-as-template?

Good catch. There isn't a "tintColor" property for NSImage, but we can something like this to set the tint https://stackoverflow.com/questions/45028530/set-image-color-of-a-template-image

macos/FluentUI/Button.swift
```diff
@@ -253,3 +479,20 @@ open class Button: NSButton {
...
+
+		let imageRect = NSRect(origin: NSZeroPoint, size: image.size)
+		imageRect.fill(using: .sourceAtop)
+
+		image.unlockFocus()
+		image.isTemplate = false
+
+		return image
+	}
+}
```
Saadnajmi, line 498: @harrieshin With this, we have tinting of NSImage's on the button working if it is a template image.
Though... maybe we should move this out to it's own file? `NSImage+TintColor.swift` ?

Screen Shot 2020-08-28 at 7 54 10 PM

macos/FluentUI/Button.swift
harrieshin: then we should have a convenience method of init that client just needs to pass in title and not have to pass in empty nsimage like NSButton initialzie. Also, it seems weird that we are foceing imageposition in intialization. i assume nsbutton, client can change imageposition even after it is created so we should be flexible here too?

macos/FluentUI/Button.swift
```diff
@@ -53,24 +66,39 @@ open class Button: NSButton {
...
 	override public var image: NSImage? {
 		willSet {
 			guard wantsLayer == true else {
 				preconditionFailure("wantsLayer must be set so that the image is rendered on the layer")
 			}
 		}
+		didSet {
+			if let image = image {
+				if image.isTemplate {
+					self.image = image.tint(color: self.textColor)
```
harrieshin, line 98: do we really need this? we can't utilize contentTintColor ?

macos/FluentUI/Button.swift
Saadnajmi: Line 25 has an init method with no image 
`@objc public init(title: String, style: ButtonStyle = .primaryFilled)`

A common bug I always ran into with NSButton is "Why is my image on top of my text?? or "where did my image go?" after I set an image but didn't set imagePosition. But since that is the expected API, I can remove it and whatever apple defaults to, we can follow

harrieshin: my bad i still get confused with github coderview gui.

macos/FluentUI/Button.swift
harrieshin: the danger with this approach in this scenario is that techincially primary color is different between dark and light mode in iOS and I am going to assume that is true for mac as well. do your image change with system settings change?

macos/FluentUI/Button.swift
Saadnajmi: Right, so NSButton defaults to imagePosition .imageOnly after you set an image. Because of the order of operations in our init, this means that our initializer with both a title and an image only shows the image. 

That + my desire to initialize buttons in one line in the test app view controller is why I went with adding it as a param.  I would like to force a new default (.imageLeading) for this one initializer, but I can remove the imagePosition param and rewrite the test app to match

macos/FluentUI/Button.swift
Saadnajmi: Well, I didn't know about that property :D 
It's only 10.14 up though, so we'll need to have a fallback, for which textColor is fine?

Saadnajmi: So this suggestion was great. I rewrote the text to also get tinted with contentTintColor, with an appropriate 10.13 fallback. Thank you :D 
Images will not be tinted on 10.13

macos/FluentUI/Button.swift
Saadnajmi: Good catch. It did not retint with dark mode change. I changed to use contentTintColor, and now it does :D

harrieshin (approved):

PR #164, opened 2020-08-01
Implement MSFTextStyle for macOS

Saadnajmi:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

At WWDC 2020, Apple revealed that they are creating default text styles for macOS. 
New HIG Guidelines [here](https://developer.apple.com/design/human-interface-guidelines/macos/visual-design/typography/), and developer documentation [here](https://developer.apple.com/documentation/appkit/nsfont/textstyle)

Let's add a new enum MSFTextStyle (similar to the one we already have on iOS) for macOS, which just calls the Apple API for macOS 11+, and mimics the style for older OS's. 
Let's also add textStyle as a property for existing controls, and as a param to their initializers.

Apple's WWDC Talk 

Screen Shot 2020-06-29 at 2 17 08 PM

Our new Text Styles

Screen Shot 2020-08-01 at 1 37 48 AM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/164)

### DIFF
--- a/macos/FluentUI/Label.swift
+++ b/macos/FluentUI/Label.swift
@@ -1,0 +1,99 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+@objc(MSFTextStyle)
+public enum TextStyle: Int, CaseIterable {
+    case largeTitle
+    case title1
+    case title2
+    case title3
+    case headline
+    case body
+    case callout
+    case subheadline
+    case footnote
+    case caption1
+    case caption2
+	
+	public func font() -> NSFont
+	{
+		switch self {
+		case .largeTitle:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .largeTitle)
+			} else {
+				return NSFont.systemFont(ofSize: 26, weight: .regular)
+			}
+		case .title1:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .title1)
+			} else {
+				return NSFont.systemFont(ofSize: 22, weight: .regular)
+			}
+		case .title2:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .title2)
+			} else {
+				return NSFont.systemFont(ofSize: 17, weight: .regular)
+			}
+		case .title3:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .title3)
+			} else {
+				return NSFont.systemFont(ofSize: 15, weight: .regular)
+			}
+		case .headline:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .headline)
+			} else {
+				return NSFont.systemFont(ofSize: 13, weight: .black)
+			}
+		case .body:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .body)
+			} else {
+				return NSFont.systemFont(ofSize: 13, weight: .regular)
+			}
+		case .callout:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .callout)
+			} else {
+				return NSFont.systemFont(ofSize: 12, weight: .regular)
+			}
+		case .subheadline:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .subheadline)
+			} else {
+				return NSFont.systemFont(ofSize: 11, weight: .regular)
+			}
+		case .footnote:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .footnote)
+			} else {
+				return NSFont.systemFont(ofSize: 10, weight: .regular)
+			}
+		case .caption1:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .caption1)
+			} else {
+				return NSFont.systemFont(ofSize:10, weight: .regular)
+			}
+		case .caption2:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .caption2)
+			} else {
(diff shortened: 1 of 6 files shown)


Saadnajmi:
Well that's fun, CI failed. Perhaps because this depends on Xcode 12 beta because I reference macOS 11.0 API's?

macos/FluentUI/Label.swift
```diff
@@ -0,0 +1,99 @@
...
+			}
+		}
+	}
+}
+
+public func Label(withString string: String, forTextStyle TextStyle: TextStyle) -> NSTextField {
+	let label = NSTextField(labelWithString: string)
+	label.font = TextStyle.font()
+	return label
+}
```
Saadnajmi, line 99: This is a public function because I saw no value in a subclass of NSTextField.. perhaps we could add textStyle as a property and have it adapt if we did go that route? Or it might help for FURN.

Saadnajmi:
Worth noting that these are the styles in FURN, and we should map these eventually...

https://developer.microsoft.com/en-us/fluentui#/controls/crossplatform/text

macos/FluentUI/Label.swift
```diff
@@ -0,0 +1,99 @@
...
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .title3)
+			} else {
+				return NSFont.systemFont(ofSize: 15, weight: .regular)
+			}
+		case .headline:
+			if #available(macOS 11.0, *) {
+				return NSFont.preferredFont(forTextStyle: .headline)
+			} else {
+				return NSFont.systemFont(ofSize: 13, weight: .black)
```
HeyImChris, line 53: I believe `headline` is bold not black

macos/FluentUI/Link.swift
```diff
@@ -110,7 +121,10 @@ open class Link: NSButton {
 	}
 	
 	private func updateTitle() {
-		let titleAttributes = (showsUnderlineWhileMouseInside && mouseEntered) ? underlinedlinkAttributes: linkAttributes
+		
+		let shouldUnderline = showsUnderlineWhileMouseInside && mouseEntered
```
HeyImChris, line 125: Why split this out?

macos/FluentUI/Link.swift
```diff
@@ -110,7 +121,10 @@ open class Link: NSButton {
...
+		
+		let titleAttributes = (shouldUnderline) ? underlinedlinkAttributes: linkAttributes
```
HeyImChris, line 127: `(shouldUnderline)` doesn't need parentheses

macos/FluentUITestViewControllers/TestLabelViewController.swift
```diff
@@ -0,0 +1,87 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+import FluentUI
+
+
```
HeyImChris, line 9: extra newline

HeyImChris: I see double newlines a lot in this PR. Super minor but I don't normally see that, but maybe it's a swift thing I don't know about?

macos/FluentUITestViewControllers/TestLabelViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+		)
+
+		containerView.addView(gridView, in: .top)
+
+		view = containerView
+	}
+	
+	func string(ForTextStyle textStyle: TextStyle) -> String {
+		switch textStyle {
+
```
HeyImChris, line 54: extra newline (we don't have one after the switch above so let's be consistent)

macos/FluentUITestViewControllers/TestLabelViewController.swift
```diff
@@ -0,0 +1,87 @@
...
+	override func loadView() {
+
+
+		let labels = TextStyle.allCases.map {textStyle -> NSTextField in
+			return Label(withString: string(ForTextStyle: textStyle), forTextStyle: textStyle)
+		}
+		
+		let buttons = TextStyle.allCases.map {textStyle -> Button in
+			return Button(title: string(ForTextStyle: textStyle), textStyle: textStyle, style: .primaryFilled)
```
harrieshin, line 20: i just feel that it is odd that we show button and link in label test code

macos/FluentUITestViewControllers/TestLabelViewController.swift
Saadnajmi: No, that's a Saad thing. I should just addthe SwiftLint build phase and be done with it...

macos/FluentUITestViewControllers/TestLabelViewController.swift
Saadnajmi: > I'm not sure I see the value in providing mappings for macOS text styles. People can and should just use them directly. As you mention, we should map to FURN (and also fluentui-ios). Adding these temporary ones in the interim seems like it's only indirection rather than values and then will make mapping to fluentui styles a behavioral and breaking change. Why not wait or do a best guess translation to fluent text styles to start with?

Only macOS Big Sur users and up can use these text syles. I wanted to add a style for the platforms we do support. Furthermore, I'm pretty sure that iOS does the same/similar thing of mimicking the system API. That being sad, If we don't want to introduce a breaking change later and just want to map to the FURN styles now, that's fine. I wanted to start the conversaton with keeping the system text styles in mind first, and this was the easiest way to do that.

macos/FluentUITestViewControllers/TestLabelViewController.swift
Saadnajmi: Yeah, it was more of a "TextStyle" test view controller, than a label view controller...

macos/FluentUI/Link.swift
Saadnajmi: It just felt slightly cleaner to me, no strong reason.

Saadnajmi:
Closing as this in't needed at this time

PR #168, opened 2020-08-04
Face stack

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Implement face stack/pile.
The face stack/pile displays several AvatarViews in a gouped UI.
Also added support for modifying the color of the AvatarView borders.

### Verification

- Tested on iPhone and iPad compact/regular.
- Tested VoiceOver accessibility. Make sure that the accessibility behavior of the face stack is identical to Apple's behavior in Messages.
- Make sure that the face stack works when the number of avatars changes (tested 0 to 15).

Screen Shot 2020-08-04 at 4 26 54 PM
Screen Shot 2020-08-04 at 4 26 57 PM
Screen Shot 2020-08-04 at 4 27 11 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/168)


ios/FluentUI/People Picker/Avatar.swift
```diff
@@ -36,6 +40,8 @@ open class AvatarData: NSObject, Avatar {
 
     @objc public var customBorderImage: UIImage?
 
+    @objc public var borderColor: UIColor?
```
harrieshin, line 43: not sure. I would have imagine the UI is more like the current Office iOS collab menu, which the ring is around the avatar and it could animate around if people is looking at the document and static or gray off if users isn't looking at the document.

ios/FluentUI/People Picker/AvatarStack.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    private func updateAvatars() {
+        removeAllSubviews()
+
+        avatarViews.removeAll()
+        var constraints: [NSLayoutConstraint] = []
+        var previousAvatarView: AvatarView?
+        let avatarOverlap = self.avatarOverlap()
+
+        for avatar in avatars {
```
harrieshin, line 81: does this logic work with RTL? shouldn't the stack go other way around?

ios/FluentUI/People Picker/Avatar.swift
harrieshin: should the bordercolor just be the same as avatar initial background color logic?

ios/FluentUI/People Picker/AvatarStack.swift
```diff
@@ -0,0 +1,160 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFAvatarStack)
+open class AvatarStack: UIView {
+    /// Array of avatars to display in the avatar stack.
+    @objc open var avatars: [Avatar] {
```
harrieshin, line 11: i feel like we need to open the door to let clients to have "+" button at the end or show "+10" somehow. it is pretty limited with just avatar array.

harrieshin: maybe we need to support overflow style of avatar shown in toolkit figma?

harrieshin: Screen Shot 2020-08-03 at 9 11 15 PM

ios/FluentUI/People Picker/Avatar.swift
MathieuKavalec: The ring animation for presence is not a requirement so let's ignore that for now.
This implementation is similar to what we have for the Office iOS collab menu. To meet the requirements of both Office iOS and Teams, I'll make this change:
Instead of a "border color", I will rename this property to simply "color", representing that this is the color of the avatar. If this color is not specified, by default, the initials color will be used and the white/black border will be used. If the color is set, this color will override the default colors for both the border and the initials. 
This is an important requirement. Having worked a lot in the collab space, we often want to override the color that was generated from the initials. In Office, presence color comes from the collab service and I think this will also be the case for Fluid and Teams.

ios/FluentUI/People Picker/AvatarStack.swift
MathieuKavalec: It mostly does, I did find a small issue. I'll send an update.

ios/FluentUI/People Picker/Avatar.swift
```diff
@@ -16,10 +16,16 @@ public protocol Avatar {
     var secondaryText: String { get }
     var image: UIImage? { get }
 
+    /// The color that represents this avatar.
+    /// This color will override the initials view's background color.
```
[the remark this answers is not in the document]
MathieuKavalec: We don't need to worry about _avatarBackgroundColor_. The API is actually logic as it is, let me explain.
If you call @objc public func setup(avatar: Avatar?), it will setup the AvatarView with the information from the object that conforms to the Avatar protocol. This could include the avatar's color.
_avatarBackgroundColor_  on the other hand is a property that can be used to manually set the avatar's background color. _avatarBackgroundColor_ is more of an override property.

You could call setup(avatar) and this will set the AvatarView's _avatarBackgroundColor_ to the avatar's color. Then, if you want to override this, you can use the _avatarBackgroundColor_ property. Works out great!

ios/FluentUI/People Picker/AvatarGroupView.swift
```diff
@@ -0,0 +1,256 @@
...
+    private func updateLayerMask() {
+        if style != .stack {
+            return
+        }
+
+        if avatarViews.count <= 1 {
+            for avatarView in avatarViews {
+                avatarView.layer.mask = nil
+            }
+
```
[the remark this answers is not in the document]
MathieuKavalec: Can you please elaborate? The formatting looks good to me and and the linter is not putting up any warnings.

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -526,3 +578,86 @@ open class AvatarView: UIView {
         set { }
     }
 }
+
+// MARK: - OverflowAvatarView
+
+class OverflowAvatarView: AvatarView {
+    @objc public init(overflowCount: UInt, avatarSize: AvatarSize, withBorder hasBorder: Bool = false) {
```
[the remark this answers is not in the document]
MathieuKavalec: No need, the initializer for this class guarantees that presence is set to .none

PR #171, opened 2020-08-04
Add default image to AvatarView

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add support for showing unauthenticated state image for avatar
Instead of showing "#" when we can't calculate the initials, show default persona filled icon.

### Verification

|dark|light|
|---|---|
|Simulator Screen Shot - iPhone 11 Pro Max - 2020-08-04 at 14 17 44| Simulator Screen Shot - iPhone 11 Pro Max - 2020-08-04 at 14 17 49|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/171)


ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -493,8 +524,42 @@ open class AvatarView: UIView {
...
+                if let window = window {
+                    imageView.backgroundColor = Colors.primary(for: window)
+                }
+                imageView.tintColor = Colors.iconOnAccent
+            }
+            isPersonIconShown = true
+        }
+    }
+
+    private func shouldShowSignOutImage() -> Bool {
```
MathieuKavalec, line 550: Is this really the way that we want to derive the signed out state?

harrieshin: do you think it should be more explicit?

MathieuKavalec: I think so but I can't think of an elegant way to do this.
Maybe create a setup method that's explicitly going to setup the AvatarView for the sign out case?

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -493,8 +524,42 @@ open class AvatarView: UIView {
         }
     }
 
+    private func updateImageViewWithPersonIcon(isSignedOut: Bool = false) {
```
markavitale, line 527: Can we add some documentation to these functions? I'm not quite sure what to expect behavior-wise, and how the input impacts what actually happens.

ios/FluentUI/People Picker/AvatarView.swift
MathieuKavalec: It might make more sense to move this behavior outside of AvatarView.
It is possible to create an avatar view with any image.
I do think we can keep this logic in fluent UI to make it easier to create an AvatarView with the signed out image. You could subclass AvatarView to have a version that displays the sign out image. Subclassing is not the only way to do this. You could expose a static function that creates an AvatarView configured for the signed out state.

ios/FluentUI/People Picker/InitialsView.swift
```diff
@@ -39,11 +39,6 @@ class InitialsView: UIView {
             initials = String(secondaryText.prefix(1))
         }
 
-        // Fallback to `#` otherwise
```
MathieuKavalec, line 42: Do we really want to get rid of this? Can InitialsView be used without AvatarView? If so, we might want to keep this logic, or not, depends.

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -109,6 +126,9 @@ public extension Colors {
         // Should use physical color because this text is shown on physical avatar background
         public static var text: UIColor = textOnAccent
         public static var border = UIColor(light: .white, dark: gray900, darkElevated: gray800)
+
+        static let signedOutImageBackground = UIColor(light: gray50, dark: gray800)
+        static let signedOutImage = UIColor(light: gray500, dark: gray200)
```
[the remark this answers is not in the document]
harrieshin: I can add more comments but it is under a color extension. adding color at the end would be Colors.Avatar.signedOutImageColor. seem bit repetitive?

ios/FluentUI/People Picker/InitialsView.swift
harrieshin: while this view could be used somewhere else, (ignoring the fact that background color is using Colors.avatarBackgroundColors) does it make sense that the character # used for other labels and stuff?

ios/FluentUI/People Picker/AvatarView.swift
harrieshin: you don't want to necessarily create a new avatar view because you are signed out. Header for example, you want to use the same view and swap out images depending on the authentication state. Avatar signed out state is part of the toolkit definition. So I do think it make sense to keep it in Fluent UI somewhere. I also, want to make sure we limit functionalities like presence but that's minor. out of all the ideas, I think setup() method with style (filled icon/outline icon) might be more abstract? what do you think?

markavitale: I still kind of feel like being signed in vs signed out isn't something the AvatarView should be concerned about. One potential use case of Avatar View is signed in/signed out representation for the active user of an app, but it's far from the only use case (see comments, coauthoring, presence bubbles, chat, etc). In iMessage, where an AvatarView equivalent exists, if I don't have a contact for someone I'm chatting with, that doesn't make them "signed out".

ios/FluentUI/People Picker/AvatarView.swift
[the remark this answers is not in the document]
markavitale: Makes sense. maybe adding foreground to the second one helps distinguish a bit?

ios/FluentUI/People Picker/InitialsView.swift
MathieuKavalec: I'm not sure why "#" was added in the first place so I don't have a good answer for that.

ios/FluentUI/People Picker/AvatarView.swift
MathieuKavalec: I've been thinking more about this and I like the idea of using a more abstract concept than sign out.
I think that the existing logic for choosing the default icon works:
- If no primary or secondary text is provided, use the "signed out" image. 
- If no initials could be generated from the primary/secondary text, use the "default" image.

It's just a matter of finding a more generic term for the case where no user information is provided to AvatarView.
Just throwing ideas out there, we could go with: "FallbackImage" (or keep "DefaultImage") for the "default" case and "EmptyImage" for the "signed out" state.

ios/FluentUI/People Picker/InitialsView.swift
harrieshin: i think this was there becuase... outlook does it that way? i can't think of good reason to keep it. lets get rid of it unless you think it will be useful for the future?

MathieuKavalec: I agree, let's get rid of it.

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -109,6 +126,9 @@ public extension Colors {
         // Should use physical color because this text is shown on physical avatar background
         public static var text: UIColor = textOnAccent
         public static var border = UIColor(light: .white, dark: gray900, darkElevated: gray800)
+
+        static let signedOutImageBackground = UIColor(light: gray50, dark: gray800)
```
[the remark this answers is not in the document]
harrieshin: I think I have address you concerns. please take a look!

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -545,6 +619,32 @@ open class AvatarView: UIView {
...
+
+            let personImage = renderer.image { _ in
+                let personImageSize = avatarSize.personImageSize
+                image.draw(at: CGPoint(x: floor((containerSize.width - personImageSize) / 2), y: floor((containerSize.height - personImageSize) / 2)))
+            }.withRenderingMode(.alwaysTemplate)
+            setupWithImage(personImage)
+
+            if let window = window {
+                imageView.backgroundColor = style.backgroundColor(for: window)
+            }
```
MathieuKavalec, line 636: Are we handle the view did move to window callback to update the image?

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -340,6 +413,7 @@ open class AvatarView: UIView {
 
     private var hasBorder: Bool = false
     private var hasCustomBorder: Bool = false
+    private var currentFallbackImageStyle: AvatarFallbackImageStyle?
```
MathieuKavalec, line 416: nit: would "fallbackImageStyle" be a simpler name? Not sure what makes it "current" more than any other property.

ios/FluentUI/People Picker/AvatarView.swift
harrieshin: if I understand your question correctly..... I think my changes in layoutSubviews() handles your scenario?

PR #179, opened 2020-08-05
Contact Collection

jonathanwangg:
### Platforms Impacted
- [X] iOS
- [ ] macOS

#### ContactCollectionViewDemoController.swift
- This is a simple view controller that uses the `samplePersonas` sample data to showcase a `ContactCollectionView`

#### ContactCollectionView.swift
- The main view which contains all the cells
- Configures properties of the collection view as well as constraints
- This class is both the data source as well as the delegate which provides the count of the number of contacts as the number of items in a section and the cell at the specified index path
- The `didSelectItemAt` method scrolls a partially visible `ContactView` that has been clicked until it is fully visible

#### ContactCollectionViewLayout.swift
- A `CollectionViewLayout` class that describes how the layout of the `ContactCollectionView` should look like
- The cells are laid out in a single section horizontally with line spacing between each cell
- The `sizeForItemAt` delegate method uses hardcoded values for the height as the height only depends on the currently `ContentSizeCategory`

#### ContactCollectionViewCell.swift
- This class represents an individual cell for ContactCollectionView and creates a new view each time a cell is created. An improvement to this would be to change the contents of the `ContactView` (i.e. title, subtitle, image) instead creating a new view each time for 
performance reasons

#### ContactViewDemoController.swift
- Added an extra example where no title, subtitle, or image is provided.

#### ContactView.swift
- Added large content viewer for the title of each `ContactView`
- Added a pressed state overlay when the user taps on a Contact (grey circle over the avatar image)

#### Other
- Fixed two typos in `PillButtonBar.swift`
- Note that in the demo app, some names are displayed on a single line instead of a double line because of the way the data was provided when creating a `ContactView`. (It was provided as a single name rather than a separate title and subtitle.)

### Verification

Verification was done using manual tests and visual inspection.

| Collection with smallest ContentSizeCateogry | smallest category |
| Collection with large(default) ContentSizeCateogry | large category |
| Collection with XXXL ContentSizeCateogry | triple extra large category |
| Collection in landscape orientation | landscape orientation |
| Collection in dark mode | dark mode |
| Scrolling through the collection | scrolling|
| Pressed state overlay | pressed state overlay |
| Autoscroll on click | auto scroll |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/179)


ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
```diff
@@ -14,6 +14,7 @@ let demos: [(title: String, controllerClass: UIViewController.Type)] = [
     ("Button", ButtonDemoController.self),
     ("Color", ColorDemoController.self),
     ("ContactView", ContactViewDemoController.self),
+    ("ContactCollectionView", ContactCollectionViewDemoController.self),
```
MathieuKavalec, line 17: One up, alphabetical order

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactCollectionViewDemoController.swift
```diff
@@ -0,0 +1,23 @@
...
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        contactCollectionView.translatesAutoresizingMaskIntoConstraints = false
+        contactCollectionView.contactList = samplePersonas
+
+        view.addSubview(contactCollectionView)
+
+        contactCollectionView.leadingAnchor.constraint(equalTo: view.leadingAnchor).isActive = true
```
MathieuKavalec, line 20: Activate all constraints at the same time like this: NSLayoutConstraints.activateConstraints([
   contactCollectionView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
   contactCollectionView.trailingAnchor.constraint(equalTo: view.trailingAnchor)
])

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactCollectionViewDemoController.swift
```diff
@@ -0,0 +1,23 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class ContactCollectionViewDemoController: DemoController {
```
MathieuKavalec, line 9: Doesn't need to inherit from DemoController does it? Could it just be UIViewController?

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+        static let extraSmallContentContactHeight: CGFloat = 115.0
+        static let smallContentContactHeight: CGFloat = 117.0
+        static let mediumContentContactHeight: CGFloat = 118.0
+        static let largeContentContactHeight: CGFloat = 121.0
+        static let extraLargeContentContactHeight: CGFloat = 125.0
+        static let extraExtraLargeContentContactHeight: CGFloat = 129.0
+        static let extraExtraExtraLargeContentContactHeight: CGFloat = 135.0
+    }
+
+    let layout: ContactCollectionViewLayout
```
MathieuKavalec, line 21: We should keep things above all else. For example, layout and struct Constants should go lower in this file.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+    @objc public weak var contactCollectionDelegate: UICollectionViewDelegate?
```
MathieuKavalec, line 22: I don't see this being used anywhere (maybe CMD+F doesn't work in github?)

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    @objc public var contactList: [PersonaData] = [] {
```
MathieuKavalec, line 24: All public APIs must be documented.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+        didSet {
+            if (oldValue.count == 0 && contactList.count > 0) || (oldValue.count > 0 && contactList.count == 0) {
```
MathieuKavalec, line 26: How about simplifying this logic to:
oldValue.count != contactList.count

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @objc public init() {
+        layout = ContactCollectionViewLayout()
+        layout.scrollDirection = .horizontal
+        super.init(frame: .zero, collectionViewLayout: layout)
+        configureCollectionView()
+        updateHeightConstraint()
+        register(ContactCollectionViewCell.self, forCellWithReuseIdentifier: ContactCollectionViewCell.identifier)
+        NotificationCenter.default.addObserver(self, selector: #selector(setNewHeightConstraint), name: UIContentSizeCategory.didChangeNotification, object: nil)
```
MathieuKavalec, line 39: nitpick: In ObjC and swift, the "set" prefix in method names is reserved for property setters (although swift does have a different mechanism to declare setters using the "set" keyword, this rule seems to still apply). A better name would be "setupHeightConstraint".

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: You're correct. I had forgotten to remove it since I initially thought that the delegate was mandatory when I was learning about `UICollectionView`s.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @objc public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    var heightConstraint: NSLayoutConstraint?
```
MathieuKavalec, line 46: Alternatively, you could do this:

lazy var heightConstraint: NSLayoutConstraint = {
  let  heightConstraint = heightAnchor.constraint(equalToConstant:0.0)
  heightConstraint.isActive = true
  return heightConstraint
}()

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+    var widthConstraint: NSLayoutConstraint?
+
+    private func updateHeightConstraint() {
+        if heightConstraint == nil {
+            heightConstraint = heightAnchor.constraint(equalToConstant: 0.0)
+            heightConstraint!.isActive = true
+            return
```
MathieuKavalec, line 53: The early return here seems odd logically. Why wouldn't we apple the logic below? 
This logic needs to be removed anyway if you follow my last comment.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+        }
+
+        var constant: CGFloat = 0.0
```
MathieuKavalec, line 56: nitpick: var constant = (contactList.count > 0) ? categoryHeight : 0

side-note, if the type can't be inferred because of the ternary expression, you'll need to add ": CGFloat".

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactCollectionViewDemoController.swift
jonathanwangg: Good catch. I agree with inheriting form `UIViewController` rather than `DemoController`. Originally I was inheriting from `DemoController` because I was using the `addRow()` method.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFContactCollectionView)
+open class ContactCollectionView: UICollectionView {
+    struct Constants {
+        static let sideInset: CGFloat = 16.0
```
MathieuKavalec, line 11: nitpick: a better name would be "verticalInset" if this is used for both the leading and trailing edge. If it is only used for the leading edge, then "leadingInset" would be a better name.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+        static let extraSmallContentContactHeight: CGFloat = 115.0
+        static let smallContentContactHeight: CGFloat = 117.0
+        static let mediumContentContactHeight: CGFloat = 118.0
+        static let largeContentContactHeight: CGFloat = 121.0
+        static let extraLargeContentContactHeight: CGFloat = 125.0
+        static let extraExtraLargeContentContactHeight: CGFloat = 129.0
+        static let extraExtraExtraLargeContentContactHeight: CGFloat = 135.0
```
MathieuKavalec, line 18: Instead of adding constants for each size category, we should extend the  enum.

extension UIContentSizeCategory {
  var contactHeight: CGFloat {
        switch self {
        case .extraSmall:
            return Constants.extraSmallContentContactHeight
        case .small:
            return Constants.smallContentContactHeight
        case .medium:
            return Constants.mediumContentContactHeight
        case .large:
            return Constants.largeContentContactHeight
        case .extraLarge:
            return Constants.extraLargeContentContactHeight
        case .extraExtraLarge:
            return Constants.extraExtraLargeContentContactHeight
        case .extraExtraExtraLarge:
            return Constants.extraExtraExtraLargeContentContactHeight
        default:
            return Constants.extraExtraExtraLargeContentContactHeight
    }
}

Later, you'll be able to do this to get the height:
UIApplication.shared.preferredContentSizeCategory.contactHeight

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    @objc public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    var heightConstraint: NSLayoutConstraint?
+    var widthConstraint: NSLayoutConstraint?
+
+    private func updateHeightConstraint() {
```
MathieuKavalec, line 49: This method and setNewHeightConstraint do the same thing. Once you address my other comments, you should combine them.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+            addInteraction(UILargeContentViewerInteraction())
+        }
+    }
+
+    @objc private func setNewHeightConstraint() {
+        let height = categoryHeight()
+        heightConstraint?.constant = height
+    }
+
+    private func categoryHeight() -> CGFloat {
```
MathieuKavalec, line 83: See my other comment about this. This should be part of an extension of UIContentSizeCategory.

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class ContactCollectionViewCell: UICollectionViewCell {
+    static var identifier: String { return "ContactCollectionViewCell" }
```
MathieuKavalec, line 9: static let identifier: String = "ContactCollectionViewCell"

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Agreed

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    func setup(contact persona: PersonaData) {
+        if persona.name != "" {
+            contactView = ContactView(identifier: persona.name)
+        } else {
+            contactView = ContactView(identifier: persona.email)
+        }
+
+        contactView.translatesAutoresizingMaskIntoConstraints = false
```
MathieuKavalec, line 24: I don't think you need this line

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
...
+    static var identifier: String { return "ContactCollectionViewCell" }
+    var contactView: ContactView
+
+    override init(frame: CGRect) {
+        contactView = ContactView(title: "", subtitle: "")
+        super.init(frame: frame)
+    }
+
+    func setup(contact persona: PersonaData) {
+        if persona.name != "" {
```
MathieuKavalec, line 18: A better check would be: persona.name.length > 0.

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
...
+            contactView = ContactView(identifier: persona.name)
+        } else {
+            contactView = ContactView(identifier: persona.email)
+        }
```
MathieuKavalec, line 22: nit:
contactView = Contactview(identifier: (persona.name.length > 0) ? persona.name : persona.email))

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,40 @@
...
+
+        contentView.addSubview(contactView)
+    }
+
+    override func prepareForReuse() {
+        contactView.removeAllSubviews()
+        super.prepareForReuse()
+    }
+
+    required init?(coder: NSCoder) {
```
MathieuKavalec, line 37: nit: should declare all initializers together. Meaning, you should move this below the other init method.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Thanks for informing me of this convention!

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: This is only used for the leading edge. I'll use your suggestion as it seems more appropriate!

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
jonathanwangg: A cleaner way of writing it!

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
```diff
@@ -0,0 +1,151 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class ContactCollectionViewLayout: UICollectionViewFlowLayout {
+    @objc weak var delegate: UICollectionViewDelegateFlowLayout?
```
MathieuKavalec, line 9: I'm very confused by this line.
A UICollectionViewDelegateFlowLayout is supposed to a UICollectionViewDelegate, which is supposed to be set on a UICollectionView. In this class, you are declaring that the layout (UICollectionViewFlowLayout) needs a UICollectionViewDelegateFlowLayout and setting the layout as its own delegate. This seems like an anti-pattern.

If you expose a delegate property, it's with the intention to have some other object be that delegate. Hence the term delegate, since you are delegating tasks to some other entity.

I think I understand the intent here.
You wanted the functionality from some of the methods exposed in the UICollectionViewDelegateFlowLayout protocol to use in this class.
Here's how you should fix this up:

- Get rid of this delegate property.
- Get rid of extension ContactCollectionViewLayout: UICollectionViewDelegateFlowLayout
- Move the methods from that extension inside the ContactCollectionViewLayout class and rearrange the parameters appropriately.

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
```diff
@@ -0,0 +1,151 @@
...
+        static let minLineSpacing: CGFloat = 14.0
+        static let maxLineSpacing: CGFloat = 16.0
+
+        static let extraSmallContentContactHeight: CGFloat = 115.0
+        static let smallContentContactHeight: CGFloat = 117.0
+        static let mediumContentContactHeight: CGFloat = 118.0
+        static let largeContentContactHeight: CGFloat = 121.0
+        static let extraLargeContentContactHeight: CGFloat = 125.0
+        static let extraExtraLargeContentContactHeight: CGFloat = 129.0
+        static let extraExtraExtraLargeContentContactHeight: CGFloat = 135.0
```
MathieuKavalec, line 22: These constants are duplicated elsewhere. As I mentioned earlier, you should add these to a UIContentSizeCategory extension. Don't duplicate the extension. You can make it available internally to the FluentUI framework and use it in both file.

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
```diff
@@ -0,0 +1,151 @@
...
+    required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) not implemented")
+    }
+
+    override var collectionViewContentSize: CGSize {
+        guard let collectionView = collectionView else {
+            return .zero
+        }
+
+        // TODO: Might not even need the setting here since the delegate method already sets it
```
MathieuKavalec, line 40: Shouldn't leave TODO comments in code like this. If there is a TODO for a later PR, then it needs a bug reference. Otherwise, either fix it now or remove comment.

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
```diff
@@ -0,0 +1,151 @@
...
+        let size = CGSize(width: totalWidth, height: height!)
+
+        return size
+    }
+
+    override func layoutAttributesForElements(in rect: CGRect) -> [UICollectionViewLayoutAttributes]? {
+        var layoutAttributes = [UICollectionViewLayoutAttributes]()
+
+        if let collectionView = collectionView {
+            numberOfItems = collectionView.numberOfItems(inSection: 0)
```
MathieuKavalec, line 59: numberOfItems shouldn't be a property, it's only used in this method, could make it a local var

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
```diff
@@ -0,0 +1,151 @@
...
+                }
+            }
+        }
+
+        return layoutAttributes
+    }
+
+    override func layoutAttributesForItem(at indexPath: IndexPath) -> UICollectionViewLayoutAttributes? {
+        let attributes = UICollectionViewLayoutAttributes(forCellWith: indexPath)
+        let frame = CGRect(x: CGFloat(indexPath.item) * Constants.itemWidth + CGFloat(indexPath.item) * minimumLineSpacing, y: 0, width: Constants.itemWidth, height: itemSize.height)
```
MathieuKavalec, line 76: nit: let's use multiple lines for this for readability.
example:
let frame = CGRect(x: CGFloat(indexPath.item) * Constants.itemWidth + CGFloat(indexPath.item) * minimumLineSpacing, 
y: 0,
...

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -99,12 +107,16 @@ open class ContactView: UIView {
             constraints.append(contentsOf: identifierLayoutConstraints())
         }
         titleLabel.translatesAutoresizingMaskIntoConstraints = false
-        labelContainer.addSubview(titleLabel)
+
+        pressedStateOverlay.translatesAutoresizingMaskIntoConstraints = false
+        avatarView.addSubview(pressedStateOverlay)
```
MathieuKavalec, line 119: We shouldn't add the pressedStateOverlay to the view hierarchy unless an action+target is set on the ContactView to handle the tap events. In other words, the overlay tap animation shouldn't be shown unless there the tap event has a handler.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -169,6 +181,17 @@ open class ContactView: UIView {
         ]
     }
 
+    private func pressedStateOverlayLayoutConstraints() -> [NSLayoutConstraint] {
+        return [
+            pressedStateOverlay.widthAnchor.constraint(equalTo: avatarView.widthAnchor),
+            pressedStateOverlay.heightAnchor.constraint(equalTo: avatarView.heightAnchor),
```
MathieuKavalec, line 192: The width and height anchor are not needed and add extra complexity to the layout. It's important to always use the smallest amount possible of constraints to layout a view. In this case, if the leading, trailing, top and bottom anchors are equal to the super view, then the width and height have to be equal too so we don't need those extra constraints.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -194,4 +217,25 @@ open class ContactView: UIView {
...
+
+    private func setupPressedStateOverlay() {
+        pressedStateOverlay.clipsToBounds = true
+        pressedStateOverlay.frame = avatarView.frame
+        pressedStateOverlay.layer.cornerRadius = avatarView.frame.width / 2
+    }
+
+    open override func touchesBegan(_ touches: Set<UITouch>, with event: UIEvent?) {
+        super.touchesBegan(touches, with: event)
+        pressedStateOverlay.backgroundColor = UIColor(red: 0.83, green: 0.83, blue: 0.83, alpha: 0.6)
```
MathieuKavalec, line 234: These colors should be constants.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -194,4 +217,25 @@ open class ContactView: UIView {
...
+    }
+
+    open override func touchesMoved(_ touches: Set<UITouch>, with event: UIEvent?) {
+        super.touchesMoved(touches, with: event)
+        pressedStateOverlay.backgroundColor = UIColor(red: 1.0, green: 1.0, blue: 1.0, alpha: 0.0)
```
MathieuKavalec, line 239: Shouldn't we set the "pressed" color back if the touches are within the bounds of the view?

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+
+    @objc public var contactList: [PersonaData] = [] {
+        didSet {
+            if (oldValue.count == 0 && contactList.count > 0) || (oldValue.count > 0 && contactList.count == 0) {
+                updateHeightConstraint()
+            }
+        }
+    }
+
+    @objc public init() {
```
MathieuKavalec, line 32: Should the initializer take a [PersonaData] array for convenience?

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
jonathanwangg: Didn't know you could use a ternary operator as an argument 😮 . Will try that approach!

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
jonathanwangg: Good point. It makes the code look more organized as well.

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
jonathanwangg: Great suggestion! I'll make the required changes.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Understood, thanks for explaining the reasoning.

ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
```diff
@@ -14,6 +14,7 @@
 		7D23482A24D89C1C00FBE057 /* AvatarGroupViewDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D23482924D89C1C00FBE057 /* AvatarGroupViewDemoController.swift */; };
 		7DC2FB2B24C0F4FD00367A55 /* TableViewCellFileAccessoryViewDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7DC2FB2A24C0F4FD00367A55 /* TableViewCellFileAccessoryViewDemoController.swift */; };
 		8AD4FF842493F2500085C076 /* ContactViewDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8AD4FF832493F2500085C076 /* ContactViewDemoController.swift */; };
```
harrieshin, line 16: contactviewdemocontroller.swift itself is not interesting. the toolkit component is meant to be horizontal scrollable. if you are creating contactcollectionviewdemocontroller separately, I would delete the other one.

harrieshin:
in the toolkit design, the horizontal scroll contact list can have 2 different avatar sizes: 72x72 avatar or 52x52 avatar with no subtitle.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,148 @@
...
+        translatesAutoresizingMaskIntoConstraints = false
+        showsHorizontalScrollIndicator = false
+        showsVerticalScrollIndicator = false
+        backgroundColor = Colors.surfacePrimary
+        dataSource = self
+        delegate = self
+        contentInset = UIEdgeInsets(top: 0, left: Constants.sideInset, bottom: 0, right: 0)
+
+        if #available(iOS 13, *) {
+            addInteraction(UILargeContentViewerInteraction())
```
harrieshin, line 74: we are supporting largecontentviewer? didn't the title size increase?

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
jonathanwangg: Good catch. Must have missed this when I was looking through it!

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Can you elaborate more on what you mean by this? From what I understand, you're saying that we should still be setting the color to the "pressed" color if the touches has moved to somewhere still within the `ContactView`? Please correct me if this isn't your intention.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: The title size does indeed increase when we change the `ContentSizeCategory`. I added the `LargeContentViewer` due to a suggestion I got from Tom. I can discuss this further with Kaan to see if this is something we would like to support.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Aren't the `touchesBegan`, `touchesMoved`, `touchesEnded` functions considered as targets or are you saying we should check if those methods have been implemented before adding `pressedStateOverlay` to the view hierarchy?

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Good to know! Are we trying to follow this https://microsoft.github.io/swift-guide/FileContentOrdering.html now?

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Gotcha. We're able to do this because the extracted code

```
let heightConstraint = heightAnchor.constraint(equalToConstant:0.0)
heightConstraint.isActive = true
return heightConstraint
```

is only run once within the `updateHeightConstraint()` function right?

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: Yes

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec: I think ContactView is missing something. If we want ContactView to be clickable, you need to expose an action selector (SEL) and a target (id). See the UIButton API. Example:
button.addTarget(self, action: action, for: .touchUpInside)
You don't need the event type param since you set the touch behavior in ContactView.

If you don't set a target/selector, then the ContactView should not have a touch down behavior at all since it's not clickable.

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec: Take a look at the button control in the demo app.
If you do a touch down, touch move (in and out of the control bounds), touch up, depending on where your at, the control will be dimmed or not dimmed. ContactView needs to have the same behavior as Buttons for the touch down/move/up behavior.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: From emails exchanged with Kaan, it has been decided that we don't need large content viewer for the Contact Collection. The relevant code for it will be moved.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Removed this code to address previous comment (https://github.com/microsoft/fluentui-apple/pull/179#discussion_r466012401).

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Great way to shorten the code!

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Don't we still need to define the constant values for the height somewhere?

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Got it.

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: No need if you follow my approach

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Awesome! Thanks for finding this 😀

ios/FluentUI/People Picker/ContactCollectionViewLayout.swift
jonathanwangg: Done

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Do we want the same behaviour as the Buttons in the demo app? I check the iOS Outlook app and it seems that when we click and move, the grey overlay disappears as well. Which behaviour should we adhere to?

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: I don't think we need the initializer to take in a `[PersonaData]` in `init()` since they can set the `contactList` 
variable. We also discussed this in the engineering plan with Miguel and mentioned that it (`[ContactView]`) shouldn't be part of the initializer. Do you think we should still include a `[PersonaData]` array in the `init()`?

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec: Let's go with what Outlook iOS does.

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: It would be mainly for convenience since anyone creating a ContactCollection would want to pass it PersonaDatas. Up to you if you want to go that direction or not. This is mostly an API design consideration.

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
jonathanwangg: I think this line is actually needed, but I misplaced it. It was missing from `ContactView.swift` and I'll move it there.

ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
jonathanwangg: I'm a bit unclear on what you mean. Are you saying we should remove the `ContactViewDemoController.swift` and any references to it since the "full control" should really be scrollable? (Therefore we only have `ContactCollectionViewDemoController.swift`?)

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactCollectionViewDemoController.swift
```diff
@@ -0,0 +1,26 @@
...
+        super.viewDidLoad()
+        view.backgroundColor = Colors.surfacePrimary
+
+        contactCollectionView.translatesAutoresizingMaskIntoConstraints = false
+        contactCollectionView.contactList = samplePersonas
+
+        view.addSubview(contactCollectionView)
+
+        NSLayoutConstraint.activate([
+            contactCollectionView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
```
MathieuKavalec, line 22: Could we add some top spacing above the collection? I think this would look nice.
contactCollectionView.topAnchor.constraint(equalTo: view.topAnchor constant:20)

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactViewDemoController.swift
```diff
@@ -32,12 +32,16 @@ class ContactViewDemoController: DemoController {
...
+        let emptyContactView = ContactView(title: "", subtitle: "")
+        emptyContactView.translatesAutoresizingMaskIntoConstraints = false
+
         addRow(text: "ContactView with image", items: [contactView], textWidth: 200)
         addRow(text: "ContactView with initials", items: [noImageContactView], textWidth: 200)
         addRow(text: "ContactView with first name", items: [firstNameContactView], textWidth: 200)
         addRow(text: "ContactView with email", items: [emailContactView], textWidth: 200)
         addRow(text: "ContactView with phone number", items: [phoneNumberContactView], textWidth: 200)
-        container.addArrangedSubview(UIView())
+        addRow(text: "ContactView with nothing", items: [emptyContactView], textWidth: 200)
```
MathieuKavalec, line 44: nit: "Empty ContactView" 
"with nothing" sounds a bit depressing 😆

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,139 @@
...
+            setContentOffset(CGPoint(x: offSet, y: contentOffset.y), animated: true)
+        }
+    }
+}
+
+extension UIContentSizeCategory {
+    var contactHeight: CGFloat {
+        switch self {
+        case .extraSmall:
+            return ContactCollectionView.Constants.extraSmallContentContactHeight
```
MathieuKavalec, line 122: We shouldn't use constants in this case. Just return the literal CGFloats.
You could consider var contactHeight as a group of constants.
In other words, remove these constants:
        static let extraSmallContentContactHeight: CGFloat = 115.0
        static let smallContentContactHeight: CGFloat = 117.0
        static let mediumContentContactHeight: CGFloat = 118.0
        static let largeContentContactHeight: CGFloat = 121.0
        static let extraLargeContentContactHeight: CGFloat = 125.0
        static let extraExtraLargeContentContactHeight: CGFloat = 129.0
        static let extraExtraExtraLargeContentContactHeight: CGFloat = 135.0

and hardcode them in here.

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class ContactCollectionViewCell: UICollectionViewCell {
+    static let identifier: String = "ContactCollectionViewCell"
+    var contactView: ContactView
```
MathieuKavalec, line 10: This should be private, same for identifier.
Throughout your PR, mark methods/constants/properties as private if you they are internal to their class.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,139 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFContactCollectionView)
+open class ContactCollectionView: UICollectionView {
```
harrieshin, line 9: there should be public documentation of this control for developers

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -93,6 +98,7 @@ open class ContactView: UIView {
     private func setupLayout() {
         var constraints = [NSLayoutConstraint]()
         constraints.append(contentsOf: avatarLayoutConstraints())
+        translatesAutoresizingMaskIntoConstraints = false
```
MathieuKavalec, line 101: In general, only the superview or whoever is adding your view to a view hierarchy should be setting translatesAutoresizingMaskIntoConstraints. Do you need this line?

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,139 @@
...
+        static let extraLargeContentContactHeight: CGFloat = 125.0
+        static let extraExtraLargeContentContactHeight: CGFloat = 129.0
+        static let extraExtraExtraLargeContentContactHeight: CGFloat = 135.0
+    }
+
+    let layout: ContactCollectionViewLayout
+    var widthConstraint: NSLayoutConstraint?
+    lazy var heightConstraint: NSLayoutConstraint = {
+        let heightConstraint = heightAnchor.constraint(equalToConstant: 0.0)
+        heightConstraint.isActive = true
```
harrieshin, line 53: this seems bit weird that the hight constraint is activated by referencing this variable.

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec: As we discussed, you will need to address this in a subsequent change.

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: I recommended this since it simplifies the logic. Height constraint will always be active. An alternative would be to activate it in the init method. Both seem fine to me.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,139 @@
...
+        setupHeightConstraint()
+        register(ContactCollectionViewCell.self, forCellWithReuseIdentifier: ContactCollectionViewCell.identifier)
+        NotificationCenter.default.addObserver(self, selector: #selector(setupHeightConstraint), name: UIContentSizeCategory.didChangeNotification, object: nil)
+    }
+
+    @objc public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    // The array of PersonaData which is used to create each ContactView. The height constraint of the ContactCollectionView is updated when the count increases from 0 or decreases to 0.
```
MathieuKavalec, line 29: nit: use three "/" for method/class/constant documentation.

Example:
/// The array of PersonaData which is used to create each ContactView. 
/// The height constraint of the ContactCollectionView is updated when the count increases from 0 or decreases to 0.

ios/FluentUI/People Picker/ContactCollectionViewCell.swift
jonathanwangg: I agree for the `ContactView`, but not for the identifier. From what I understand, I need to use the identifier in `ContactCollectionView.swift` when registering `ContactCollectionViewCell` using its identifier. I've also check other "cell classes" and the access control for the identifier is `static` or `class` level.

ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
jonathanwangg: Resolved via messages. Will be removing `ContactViewDemoController.swift` and any references to it.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ContactViewDemoController.swift
jonathanwangg: 🌞  Will try to brighten up the code 🌞

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Originally I placed this in `ContactCollectionViewCell.swift`, but you commented that it was probably not needed. I thought it would be better placed in `ContactView.swift` instead. I'll move it back to `ContactCollectionViewCell.swift`.

Yes, this line is needed. Without it, the `ContactView`'s don't show up properly.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,138 @@
...
+            return 118.0
+        case .large:
+            return 121.0
+        case .extraLarge:
+            return 125.0
+        case .extraExtraLarge:
+            return 129.0
+        case .extraExtraExtraLarge:
+            return 135.0
+        default:
```
harrieshin, line 134: let's not have default defined so it will show compiler error when new enum is created

jonathanwangg: Got it!

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,22 +9,18 @@ import UIKit
 
 public extension Colors {
     struct Contact {
-        public static var title: UIColor = textPrimary
-        public static var subtitle: UIColor = textSecondary
+        public static let title: UIColor = textPrimary
+        public static let subtitle: UIColor = textSecondary
+        public static let pressedState = UIColor(red: 0.83, green: 0.83, blue: 0.83, alpha: 0.6)
```
harrieshin, line 14: we should be using semantic colors here. if it isn't defined already, it should be declared in colors.swift or asset catlog

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,22 +9,18 @@ import UIKit
...
+        public static let unpressedState = UIColor(red: 1.0, green: 1.0, blue: 1.0, alpha: 0.0)
```
harrieshin, line 15: white with alpha?

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: What do you mean by semantic colours? The colours I chose aren't defined yet in `Colors.swift` as I wasn't given a specification on which colours to use. I can define them in `Colors.swift` if that is what you're reccomending.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: I'm not too familiar with colours, so I want to confirm that you want me to use `UIColor(white: CGFloat, alpha: CGFloat)` instead of what I currently have?

ios/FluentUI/People Picker/ContactCollectionView.swift
harrieshin: i think activate in initi seems clearer to me

ios/FluentUI/People Picker/ContactView.swift
harrieshin: all fluent gray colors needs to be from fluent gray color pallet . gray10 - gray900. if this defers from that we need to talk to design.

ios/FluentUI/People Picker/ContactCollectionView.swift
harrieshin: yes

ios/FluentUI/People Picker/ContactView.swift
harrieshin: but then again do we need this color? if you aren't pressing, isn't it just clear color?

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: I can try to find a similar colour using the grey colour pallet.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Just to confirm, we're keeping the default right?

harrieshin: no remove default. we should be explicit for all the other cases thatit will use .extraExtraExtraLarge by design.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: This colour is to set the view back to invisible when the contact view isn't being tapped. I think we need it since if we don't set it back, the avatar image will still have a grey overlay when it isn't being tapped.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: After discussing with @MathieuKavalec, there seems to be an issue with Xcode not correctly understanding that a switch statement is already exhaustive and still requires a `default` (otherwise building will fail). Due to this, I've grouped the `.unspecified` category into `default`.

Currently there aren't any redlines for the accessibility sizing, so I've grouped them under the `.extraExtraExtraLarge` case since those have the closest height. Perhaps we can update the switch statement when designs for the missing sizes are finalized?

ios/FluentUI/People Picker/ContactView.swift
MathieuKavalec: Originally, I was curious to know why it was needed since I didn't see where we were adding layout constraints to this view. This does need to go where the view is being added to another subview. Sorry for the miscommunication.

jonathanwangg: All good. Thanks for the follow up 👍

harrieshin:
talked offline but need to add all the new controls to static library target

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,161 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+// MARK: ContactCollectionViewDelegate
+
+public protocol ContactCollectionViewDelegate: AnyObject {
```
MathieuKavalec, line 10: Need to expose this to ObjC. (don't forget the MSF prefix)

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,161 @@
...
+    func didTapOnContactViewAtIndex(index: Int, personaData: PersonaData)
```
MathieuKavalec, line 11: This function needs to be exposed to objc (@objc).

MathieuKavalec: Also, in general, delegate methods are optional.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -199,4 +215,35 @@ open class ContactView: UIView {
...
+        pressedStateOverlay.isHidden = false
+    }
+
+    open override func touchesMoved(_ touches: Set<UITouch>, with event: UIEvent?) {
+        super.touchesMoved(touches, with: event)
+        pressedStateOverlay.isHidden = true
+    }
+
+    open override func touchesEnded(_ touches: Set<UITouch>, with event: UIEvent?) {
+        super.touchesEnded(touches, with: event)
```
MathieuKavalec, line 245: I don't think that the touchesEnded is called necessarily because a touchUpInside event happened.
Generally, a tap event is considered a touch up inside event meaning that the touch up event happened within the bounds of the view. Let's make sure that the tap event is only triggered when the user lifts his finger above the view (or follow whatever Outlook is doing if it differs).

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,23 +9,22 @@ import UIKit
...
-        public static var subtitle: UIColor = textSecondary
+        public static let title: UIColor = textPrimary
+        public static let subtitle: UIColor = textSecondary
+        public static let pressedState: UIColor = Colors.gray200.withAlphaComponent(0.6)
     }
 }
 
+// MARK: ContactViewDelegate
+
+public protocol ContactViewDelegate: AnyObject {
```
MathieuKavalec, line 20: Expose this protocol to objc

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,23 +9,22 @@ import UIKit
...
+    func didTapContactView(_ contact: ContactView)
```
MathieuKavalec, line 21: Delegate methods are optional and method needs to be exposed to objc (@objc)

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -40,6 +39,8 @@ open class ContactView: UIView {
     private var titleLabel: UILabel
     private var subtitleLabel: UILabel?
     private var labelContainer: UIView
+    private let pressedStateOverlay: UIView
+    open weak var contactViewDelegate: ContactViewDelegate?
```
MathieuKavalec, line 43: open/public stuff goes above private stuff

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -106,12 +109,16 @@ open class ContactView: UIView {
             constraints.append(contentsOf: identifierLayoutConstraints())
         }
         titleLabel.translatesAutoresizingMaskIntoConstraints = false
-        labelContainer.addSubview(titleLabel)
+
+        pressedStateOverlay.translatesAutoresizingMaskIntoConstraints = false
+        avatarView.addSubview(pressedStateOverlay)
+        constraints.append(contentsOf: pressedStateOverlayLayoutConstraints())
```
MathieuKavalec, line 115: nit: would be cleaner to just pass the array of constraints here instead of having a separate method in this case, especially considering that the method only gets called once.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,161 @@
...
+}
+
+extension ContactCollectionView: UICollectionViewDelegate {
+    // If the Contact is not fully visible in the scroll frame, scrolls the view by an offset large enough
+    // so that the tapped Contact is fully visible.
+    @objc public func collectionView(_ collectionView: UICollectionView, didSelectItemAt indexPath: IndexPath) {
+        guard let cell = cellForItem(at: indexPath) else {
+            return
+        }
+        currentTappedIndex = indexPath.item
```
MathieuKavalec, line 113: We should be going through ContactViewDelegate to detect tap events on the ContactViews. 
ContactCollectionView should set itself as the delegate of each of its ContactViews.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,161 @@
...
+        let heightConstraint = heightAnchor.constraint(equalToConstant: 0.0)
+        heightConstraint.isActive = true
+        return heightConstraint
+    }()
+    private var currentTappedIndex: Int?
+}
+
+extension ContactCollectionView: ContactViewDelegate {
+    public func didTapContactView(_ contact: ContactView) {
+        if let contactCollectionViewDelegate = contactCollectionViewDelegate, let currentTappedIndex = currentTappedIndex {
```
MathieuKavalec, line 86: There must be a better way calculate the tapped index from the ContactView reference.
You should get rid of the currentTappedIndex property. There should only be once source of truth for the tapped contact view. Currently, you are mixing the ContactView's tap handling and the UICollectionView's tap callback.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: We want to expose the functions in the protocol to objc because other classes written in objc may want to be the delegate right?

MathieuKavalec: right

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: That was what I originally was thinking. After some more thought about it, I don't think there is a way to calculate the tapped index from the `ContactView` reference because a `ContactView` shouldn't have a concept of an index since it doesn't have to be placed inside  a `ContactCollectionView`.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Got it. Matching Outlook requires us to use `touchUpInside`.

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: Ah, I forgot that there isn't a way to just get a list of collection view cells. Ignore my comment, this'll work.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,162 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+// MARK: ContactCollectionViewDelegate
+
+public typealias MSContactCollectionViewDelegate = ContactCollectionViewDelegate
```
MathieuKavalec, line 10: Use this instead:
@objc(MSContactCollectionViewDelegate)

side-note: the prefix is MSF, not just MS.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -0,0 +1,162 @@
...
+        }
+    }
+
+    private struct Constants {
+        static let leadingInset: CGFloat = 16.0
+        static let amountOfNextContactToShow: CGFloat = 20.0
+    }
+
+    private let layout: ContactCollectionViewLayout
+    private var widthConstraint: NSLayoutConstraint?
```
MathieuKavalec, line 106: widthConstraint is unused

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,23 +9,22 @@ import UIKit
 
 public extension Colors {
     struct Contact {
-        public static var title: UIColor = textPrimary
-        public static var subtitle: UIColor = textSecondary
+        public static let title: UIColor = textPrimary
+        public static let subtitle: UIColor = textSecondary
```
MathieuKavalec, line 13: nit: don't need to create a second constant for textPrimary and textSecondary. Just use those inline directly.
Color constants would be useful if you declared colors using random RGB values.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -9,23 +9,22 @@ import UIKit
...
+        public static let pressedState: UIColor = Colors.gray200.withAlphaComponent(0.6)
     }
 }
 
+// MARK: ContactViewDelegate
+public typealias MSContactViewDelegate = ContactViewDelegate
```
MathieuKavalec, line 19: Use this instead:
@objc(MSFContactViewDelegate)

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -106,12 +115,22 @@ open class ContactView: UIView {
...
+        avatarView.addSubview(pressedStateOverlay)
+        constraints.append(contentsOf: [
+            pressedStateOverlay.leadingAnchor.constraint(equalTo: avatarView.leadingAnchor),
+            pressedStateOverlay.trailingAnchor.constraint(equalTo: avatarView.trailingAnchor),
+            pressedStateOverlay.topAnchor.constraint(equalTo: avatarView.topAnchor),
+            pressedStateOverlay.bottomAnchor.constraint(equalTo: avatarView.bottomAnchor)
+        ])
 
         labelContainer.translatesAutoresizingMaskIntoConstraints = false
+        labelContainer.isUserInteractionEnabled = false
```
MathieuKavalec, line 129: Are these isUserInteractionEnabled calls useful? Curious to know why.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -199,4 +218,34 @@ open class ContactView: UIView {
...
+        pressedStateOverlay.isHidden = false
+    }
+
+    @objc private func touchUpInsideHandler() {
+        contactViewDelegate?.didTapContactView?(self)
+        pressedStateOverlay.isHidden = true
+    }
+
+    @objc private func touchMovedHandler() {
+        pressedStateOverlay.isHidden = true
```
MathieuKavalec, line 241: I don't think we want to hide the overlay just because the user moved his finder? Typically, the dimmed state of a tappable control goes away when the touch event exits the bounds of the control's view.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -199,4 +218,34 @@ open class ContactView: UIView {
...
+    }
+
+    private struct Constants {
+        static let labelMinimumHeight: CGFloat = 16.0
```
MathieuKavalec, line 245: labelMinimumHeight isn't being used

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec, line 106: Is widthConstraint being used anywhere? Can't seem to find any usage.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -199,4 +218,34 @@ open class ContactView: UIView {
...
+        static let titleLabelMaximumHeight: CGFloat = 28.0
```
MathieuKavalec, line 246: I don't think titleLabelMaximumHeight is used anywhere

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -199,4 +218,34 @@ open class ContactView: UIView {
...
+        static let subtitleMaximumHeight: CGFloat = 24.0
```
MathieuKavalec, line 247: Same for subtitleMaximumHeight

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: The `isUserInteractionEnabled` for both the `labelContainer` and `avatarView` are needed for the `ContactView` to receive the touch events. When `isUserInteractionEnabled` is set to true (i.e. by default), the `ContactView` is not able to receive any touch events and the handlers in turn will not be called.

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: That is what I originally thought as well. After playing around with Outlook on iOS, it seems like the the dimmed state goes away when I tap the contact and drag.

Example: Tap a contact by the picture and then drag to the name of the contact. The dimmed state disappears. (The same behaviour occurs if tap and drag from the name to the picture of the same contact)

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: You're correct, it's not being used anywhere. I'll remove it. It's weird that I tried using the "Static Analyzer" in Xcode which is supposed to show unused variables, but it didn't report it as being unused.

https://developer.apple.com/library/archive/documentation/ToolsLanguages/Conceptual/Xcode_Overview/AnalyzingYourCode.html

> Dead store (unused variable) flaws

ios/FluentUI/People Picker/ContactView.swift
jonathanwangg: Got it. I don't think I was the one who created the color variables here. I was just changing them to be constants rather than variables since their values never changed. I believe it was part of the effort of moving all the color definitions of components out of `Colors.swift` and into their individual files. Nevertheless I can use the values inline directly as you suggested.

ios/FluentUI/People Picker/ContactCollectionView.swift
jonathanwangg: Discussed offline with Tom and it appears that it isn't able to discover unused **member** variables.

PR #180, opened 2020-08-05
Add MSF prefix for ObjC name

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Class is missing "MSF" prefix for ObjC name.

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/180)


MathieuKavalec:
> Should we do this using deprecation to avoid it being a breaking change for clients?

No need, this was very recently added to the framework and no client is consuming this yet.

PR #181, opened 2020-08-05
Add large content interaction for buttons in table view cell file accessory views

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added a large content interaction for buttons in table view cell file accessory views.

Note that large text accessibility is not supported in the demo controller for these cells because the cells are not being added to a UITableView. The cells do support large text accessibility.

Screen Shot 2020-08-05 at 4 56 52 PM

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/181)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
+++ b/ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
@@ -104,6 +104,11 @@ open class TableViewCellFileAccessoryView: UIView {
         columnStackView.addArrangedSubview(actionsStackView)
 
         updateSharedStatus()
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(updateLayout),
+                                               name: UIContentSizeCategory.didChangeNotification,
+                                               object: nil)
     }
 
     @available(*, unavailable)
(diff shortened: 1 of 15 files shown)


ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
```diff
@@ -380,6 +380,13 @@ private class FileAccessoryViewActionView: UIButton {
             widthAnchor.constraint(equalToConstant: FileAccessoryViewActionView.size.width),
             heightAnchor.constraint(greaterThanOrEqualToConstant: FileAccessoryViewActionView.size.height)
         ])
+
+        if #available(iOS 13, *) {
+            addInteraction(UILargeContentViewerInteraction())
```
harrieshin, line 385: shouldn't this be in the container view of all buttons and the button just need to set the scale image and title? https://developer.apple.com/videos/play/wwdc2019/261/

PR #190, opened 2020-08-14
Added support for app tinted TabBarItems in unselected state

MathieuKavalec:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

TabBarItems are app tinted in the selected state, but not the default unselected state.
In some cases, we want the TabBarItems to also be app tinted in the unselected state.
Let's add a parameter to allow this.

Since default parameter values don't exist in ObjC and our designated initializer for TabBarItem is getting crowded with a lot of parameters, let's create other convenience initializers without the default parameters.

Screen Shot 2020-08-13 at 5 23 48 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/190)


harrieshin:
is there a reason why the image isn't just passed in with the tinted color (aka not templated)?

MathieuKavalec:
> is there a reason why the image isn't just passed in with the tinted color (aka not templated)?

We do tinting internally inside the SideTabBar for the selected state. It makes more sense to me to provide an easy way to indicate that the unselected state also needs to be tinted.
Side-note, the utility method to tint an image is private to the FluentUI library. Should we make this public and available to ObjC?

PR #193, opened 2020-08-14
Contact collection large text accessibility support, bug fixes and small size

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Small size support for contact collection.
- Don't show overlay if the contact view is not tappable.
- Add support for first/last name in the PersonaData so names can be separated in the contact collection view.
- Large text accessibility support.
- Fix spacing issues. Added insets in the collection view. Added trailing spacing at the end of the collection to match the leading spacing.

Screen Shot 2020-08-14 at 4 00 56 PM
Screen Shot 2020-08-14 at 4 00 47 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/193)


ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -117,47 +165,67 @@ extension ContactCollectionView: ContactViewDelegate {
...
+        case .accessibilityExtraExtraLarge:
+            height = 178
+        case .accessibilityExtraExtraExtraLarge:
+            height = 194
         default:
-            return 135.0
+            break
         }
+
+        if size == .small {
```
rdeassis, line 225: What's the reason behind this logic? I'm wondering why the small size subtracts a fixed constant even though the heights are variable.
Maybe worth a comment explaining it?

ios/FluentUI/People Picker/Persona.swift
```diff
@@ -62,4 +76,20 @@ open class PersonaData: NSObject, Persona {
         self.presence = presence
         self.color = color
     }
+
+    @objc public init(surname: String = "",
```
rdeassis, line 80: nit: Xcode style comments on public initializers?

ios/FluentUI/People Picker/ContactCollectionView.swift
MathieuKavalec: The difference is height is constant since we remove a label for the smaller version.

rdeassis: Got it. I think it's worth a comment explaining it and/or moving the 38 value to a named constant (subtitleLabelHeight for example).

MathieuKavalec: Will add a comment. As far as a named constraint, the idea here was to merge all these constants into one big contactHeight method.

MathieuKavalec:
> LGTM aside from the `surname` parameter naming in `Persona.swift`.

Agreed, I will rename this to first name.

ios/FluentUI/People Picker/ContactCollectionView.swift
rdeassis: Just to be explicit on the reason why it's taking out from the height only on the smaller size.
I think this should be more like: "// remove/compensate the hight of the subtitle label that is not added/used in the smaller size".

MathieuKavalec:
> Signing off on the PR. I think it'd be nice to be a bit more descriptive on the comment I unresolved.
> Other than that, looks good. Thanks!

Thanks, I'll add a more detailed comment!

PR #194, opened 2020-08-18
Added new expand resizingBehaviour and passthrough presentationBackground

aamittal1989:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

DrawerController.swift
- Added new expand resizingBehaviour and passthrough presentationBackground
- Handling of non-dismissable behaviour in case of new expand resizingBehaviour
- Added new delegateeTouchView where touch events should be delegated in case of passthrough presentationBackground

DrawerPresentationController.swift
- Added DelegatorTouchView to delegate the touch events outside drawer to delegateeTouchView in case of passthrough presentationBackground

DrawerDemoController.swift
- Added new button for launching drawer with new expand resizingBehaviour and passthrough presentationBackground

### Verification

Added new button for launching drawer with new expand resizingBehaviour and passthrough presentationBackground in drawer demo screen.

Did sanity around Tooltip and HUD as they are using TouchForwardingView (which I modified in this PR).

| No expand resizingBehaviour  | New expand resizingBehaviour where dismissal of drawer is restricted with drag |
| No passthrough presentationBackground  | New passthrough presentationBackground which can delegate the touch events to any delegateeTouchView provided by presenting view controller |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/194)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -892,7 +909,7 @@ extension DrawerController: UIViewControllerTransitioningDelegate {
             if #available(iOS 13.0, *) {
                 useNavigationBarBackgroundColor = (direction.isVertical && source.traitCollection.userInterfaceLevel == .elevated)
             }
-            return DrawerPresentationController(presentedViewController: presented, presenting: presenting, source: source, sourceObject: sourceView ?? barButtonItem, presentationOrigin: presentationOrigin, presentationDirection: direction, presentationOffset: presentationOffset, presentationBackground: presentationBackground, adjustHeightForKeyboard: adjustsHeightForKeyboard, shouldUseWindowFullWidthInLandscape: shouldUseWindowFullWidthInLandscape)
+            return DrawerPresentationController(presentedViewController: presented, presenting: presenting, source: source, sourceObject: sourceView ?? barButtonItem, presentationOrigin: presentationOrigin, presentationDirection: direction, presentationOffset: presentationOffset, presentationBackground: presentationBackground, adjustHeightForKeyboard: adjustsHeightForKeyboard, shouldUseWindowFullWidthInLandscape: shouldUseWindowFullWidthInLandscape, delegateeTouchView: delegateeTouchView)
```
harrieshin, line 912: why is delegateeeTouchView needed? can't we just use TouchForwardingView to forward the touch but not handling touches within the presentationcontroller? this will help us handle darkening the background when bottomsheet is expanded?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -53,8 +54,16 @@ class DrawerDemoController: DemoController {
...
+    private func presentDrawer(sourceView: UIView? = nil, barButtonItem: UIBarButtonItem? = nil, presentationOrigin: CGFloat = -1, presentationDirection: DrawerPresentationDirection, presentationStyle: DrawerPresentationStyle = .automatic, presentationOffset: CGFloat = 0, presentationBackground: DrawerPresentationBackground = .black, presentingGesture: UIPanGestureRecognizer? = nil, permittedArrowDirections: UIPopoverArrowDirection = [.left, .right], contentController: UIViewController? = nil, contentView: UIView? = nil, resizingBehavior: DrawerResizingBehavior = .none, adjustHeightForKeyboard: Bool = false, animated: Bool = true, customWidth: Bool = false, delegateeTouchView: UIView? = nil) -> DrawerController {
```
rdeassis, line 66: This func has way too many params. We should at least put one per line to make this more readable.

ios/FluentUI/Drawer/DrawerController.swift
rdeassis, line 912: Same here regarding one parameter per line? It's very hard to read all in one line.

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -736,6 +739,20 @@ open class DrawerController: UIViewController {
                 offset = min(offset, 0)
             }
         }
+
+        // This resizingBehavior(expand) avoid the user to dismiss the drawer
+        if resizingBehavior == .expand && presentationDirection == .up {
```
rdeassis, line 744: What about the scenario where the presentation direction is .down?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -32,6 +32,7 @@ class DrawerDemoController: DemoController {
 
         addTitle(text: "Bottom Drawer")
         container.addArrangedSubview(createButton(title: "Show resizable", action: #selector(showBottomDrawerButtonTapped)))
+        container.addArrangedSubview(createButton(title: "Show resizable with expand state and passthrough presentation background", action: #selector(showBottomDrawerWithExpandStateAndPassthroughBackgroundButtonTapped)))
```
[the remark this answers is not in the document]
aamittal1989: Yeah make sense! Actually, i tried that behaviour in landscape mode. 
There is one more option that I add some blank views to make the page scrollable. Or should I add different view controller to demo this type of drawer?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -53,8 +54,16 @@ class DrawerDemoController: DemoController {
         view.addGestureRecognizer(trailingEdgeGesture)
     }
 
+    override func viewWillDisappear(_ animated: Bool) {
```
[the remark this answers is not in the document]
aamittal1989: When we make the background passthrough, then user can cancel the presenting view controller from navigation bar. So this logic will dismiss the DrawerContoller presented on top of it. Without this logic DrawerController will remain on the screen.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
aamittal1989: Sure.

ios/FluentUI/Drawer/DrawerController.swift
aamittal1989: delegateeTouchView is required so that Presented view can decide where the touch events should get delegated (NavigationViewController root view or presented view controller root view etc).

**can't we just use TouchForwardingView to forward the touch but not handling touches within the presentationcontroller?**

Can you please elaborate?  Are you asking to use TouchForwardingView instead of DelegatorTouchView? In this case, we want to delegate the touch events from outside tappable area of the drawer to the delegateeTouchView passed by presentedController. DelegateeTouchView also should get the touch events in the respective subview where the user tries to touch.

**this will help us handle darkening the background when bottomsheet is expanded?**
Right now in case of passThrough background, we want the background to be transparent.

ios/FluentUI/Drawer/DrawerController.swift
aamittal1989: My bad, I will handle it. 

Actually tried the scenario with down direction. In that case, offsetWithResistance is going beyond 30 (maximum offset to dismiss). So I am thinking to make the offset as 0 in case of down direction like handled in case of expand. Thoughts?

ios/FluentUI/Drawer/DrawerController.swift
harrieshin: I saw design mocks of bottom sheet that when it is shown in its full extent it shows dark background. My suggestion is that control will have its passthrough uiview. TouchForwardingUIView by default doesn't handle "open override func point(inside point: CGPoint, with event: UIEvent?) -> Bool" in theory, whatever UIView that is shown by the client should receive the event.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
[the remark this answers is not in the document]
rdeassis: I think adding a new view controller that with the purpose of showcasing the interaction is optimal. This way you prevent the tap on other buttons to display the other drawer demos. And the new view controller can be more optimized to show that capability.

ios/FluentUI/Drawer/DrawerController.swift
rdeassis: Why is the offsetWithResistance is going beyond 30?
I'm not sure I understand what you meant by "make the offset as 0 in case of down direction like handled in case of expand"...

aamittal1989: Because in case of .down direction, offset is going beyond 300. I debugged in case .dismiss behavior as well. We are not using offsetWithResistance in case of .down direction.

ios/FluentUI/Drawer/DrawerController.swift
aamittal1989: Did changes for same. Please check.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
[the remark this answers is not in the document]
aamittal1989: Added new view controller. Please check.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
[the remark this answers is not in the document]
aamittal1989: Removed this code and moved to new view controller.

ios/FluentUI/Drawer/DrawerController.swift
aamittal1989: Fixed!

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -143,6 +144,10 @@ class DrawerDemoController: DemoController {
         presentDrawer(sourceView: sender, presentationDirection: .up, contentView: containerForActionViews(), resizingBehavior: .dismissOrExpand)
     }
 
+    @objc private func showBottomDrawerWithExpandStateAndPassthroughBackgroundButtonTapped(sender: UIButton) {
+        navigationController?.pushViewController(PassThroughDrawerDemoController(), animated: true)
```
harrieshin, line 148: In darkmode: can you make sure it looks good for passthrough drawercontroller? right now it is hard to tell between background and drawercontroller.

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -22,38 +22,46 @@ class DrawerPresentationController: UIPresentationController {
...
         // Workaround for a bug in iOS: if the resizing handle happens to be in the middle of the backgroundView, VoiceOver will send touch event to it (according to the backgroundView's accessibilityActivationPoint) even though it's not parented in backgroundView or even interactable - this will prevent backgroundView from receiving touch and dismissing controller
+
+        view.gestureRecognizers = [UITapGestureRecognizer(target: self, action: #selector(handleBackgroundViewTapped(_:)))]
         view.onAccessibilityActivate = { [unowned self] in
             self.presentingViewController.dismiss(animated: true)
         }
         return view
     }()
+
     private lazy var dimmingView: DimmingView = {
```
harrieshin, line 65: any we need to change in the dimmingView bound to make sure the rounded corners of the drawerControlller's view is shown?

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -207,6 +208,8 @@ open class DrawerController: UIViewController {
     }
     @objc open var presentationBackground: DrawerPresentationBackground = .black
 
+    @objc open weak var passThroughView: UIView?
```
harrieshin, line 211: there needs to be documentation about this variable

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -143,6 +144,10 @@ class DrawerDemoController: DemoController {
         presentDrawer(sourceView: sender, presentationDirection: .up, contentView: containerForActionViews(), resizingBehavior: .dismissOrExpand)
     }
 
+    @objc private func showBottomDrawerWithExpandStateAndPassthroughBackgroundButtonTapped(sender: UIButton) {
```
rdeassis, line 147: In landscape mode it is also hard to interact with the underlying view controller. There is just a minimal gap that allows the interaction.

One potential solution I can think of for this scenario is to make the 4 buttons a grid (2x2) instead of a stack of 4 buttons.

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -736,6 +739,20 @@ open class DrawerController: UIViewController {
...
+        // This resizingBehavior(expand) avoid the user to dismiss the drawer
+        if resizingBehavior == .expand {
+            // This is to avoid drawer to get dismissed when user drag the drawer from expanded state to normal state
+            if isExpanded {
+                offset = max(offset, -(originalDrawerHeight - normalDrawerHeight))
+            }
+
+            // This is to restrict dismiss behaviour of drawer
+            if !isExpanded && offset < 0 && !mayResizeViaContentScrolling {
+                offset = presentationDirection == .up ? offsetWithResistance(for: offset) : 0
```
rdeassis, line 752: I think it's worth adding a Top Drawer scenario to validate this behavior.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -32,6 +32,7 @@ class DrawerDemoController: DemoController {
 
         addTitle(text: "Bottom Drawer")
         container.addArrangedSubview(createButton(title: "Show resizable", action: #selector(showBottomDrawerButtonTapped)))
+        container.addArrangedSubview(createButton(title: "Show resizable with expand state and passthrough behavior", action: #selector(showBottomDrawerWithExpandStateAndPassthroughBackgroundButtonTapped)))
```
rdeassis, line 35: How about just "Show with underlying interactable content view" as the description?

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -22,38 +22,46 @@ class DrawerPresentationController: UIPresentationController {
     private let presentationOrigin: CGFloat?
     private let presentationOffset: CGFloat
     private let presentationBackground: DrawerPresentationBackground
+	private weak var passThroughView: UIView?
 
-    init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?, source: UIViewController, sourceObject: Any?, presentationOrigin: CGFloat?, presentationDirection: DrawerPresentationDirection, presentationOffset: CGFloat, presentationBackground: DrawerPresentationBackground, adjustHeightForKeyboard: Bool, shouldUseWindowFullWidthInLandscape: Bool) {
+	init(presentedViewController: UIViewController, presenting presentingViewController: UIViewController?, source: UIViewController, sourceObject: Any?, presentationOrigin: CGFloat?, presentationDirection: DrawerPresentationDirection, presentationOffset: CGFloat, presentationBackground: DrawerPresentationBackground, adjustHeightForKeyboard: Bool, shouldUseWindowFullWidthInLandscape: Bool, passThroughView: UIView?) {
```
rdeassis, line 27: Can you make a single parameter per line?

ios/FluentUI/Drawer/DrawerPresentationController.swift
aamittal1989: I didn't change the dimming view. Its bounds are already changing. Please let me know if you meant something else.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
aamittal1989: Fixed!

ios/FluentUI/Drawer/DrawerController.swift
aamittal1989: Fixed!

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
aamittal1989: Fixed!

ios/FluentUI/Drawer/DrawerPresentationController.swift
harrieshin: Screen Shot 2020-08-31 at 10 30 30 AM
There isn't much shadow around the drawer that I can't see the rounded corner or the difference between the passthrough view and the drawercontroller view.

ios/FluentUI/Controls/TouchForwardingView.swift
```diff
@@ -27,11 +27,33 @@ open class TouchForwardingView: UIView {
...
             }
         }
         return super.point(inside: point, with: event)
     }
+
+    open override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        var hitTestView = super.hitTest(point, with: event)
+
+        // If onPassthroughViewTouches is non-nil then don't pass the touch events to passthroughView view
+        if onPassthroughViewTouches != nil {
```
rdeassis, line 46: Can you use guard ... else here instead?

From the swift guidelines (https://github.com/microsoft/swift-guide/blob/gh-pages/EarlyReturn.md):
"It's recommended to use guard ... else statements and early exits on top of a function to validate parameters or other variables."

ios/FluentUI/Controls/TouchForwardingView.swift
```diff
@@ -27,11 +27,33 @@ open class TouchForwardingView: UIView {
...
                 return false
             }
         }
         return super.point(inside: point, with: event)
     }
+
+    open override func hitTest(_ point: CGPoint, with event: UIEvent?) -> UIView? {
+        var hitTestView = super.hitTest(point, with: event)
+
+        // If onPassthroughViewTouches is non-nil then don't pass the touch events to passthroughView view
```
rdeassis, line 45: This comment can provide a more meaningful explanation. I suggest something like:

"Prevents the underlying (passthrough) view from getting touch events if there's custom code already handling it (onPassthroughViewTouches closure)"

rdeassis: Same in line 30.

ios/FluentUI/Controls/TouchForwardingView.swift
harrieshin: also it should be more like let foo = fooo {} rather than nil check

harrieshin:
in order for this PR to merge to master , the control needs to be somewhat usable in iPad. right now anytime there is content size change like split mode to compact mode we dismiss the drawer which defeats the purpose of this control.

ios/FluentUI/Controls/TouchForwardingView.swift
aamittal1989: @harrieshin  @rdeassis we don't want to use onPassthroughViewTouches that's why used if check here.

I can use guard for early return something like this:
guard onPassthroughViewTouches == nil else {
        return hitTestView
}

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerDemoController.swift
```diff
@@ -16,6 +16,7 @@ class DrawerDemoController: DemoController {
 
         addTitle(text: "Top Drawer")
         container.addArrangedSubview(createButton(title: "Show resizable", action: #selector(showTopDrawerButtonTapped)))
+        container.addArrangedSubview(createButton(title: "Show non dismissable", action: #selector(showTopDrawerNotDismissableButtonTapped)))
         container.addArrangedSubview(createButton(title: "Show with no animation", action: #selector(showTopDrawerNotAnimatedButtonTapped)))
```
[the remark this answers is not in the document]
aamittal1989: Looks like its an existing issue. Same problem is happening for resizable top drawer as well.

aamittal1989:
> in order for this PR to merge to master , the control needs to be somewhat usable in iPad. right now anytime there is content size change like split mode to compact mode we dismiss the drawer which defeats the purpose of this control.

In case of Ipad if we have presentation style as slideover, I think then this mode should be usable. Thoughts?

ios/FluentUI/Drawer/DrawerPresentationController.swift
aamittal1989: There is DrawerShadowView which is being used in the drawer for shadow. Is that the right place to do the changes for shadow?

aamittal1989: I checked, DrawerShadowView is supposed to provide the shadow to the bottom sheet. Looks like adding below line in one of the commit (4cf4800) break the shadow view.
**contentView.clipsToBounds = true** 

**4cf4800**: add a white mask to MSDrawerPresentationController so that during presentation it won't slide over the navigation bar.

Right now shadowView is the part of contentview (where we are setting clipToBounds to be true). What we can do is to add shadowView in the container view. It fixes the issue. Thoughts?

ios/FluentUI/Controls/TouchForwardingView.swift
rdeassis: Yeah. I think that the fact that you don't need onPassthroughViewTouches in the logic of this method, it's okay to do the nil check, but still with the guard statement as you mentioned:

        guard onPassthroughViewTouches == nil else {
            return hitTestView
        }

ios/FluentUI/Drawer/DrawerController.swift
rdeassis: I think the offset going beyond the resizingThreshold is not the issue per se. We could write logic around that.
The problem is that the drawer is presented from a custom base and the resistance is currently just moving the entire drawer on top of the content underneath (instead of "compressing" the drawer to make it look like it was being dismissed into its custom base).

resistance_custom_base

We can treat that as a separate issue since it's not introduced by these changes.

ios/FluentUI/Controls/TouchForwardingView.swift
aamittal1989: fixed!

ios/FluentUI/Controls/TouchForwardingView.swift
aamittal1989: Fixed!

ios/FluentUI/Drawer/DrawerPresentationController.swift
aamittal1989: Fixed!

rdeassis (commented):
I think the shadow is still looking a bit odd because it appears abruptly after the presentation animation is finished.... The drawer also does not have a shadow while the animation is happening....

I have experimented some other options and I still believe that a good route to explore is to add the shadowview to the content view and set the owner as the presentedViewController...
That approach will have the contentView clipping the shadow without any additional changes, but you can work with the content view height to be increased and change the frame of the presentedViewController to offset the additional height.
 
This way you present the drawer at the same height and still not clipping the shadow on the top.

You can offset the additional contentView frame height by changing the Y position on the presentedViewRectPresented func in the DrawerTransitionAnimator and set the same Y position on dismissalTransitionWillBegin for the dismissal to be smooth....

ios/FluentUI/Drawer/DrawerShadowView.swift
```diff
@@ -9,7 +9,22 @@ class DrawerShadowView: UIView {
     private struct Constants {
         static let shadowRadius: CGFloat = 4
         static let shadowOffset: CGFloat = 2
-        static let shadowOpacity: Float = 0.05
+        static let shadowOpacity: Float = 0.1
```
rdeassis, line 12: Is there any design guidance on what the value of the opacity should be?
It's good to check with them if we're changing the value.

ios/FluentUI.Demo/FluentUI.Demo.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
```diff
@@ -0,0 +1,8 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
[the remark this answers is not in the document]
aamittal1989: It was not intentional. Let me remove it in next iteration. Thanks for pointing out.

ios/FluentUI/Drawer/DrawerShadowView.swift
aamittal1989: I got go-ahead from design for it.

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -883,14 +888,14 @@ open class DrawerController: UIViewController {
 extension DrawerController: UIViewControllerTransitioningDelegate {
     public func animationController(forPresented presented: UIViewController, presenting: UIViewController, source: UIViewController) -> UIViewControllerAnimatedTransitioning? {
         if presentationStyle(for: source) == .slideover {
-            return DrawerTransitionAnimator(presenting: true, presentationDirection: presentationDirection(for: source.view))
+            return DrawerTransitionAnimator(presenting: true, presentationDirection: presentationDirection(for: source.view), shadowOffset: shadowOffset)
```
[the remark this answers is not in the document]
aamittal1989: Fixed!

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -169,7 +172,9 @@ class DrawerPresentationController: UIPresentationController {
         case resize
     }
 
-    override var frameOfPresentedViewInContainerView: CGRect { return contentView.frame }
+    // As content view is clipped 'clipsToBounds = true' to prevent any animation bug sliding over the navigation bar, shadow was also getting clipped.
```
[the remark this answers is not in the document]
aamittal1989: Fixed!

ios/FluentUI.Demo/FluentUI.Demo.xcworkspace/xcshareddata/IDEWorkspaceChecks.plist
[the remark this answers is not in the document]
aamittal1989: Fixed!

ios/FluentUI/Drawer/DrawerShadowView.swift
aamittal1989: This change got reverted!

rdeassis (approved):

PR #196, opened 2020-08-19
Allow clients to anchor popovers on titlebar avatar, navigation tab buttons

jpalmr:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Expose `UIView` for:
* `PersonaView` inside `NavigationBar`
* `TabBarItemView`s inside `TabBarView`

so clients can anchor `UIPopoverPresentationController`s on them.

### Verification

Manually validated by presenting popovers in client application. There is no change to the functionality or visual appearance of any controls.

### Pull request checklist

This PR has considered:

None of the above are applicable

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/196)


ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -70,6 +70,10 @@ class LargeTitleView: UIView {
 
     var onAvatarTapped: (() -> Void)? // called in response to a tap on the MSAvatarView
 
+    var avatarUIView: UIView? {
```
harrieshin, line 73: this is incorrect. largetitleview has 2 avatar views currently (avatarView, smallMorphingAvatarView). Small one when it is collapsed header, and large one.

ios/FluentUI/Tab Bar/TabBarView.swift
```diff
@@ -64,6 +64,10 @@ open class TabBarView: UIView {
         }
     }
 
+    @objc public func itemView(item: TabBarItem) -> UIView? {
```
harrieshin, line 67: why are you creating extra function? why won't you just expose the private one?

jpalmr: I tried this first but got "Method cannot be declared public because its result uses an internal type" - `class TabBarItemView` is not `public` - should we make it public it or keep this `UIView?` wrapper?

ios/FluentUI/Navigation/Views/LargeTitleView.swift
jpalmr: is only one of these views visible at a time? should this return the currently visible one?

jpalmr: Addressed this in the most recent commit, also validated that in both expanded and compact mode the `UIPopoverPresentationController` anchors to the correct view

harrieshin: "avatarUIView" doesn't sound very swift friendly name. Also, there might not be any avatar view visible depending on how navigation bar is set up

jpalmr: agreed, unfortunately can't use `avatarView` since it collides with the name of the internal property. if both avatars are nil this will return nil as intended. is there a case where avatar data is set but it is not visible?

harrieshin: you would have to test it. it isn't the exact scenario but in demo app > navigationcontroller > "show without avatar" > "select" button. You can imagine that there are scenarios where client lets users to do cell selection operation while hiding avatar view. either it is a var with different name(like visiableAvatarView) or i suggest making a function like visibleAvatar()->UIView?  to make it clearer that it will return nil when showsProfileButton is false.

jpalmr: great points, will test invisible avatar scenario and push changes

jpalmr: i've validated nil is returned when avatar is unset/nil + when cell selection/search bar focus hides the title view

PR #201, opened 2020-08-24
SideTabBar spacing should not only rely on safe layout guides

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Currently, the SideTabBar layout assumes that safe layout guides are not at the edges of the screen.
This causes the layout to look bad on older devices that don't have spacing for safe layout guides.
Let's set the minimum top and bottom spacing according to the spacing in the red lines for the side tab bar.
In case the safe layout guides have a lot of space, let's keep the old constraints that add space relative to the layout guides. We'll lower the priority of the constraints.

### Verification
- Tested on many different iPad devices with/without bezels.
- Tested on iPhone (even though SideTabBar is not meant to be used on iPhone)

Before:
before

After:
after

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/201)

### DIFF
--- a/ios/FluentUI/Tab Bar/SideTabBar.swift
+++ b/ios/FluentUI/Tab Bar/SideTabBar.swift
@@ -177,6 +177,9 @@ open class SideTabBar: UIView {
         }
 
         if let avatarView = avatarView {
+            // The avatar view's distance from the top of the side tab bar depends on safe layout guides.
+            // There is a minimum spacing. If the layout guide spacing is large than the minimum spacing,
+            // then the spacing will be layoutGuideSpacing + safeTopSpacing.
             let topSafeConstraint = avatarView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: Constants.avatarViewSafeTopSpacing)
             topSafeConstraint.priority = .defaultHigh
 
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -132,16 +132,18 @@ open class SideTabBar: UIView {
...
-        static let topItemSize: CGFloat = 28.0
-        static let bottomItemSize: CGFloat = 24.0
-        static let badgeTopSectionPadding: CGFloat = 2.0
-        static let badgeBottomSectionPadding: CGFloat = 4.0
+        static let viewWidth: CGFloat = 62
+        static let avatarViewSafeTopSpacing: CGFloat = 18
+		static let avatarViewMinTopSpacing: CGFloat = 36
+        static let avatarViewTopStackViewSpacing: CGFloat = 34
+        static let bottomStackViewSafeSpacing: CGFloat = 14
+		static let bottomStackViewMinSpacing: CGFloat = 24
```
harrieshin, line 140: nit: is there inconsistency in indentation?

MathieuKavalec: fixed

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -175,21 +177,29 @@ open class SideTabBar: UIView {
         }
 
         if let avatarView = avatarView {
+            let topSafeConstraint = avatarView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: Constants.avatarViewSafeTopSpacing)
```
harrieshin, line 180: can we comment why you are doing this? i am not sure it is obvious at first sight.

PR #202, opened 2020-08-25
Top accessory view for navigation bar

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add support for a top accessory view in the FluentUI navigation bar.
For now, this top accessory view will always be centered and uses a new class called NavigationBarTopAccessoryViewAttributes to configure its layout. It should be relatively straight forward in the future to add more to this class to configure the position of the top accessory view differently.

The immediate need for the top accessory view is to support a SearchBar that can move between the top section of the navigation bar and the default bottom section depending on the screen's width.
The demo controller implements this behavior with a new option called "Show with top search bar for large screen width".

To support the search bar scenario, we also need to add a property in SearchBar to prevent it from hiding the navigation bar when a search is in progress. We only want to hide the navigation bar if the SearchBar is in the default bottom section.

Side change:
When the custom UINavigationItem properties are changed, we can reflect those change in the navigation bar by sending/handling a notification. Otherwise, changes to these properties are not reflected until the view controller is re-presented.

### Verification

- Tested on iPhone and iPad.
- Test iPad compact/regular size transitions.
- Make sure that nothing has changed in the navigation bar when the top accessory view is nil.
- Start a search while the search bar is in the top section. Reduce the size of the window. Observe that the search bar moves to the bottom section and maintains the same state. Make sure that the navigation bar's title hides as expected. 
- Test the opposite transition from the bottom section to top section.

Screen Shot 2020-08-24 at 10 30 02 PM

Untitled

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/202)

### DIFF
--- a/ios/FluentUI/Controls/SearchBar.swift
+++ b/ios/FluentUI/Controls/SearchBar.swift
@@ -156,9 +156,9 @@ open class SearchBar: UIView {
         didSet {
             if oldValue != hidesNavigationBarDuringSearch && isActive {
                 if hidesNavigationBarDuringSearch {
-                    hideNavigationBar()
+                    hideNavigationBar(animated: false)
                 } else {
-                    unhideNavigationBar()
+                    unhideNavigationBar(animated: false)
                 }
             }
         }
@@ -287,7 +287,11 @@ open class SearchBar: UIView {
         if hidesNavigationBarDuringSearch {
             originalIsNavigationBarHidden = navigationController?.isNavigationBarHidden ?? false
 
-            hideNavigationBar()
+            // Using delayed async to work around a bug on iOS when it restores responder status for the text field when controller appears (due to navigation controller's pop action) even though text field resigned responder status before a detail controller was pushed
+            let isTransitioning = navigationController?.transitionCoordinator != nil
+            DispatchQueue.main.asyncAfter(deadline: .now() + (isTransitioning ? Constants.navigationBarTransitionHidingDelay : 0)) {
+                self.hideNavigationBar(animated: true)
+            }
         }
 
         isActive = true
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Controls/SearchBar.swift
```diff
@@ -287,7 +287,11 @@ open class SearchBar: UIView {
         if hidesNavigationBarDuringSearch {
             originalIsNavigationBarHidden = navigationController?.isNavigationBarHidden ?? false
```
harrieshin, line 289: so why did the delayasync get moved?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -192,7 +214,9 @@ open class NavigationBar: UINavigationBar {
     private let contentStackView = ContentStackView() //used to contain the various custom UI Elements
     private let rightBarButtonItemsStackView = UIStackView()
     private let leftBarButtonItemsStackView = UIStackView()
-    private let spacerView = UIView() //defines the space between the left and right barbuttonitems stack
+    private let leadingSpacerView = UIView()
```
harrieshin, line 217: we should still have comments. don't think it is that obvious. Also, why have private let for leadingspacerview but create it locally for trailingspacerview?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -398,12 +462,29 @@ open class NavigationBar: UINavigationBar {
...
         updateColors(for: actualItem)
         showsLargeTitle = navigationItem.usesLargeTitle
         updateShadow(for: navigationItem)
+        updateTopAccessoryView(for: navigationItem)
+
+        NotificationCenter.default.removeObserver(self,
+                                                  name: UINavigationItem.navigationItemDidChangeNotification,
+                                                  object: nil)
+
+        NotificationCenter.default.addObserver(self,
```
harrieshin, line 484: I am bit confusing. what are we lisenting on top of leftBarButtonItemObserver, rightbarButtonItemsObserver etc?

ios/FluentUI/Controls/SearchBar.swift
MathieuKavalec: The delay async is only needed when the user cancels/exits search using the UI. It is not needed when setting hidesNavigationBarDuringSearch since the hide navigation bar isn't animated.

ios/FluentUI/Navigation/NavigationBar.swift
MathieuKavalec: To be consistent, I'll add trailingspacerview here. I'll also add back the comment. The reason why I didn't include trailingspacerview was because it isn't referenced anywhere else than the method where it is added to the view hierarchy.

ios/FluentUI/Navigation/NavigationBar.swift
MathieuKavalec: Notifications or observers end up doing the same thing here.
Because we are using property observers for other things, I update the logic to use those instead of notifications.
I tested that each property updates correctly after being set to new values after the nav bar is presented.

PR #204, opened 2020-08-26
Custom padding for table view cells

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add properties to allow customization of the leading and trailing padding of TableViewCells.
This will make it a lot easier to customize the cell insets of a table view that uses TableViewCells from FluentUI.
The UITableView contentInset property cannot achieve this, and it's possible that the header views and section headers don't want this inset.

### Verification

- Test on iPhone and iPad.
- Test iPad compact/regular size transitions.
- Customize the padding on the fly and make sure that the cells update.
- Make sure that nothing changed with the padding is not customized.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/204)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -202,11 +202,6 @@ open class TableViewCell: UITableViewCell {
     }
 
     private struct Constants {
-        static let horizontalSpacing: CGFloat = 16
-
-        static let paddingLeading: CGFloat = horizontalSpacing
-        static let paddingTrailing: CGFloat = horizontalSpacing
-
         static let labelAccessoryViewMarginLeading: CGFloat = 8
         static let labelAccessoryViewMarginTrailing: CGFloat = 8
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -283,14 +287,16 @@ open class TableViewCell: UITableViewCell {
...
+    @objc public class func height(title: String, subtitle: String = "", footer: String = "", titleLeadingAccessoryView: UIView? = nil, titleTrailingAccessoryView: UIView? = nil, subtitleLeadingAccessoryView: UIView? = nil, subtitleTrailingAccessoryView: UIView? = nil, footerLeadingAccessoryView: UIView? = nil, footerTrailingAccessoryView: UIView? = nil, customViewSize: CustomViewSize = .default, customAccessoryView: UIView? = nil, accessoryType: TableViewCellAccessoryType = .none, titleNumberOfLines: Int = 1, subtitleNumberOfLines: Int = 1, footerNumberOfLines: Int = 1, customAccessoryViewExtendsToEdge: Bool = false, containerWidth: CGFloat = .greatestFiniteMagnitude, isInSelectionMode: Bool = false, paddingLeading: CGFloat = defaultPaddingLeading, paddingTrailing: CGFloat = defaultPaddingTrailing) -> CGFloat {
```
rdeassis, line 293: This func has a lot of params. It'd be more readable if you put each one on a single line.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -247,19 +242,28 @@ open class TableViewCell: UITableViewCell {
     /// A constant representing the number of lines for a label in which no change will be made when the `preferredContentSizeCategory` returns a size greater than `.large`.
     @objc public static let defaultNumberOfLinesForLargerDynamicType: Int = -1
 
+    /// The default horizontal spacing in the cell.
+    @objc public static let horizontalSpacing: CGFloat = 16
+
```
rdeassis, line 247: Like I mentioned on the other review: do we need to expose those default values as part of the API?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -332,8 +338,10 @@ open class TableViewCell: UITableViewCell {
...
+    @objc public class func preferredWidth(title: String, subtitle: String = "", footer: String = "", titleLeadingAccessoryView: UIView? = nil, titleTrailingAccessoryView: UIView? = nil, subtitleLeadingAccessoryView: UIView? = nil, subtitleTrailingAccessoryView: UIView? = nil, footerLeadingAccessoryView: UIView? = nil, footerTrailingAccessoryView: UIView? = nil, customViewSize: CustomViewSize = .default, customAccessoryView: UIView? = nil, accessoryType: TableViewCellAccessoryType = .none, customAccessoryViewExtendsToEdge: Bool = false, isInSelectionMode: Bool = false, paddingLeading: CGFloat = defaultPaddingLeading, paddingTrailing: CGFloat = defaultPaddingTrailing) -> CGFloat {
```
rdeassis, line 344: Same here: one parameter per line would make this way easier to read.

ios/FluentUI/Table View/TableViewCell.swift
MathieuKavalec: The idea is to allow the customer to reset the properties to their default values.

MathieuKavalec: Also, in this case, the default values do need to be exposed since they are default params for the static functions above.

PR #206, opened 2020-08-26
Allow customization of NavigationBar's content margin

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The content margin of the navigation bar often varies based on the design. 
FluentUI needs to be more flexible and allow the customization of its NavigationBar's content margins (leading and trailing).

### Verification

- Test on iPhone and iPad (compact/regular).
- Set custom content margins and make sure that they are respected.
- Set the custom content margins after the navigation bar is presented. Make sure that the layout updates correctly.
- Make sure that nothing changed when the content margins are not customized.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/206)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -118,9 +118,6 @@ open class NavigationBar: UINavigationBar {
         static let leftBarButtonItemLeadingMargin: CGFloat = 8
         static let rightBarButtonItemHorizontalPadding: CGFloat = 10
 
-        static let contentLeadingMargin: CGFloat = 8
-        static let contentTrailingMargin: CGFloat = 6
-
         static let obscuringAnimationDuration: TimeInterval = 0.12
         static let revealingAnimationDuration: TimeInterval = 0.25
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -186,6 +183,30 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+     /// The navigation bar's default leading content margin.
+    @objc public static let defaultContentLeadingMargin: CGFloat = 8
```
rdeassis, line 187: Do the default values need to be public? Wouldn't they just be the value set to the properties themselves?

MathieuKavalec: Yes, they do need to be public.
The idea is to allow the customer to reset these properties to their default values.
For example, on iPad, we want to customize the nav bar's margins only when the screen is greater a certain width. Otherwise, we want to use the default values so we have to switch between a custom value and the default value.

PR #213, opened 2020-09-01
[FluentUI vNext - Design Token System] Add sGen source code to vNext branch so it can be further modified

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

sGen.app is the tool that converts YAML files representing design tokens into Swift code.
Currently the source code of this tool used by Stardust lives on Skype repo.

This PR aims to just add the sGen code as is so we can start modifying it to serve the requirements of FluentUI v-next.

We'll wait on the OSS compliance approval before merging this change to the vnext-prototype branch.

### Verification

Built the tool locally after granting execution permission on the Pods-sgen-frameworks.sh script.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/213)


anandrajeswaran:
Do we actually want a copy of these sources in the repo or do we want to just consume it as a dependency through cocoapods/carthage/eventually SPM? It looks like we're also trying to commit all the pods of this tool which we probably don't want to do?

rdeassis:
> Do we actually want a copy of these sources in the repo or do we want to just consume it as a dependency through cocoapods/carthage/eventually SPM? It looks like we're also trying to commit all the pods of this tool which we probably don't want to do?

I believe there's value in having the Generator's source code in FluentUI's repository if we're allowing the consumers to customize the tokens and potentially the code generated by the tool.

We discussed about it in the engineering sync and the idea is to fork what we have in the skype repository and start making the changes from that version. Which is why there are a lot of files we don't need and will start pruning from here.

rdeassis:
Closing this pull request because a newer version of sGen (from the symbolAsset branch) is available.
A new pull request was created: #252

PR #219, opened 2020-09-03
Added new property shimmersLeafViews to ShimmerView

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We want to use the ShimmerView in FluentUI Rect Native. React Native however, uses flexbox for it's layout, which likes to wrap views inside views inside views for it's layout. This is incompatible with ShimmerView's current implementation, which only shimmers the top level subviews of it's containerview.

In order to support React Native's flexbox, we need to make 2 changes:

1. Add a new property "shimmersLeafViews", which will determine whether we shimmer the top level subviews, or the leaf views of containerview's view heirarchy. This will be by default off, and can be set to true when we create an instance for React Native.

2. Allow the containerView to be self. ShimmerView normally expects to have a container view passed to it on initialization, who's subviews it will shimmer. Because of how RCTViewManager in React Native works, the subviews in React's DOM will become subviews of ShimmerView, so we need to allow shimmerview's containerview to be "self". In other words, Shimmer view should be able to shimmer it's own subviews. This is simply dont with a nil coalesce check.

### Verification

Copied the ShimmerView implementation to FURN to test locally in my [FluentUI React Native PR ](https://github.com/microsoft/fluentui-react-native/pull/372) to wrap shimmer.

Made sure no changes happened to the default shimmer (when `shimmersLeafViews = false`)

Screen Shot 2020-09-03 at 9 14 01 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/219)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -9,7 +9,7 @@ import FluentUI
 // MARK: TableViewCellDemoController
 
 class TableViewCellDemoController: DemoController {
-    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections 
+    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections
 
     private var isGrouped: Bool = false {
         didSet {
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -9,7 +9,7 @@ import FluentUI
 // MARK: TableViewCellDemoController
 
 class TableViewCellDemoController: DemoController {
-    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections
+    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections 
 
     private var isGrouped: Bool = false {
         didSet {
--- a/ios/FluentUI/Shimmer/ShimmerView.swift
+++ b/ios/FluentUI/Shimmer/ShimmerView.swift
@@ -306,7 +306,7 @@ open class ShimmerView: UIView {
 	private static let defaultUsesTextHeightForLabels: Bool = false
 	private static let defaultLabelHeight: CGFloat = 11
 
-	private let defaultShimmersLeafViews: Bool = false
+	private static let defaultShimmersLeafViews: Bool = false
 }
 
 public extension Colors {
--- a/ios/FluentUI/Shimmer/ShimmerView.swift
+++ b/ios/FluentUI/Shimmer/ShimmerView.swift
@@ -32,7 +32,7 @@ open class ShimmerView: UIView {
 		}
 	}
 
-	/// the wirth of the gradient in the animation
+	/// the width of the gradient in the animation
 	@objc open var shimmerWidth: CGFloat = defaultWidth {
 		didSet {
 			setNeedsLayout()
@@ -109,6 +109,13 @@ open class ShimmerView: UIView {
 		}
 	}
 
+	/// Determines whether we shimmer the top level subviews, or the leaf nodes of the view heirarchy
+	@objc open var shimmersLeafViews: Bool = defaultShimmersLeafViews {
+		didSet {
+			setNeedsLayout()
+		}
+	}
+
 	/// Optional synchronizer to sync multiple shimmer views
 	@objc open weak var animationSynchronizer: AnimationSynchronizerProtocol?
 
@@ -145,14 +152,12 @@ open class ShimmerView: UIView {
 
 		updateViewCoverLayers()
 
-		guard let containerView = containerView else {
-			return
-		}
+		let viewToCover = containerView ?? self
 
-		shimmeringLayer.frame = CGRect(x: -containerView.frame.width,
+		shimmeringLayer.frame = CGRect(x: -viewToCover.frame.width,
 									   y: 0.0,
-									   width: containerView.bounds.width + 2 * containerView.frame.width,
-									   height: containerView.frame.height)
+									   width: viewToCover.bounds.width + 2 * viewToCover.frame.width,
+									   height: viewToCover.frame.height)
 
 		updateShimmeringLayer()
 		updateShimmeringAnimation()
@@ -165,13 +170,19 @@ open class ShimmerView: UIView {
 
 	/// Update the frame of each layer covering views in the containerView
 	func updateViewCoverLayers() {
-		guard let containerView = containerView else {
-			return
-		}
+		let viewToCover = containerView ?? self
 
 		viewCoverLayers.forEach { $0.removeFromSuperlayer() }
 
-		let subviews = Set(containerView.subviews).subtracting(Set(excludedViews))
+		let subviews: Set<UIView> = {
+			if shimmersLeafViews {
+				var leaves = [UIView]()
+				searchLeaves(in: viewToCover, output: &leaves)
+				return Set(leaves).subtracting(Set(excludedViews))
+			} else {
+				return Set(viewToCover.subviews).subtracting(Set(excludedViews))
+			}
+		}()
 
 		viewCoverLayers = subviews.filter({ !$0.isHidden && !($0 is ShimmerView) }).map { subview in
 			let coverLayer = CALayer()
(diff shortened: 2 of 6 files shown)


Saadnajmi:
@HeyImChris as an FYI, shimmer's view manager would now look like this:
```
@objc(MSFShimmerViewManager)
class ShimmerViewManager: RCTViewManager {
	
  override func view() -> UIView! {
	let shimmerView = ShimmerView()
	shimmerView.shimmersLeafViews = true
	return shimmerView
  }

  override class func requiresMainQueueSetup() -> Bool {
	return true
  }
}
```

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -145,14 +152,12 @@ open class ShimmerView: UIView {
 
 		updateViewCoverLayers()
 
-		guard let containerView = containerView else {
-			return
-		}
+		let viewToCover = containerView ?? self
```
Saadnajmi, line 155: containerView is set on init, and we can't do `containerView = self` inside an init, so I do this instead

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -165,13 +170,19 @@ open class ShimmerView: UIView {
...
-		let subviews = Set(containerView.subviews).subtracting(Set(excludedViews))
+		let subviews: Set<UIView> = {
+			if shimmersLeafViews {
+				var leaves = [UIView]()
+				searchLeaves(in: viewToCover, output: &leaves)
+				return Set(leaves).subtracting(Set(excludedViews))
+			} else {
+				return Set(viewToCover.subviews).subtracting(Set(excludedViews))
+			}
+		}()
```
Saadnajmi, line 185: yay for a let with a boolean on the inside! This is instead of writing a super weird looking ternary statement, or making subviews be a var 😄

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -272,6 +283,30 @@ open class ShimmerView: UIView {
...
+	private static let defaultSpeed: CGFloat = 350
+	private static let defaultDelay: TimeInterval = 0.4
+
+	private static let defaultViewTintColor: UIColor = Colors.Shimmer.tint
+	private static let defaultCornerRadius: CGFloat = 4.0
+	private static let defaultLabelCornerRadius: CGFloat = 2.0
+	private static let defaultUsesTextHeightForLabels: Bool = false
+	private static let defaultLabelHeight: CGFloat = 11
+
+	private let defaultShimmersLeafViews: Bool = false
```
Saadnajmi, line 309: SwiftLint doesn't like fileprivate, even though that's what our Microsoft Swift Guide says to use. Which do we change?
https://github.com/microsoft/swift-guide/blob/gh-pages/FileContentOrdering.md

Also, I might need to make these public so I can expose them as constants to React Natice

harrieshin: yeah update the lint!

ios/FluentUI/Shimmer/ShimmerView.swift
harrieshin: this is bit of behavioral change right? what are the native cases that containerView could be nil?

harrieshin: should we just check shimmersLeafViews then viewToCover = self?

ios/FluentUI/Shimmer/ShimmerView.swift
harrieshin, line 309: something about this variable name makes me think it is going to be array of ivies rather than bool....

ios/FluentUI/Shimmer/ShimmerView.swift
Saadnajmi: Theoretically you could have a containerView, AND have shimmersLeafViews = true, so I don't think we would want that.

This is a behavioral change. I don't think there was ever a use case for shimmerView without a containerview. ShimmerLinesView does that, and remains unaffected

ios/FluentUI/Shimmer/ShimmerView.swift
Saadnajmi: I don't particularly like the constants prefix. I wasn't sure if we decided to no longer use the `struct Constants` pattern anymore, but that would look somewhat cleaner

ios/FluentUI.Demo/FluentUI.Demo/Demos/ShimmerLinesViewDemoController.swift
```diff
@@ -5,11 +5,61 @@
 
 import UIKit
 
-class ShimmerLinesViewDemoController: DemoController {
-    override func viewDidLoad() {
-        super.viewDidLoad()
+class ShimmerViewDemoController: DemoController {
+	override func viewDidLoad() {
+		super.viewDidLoad()
```
HeyImChris, line 10: Is there a linter we're supposed to run? Looks like it used to be tabs but your settings have it set to spaces

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
```diff
@@ -9,7 +9,7 @@ import FluentUI
 // MARK: TableViewCellDemoController
 
 class TableViewCellDemoController: DemoController {
-    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections
+    let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections
```
HeyImChris, line 12: nit- extraneous space

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -165,13 +170,19 @@ open class ShimmerView: UIView {
...
+		let viewToCover = containerView ?? self
 
 		viewCoverLayers.forEach { $0.removeFromSuperlayer() }
 
-		let subviews = Set(containerView.subviews).subtracting(Set(excludedViews))
+		let subviews: Set<UIView> = {
+			if shimmersLeafViews {
+				var leaves = [UIView]()
+				searchLeaves(in: viewToCover, output: &leaves)
+				return Set(leaves).subtracting(Set(excludedViews))
```
HeyImChris, line 181: Shouldn't have early returns

HeyImChris (changes requested):

ios/FluentUI/Shimmer/ShimmerView.swift
Saadnajmi: This is not a return of the entire function. This is a closure that lets me evaluate `subviews` as a `let` instead of a var

ios/FluentUI.Demo/FluentUI.Demo/Demos/ShimmerLinesViewDemoController.swift
```diff
@@ -5,11 +5,61 @@
...
 
-        // Default Shimmer
-        container.addArrangedSubview(ShimmerLinesView())
-    }
+		let contentView = { () -> UIStackView in
+			let label1 = UILabel()
```
harrieshin, line 13: Label() instead of UILabel() to make sure larger text size ....would also increase the shimmer re-sizes correctly?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ShimmerLinesViewDemoController.swift
Saadnajmi: We have SwiftLint on iOS that runs if it's installed. I do get lots of lint warnings. However, I don't think we ever specificied a SwiftLint formatting file, and just use the defaults. 

Furthermore, I'm pretty sure most engineers don't have SwiftLint installed AKA it never runs

Saadnajmi: Proper linting on FUA is still an active work item IMO, we should use Swift-Lint by apple like what @cy255 did earlier

ios/FluentUI.Demo/FluentUI.Demo/Demos/ShimmerLinesViewDemoController.swift
Saadnajmi: Nice catch =). 

Screen Shot 2020-09-08 at 9 57 20 AM

HeyImChris (approved):

PR #220, opened 2020-09-04
Mechanism to auto-launch a demo controller on boot of the demo app

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

It can be repetitive to scroll down the list of demo controllers every time the demo app reboots, especially when working on a specific demo controller.
Let's add a simple mechanism that can be used to setup a demo controller for auto-launch on boot.
The last opened demo controller will automatically launch.
We will save whichever demo controller is opened to the user defaults.
If the user navigates out of a demo controller, then nothing will auto-launch on the next boot.

### Verification

Launch first time, nothing auto-launches.
Open a demo controller, reboot the app, make sure that this demo controller automatically reopens.
Open a demo controller, close it, reboot the app, make sure that nothing auto-launches.
Open a demo controller, close it, open a different demo controller, reboot the app, make sure that this last demo controller automatically reopens.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/220)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -8,9 +8,6 @@ import FluentUI
 
 class DemoListViewController: UITableViewController {
 
-    /// Set this to a non-zero value to automatically launch the demo controller at this index when the demo app launches.
-    private var autoLaunchDemoIndex: Int = 0
-
     static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
         if let colorProvider = window as? ColorProviding {
             Colors.setProvider(provider: colorProvider, for: window)
@@ -59,9 +56,15 @@ class DemoListViewController: UITableViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
-        if autoLaunchDemoIndex != 0 {
-            tableView(tableView, didSelectRowAt: IndexPath(row: autoLaunchDemoIndex - 1, section: 0))
-            autoLaunchDemoIndex = 0
+        if didAutoLaunch {
+            UserDefaults.standard.set(0, forKey: DemoListViewController.lastDemoControllerKey)
+        } else {
+            let lastDemoController = UserDefaults.standard.integer(forKey: DemoListViewController.lastDemoControllerKey)
+            if lastDemoController != 0 {
+                tableView(tableView, didSelectRowAt: IndexPath(row: lastDemoController - 1, section: 0))
+            }
+
+            didAutoLaunch = true
         }
     }
 
@@ -83,7 +86,11 @@ class DemoListViewController: UITableViewController {
         let demoController = demo.controllerClass.init(nibName: nil, bundle: nil)
         demoController.title = demo.title
         navigationController?.pushViewController(demoController, animated: true)
+
+        UserDefaults.standard.set(indexPath.row + 1, forKey: DemoListViewController.lastDemoControllerKey)
     }
 
     let cellReuseIdentifier: String = "TableViewCell"
+    private var didAutoLaunch: Bool = false
+    private static let lastDemoControllerKey: String = "LastDemoController"
 }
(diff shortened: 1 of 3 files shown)


anandrajeswaran:
Thoughts on just using NSUserDefaults so the default behavior is to launch to the previously opened view controller on that device? Obviously iOS is a bit more tricky than macOS in case you're testing on different simulators, but someone (I think it was @Saadnajmi) added that behavior for fluentui-macos and I love it.

Saadnajmi:
@anandrajeswaran pretty sure it was @markavitale

MathieuKavalec:
> Thoughts on just using NSUserDefaults so the default behavior is to launch to the previously opened view controller on that device? Obviously iOS is a bit more tricky than macOS in case you're testing on different simulators, but someone (I think it was @Saadnajmi) added that behavior for fluentui-macos and I love it.

I like the idea. I'll go ahead and implement that.

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -53,6 +53,21 @@ class DemoListViewController: UITableViewController {
...
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        if didAutoLaunch {
+            UserDefaults.standard.set(0, forKey: DemoListViewController.lastDemoControllerKey)
+        } else {
+            let lastDemoController = UserDefaults.standard.integer(forKey: DemoListViewController.lastDemoControllerKey)
+            if lastDemoController != 0 {
+                tableView(tableView, didSelectRowAt: IndexPath(row: lastDemoController - 1, section: 0))
```
rdeassis, line 64: I'm thinking about the out of bounds scenario here and/or when demos list changes.

I think storing the title in the demos list could be a better approach because:
 1. you can check whether the item is present on the list (in case the app updates and the item was renamed or removed)
 2. you can remove the UserDefaults value and not load any other demo view controller in case it was not found
 3. If new items are added to the list (before the current default index), finding it by the title won't cause a different one to be presented.

Then, instead of calling func tableView(_ tableView: UITableView, didSelectRowAt indexPath: IndexPath), you can just use its logic once you have the title of the demo controller you want.

MathieuKavalec: Updated to using the demo controller name instead.

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -71,7 +92,11 @@ class DemoListViewController: UITableViewController {
         let demoController = demo.controllerClass.init(nibName: nil, bundle: nil)
         demoController.title = demo.title
         navigationController?.pushViewController(demoController, animated: true)
+
+        UserDefaults.standard.set(demo.title, forKey: DemoListViewController.lastDemoControllerKey)
     }
 
     let cellReuseIdentifier: String = "TableViewCell"
+    private var didAutoLaunch: Bool = false
```
rdeassis, line 100: nit:
This variable tracks the whether or not it's the first launch... not if the auto launch of a view controller happened.
Can we rename it to isFirstLaunch and default it to true (inverting the logic on viewDidAppear)?

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -53,6 +53,27 @@ class DemoListViewController: UITableViewController {
...
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        if didAutoLaunch {
+            UserDefaults.standard.set(nil, forKey: DemoListViewController.lastDemoControllerKey)
+        } else {
+            let lastDemoController = UserDefaults.standard.string(forKey: DemoListViewController.lastDemoControllerKey)
+
+            if lastDemoController?.count ?? -1 > 0 {
```
rdeassis, line 64: How about this logic to find the index? (considering the isFirstLaunch variable name)

        if !isFirstLaunch {
            UserDefaults.standard.set(nil, forKey: DemoListViewController.lastDemoControllerKey)
        } else {
            isFirstLaunch = false

            if let lastDemoController = UserDefaults.standard.string(forKey: DemoListViewController.lastDemoControllerKey),
               let index = demos.firstIndex(where: { $0.title == lastDemoController }) {
                tableView(tableView, didSelectRowAt: IndexPath(row: index, section: 0))
            }
        }

rdeassis (approved):

PR #224, opened 2020-09-08
Pointer interaction for TableViewCell accessory view buttons

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Accessory view buttons in TableViewCell should use the default pointer interaction for UIButtons.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/224)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1488,6 +1488,11 @@ internal class TableViewCellAccessoryView: UIView {
         button.accessibilityLabel = "Accessibility.TableViewCell.MoreActions.Label".localized
         button.accessibilityHint = "Accessibility.TableViewCell.MoreActions.Hint".localized
         button.addTarget(self, action: #selector(handleOnAccessoryTapped), for: .touchUpInside)
+
+        if #available(iOS 13.4, *) {
+            button.isPointerInteractionEnabled = true
+        }
+
         return button
     }()
 


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1488,6 +1488,11 @@ internal class TableViewCellAccessoryView: UIView {
         button.accessibilityLabel = "Accessibility.TableViewCell.MoreActions.Label".localized
         button.accessibilityHint = "Accessibility.TableViewCell.MoreActions.Hint".localized
         button.addTarget(self, action: #selector(handleOnAccessoryTapped), for: .touchUpInside)
+
+        if #available(iOS 13.4, *) {
+            button.isPointerInteractionEnabled = true
```
markavitale, line 1493: Is this actually part of our design for table view cells? Note that Apple doesn't actually have these enabled for any table view cells in the system that I can find:

Tested Contacts and Settings, "Information" disclosure button, "Call" and "text" buttons in Contacts, and Switch accessory views.

markavitale (changes requested):

ios/FluentUI/Table View/TableViewCell.swift
markavitale: Basically, the code looks fine if we want to be different than the system on this particular interaction, but I'm not convinced that's intentional.

MathieuKavalec: The pointer interaction will only be enabled on out "..." button in the TableViewCell. 
This is what the designers have asked for. There's a Teams discussion about this, let me know if you want to join the conversation.
For this particular interaction, we are okay with being different from the system.

markavitale (approved):

PR #225, opened 2020-09-09
DrawerController should respect safe area when presenting vertically.

rdeassis:
The Drawer controller currently uses all the window width to present itself by default.
On devices with the notch (landscape mode) this can be a problem because the entire window width does consider the safe area.

no_safe_area

The design guidelines for some cases define that the entire drawer should limit itself to the safe area. So, we need to add support for that.

### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change adds this behavior, but the default is still set to take the entire width space. The client code is able to configure the drawer to respect the safe areas through a property.

### Verification

| Screen Shot 2020-09-09 at 12 47 26 PM | Screen Shot 2020-09-09 at 12 49 10 PM |
| Screen Shot 2020-09-09 at 12 47 42 PM | Screen Shot 2020-09-09 at 12 49 19 PM |

Screen Shot 2020-09-09 at 12 51 20 PM
Screen Shot 2020-09-09 at 12 49 54 PM
Screen Shot 2020-09-09 at 12 50 03 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/225)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -334,6 +334,9 @@ open class DrawerController: UIViewController {
     /// For `vertical` presentation shown when horizontal size is `.compact`, the content width will be the full width of the presenting window. If set to false, the `preferredContentSize.width` will be used for calculation in landscape mode.
     @objc open var shouldUseWindowFullWidthInLandscape: Bool = true
 
+    /// Limits the full window width to its safe area for `vertical` presentation.
+    @objc open var shouldRespectSafeAreaForWindowFullWidth: Bool = true
+
     // Override to provide the preferred size based on specifics of the concrete drawer subclass (see popup menu, for example)
     open var preferredContentWidth: CGFloat { return 0 }
     open var preferredContentHeight: CGFloat { return 0 }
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -334,6 +334,9 @@ open class DrawerController: UIViewController {
     /// For `vertical` presentation shown when horizontal size is `.compact`, the content width will be the full width of the presenting window. If set to false, the `preferredContentSize.width` will be used for calculation in landscape mode.
     @objc open var shouldUseWindowFullWidthInLandscape: Bool = true
 
+    /// Limits the full window width to its safe area for `vertical` presentation.
+    @objc open var shouldRespectSafeAreaForWindowFullWidth: Bool = true
```
harrieshin, line 338: we should change this default to false. because lot of client depends on it.

harrieshin: and lets add a demo case.

harrieshin (approved):

PR #226, opened 2020-09-09
Pointer interactions for PillButtonBar

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Implement pointer interaction for PillButtonBar.
Followed Apple's guidelines from the WWDC demo: https://developer.apple.com/videos/play/wwdc2020/10093/.
For group of continuous items, it is recommended to add the pointer interaction on the container view, the stack view in this case.
Then, add a pointer region for each item, PillButton in this case.
The lift style is used for the interaction style.
When a pill button is hovered, its background color needs to be slightly adjusted by being a bit darker in light mode and a bit lighter in dark mode. Otherwise, the highlight effect makes the pill button less visible.

Also included in this change: set the avatar view target preview to the window. That way, no matter in which subview the avatar view will end up, its lift effect will never get cutoff since the preview will render in the top superview, the window.

demo

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/226)

### DIFF
--- a/ios/FluentUI/People Picker/AvatarView.swift
+++ b/ios/FluentUI/People Picker/AvatarView.swift
@@ -829,15 +829,15 @@ class OverflowAvatarView: AvatarView {
 extension AvatarView: UIPointerInteractionDelegate {
     @available(iOS 13.4, *)
     public func pointerInteraction(_ interaction: UIPointerInteraction, styleFor region: UIPointerRegion) -> UIPointerStyle? {
-        if let superview = window {
-            let frame = superview.convert(bounds, from: self)
-            let target = UIPreviewTarget(container: superview, center: CGPoint(x: frame.midX, y: frame.midY))
-            let preview = UITargetedPreview(view: self, parameters: UIPreviewParameters(), target: target)
-            let pointerEffect = UIPointerEffect.lift(preview)
-
-            return UIPointerStyle(effect: pointerEffect, shape: nil)
+        guard let superview = window else {
+            return nil
         }
 
-        return nil
+        let frame = superview.convert(bounds, from: self)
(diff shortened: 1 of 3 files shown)


ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -829,7 +829,15 @@ class OverflowAvatarView: AvatarView {
 extension AvatarView: UIPointerInteractionDelegate {
     @available(iOS 13.4, *)
     public func pointerInteraction(_ interaction: UIPointerInteraction, styleFor region: UIPointerRegion) -> UIPointerStyle? {
-        let pointerEffect = UIPointerEffect.lift(.init(view: self))
-        return UIPointerStyle(effect: pointerEffect, shape: nil)
+        if let superview = window {
```
rdeassis, line 832: Use guard .. else for early returns?

guard let superview = window else {
        return nil
 }

ios/FluentUI/Pill Button Bar/PillButtonBar.swift
```diff
@@ -450,3 +455,62 @@ open class PillButtonBar: UIScrollView {
...
+            }
+        }
+
+        return region
+    }
+
+    @available(iOS 13.4, *)
+    public func pointerInteraction(_ interaction: UIPointerInteraction, styleFor region: UIPointerRegion) -> UIPointerStyle? {
+        let index = region.identifier as! Int
+        if let superview = window, index < buttons.count {
```
rdeassis, line 484: Same here for the use of guard .. else for early returns...

rdeassis (approved):
Just minor nits to use guard ... else for early returns.

PR #227, opened 2020-09-12
Support attributed strings for Label Control

chiuam:
### Platforms Impacted
- [X] iOS
- [] macOS

### Description of changes
This change fixes a bug in label control where attributed string text color is disregarded. When a label attributed string is set with custom UIColor, the colorStyle overrides the custom color because we set the label color after the UIWindow is loaded. The fix is to set the text color before UIWindow is loaded by making it an optional.
### Verification
Created a test case for attributed strings in the demo app.
| Before                                                                  | After                                                                 |
| image | image |
| image | image|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/227)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/LabelDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/LabelDemoController.swift
@@ -26,6 +26,11 @@ class LabelDemoController: DemoController {
 
         container.addArrangedSubview(UIView())  // spacer
 
+        addLabel(text: "Test attributed strings", style: .headline, colorStyle: .regular).textAlignment = .center
+        for colorStyle in TextColorStyle.allCases {
+            addAttributedStringLabel(text: "Test attributed strings", substring: "attributed", style: .footnote, colorStyle: colorStyle)
+        }
+
         NotificationCenter.default.addObserver(self, selector: #selector(handleContentSizeCategoryDidChange), name: UIContentSizeCategory.didChangeNotification, object: nil)
     }
 
@@ -39,6 +44,21 @@ class LabelDemoController: DemoController {
         }
         container.addArrangedSubview(label)
         return label
+    }
+
+    @discardableResult
+    func addAttributedStringLabel(text: String, substring: String, style: TextStyle, colorStyle: TextColorStyle) -> Label {
+        let label = Label(style: style, colorStyle: colorStyle)
+        label.numberOfLines = 0
+        let range = (text as NSString).range(of: substring)
+        let attributedString = NSMutableAttributedString(string:text)
+
+        attributedString.addAttribute(NSAttributedString.Key.foregroundColor, value: UIColor.red , range: range)
+        label.attributedText = attributedString
+
+        container.addArrangedSubview(label)
+        return label
+
     }
 
     @objc private func handleContentSizeCategoryDidChange() {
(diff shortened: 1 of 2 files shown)


anandrajeswaran (changes requested):
Is there actually a timing aspect to this or can we just avoiding setting textColor if attributedText != nil in the existing implementation without any of the other changes to keep all logic centralized and avoid making the non-optional color func optional.

Aside: do we really need to support attributedText in FluentUI labels? Feels like in that case someone just shouldn't be using a FluentUI label if they don't want FluentUI styling?

harrieshin:
> Is there actually a timing aspect to this or can we just avoiding setting textColor if attributedText != nil in the existing implementation without any of the other changes to keep all logic centralized and avoid making the non-optional color func optional.
> 
> Aside: do we really need to support attributedText in FluentUI labels? Feels like in that case someone just shouldn't be using a FluentUI label if they don't want FluentUI styling?

yeah I was also back and forth about it. I think the usage case that Amanda is trying to fix was mostly fluent colors except she wants to insert hyperlink (which I don't think the demo case doesn't really highlight correctly.) We we decide to not support attributed string for all the label style then we do need to write some documentation about it similar to UILabel

chiuam:
> > Is there actually a timing aspect to this or can we just avoiding setting textColor if attributedText != nil in the existing implementation without any of the other changes to keep all logic centralized and avoid making the non-optional color func optional.
> > Aside: do we really need to support attributedText in FluentUI labels? Feels like in that case someone just shouldn't be using a FluentUI label if they don't want FluentUI styling?
> 
> yeah I was also back and forth about it. I think the usage case that Amanda is trying to fix was mostly fluent colors except she wants to insert hyperlink (which I don't think the demo case doesn't really highlight correctly.) We we decide to not support attributed string for all the label style then we do need to write some documentation about it similar to UILabel



> > Is there actually a timing aspect to this or can we just avoiding setting textColor if attributedText != nil in the existing implementation without any of the other changes to keep all logic centralized and avoid making the non-optional color func optional.
> > Aside: do we really need to support attributedText in FluentUI labels? Feels like in that case someone just shouldn't be using a FluentUI label if they don't want FluentUI styling?
> 
> yeah I was also back and forth about it. I think the usage case that Amanda is trying to fix was mostly fluent colors except she wants to insert hyperlink (which I don't think the demo case doesn't really highlight correctly.) We we decide to not support attributed string for all the label style then we do need to write some documentation about it similar to UILabel

chiuam:
> > Is there actually a timing aspect to this or can we just avoiding setting textColor if attributedText != nil in the existing implementation without any of the other changes to keep all logic centralized and avoid making the non-optional color func optional.
> > Aside: do we really need to support attributedText in FluentUI labels? Feels like in that case someone just shouldn't be using a FluentUI label if they don't want FluentUI styling?
> 
> yeah I was also back and forth about it. I think the usage case that Amanda is trying to fix was mostly fluent colors except she wants to insert hyperlink (which I don't think the demo case doesn't really highlight correctly.) We we decide to not support attributed string for all the label style then we do need to write some documentation about it similar to UILabel

The issue is that the text color doesn't get set until we call didMoveToWindow, which is after we initialize the label with custom styling. I agree that FluentUI Labels shouldn't support attributed strings, so I'll make a fix on the client side instead. 

@harrieshin , yeah we should update the documentation.

PR #231, opened 2020-09-16
Fix consumption of fluentui-ios (built with Xcode 11) when building with Xcode 12

anandrajeswaran:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
A new swift compiler version is used between Xcode 11 and Xcode 12. When building with Xcode 12 and trying to import the FluentUILib module, the compiler complains about a few things

**NavigationBar.swift**
NavigationBarTopSearchBarAttributes subclasses from NavigationBarTopAccessoryViewAttributes. The superclass has an init with multiple parameters whereas the subclass has an init with no parameters. The subclass does not declare its init as an override even though it is technically overriding from a few levels up in the hierarchy. Consuming the swift interface complains about that, but we can't just mark it as override because its superclass does not have that method.

Ultimately the best solution I can think of is to add a dummy init to NavigationBarTopAccessoryViewAttributes with no parameters so overriding init is defined behavior

**PopupMenuController.swift**
**PopupMenuProtocols.swift**
PopupMenuItemTemplateCell is a protocol requires conformers to be UITableViewCell. Unfortunately the generated swift-interface loses this distinction and just tries to make the protocol inherit from the class which is not ok. That syntax (public protocol Foo where Self:Bar) appears to be new and not quite settled. Let's just skip doing that for now and put the onus on usages to also specify the conformer must be a UITableViewCell. There's only one instance here

### Verification
Manual inspection before/after of swift interface and objC interfaces

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/231)

### DIFF
--- a/ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
@@ -38,7 +38,7 @@ public protocol PopupMenuTemplateItem: AnyObject {
  You can use `UITableViewCell` conforms to this protocol for customization.
 */
 @objc(MSFPopupMenuItemTemplateCell)
-public protocol PopupMenuItemTemplateCell where Self: UITableViewCell {
+public protocol PopupMenuItemTemplateCell {
     /// `PopupMenuController` will notify that one or more items in the list contain image(s)
     @objc var preservesSpaceForImage: Bool { get set }
     /// `PopupMenuController` will notify the custom separatorColor.
--- a/ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
@@ -35,7 +35,7 @@ public protocol PopupMenuTemplateItem: AnyObject {
 /**
 `PopupMenuItemTemplateCell` represents a template cell protocol inside `PopupMenuController`.
  The built-in type is `PopupMenuItemCell`.
- You can use `UITableViewCell` conforms to this protocol for customization.
+ If you are making a custom object conform to this protocol, it must be a `UITableViewCell` for the Pop Up Menu to work properly
 */
 @objc(MSFPopupMenuItemTemplateCell)
 public protocol PopupMenuItemTemplateCell {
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
```diff
@@ -38,7 +38,7 @@ public protocol PopupMenuTemplateItem: AnyObject {
  You can use `UITableViewCell` conforms to this protocol for customization.
 */
 @objc(MSFPopupMenuItemTemplateCell)
-public protocol PopupMenuItemTemplateCell where Self: UITableViewCell {
```
markavitale, line 41: Did we ever figure out why this in particular wasn't valid? It still doesn't quite make sense why this isn't allowed. My best guess is that it doesn't work with the `@objc` attribute, but I haven't found explicit documentation for that limitation anywhere.

ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
```diff
@@ -38,7 +38,7 @@ public protocol PopupMenuTemplateItem: AnyObject {
  You can use `UITableViewCell` conforms to this protocol for customization.
```
markavitale, line 38: Should we improve this documentation now that we don't enforce that it's of type UITableViewCell to explain it needs to be?

markavitale (approved):

ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
anandrajeswaran: Because it emits this code in the swift interface which is invalid. Probably worth a bug on Apple?
```
@objc(MSFPopupMenuItemTemplateCell) public protocol PopupMenuItemTemplateCell : UITableViewCell {
```

ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
anandrajeswaran: I think the comment already tries to say that, but tried to make it more explicit

ios/FluentUI/Popup Menu/PopupMenuProtocols.swift
markavitale: But I guess, I still don't understand why that's invalid. Those are two different ways of expressing essentially the same thing in Swift.

PR #234, opened 2020-09-17
Add an option to set minimum number of actions in file accessory view

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add an option to set minimum number of actions in file accessory view.
This will allow columns to be aligned when the number of actions can vary between rows.
Added some options in the demo controller to test this change.

After:
Screen Shot 2020-09-17 at 2 20 10 PM

Before:
Screen Shot 2020-09-17 at 2 20 12 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/234)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
+++ b/ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
@@ -86,6 +86,7 @@ open class TableViewCellFileAccessoryView: UIView {
 
     /// The minimum count of actions.
     /// If there are fewer actions to display than this count, empty spaces will be reserved for those missing actions.
+    /// This property is useful to align columns between cells that display a different number of actions.
     /// Setting this value too high could result in a broken layout.
     @objc public var minimumActionsCount: UInt = 0 {
         didSet {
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
```diff
@@ -84,6 +84,17 @@ open class TableViewCellFileAccessoryView: UIView {
         }
     }
 
+    /// The minimum count of actions.
```
rdeassis, line 87: I think it's important to mention on the documentation here that this property is used to force the width of the accessory view by adding shimming empty views to it. It will also help explain why the high value may cause a broken layout.

And maybe a recommendation like: "use this property when you have variable action counts across cells' accessory views and need to have them all aligned".

rdeassis (approved):
Just a minor comment on the explanation of the property on the documentation comment.

PR #237, opened 2020-09-18
New property in the file accessory cell view for action buttons that shouldn't offset the columns

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In the file accessory table view cell, clients can specify a set of actions that will be displayed in the trailing part of the cell's view. Clients will want columns to always be aligned between cells. In some cases, there are actions that will rarely display. For example, in Office, we need an action to handle cache errors. These errors are rare. minimumActionsCount is a useful property that can be used to set the minimum count of actions per cell. In this case, we don't want to use minimumActionsCount otherwise we would have to reserve an empty space for an action that rarely shows. This will move the alignment of the columns towards the left. Instead, we want that last action to overlap the columns section. We don't actually want the UIViews to overlap but we want to create that visual effect. 
The solution: add a property that indicates how many actions should overlap the other columns. If this property is set to a value greater than 0, we will recenter the column's contents towards the right based on this property. The bounds of the columns will not change.

### Verification

- Test with large and small screen sizes to test the different spacing configurations.
- Test with only the shared column visible.
- Test with only the date column visible.
- Test with no column visible.
- Test with the shared and date columns visible.
- Test with various configurations of action buttons.
- Test with different values of this new property.
- Make sure that the columns are always aligned when the property is set correctly.

Before:
Screen Shot 2020-09-18 at 2 40 19 PM
Screen Shot 2020-09-18 at 2 40 14 PM
Screen Shot 2020-09-18 at 2 40 22 PM

After:
Screen Shot 2020-09-18 at 2 40 01 PM
Screen Shot 2020-09-18 at 2 40 10 PM
Screen Shot 2020-09-18 at 2 39 58 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/237)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
@@ -92,6 +92,7 @@ class TableViewCellFileAccessoryViewDemoController: DemoController {
         static let cellPaddingThreshold: CGFloat = 768
         static let largeCellPadding: CGFloat = 16
         static let smallCellPadding: CGFloat = 8
+        static let plusMinusButtonWidth: CGFloat = 40
     }
 
     private var topAccessoryViews: [TableViewCellFileAccessoryView] = []
@@ -353,6 +354,26 @@ class TableViewCellFileAccessoryViewDemoController: DemoController {
         }
     }
 
+    private var topActionsOverlap: UInt = 0 {
+        didSet {
+            if oldValue != topActionsOverlap {
+                for accessoryView in topAccessoryViews {
+                    accessoryView.actionsColumnOverlap = topActionsOverlap
+                }
+            }
+        }
+    }
+
+    private var bottomActionsOverlap: UInt = 0 {
+        didSet {
+            if oldValue != bottomActionsOverlap {
+                for accessoryView in bottomAccessoryViews {
+                    accessoryView.actionsColumnOverlap = bottomActionsOverlap
+                }
+            }
+        }
+    }
+
     private lazy var settingsView: UIView = {
         let settingsView = UIStackView(frame: .zero)
         settingsView.axis = .horizontal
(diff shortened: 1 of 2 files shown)


rdeassis (commented):
It seems like these properties are adding a lot of complexity for the calling code to set.

1. What would be the correct approach in case there are more than 2 rarely displayed actions? Increase the minimum actions by one to prevent a potential overlap (or a very short distance) between the text and the action buttons?

2. Couldn't the calling code just set the minimum actions based on whether the rare error icon is going to be displayed for any of the cells in the table?

For example, if the typical number of actions is 3. The calling code set the minimum to 4 actions only if any of the entries has an error, otherwise you keep it to 3.

Screen Shot 2020-09-19 at 12 54 08 PM
Screen Shot 2020-09-19 at 12 54 40 PM

MathieuKavalec:
> It seems like these properties are adding a lot of complexity for the calling code to set.
> 
> 1. What would be the correct approach in case there are more than 2 rarely displayed actions? Increase the minimum actions by one to prevent a potential overlap (or a very short distance) between the text and the action buttons?
> 2. Couldn't the calling code just set the minimum actions based on whether the rare error icon is going to be displayed for any of the cells in the table?
> 
> For example, if the typical number of actions is 3. The calling code set the minimum to 4 actions only if any of the entries has an error, otherwise you keep it to 3.
> 
> Screen Shot 2020-09-19 at 12 54 08 PM
> Screen Shot 2020-09-19 at 12 54 40 PM

Minimum action count and overlap are two different things. Let me explain.
The minimum action count will always guarantee that a number of slots are reserved for actions. This is used when rows can have varying number of actions but we need to keep columns aligned.
The minimum action count should never be changed after the cells have been displayed. It would look silly to shift all columns when an error appears in one of the rows.
The minimum action count will be used once we add the keep offline action. Keep offline will only be shown for some files, and not others. Since this action will be displayed prominently, we will reserve a slot for it and shift the columns towards the left.

The overlap property is not used for alignment columns. It is used to prevent misalignment when a rare action is shown, like the error action. We don't want to reserve a slot for that action. Most users don't have cache errors in their files. If we reserved a slot for that, the columns would look uncentered because we will have an empty space in each row reserved for an action that rarely ever shows.

rdeassis:
> > It seems like these properties are adding a lot of complexity for the calling code to set.
> > 
> > 1. What would be the correct approach in case there are more than 2 rarely displayed actions? Increase the minimum actions by one to prevent a potential overlap (or a very short distance) between the text and the action buttons?
> > 2. Couldn't the calling code just set the minimum actions based on whether the rare error icon is going to be displayed for any of the cells in the table?
> > 
> > For example, if the typical number of actions is 3. The calling code set the minimum to 4 actions only if any of the entries has an error, otherwise you keep it to 3.
> > Screen Shot 2020-09-19 at 12 54 08 PM
> > Screen Shot 2020-09-19 at 12 54 40 PM
> 
> Minimum action count and overlap are two different things. Let me explain.
> The minimum action count will always guarantee that a number of slots are reserved for actions. This is used when rows can have varying number of actions but we need to keep columns aligned.
> The minimum action count should never be changed after the cells have been displayed. It would look silly to shift all columns when an error appears in one of the rows.
> The minimum action count will be used once we add the keep offline action. Keep offline will only be shown for some files, and not others. Since this action will be displayed prominently, we will reserve a slot for it and shift the columns towards the left.
> 
> The overlap property is used to prevent misalignment when a rare action is shown, like the error action. We don't want to reserve a slot for that action. Most users don't have cache errors in their files. If we reserved a slot for that, the columns would look uncentered because we will have an empty space in each row reserved for an action that rarely ever shows.

Yes. I understand the goal of the property. The only missed point of the earlier suggestion I made is that rows can be updated after the table is already presented and display the cache error.

rdeassis (approved):

PR #243, opened 2020-09-23
Disable ShimmerView animation if OS has "reduce motion" accessibility…

rdeassis:
… setting turned ON.

### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The design team requested that for usability reasons, the ShimmerView should not animate when the OS "reduce motion" accessibility setting is turned on.

### Verification

Validated the animation is being executed when the setting is disabled and the shimmer is static when it's turned on:

Screen Shot 2020-09-22 at 9 07 13 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/243)

### DIFF
--- a/ios/FluentUI/Shimmer/ShimmerView.swift
+++ b/ios/FluentUI/Shimmer/ShimmerView.swift
@@ -141,6 +141,11 @@ open class ShimmerView: UIView {
 		self.excludedViews = excludedViews
 		self.animationSynchronizer = animationSynchronizer
 		super.init(frame: CGRect(origin: .zero, size: containerView?.bounds.size ?? .zero))
+
+        NotificationCenter.default.addObserver(self,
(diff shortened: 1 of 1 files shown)


Saadnajmi:
So it freezes with the shimmer in the middle of the control? Not just a uniform grey all around?

Saadnajmi (approved):

rdeassis:
> So it freezes with the shimmer in the middle of the control? Not just a uniform grey all around?

Yes. The design team liked the static gradient.
On a real use case scenario, the user would just top seeing the animation if the setting change is made while the shimmer is animating.

PR #245, opened 2020-09-24
Add concealing shimmer style to ShimmerView

cy255:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added ShimmerStyle to distinguish between revealing shimmer and concealing shimmer style. Concealing shimmer is a shimmer style where the gradient conceals parts of the subviews as it moves leaving most parts of the subviews unblocked. Revealing shimmer is a shimmer style where the gradient reveals parts of the subviews as it moves leaving most parts of the subview blocked.

The current ShimmerView is of the "revealing shimmer" style which works great when used on labels, ShimmeringLines etc. However, it does not work great on images as a shimmering effect because the image contents are mostly blocked under the "revealing shimmer" style. 

### Verification

Added ShimmerStyle related tests to the test app and made sure the shimmering effects are as expected.

A screenshot of the test app showcasing the two different shimmer styles side by side:
Screen Shot 2020-09-24 at 3 55 42 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/245)


cy255:
Fixing white spaces in the next commit.

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -5,6 +5,35 @@
...
+			return 0.4
+        case .revealing:
+			return 0.6
+        }
+    }
+	
+	var defaultTintColor: UIColor {
+		switch self {
+		case .concealing:
+			return .white
```
rdeassis, line 30: This needs to be a constant on the colors with defined values for both light and dark modes.

Screen Shot 2020-09-24 at 11 07 08 PM

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -227,7 +269,7 @@ open class ShimmerView: UIView {
 		let dark = UIColor.black.cgColor
```
rdeassis, line 269: Not part of your change, but I think it'd be nicer to have these light and dark colors in the constants too.

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -227,7 +269,7 @@ open class ShimmerView: UIView {
...
 		let isRTL = effectiveUserInterfaceLayoutDirection == .rightToLeft
 
-		shimmeringLayer.colors = [dark, light, dark]
+		shimmeringLayer.colors = shimmerStyle == .concealing ? [light, dark, light] : [dark, light, dark]
```
rdeassis, line 272: If the light and dark shimmering layer colors are moved to constants, this could be another property of the enum to avoid the ternary operator here?

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -5,6 +5,35 @@
...
 import UIKit
 
+/// Shimmer style can be either concealing or revealing
+/// The style affects the default shimmer alpha value and the default shimmer tint color
+@objc(MSFShimmerViewStyle)
+public enum ShimmerStyle: Int, CaseIterable {
+    /// Concealing shimmer: the gradient conceals parts of the subviews as it moves leaving most parts of the subviews unblocked.
+    case concealing
+
+    /// Revealing shimmer: the gradient reveals parts of the subviews as it moves leaving most parts of the subview blocked.
```
rdeassis, line 15: It seems like the revealing style is more challenging since it has to match the background of the view that is being shimmed. It makes me think that the default tint color should be a property that can be set on the control.

But could also need to support scenarios where the background is a gradient like in OfficeMobile or even image backgrounds.

Do you think we could make a transparent gradient mask and animate it?
I believe it'd be the approach that supports most of the cases.

ios/FluentUI/Shimmer/ShimmerView.swift
cy255: Good point, moved the layer colors to Shimmer Style

ios/FluentUI/Shimmer/ShimmerView.swift
cy255: I think you are referring to concealing shimmer? The current design of concealing shimmer actually does show most of the image by default (regardless of light or dark mode) with transparent gradient layers and a default gradient center color white. The user can set a different color with viewTintColor property as well to best match their images.

ios/FluentUI/Shimmer/ShimmerView.swift
cy255: Moved the dark color to the constant. Light color unfortunately cannot be part of constant because it depends on shimmerAlpha and is created dynamically.

ios/FluentUI/Shimmer/ShimmerView.swift
cy255: Moved the dark color to the constant. Light color unfortunately cannot be part of constant because it depends on shimmerAlpha and is created dynamically. So unfortunately can't remove the ternary operator here.

ios/FluentUI/Shimmer/ShimmerView.swift
cy255: Added dark variants to the demo images and dark mode colors.

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -322,6 +360,8 @@ open class ShimmerView: UIView {
 
 public extension Colors {
 	struct Shimmer {
+		public static var darkGradient = UIColor.black
```
harrieshin, line 363: super nit: ithink it darkGradient: UIColor = .black  would be more consistent with other files

harrieshin (approved):

PR #246, opened 2020-09-25
Navigation bar subtitle and title/subtitle interactions

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Included in this change:
- Subtitle support in NavigationBar. The subtitle in positioned under the title.
- Callbacks to handle tap events on the title and the subtitle.
- Chevron configuration for the title and subtitle.
- A property to disable the behavior that expands the Navigation

Before this change, we only used LargeTitleView in NavigationBar when usesLargeTitle is set to true.
Additionally, LargeTitleView only supports aligning the content along the leading edge.
In this change, LargeTitleView was customized to support centered content and subtitles.
The layout of LargeTitleView was also modified to allow for the presentation of chevrons next to the title and subtitle.

Screen Shot 2020-09-24 at 7 14 51 PM
Screen Shot 2020-09-24 at 7 14 49 PM
Screen Shot 2020-09-24 at 7 14 42 PM
Screen Shot 2020-09-24 at 7 14 40 PM
Screen Shot 2020-09-24 at 10 56 07 PM
Screen Shot 2020-09-24 at 7 14 33 PM
Screen Shot 2020-09-24 at 10 55 57 PM
Screen Shot 2020-09-24 at 7 14 27 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/246)


ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
```diff
@@ -15,6 +30,14 @@ import UIKit
         static var navigationBarShadow: String = "navigationBarShadow"
         static var usesLargeTitle: String = "usesLargeTitle"
         static var customNavigationBarColor: String = "customNavigationBarColor"
+        static var subtitle: String = "subtitle"
+        static var expandsNavigationBarOnTitleAreaTap: String = "expandsNavigationBarOnTitleAreaTap"
+        static var titleChevronStyle: String = "titleChevronStyle"
```
harrieshin, line 35: is it possible to reuse/converge with @objc(MSFTwoLineTitleViewAccessoryType)  public enum AccessoryType: Int {

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -267,16 +280,7 @@ open class NavigationBar: UINavigationBar {
     private let trailingSpacerView = UIView() //defines the trailing space between the left and right barbuttonitems stack
     private var topAccessoryView: UIView?
     private var topAccessoryViewConstraints: [NSLayoutConstraint] = []
-
-    private var showsLargeTitle: Bool = true {
-        didSet {
-            if showsLargeTitle == oldValue {
-                return
-            }
-            updateAccessibilityElements()
```
harrieshin, line 276: no consequences on accessibility updates by deleting these lines?

ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -114,12 +179,164 @@ class LargeTitleView: UIView {
     private var smallMorphingAvatarViewAnimator: UIViewPropertyAnimator? // responsible for animating the transform of smallMorphingAvatarView
     private var smallAnimatorRunningObserver: NSKeyValueObservation? // observes the running property, done to accomplish a "completion" for a pausesOnCompletion = YES animator
 
-    private let titleButton = UIButton() // button used to display the title of the current navigation item
+    private let titleAreaView = UIView(frame: .zero)
```
harrieshin, line 182: i think we should try to reuse TwoLineTitleView. I think the class is intended that only one of the chevron can exists either title or subtitle. It uses MSFEasyTapButton so that the taparea target is much bigger than its icon to match apple HIG

ios/FluentUI/Navigation/NavigationBar.swift
MathieuKavalec: No consequences, no. showsLargeTitle is a private var that is only set in one place. updateAccessibilityElements() now gets called in that other place.

PR #248, opened 2020-09-25
Fluent colors for Mac

chiuam:
### Platforms Impacted
- [X] iOS
- [X] macOS

### Description of changes
Changes:

1. Moved fluent colors to a new asset catalog /apple/Resources/Assets.xcassets
2. Copied iOS color logic to Mac
3. Renamed the gray colors

### Verification
image
image

Created a test pod locally to validate podspec changes.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/248)


harrieshin (approved):

rdeassis (approved):

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+		view = containerView
+	}
+}
+
+class ColorRectView: NSView {
+	
+	var color: NSColor = {
+		let color = NSColor()
+		return color
+	}()
```
anandrajeswaran, line 59: Why is this needed?

macos/FluentUITestViewControllers/TestViewControllers.swift
```diff
@@ -20,4 +20,6 @@ public let testViewControllers = [TestViewController(title: "Avatar View",
 								  TestViewController(title: "Link",
 													 type: TestLinkViewController.self),
 								  TestViewController(title: "Separator",
-													 type: TestSeparatorViewController.self)]
+													 type: TestSeparatorViewController.self),
+								  TestViewController(title: "Color",
```
anandrajeswaran, line 24: Should keep these alphabetized

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+		containerView.addSubview(scrollView)
+		
+		let colorsStackView = NSStackView()
+		colorsStackView.translatesAutoresizingMaskIntoConstraints = false
+		colorsStackView.orientation = .vertical
+		colorsStackView.alignment = .leading
+		
+		for color in Colors.Palette.allCases {
+			let colorView = ColorRectView(color: color.color)
+			colorView.invalidateIntrinsicContentSize()
```
anandrajeswaran, line 24: should not be needed

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -0,0 +1,79 @@
...
+		let documentView = NSView()
+		documentView.translatesAutoresizingMaskIntoConstraints = false
+		scrollView.documentView = documentView
+		documentView.addSubview(colorsStackView)
+		
+		NSLayoutConstraint.activate([
+			containerView.heightAnchor.constraint(equalTo:scrollView.heightAnchor),
+			containerView.widthAnchor.constraint(equalTo:scrollView.widthAnchor),
+
+			colorsStackView.topAnchor.constraint(equalTo: documentView.topAnchor, constant: 10.0),
```
anandrajeswaran, line 44: minor: split the 10 out into a constant - avoid duplication and assign meaning to the value by giving it a name

macos/FluentUI/Colors.swift
```diff
@@ -0,0 +1,304 @@
...
+		
+		public var color: NSColor {
+			if let fluentColor = NSColor(named: "FluentColors/" + self.name, bundle: FluentUIResources.resourceBundle) {
+				return fluentColor
+			} else {
+				preconditionFailure("invalid fluent color")
+			}
+		}
+		
+		public var name: String {
```
anandrajeswaran, line 99: why does name need to be public?

macos/FluentUI/Colors.swift
```diff
@@ -0,0 +1,304 @@
...
+				return "presenceOof"
+			case .presenceUnknown:
+				return "presenceUnknown"
+			}
+		}
+	}
+	
+	// MARK: Physical - grays
+	
+	@objc public static let gray950: NSColor = Palette.gray950.color
```
anandrajeswaran, line 253: wy do we need the duplication of the palette and these colors. Can we just Palette and have clients call that directly?

macos/FluentUI/Colors.swift
```diff
@@ -0,0 +1,304 @@
...
+		Palette.gray40.color,
+		Palette.green20.color,
+		Palette.blueMagenta20.color,
+		Palette.pinkRed10.color,
+		Palette.gray30.color,
+		Palette.blueMagenta30.color,
+		Palette.gray20.color,
+		Palette.cyan30.color,
+		Palette.orange30.color
+	]
```
anandrajeswaran, line 293: shouldn't this be in AvatarView instead?

macos/FluentUI/Colors.swift
```diff
@@ -0,0 +1,304 @@
...
+	/// Used for hyperlinks
+	@objc public static let communicationBlue: NSColor = Palette.communicationBlue.color
+	
+	@objc public static func color(from palette: Palette) -> NSColor {
+		return palette.color
+	}
+}
+
+/// Make palette enum CaseIterable for unit testing purposes
+extension Colors.Palette: CaseIterable {}
```
anandrajeswaran, line 304: Can this happen in test code instead of production code?

macos/FluentUITestViewControllers/TestColorViewController.swift
chiuam: It's a placeholder to fill the rect view.

macos/FluentUITestViewControllers/TestColorViewController.swift
chiuam: I'm creating a rect view and a text view for each color in the palette. What's the alternative way to achieve that?

macos/FluentUI/Colors.swift
chiuam: The test app needs it, though it's unlikely other clients would need it.

macos/FluentUI/Colors.swift
chiuam: I think it's for accessibility?

macos/FluentUI/Colors.swift
chiuam: Agreed, though I'm just gonna remove these for now until they're needed.

macos/FluentUI/Colors.swift
chiuam: yes

chiuam: actually, we might need a workaround for that.

macos/FluentUITestViewControllers/TestColorViewController.swift
anandrajeswaran: Why not just call NSColor() directly? What does the closure get us?

macos/FluentUITestViewControllers/TestColorViewController.swift
anandrajeswaran: Sorry the highlight was off - we do not need invalidateIntrinsicContentSize. The intrinsicContentSize will always be queried on initial layout and only needs to be invalidated if something about the object changes later that could change its intrinsicContentSize

macos/FluentUI/Colors.swift
anandrajeswaran: Sorry not why do we need this specific color - why do we need to expose both this gray950 and Palette.gray950? Those are the same thing?

chiuam: Talked to @harrieshin and @rdeassis. I learned that storing the color ensures we call the resourceBundle only once, which improves perf especially for colors that are frequently used. However, Rafael is working on moving the color palettes to the YAML file, so I'll remove the constants now to align with his work.

anandrajeswaran: Regardless of the move to the yml file, I'd imagine we can ensure good perf and not over-loading the bundle without exposing 2 separate identical API? We can make it an implementation detail of the one API

macos/FluentUI/Colors.swift
harrieshin: it should be needed for Mac AvatarView ....

chiuam: I don't see it being used anywhere in the code...

harrieshin: static let avatarBackgroundColors: [NSColor] = [] in AvatarView.swit should be replace with this new list so that mac and iOS have the same avatar colors.

macos/FluentUI/Colors.swift
anandrajeswaran: For app size every bit counts, can we just get the name using Mirrors in the test app so there's no impact to product code? https://developer.apple.com/documentation/swift/mirror

anandrajeswaran (changes requested):

ios/FluentUI/Core/Colors.swift
```diff
@@ -60,18 +60,18 @@ public final class Colors: NSObject {
...
-        case gray9
-        case gray8
-        case gray7
-        case gray6
-        case gray5
-        case gray4
-        case gray3
-        case gray2
-        case gray1
+        case gray25
```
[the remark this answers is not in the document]
chiuam: yeah, I'll keep them in descending order.

anandrajeswaran (approved):

PR #255, opened 2020-10-08
[sGen code generator] Adding support for FluentUI asset catalog colors.

rdeassis:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

Adding support for FluentUI asset catalog colors in the YAML language (borderColor on the example below).
Currently colors can only be expressed with a hex literal value (like the backgroundColor on the example below).

AP_ButtonVNext:
  borderColor: **NamedColor(communicationBlue)**
  backgroundColor: "#00FF00"

### Verification

Generated swift files from a YAML with the new NamedColor construct and built the generated file.

```
		//MARK: borderColor 
		public var _borderColor: UIColor?
		open func borderColorProperty(_ traitCollection: UITraitCollection? = UIScreen.main.traitCollection) -> UIColor {
			if let override = _borderColor { return override }
			return UIColor(named: "FluentColors/communicationBlue", in: FluentUIFramework.resourceBundle, compatibleWith: nil)!
			}
		public var borderColor: UIColor {
			get { return self.borderColorProperty() }
			set { _borderColor = newValue }
		}
```

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/255)

### DIFF
--- a/tools/sgen/Sources/SGen/Generatable/RhsValue.swift
+++ b/tools/sgen/Sources/SGen/Generatable/RhsValue.swift
@@ -41,6 +41,9 @@ enum RhsValue {
     /// A font value.
     case font(font: Rhs.Font)
     
+    /// A named color in an asset catalog.
+    case namedColor(namedColor: Rhs.NamedColor)
+    
     /// A color value.
     case color(color: Rhs.Color)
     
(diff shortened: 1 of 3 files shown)


harrieshin:
it might be out of scope for this PR but "open func borderColorProperty(_ traitCollection: UITraitCollection? = UIScreen.main.traitCollection" why does it need the trait collection input param?

rdeassis:
> it might be out of scope for this PR but "open func borderColorProperty(_ traitCollection: UITraitCollection? = UIScreen.main.traitCollection" why does it need the trait collection input param?

Yes, I agree it's out of the scope of this PR. This is part of the current Stardust design that is reflected in the sGen tool. This will be revisited once we redesign the StylesheetManager generated class that relies on explicit calls for theme switching.

harrieshin (approved):

PR #258, opened 2020-10-14
Empty tab bar items for inter-item spacing

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added a new initializer in TabBarItem for inter-item spacing items.
These spacing items can be added to TabBarView or SideTabBar to add empty space between two other TabBarItems.
If the spacing TabBarItem is added first, it won't have any effect.
If two spacing TabBarItem are added in a row, their spacing will be combined.

### Verification

- Added empty spacing items in the SideTabBar and the TabBarView to make sure that the spacing works.
- Used the view hierarchy debugger to verify that the spacing is added correctly.
- Added two spacing items in a row. Make sure that their spacing is added.
- Tested on iPad and iPhone.
- Tested RTL.
- Tested with screen rotation and size transitions on iPad.

Screen Shot 2020-10-14 at 12 19 56 PM
Screen Shot 2020-10-14 at 12 14 20 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/258)

### DIFF
--- a/ios/FluentUI/Tab Bar/TabBarItem.swift
+++ b/ios/FluentUI/Tab Bar/TabBarItem.swift
@@ -65,20 +65,6 @@ open class TabBarItem: NSObject {
                   accessibilityLabelBadgeFormatString: nil)
     }
 
-    /// Initializes `TabBarItem` for inter-item spacing.
-    /// This tab bar item does not have an image or title, it is used to add space between two other tab bar items.
-    @objc public init(emptySpace: CGFloat) {
-        self.image = nil
-        self.selectedImage = nil
-        self.title = ""
-        self.largeContentImage = nil
-        self.landscapeImage = nil
-        self.landscapeSelectedImage = nil
-        self.accessibilityLabelBadgeFormatString = nil
-        self.emptySpace = emptySpace
-        super.init()
-    }
-
     /// Initializes `TabBarItem`
     /// - Parameter title: Used for tabbar item view's label and for its accessibilityLabel.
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Tab Bar/TabBarItem.swift
```diff
@@ -65,6 +65,20 @@ open class TabBarItem: NSObject {
                   accessibilityLabelBadgeFormatString: nil)
     }
 
+    /// Initializes `TabBarItem` for inter-item spacing.
+    /// This tab bar item does not have an image or title, it is used to add space between two other tab bar items.
+    @objc public init(emptySpace: CGFloat) {
```
harrieshin, line 70: seems weird that empty space is the only param in initializer

MathieuKavalec: Can you elaborate on that?

harrieshin: so, why aren't we taking a similar approach like UIStackView where you can define the padding per item rather than creating an empty item? initializing tabbaritem with empty space value seems bit weird to me.

MathieuKavalec: We can also take this approach too. Take a look at the latest iteration.

harrieshin (approved):

PR #260, opened 2020-10-15
Fixing height for badgefield when there is text and keyboard is hidden

kiagrawal:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

On keyboard hide in badge field, if there is text in textField, that is ignored and view frame height is calculated without it. Fixing it as list app uses keyboard hide on scroll and the height of UI view gets distorted because of this.

### Verification

Texted in FluentUI using demo

| ezgif com-video-to-gif-4| ezgif com-video-to-gif-5
 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/260)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -453,7 +453,7 @@ open class BadgeField: UIView {
             calculateBadgeFrame(badge: badge, badgeIndex: index, lineIndex: &lineIndex, left: &left, topMargin: 0, boundingWidth: bounds.width)
         }
 
-        let isFirstResponderOrHasTextFieldContent = isFirstResponder || (!isFirstResponder && !textFieldContent.isEmpty)
+        let isFirstResponderOrHasTextFieldContent = isFirstResponder || !textFieldContent.isEmpty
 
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -453,7 +453,9 @@ open class BadgeField: UIView {
             calculateBadgeFrame(badge: badge, badgeIndex: index, lineIndex: &lineIndex, left: &left, topMargin: 0, boundingWidth: bounds.width)
         }
 
-        if isEditable && isFirstResponder && left + Constants.textFieldMinWidth > boundingWidth {
+        let isFirstResponderOrHasTextFieldContent = isFirstResponder || (!isFirstResponder && !textFieldContent.isEmpty)
```
rdeassis, line 456: if the condition is just (isFirstResponder || !textFieldContent.isEmpty), we'd have the same results, right?

If we're evaluating the right side of the || operation, it already means that isFirstResponder is false.

rdeassis (approved):

PR #262, opened 2020-10-19
Prevent text label from being displayed outside of the TabBarItemView

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

TabBarItemView does not constrain its label to its bounds.
If the item's title is long enough, it will be displayed outside of the view's bounds.

We can fix this by limiting the TabBarItemView's content stack view width to its view width.
Additionally, for the SideTabBar, we want the title to be displayed on 2 lines if it doesn't fit on one line, since SideTabBar has less vertical space than TabBarView. We can achieve this by adding a property in TabBarItemView that allows us to modify the number of lines for the title label.

### Verification

- Test on iPhone and iPad.
- Tested iPad compact and regular. 
- Tested size class transitions. Rotation and iPad split-view.
- Tested in SideTabBar and TabBarView.

Before:
Screen Shot 2020-10-19 at 2 49 22 PM
Screen Shot 2020-10-19 at 2 49 35 PM

After:
Screen Shot 2020-10-19 at 2 44 54 PM
Screen Shot 2020-10-19 at 2 45 14 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/262)

### DIFF
--- a/ios/FluentUI/Tab Bar/SideTabBar.swift
+++ b/ios/FluentUI/Tab Bar/SideTabBar.swift
@@ -145,6 +145,7 @@ open class SideTabBar: UIView {
         static let bottomItemSize: CGFloat = 24
         static let badgeTopSectionPadding: CGFloat = 2
         static let badgeBottomSectionPadding: CGFloat = 4
+        static let numberOfTitleLines = 2
     }
 
     private var layoutConstraints: [NSLayoutConstraint] = []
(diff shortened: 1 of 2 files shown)


harrieshin (approved):
:shipit:

PR #271, opened 2020-10-22
Bug fix: Drawer Controller does not behave correctly when being dismissed by gesture

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When presented from a base that is not the edge of the screen, the drawer does not get dismissed in the same way as it was presented (by going under the presentation base).

This issue reproduces only when the drawer is being dragged towards dismissal. Tapping outside of the drawer dismisses it correctly.

Previously the content view that clips the drawer on the custom base dismissal was being moved when the user dragged. The change makes the drawer view itself to be moved while the content view remains in place.

### Verification

Exercised all demo controller cases and ensured no regression was introduced on both portrait and landscape modes.

| drawer_bug | bug_drawer_fixed |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/271)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -145,7 +145,7 @@ class DrawerPresentationController: UIPresentationController {
             // Horizontally presented drawers must be inside containerView in order for device rotation animation to work correctly
             if presentationDirection.isHorizontal {
                 containerView?.addSubview(presentedViewController.view)
-                presentedViewController.view.frame = frameForPresentedViewController(in: contentView.bounds)
+                presentedViewController.view.frame = frameForPresentedViewController(in: frameOfPresentedViewInContainerView)
                 focusElement = containerView
             } else {
                 focusElement = contentView
@@ -303,11 +303,7 @@ class DrawerPresentationController: UIPresentationController {
         contentView.frame = frame
 
         if let presentedView = presentedView {
-            presentedView.frame = frameForPresentedViewController(in: contentView.bounds)
-
-            if presentedView.superview == containerView {
-                presentedView.frame = frameForPresentedViewController(in: frameOfPresentedViewInContainerView)
-            }
+            presentedView.frame = frameForPresentedViewController(in: presentedView.superview == containerView ? frameOfPresentedViewInContainerView : contentView.bounds)
         }
 
         if separator.superview != nil {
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -126,7 +126,7 @@ class DrawerPresentationController: UIPresentationController {
         // For animated presentations presented view must be inside contentView to not slide over navigation bar/toolbar
         if presentingViewController.transitionCoordinator?.isAnimated == true {
             // Avoiding content animation due to showing of the keyboard (when presented view contains the first responder)
-            presentedViewController.view.frame = contentView.bounds
+            presentedViewController.view.frame = frameForPresentedViewController(in: contentView.bounds)
```
harrieshin, line 129: can you just verify side drawers(left/right) when they rotate it doesn't show weird black box when animating?

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -438,13 +436,27 @@ class DrawerPresentationController: UIPresentationController {
         return margins
     }
 
-    private func frameForSeparator(in bounds: CGRect, withThickness thickness: CGFloat) -> CGRect {
-        var bounds = bounds
-        // Separator should stay fixed even when content view is moving - compensating for move
-        if extraContentSize < 0 && extraContentSizeEffectWhenCollapsing == .move {
-            bounds.origin.y += presentationDirection == .down ? -extraContentSize : extraContentSize
+    private func frameForPresentedViewController(in bounds: CGRect) -> CGRect {
+        var frame = CGRect(x: bounds.origin.x,
```
harrieshin, line 440: we need to recreate the CGRect like this?

ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -438,13 +436,27 @@ class DrawerPresentationController: UIPresentationController {
...
+        if presentationDirection.isVertical {
+            frame.origin.y += presentationDirection == .down ? gestureOffset - shadowOffset : -gestureOffset + shadowOffset
+        } else {
+            frame.origin.x += presentationDirection == .fromLeading ? gestureOffset - shadowOffset : -gestureOffset + shadowOffset
         }
 
+        return frame
+    }
+
+    private func frameForSeparator(in bounds: CGRect, withThickness thickness: CGFloat) -> CGRect {
```
harrieshin, line 459: seems like a logic change for separator? is it ok?

rdeassis: Yes. The reason for this change is because the previous logic had the content view (the one that contains the drawer view) moving down with the drawer inside of it. So, the separator needed to remain fixed in place relative to the content position and this logic was intended to compensate for that.

Now that we move the inner view (the drawer inside the content view) we don't need to compensate the separator anymore, so it remains in a fixed place with the content view.

ios/FluentUI/Drawer/DrawerPresentationController.swift
rdeassis: It seems the GIF does not show all the validation. But I did validate rotating from both sides while launching the drawer from both orientations (vertical/horizontal).

harrieshin (approved):

PR #272, opened 2020-10-23
Fix for crashes while adding UIPointerInteraction instance

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Crashes were reported for pointer interaction instances being added to views.
We found out that some beta versions of iOS 13.4 and later don't include the pointer interaction APIs.
So, we're adding a workaround by checking the UIButton's isPointerInteractionEnabled property in order prevent crashes due to an exception thrown while trying to add the UIPointerInteraction instance while running on those iOS versions:

**"uncaught ObjC exception, reason: Invalid parameter not satisfying: interaction"**

### Verification

- Ensured that the arePointerInteractionAPIsAvailable API returns false in iOS versions that don't support Pointer Interactions (iOS 12 for example)
- Exercised the controls in FluentUI that use pointer interactions to make sure the functionality is not regressed.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/272)

### DIFF
--- a/ios/FluentUI/Extensions/UIView+Extensions.swift
+++ b/ios/FluentUI/Extensions/UIView+Extensions.swift
@@ -200,16 +200,3 @@ extension UIView {
         NSLayoutConstraint.contain(view: view, in: self, withInsets: insets, respectingSafeAreaInsets: respectsSafeAreaInsets)
     }
 }
-
-// MARK: - Pointer Interaction APIs check
-
-extension UIView {
-    /// Some beta versions of iOS 13.4 and later don't include the pointer interaction APIs.
-    /// Checking the UIButton's isPointerInteractionEnabled property is a workaround that prevents a crash due to an
-    /// exception thrown while trying to add the UIPointerInteraction instance while running on those iOS versions:
-    /// "uncaught ObjC exception, reason: Invalid parameter not satisfying: interaction"
-    /// - Returns: true if Pointer Interaction APIs are available, otherwise false
-    func arePointerInteractionAPIsAvailable() -> Bool {
-        return UIButton.instancesRespond(to: Selector(("isPointerInteractionEnabled")))
-    }
-}
--- a/ios/FluentUI/Extensions/UIView+Extensions.swift
+++ b/ios/FluentUI/Extensions/UIView+Extensions.swift
@@ -200,3 +200,17 @@ extension UIView {
         NSLayoutConstraint.contain(view: view, in: self, withInsets: insets, respectingSafeAreaInsets: respectsSafeAreaInsets)
     }
 }
+
+// MARK: - Pointer Interaction APIs check
+
+extension UIView {
+    
+    /// Some beta versions of iOS 13.4 and later don't include the pointer interaction APIs.
+    /// Checking the UIButton's isPointerInteractionEnabled property is a workaround that prevents a crash due to an
+    /// exception thrown while trying to add the UIPointerInteraction instance while running on those iOS versions:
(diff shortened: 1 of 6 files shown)


ios/FluentUI/Extensions/UIView+Extensions.swift
```diff
@@ -200,3 +200,17 @@ extension UIView {
...
+    
+    /// Some beta versions of iOS 13.4 and later don't include the pointer interaction APIs.
+    /// Checking the UIButton's isPointerInteractionEnabled property is a workaround that prevents a crash due to an
+    /// exception thrown while trying to add the UIPointerInteraction instance while running on those iOS versions:
+    /// "uncaught ObjC exception, reason: Invalid parameter not satisfying: interaction"
+    /// - Returns: true if Pointer Interaction APIs are available, otherwise false
+    func arePointerInteractionAPIsAvailable() -> Bool {
+        return UIButton.instancesRespond(to: Selector(("isPointerInteractionEnabled")))
+    }
+}
```
anandrajeswaran: Does this need to be an extension on UIView? Especially since it doesn't complement or use any of the built-in UIView behavior can we just have a standalone func?

ios/FluentUI/Extensions/UIView+Extensions.swift
```diff
@@ -200,3 +200,17 @@ extension UIView {
...
+
+extension UIView {
+    
+    /// Some beta versions of iOS 13.4 and later don't include the pointer interaction APIs.
+    /// Checking the UIButton's isPointerInteractionEnabled property is a workaround that prevents a crash due to an
+    /// exception thrown while trying to add the UIPointerInteraction instance while running on those iOS versions:
+    /// "uncaught ObjC exception, reason: Invalid parameter not satisfying: interaction"
+    /// - Returns: true if Pointer Interaction APIs are available, otherwise false
+    func arePointerInteractionAPIsAvailable() -> Bool {
+        return UIButton.instancesRespond(to: Selector(("isPointerInteractionEnabled")))
```
anandrajeswaran, line 213: No practical impact, just a potential optimization that could help us detect and delete this when possible - thoughts on marking either the impl or the declaration #available with iOS 13.4 as well? That way we don't need to individually track such issues and can ensure it will get cleaned up as part of regularly scheduled combing for no longer relevant code when we do version bumps. Or is swift too smart and won't let us do that?

anandrajeswaran: actually yeah we could mark this func as available for 13.4 since all call sites currently have iOS 13.4 availability checks already out of necessity

anandrajeswaran (changes requested):

ios/FluentUI/Extensions/UIView+Extensions.swift
rdeassis: The reasoning behind putting it on a UIView extension is because the addInteraction function is also in a UIView extension. 
But I agree that a standalone function is a better approach.

ios/FluentUI/Extensions/UIView+Extensions.swift
rdeassis: Good point. The func is now marked @available for iOS 13.4.

MathieuKavalec (approved):

anandrajeswaran (approved):

Issue #274, opened 2020-10-27
Podspec in 0.1.16 specifies the iOS bundle for the Mac target


toadgee:
### Environment Information

- **Platform**:
  - [ ] iOS
  - [x] macOS
- **Package version(s)**: CocoaPods, podspec 0.1.16
- **Xcode and OS versions**: 12.2, 10.15.7

### Please provide a reproduction of the bug:

It looks like when I build 0.1.16 from cocoapods that it embeds the iOS bundle in the Mac framework, which then crashes when you try to use it.

#### Actual behavior:

crashes, looking for the fluentuiresources-mac bundle.

#### Expected behavior:

Not crashing

### Priorities and help requested:

Are you willing to submit a PR to fix? (Yes, No) Yep

Requested priority: (Blocking, High, Normal, Low) Normal

Products/applications affected: (if applicable)

Saadnajmi:
This should be fixed with 0.1.17

toadgee:
yep, fixed thanks @Saadnajmi  and @HeyImChris

PR #278, opened 2020-10-28
Remove macOS resources bundle and collapse into framework

anandrajeswaran:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
Uploading to App Store is reporting some errors because the resource bundle has an invalid code signature. It's still unclear what about the bundle is invalid as you can't really test it without actually uploading to App Store.

Trying to remove the Info .plist for the bundle would definitely stop it from being signed, but sadly breaks any sort of fetching from the asset catalog. Instead let's try just removing the bundle and adding the resources directly to the framework as that's a more standard approach Apple would be expecting.

While we're at it, don't expose our resource bundle publicly - that should make future iteration easier.

Note #1: This seems like it explicitly goes against what https://github.com/microsoft/fluentui-apple/pull/18 was trying to fix. I think/hope this will just force cocoapods consumers to use frameworks for now, but I'm unsure how to test that without publishing

Note #2: Long term we may want to do a slightly different approach of not having a framework and just using a static library and bundle, but that should still be a trivial change and this current approach has the advantage of being non-breaking.

### Verification
- Added a print statement in the resourceBundle function to ensure it only gets hit once even when switching between date picker and color picker.
- Verified colors appear correctly and date picker announces labels with VO on
- Verified Assets.car and .lproj folders appear in .framework build output
- TBD: How to validate cocoapods without checking in and publishing a version. Can I run pod-publish locally?

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/278)


harrieshin:
could we make the pod version update separate? it shouldn't matter for your office build testing right?

Saadnajmi:
Is this not related to the bug we had in 0.1.16 where the macOS resources bundle was incorrectly specified in the podfile? 
https://github.com/microsoft/fluentui-apple/issues/274
That should be fixed with 0.1.17

Saadnajmi:
Also it's not enough to just update the podfile version, there's 3-4 other strings that need to be updated, plus a tag and a release through github. We usually leave version updates as a seprate PR

Saadnajmi (dismissed):
On the podfile version update

anandrajeswaran:
> Is this not related to the bug we had in 0.1.16 where the macOS resources bundle was incorrectly specified in the podfile?
> #274
> That should be fixed with 0.1.17

Nope. This is not a bug when importing with cocoa pods. This is a bug when importing through the Nuget pipeline with a prebuilt framework with a resource bundle jammed in

macos/FluentUI/Core/FluentUIResources.swift
```diff
@@ -5,17 +5,12 @@
...
 	
-	/// The resource bundle contained within the FluentUI framework
-	@objc public static let resourceBundle: Bundle = {
-		guard let url = bundle.resourceURL?.appendingPathComponent("FluentUIResources-macos.bundle", isDirectory: true), let bundle = Bundle(url: url) else {
-			preconditionFailure("FluentUI resource bundle is not found")
-		}
-		
-		return bundle
+	/// The bundle of the FluentUI framework
+	static let resourceBundle: Bundle = {
```
anandrajeswaran, line 13: I need to preserve some fork of this logic for importing through cocoapods to actually still work.

Saadnajmi:
For testing cocoapods locally, you could take a project that depends on FluentUI Apple (like FluentUI React Native), do a `pod install` and then go into the pod folder and replace whatever was downloaded with your own framework, which might work? FURN imports it as a static library though

Side note: I had issues with using FluentUI Apple through cocoapods as a static library where the header "FluentUI-Swift.h" was not copied to the right Header Search Path, and I could only use the static library through Swift. I should probably file a bug for that.

anandrajeswaran:
> For testing cocoapods locally, you could take a project that depends on FluentUI Apple (like FluentUI React Native), do a `pod install` and then go into the pod folder and replace whatever was downloaded with your own framework, which might work? FURN imports it as a static library though
> 
> Side note: I had issues with using FluentUI Apple through cocoapods as a static library where the header "FluentUI-Swift.h" was not copied to the right Header Search Path, and I could only use the static library through Swift. I should probably file a bug for that.

Yup thanks! I've been verifying locally by changing the podspec to list a local :path instead of a version and that's working great. Just need to figure out a solution for the bundle lookup using the hardcoded name if in cocoapods or the framework's bundle directly if not

anandrajeswaran:
Found out how to just sign anyway and this is no longer a problem. Dropping this PR, but definitely something we can revisit for more standardization between different package management

PR #285, opened 2020-10-29
Access table cell label

nikitawani07:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Since title and subtitle are private, hence made them public.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/285)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -864,13 +864,13 @@ open class TableViewCell: UITableViewCell {
         }
     }
 
-    let titleLabel: Label = {
+    public let titleLabel: Label = {
         let label = Label(style: TextStyles.title)
         label.lineBreakMode = .byTruncatingTail
         return label
     }()
 
-    let subtitleLabel: Label = {
+    public let subtitleLabel: Label = {
         let label = Label(style: TextStyles.subtitleTwoLines)
         label.lineBreakMode = .byTruncatingTail
         label.isHidden = true


harrieshin:
why do you need access? tableviewcell has been carefully design to make a lot of the info private

nikitawani07:
> why do you need access? tableviewcell has been carefully design to make a lot of the info private

To show tooltip for the labels

harrieshin:
> To show tooltip for the labels

I am not sure this is the right approach. these are static views and not interact-able. why would we show a tooltip?

nikitawani07:
> > To show tooltip for the labels
> 
> I am not sure this is the right approach. these are static views and not interact-able. why would we show a tooltip?

For truncated file names.
I need access to label so that tooltip can be enabled on it. Since it was private, made them public.
Is there any other approach?

anandrajeswaran:
> > > To show tooltip for the labels
> > 
> > 
> > I am not sure this is the right approach. these are static views and not interact-able. why would we show a tooltip?
> 
> For truncated file names.
> I need access to label so that tooltip can be enabled on it. Since it was private, made them public.
> Is there any other approach?

What's the interaction that will show the tooltip? Static text labels are not usually interactable elements so I'm not sure users would really benefit from it, but if we're adding some way to dynamically "expand" a truncated label it probably should be added in the toolkit and FluentUI itself?

nikitawani07:
> > > > To show tooltip for the labels
> > > 
> > > 
> > > I am not sure this is the right approach. these are static views and not interact-able. why would we show a tooltip?
> > 
> > 
> > For truncated file names.
> > I need access to label so that tooltip can be enabled on it. Since it was private, made them public.
> > Is there any other approach?
> 
> What's the interaction that will show the tooltip? Static text labels are not usually interactable elements so I'm not sure users would really benefit from it, but if we're adding some way to dynamically "expand" a truncated label it probably should be added in the toolkit and FluentUI itself?

Enabling the user interaction for the label and adding fluent Tooltip as UILongPressGestureRecognizer action.

harrieshin:
usually Apple cells just show more lines of the label to avoid truncated text. I am not sure users will know they need to long press to see more? also that means the first responder chain gets different by showing the tooltip right?

nikitawani07:
> usually Apple cells just show more lines of the label to avoid truncated text. I am not sure users will know they need to long press to see more? also that means the first responder chain gets different by showing the tooltip right?

Tooltip will be shown for all the truncated texts in the app. So, it would be intuitive for user to long press in the cell as well.

anandrajeswaran:
> > usually Apple cells just show more lines of the label to avoid truncated text. I am not sure users will know they need to long press to see more? also that means the first responder chain gets different by showing the tooltip right?
> 
> Tooltip will be shown for all the truncated texts in the app. So, it would be intuitive for user to long press in the cell as well.

Sorry just catching up so apologies if this has already happened, but we should start a conversation with PM/design pushing for this experience. Apple doesn't show tooltips for truncated text elsewhere on the problem so I'd be surprised if users were able to discover this behavior. Per @harrieshin's comments we should probably be handling this the way Apple does in their first party apps to ensure we meet user expectations

PR #287, opened 2020-10-30
Overriding BadgeView background color to set transparent background

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Override background color for base view
Set it to transparent color and update color for front view (badge view background)

### Verification

Override the background Color for badge view instance.
```swift
       badge.backgroundColor = Colors.Palette.blueMagenta30.color
```

| Screen Shot 2020-10-30 at 12 00 00 PM | Screen Shot 2020-10-30 at 11 59 21 AM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/287)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -262,6 +262,13 @@ open class BadgeView: UIView {
         let labelWidth = max(minLabelWidth, min(maxLabelWidth, fittingLabelWidth))
         label.frame = CGRect(x: size.horizontalPadding, y: size.verticalPadding, width: labelWidth, height: labelHeight)
     }
+    
+    open override var backgroundColor: UIColor? {
+        didSet {
+            super.backgroundColor = .clear
+            updateBackgroundColor()
+        }
+    }
 
     open func reload() {
         label.text = dataSource?.text
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -38,6 +38,7 @@ public protocol BadgeViewDelegate {
 
 public extension Colors {
     struct Badge {
+        public static var background: UIColor = .clear
         public static var backgroundDisabled = UIColor(light: surfaceSecondary, dark: gray700)
         public static var backgroundError = UIColor(light: Palette.dangerTint40.color, dark: Palette.dangerTint30.color)
         public static var backgroundErrorSelected: UIColor = error
@@ -174,17 +175,21 @@ open class BadgeView: UIView {
 
     @objc open var isActive: Bool = true {
         didSet {
-            updateBackgroundColor()
-            updateLabelTextColor()
+            update()
             isUserInteractionEnabled = isActive
         }
     }
 
     @objc open var isSelected: Bool = false {
         didSet {
-            updateBackgroundColor()
-            updateLabelTextColor()
+            update()
             updateAccessibility()
+        }
+    }
+    
+    open var labelTextColor: UIColor? {
+        didSet {
+            update()
         }
     }
 
@@ -200,8 +205,7 @@ open class BadgeView: UIView {
 
     private var style: Style = .default {
         didSet {
-            updateBackgroundColor()
-            updateLabelTextColor()
+            update()
         }
     }
 
@@ -265,8 +269,9 @@ open class BadgeView: UIView {
     
     open override var backgroundColor: UIColor? {
         didSet {
-            super.backgroundColor = .clear
-            updateBackgroundColor()
+            if backgroundColor != oldValue {
+                update()
+            }
         }
     }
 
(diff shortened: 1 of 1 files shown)


harrieshin:
we should add a test case for this custom color

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -262,6 +262,13 @@ open class BadgeView: UIView {
         let labelWidth = max(minLabelWidth, min(maxLabelWidth, fittingLabelWidth))
         label.frame = CGRect(x: size.horizontalPadding, y: size.verticalPadding, width: labelWidth, height: labelHeight)
     }
+    
+    open override var backgroundColor: UIColor? {
+        didSet {
+            super.backgroundColor = .clear
```
harrieshin, line 268: this doesn't seems right to me. if user wants to set the background color custom, it should show that in the background color right? I don't see the difference in your screenshots in the pr description? also, when user change the text or selection state/active nonactive, wouldn't it fall back to the default style color?

kubalani: The PR doesn't allow custom color. My understanding was that we can only set to default presets color based on style. I will update the PR to keep the background color of badge in sync with the base view background color. 

The difference in the snapshot is extra color around the corner edges. This happens because the background color changes the color of the underlying base view color for badge. 

Screen Shot 2020-10-30 at 12 35 56 PM

I haven't tested the behavior with active/nonactive state but based on my understanding of the [code](https://github.com/microsoft/fluentui-apple/blob/37b39d0f8cb9fa6f71c90ad30d16576a3430c7b4/ios/FluentUI/Badge%20Field/BadgeView.swift#L301 ) the `updateBackground` method handles the underlying state while updating the setting the background color. 

Are there any specific test case you have in mind, that needs to be tested?

Saadnajmi:
This change is on a branch of the repo, instead of in a fork. As per our contribution guidelines, we should move the change to a fork and have an upstream pull request that way

ios/FluentUI/Badge Field/BadgeView.swift
harrieshin: the demo app should have a case where it shows that client code can change colors of the badgeview

rdeassis:
> This change is on a branch of the repo, instead of in a fork. As per our contribution guidelines, we should move the change to a fork and have an upstream pull request that way

Just for convenience, here's the contribution guidelines Saad mentioned:

https://github.com/microsoft/fluentui-apple#developing-in-the-repo

Developing in the repo
Fluent UI Apple requires all pull requests to come from forks of the repository. Please see Fork a Repo - GitHub Help for more details on how to set up a fork of Microsoft/fluentui-apple, keep it up-to-date, and submit pull requests back to this repository.

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -292,16 +303,23 @@ open class BadgeView: UIView {
             accessibilityHint = "Accessibility.Select.Hint".localized
         }
     }
+    
+    private func update() {
```
rdeassis, line 307: I think "update" is too generic. How about "updateColors" instead?

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -292,16 +303,23 @@ open class BadgeView: UIView {
...
+        updateBackgroundColor()
+        updateLabelTextColor()
+    }
 
     private func updateBackgroundColor() {
         if let window = window {
-            backgroundView.backgroundColor = backgroundColor(for: window, style: style, selected: isSelected, enabled: isActive)
+            let color = backgroundColor ?? backgroundColor(for: window, style: style, selected: isSelected, enabled: isActive)
```
rdeassis, line 314: Can we assign it directly to the background color since the color constant is not being used anywhere else? (just like you did in the text color)

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -262,6 +266,14 @@ open class BadgeView: UIView {
         let labelWidth = max(minLabelWidth, min(maxLabelWidth, fittingLabelWidth))
         label.frame = CGRect(x: size.horizontalPadding, y: size.verticalPadding, width: labelWidth, height: labelHeight)
     }
+    
+    open override var backgroundColor: UIColor? {
+        didSet {
+            if backgroundColor != oldValue {
```
rdeassis, line 272: Why do you do this check on backgroundColor but not on the LabelTextColor?

rdeassis (commented):
Could you provide a bit more context on what's the problem this change solves in the description of the PR?
Is the problematic scenario not covered in the Demo app? Is that's the case, we should add the scenario and have screenshots of before and after based on that scenario.

PR #288, opened 2020-11-02
Switch Xcode12-beta pipeline to use Xcode 12.2

anandrajeswaran:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

ADO has recently updated their images to account for the fact that there are 3 versions of Xcode in circulation (12.0.1, 12.1, 12.2). As such they abandoned the Xcode_12_beta naming convention in favor of being more explicit.

Since then, our beta pipeline publishes have tried to xcode-select Xcode_12_beta, failed silently (ugh) and fallen back to Xcode_12.app (12.0.1). Notably for macOS this means no Apple silicon slices were being produced

### Verification

Kicked off a manual publish adding this variable:
```
xcode_path_override : /Applications/Xcode_12.2.app
```

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/288)

### DIFF
--- a/.ado/templates/fluentui-apple-publish-nuget-job-xcode12-beta.yml
+++ b/.ado/templates/fluentui-apple-publish-nuget-job-xcode12-beta.yml
@@ -7,7 +7,7 @@ jobs:
   cancelTimeoutInMinutes: 5 # how much time to give 'run always even if cancelled tasks' before killing them
   
   steps:
-  # Select Xcode 12 beta version
+  # Select Xcode 12.2 beta version
   - template: apple-xcode-select.yml
     parameters:
       xcode_path_override: '/Applications/Xcode_12.2.app/Contents/Developer'
--- a/.ado/templates/fluentui-apple-publish-nuget-job-xcode12-beta.yml
+++ b/.ado/templates/fluentui-apple-publish-nuget-job-xcode12-beta.yml
@@ -10,7 +10,7 @@ jobs:
   # Select Xcode 12 beta version
   - template: apple-xcode-select.yml
     parameters:
-      xcode_path_override: '/Applications/Xcode_12_beta.app/Contents/Developer'
+      xcode_path_override: '/Applications/Xcode_12.2.app/Contents/Developer'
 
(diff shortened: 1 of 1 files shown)


harrieshin (approved):

HeyImChris (approved):
Assuming this means we're broken in all our FURN and SDX-Platform pipelines as well?

anandrajeswaran:
> can we just delete this yml file and just use fluentui-apple-publish-nuget.yml?

Not yet, Xcode 12.2 is not at GM yet. Hoping for tomorrow...

anandrajeswaran:
> Assuming this means we're broken in all our FURN and SDX-Platform pipelines as well?

They probably will have the core issue, but may not be a problem in practice yet - going one-by-one through linker errors today. FURN will not register there due to symbols not being used yet, and sdx-platform may also not register if we haven't actually updated consumed version in Office since ADO made this change.

HeyImChris:
> > Assuming this means we're broken in all our FURN and SDX-Platform pipelines as well?
> 
> They probably will have the core issue, but may not be a problem in practice yet - going one-by-one through linker errors today. FURN will not register there due to symbols not being used yet, and sdx-platform may also not register if we haven't actually updated consumed version in Office since ADO made this change.

Sounds good. I'm making a PR now to fix our pipeline in SDX-Platform for RN

Saadnajmi:
FURN uses Xcode 12, and no beta vesrions of xcode, so that should be fine

anandrajeswaran:
> FURN uses Xcode 12, and no beta vesrions of xcode, so that should be fine

We should probably make sure that's on Xcode 12.2 to start with so we don't break Apple Silicon when we turn it on in Office

Saadnajmi:
@anandrajeswaran do we want to make the default pipeline use 12.2, or have two separate pipelines like we have here in `FluentUI Apple` ?

PR #292, opened 2020-11-03
fix for voice over issue in case of passthrough drawercontroller

aamittal1989:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When we make the drawerController passthrough, Voice Over doesn't work on the underlying screen. So as a fix, added passthrough view as accessibility elements.

### Verification

1) Tap on **Show with underlying interactable content View**
2) Enable Voice over and check.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/292)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -77,12 +77,15 @@ class DrawerPresentationController: UIPresentationController {
         // Pass the passthrough view in touch forwarding view
         if let passThroughView = self.passThroughView {
             view.passthroughView = passThroughView
+            view.accessibilityElements = [passThroughView]
+        } else {
+            view.isAccessibilityElement = true
+            view.accessibilityLabel = "Accessibility.Dismiss.Label".localized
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -77,12 +77,15 @@ class DrawerPresentationController: UIPresentationController {
         // Pass the passthrough view in touch forwarding view
         if let passThroughView = self.passThroughView {
             view.passthroughView = passThroughView
+            view.accessibilityElements = [passThroughView]
```
harrieshin, line 80: Should passthroughview even be part of the accessibilityelements?

aamittal1989: @harrieshin Lets suppose we have passthroughView as the multi-select grid view. In that case through voice over user will not be able to access the passthroughView. After adding it to accessibilityElements, it can be accessed.

harrieshin (approved):

PR #294, opened 2020-11-03
Voiceover sequence enforced for avatar view

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Bugfix: Focus order is not in sequential order in `AvatrViewDemoController.swift`.

Repro Steps: 

1.  Open FluentUI application.
2. Navigate to 'Avtar View' button using swipe gesture.
3. Navigate to ‘Circle style for the person’ using swipe gesture.
4.  Using right swipe navigate in forward direction and observe the issue.

 Fix: Explicitly enforce ordering of views for accessibility of stack view.

### Verification

1. Open AvatarDemoViewController and select avatar view. 
2. Swipe right after Circle style for person heading.
3. Observe the issue i.e instead of going on the Extra Extra large it is going on the image

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/294)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoController.swift
@@ -84,6 +84,7 @@ class DemoController: UIViewController {
 
         items.forEach { itemRow.addArrangedSubview($0) }
         itemsContainer.addArrangedSubview(itemRow)
+        itemRow.accessibilityElements = itemRow.arrangedSubviews
         container.addArrangedSubview(itemsContainer)
     }
 


harrieshin:
there isn't anything about the component itself that should ensure the accessibility order is correct? your change is in demo app code only. does that mean it was a demo app only problem? or any client that consumes avatargroupview they need to re-arrange the subviews?

kubalani:
> there isn't anything about the component itself that should ensure the accessibility order is correct? your change is in demo app code only. does that mean it was a demo app only problem? or any client that consumes avatargroupview they need to re-arrange the subviews?

Yes. The bug filed in VSTS isn't related to component but its voice over sequence which is determined by the layout in `StackView`. 

If the client uses `AvatarGroupView` they then voice over sequence within the group view is works as expected but if the client places the `AvatarGroupView` in another container then defining the accessibility of the container is clients responsibility (which in this case is DemoController)

harrieshin (approved):

PR #300, opened 2020-11-05
Updating AvatarView Components

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated borders and overflow for AvatarView.

### Verification

Light and dark mode was tested.

| Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-04 at 18 31 37 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-04 at 20 20 58 |
| Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-04 at 18 31 54 | Simulator Screen Shot - iPhone 11 Pro Max - 2020-11-04 at 20 21 34 |
| BeforeCustom | AfterCustom |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/300)


ios/FluentUI/Resources/Localization/ar.lproj/Localizable.strings
```diff
@@ -186,5 +186,5 @@
 /* Presence - Blocked status */
 "Presence.Blocked" = "ممنوعة";
 
-/* AvatarView - Format string for overflow avatar. Format: "+<OverflowCount>". Example: "+5" */
-"AvatarView.OverflowFormat" = "+%@";
+/* AvatarView - Format string for overflow avatar. Format: "<OverflowCount>". Example: "5" */
+"AvatarView.OverflowFormat" = "%@";
```
harrieshin, line 190: I don't think we need this any more. can we just remove the string?

harrieshin:
looking at your screenshot, it seems like your extra small case, the initials are being cut off when the face stack has borders

harrieshin:
does your border change impact demo app > avatar view circle style with image-based border and just regular border?

sophialee0416:
You're right, I'll go over that again

sophialee0416:
> does your border change impact demo app > avatar view circle style with image-based border and just regular border?

It has updated image-based borders as well for consistency but I can revert if needed. Updated PR with images

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -23,7 +23,7 @@ public enum AvatarSize: Int, CaseIterable {
     public var font: UIFont {
         switch self {
         case .extraSmall:
-            return UIFont.systemFont(ofSize: 9)
+            return UIFont.systemFont(ofSize: 8)
```
harrieshin, line 26: but it should be 9 for regular case with out borders according to design?

sophialee0416: Ok, should the bordered case remain as a smaller font or should the internal border width change instead? The internal border is currently matching the design right now which I think cut off the extra small font

harrieshin: that's what the figma redline shows. double check with Victor and Jackson who are the designers of this control.

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -66,6 +66,15 @@ public enum AvatarSize: Int, CaseIterable {
         }
     }
 
+    var insideBorder: CGFloat {
```
harrieshin, line 69: we should document what this means

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -770,6 +786,16 @@ class OverflowAvatarView: AvatarView {
     open override func layoutSubviews() {
         super.layoutSubviews()
 
+        initialsView.setFontColor(UIColor(light: Colors.gray500, dark: Colors.gray100))
```
harrieshin, line 789: seems like this block of code can be grouped to "updateColors()" function

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -796,16 +822,16 @@ class OverflowAvatarView: AvatarView {
         var avatarFrame = bounds
         if hasBorder {
             let borderOffset = Constants.borderSize - borderWidth
-            avatarFrame.origin.x -= borderOffset
-            avatarFrame.origin.y -= borderOffset
-            avatarFrame.size.width += -borderOffset * 2
-            avatarFrame.size.height += -borderOffset * 2
+            avatarFrame.origin.x -= borderOffset * 1.75
```
harrieshin, line 825: where is this 1.75 coming from?

ios/FluentUI/People Picker/InitialsView.swift
```diff
@@ -131,7 +133,11 @@ class InitialsView: UIView {
 
     override func layoutSubviews() {
         super.layoutSubviews()
-        initialsLabel.frame = bounds
+        let width = bounds.width - (avatarSize.insideBorder * 2.5)
```
harrieshin, line 136: where is this 2.5 and 1.25 coming from?

sophialee0416: Currently, the text is not adjusting because it is just barely within the frame for the extra small case (if you zoom in you can see that there is no overlap, but the text is just touching the interior of the bounds. To accommodate for that, the 0.5 provides slight buffer so that the text can adjust accordingly. 1.25 is half of 2.5 which is in consequence of the width and height adjustment

ios/FluentUI/People Picker/AvatarView.swift
sophialee0416: It was to adjust for the new inside border; replaced with existing variables for better readability

ios/FluentUI/People Picker/InitialsView.swift
harrieshin: if these paddings are coming from design then the number should be define as constants then do the 1.25/2 and such for readability for the next person? I also think we should write comments for more clarity for the future engineer reading this code

harrieshin:
are the image of the pr description updated?

ios/FluentUI/Resources/Localization/he.lproj/Localizable.strings
```diff
@@ -181,7 +181,4 @@
 "Presence.Unknown" = "לא ידוע";
 
 /* Presence - Blocked status */
-"Presence.Blocked" = "חסום/ה";
-
-/* AvatarView - Format string for overflow avatar. Format: "+<OverflowCount>". Example: "+5" */
-"AvatarView.OverflowFormat" = "ועוד %@";
+"Presence.Blocked" = "חסום/ה";
```
harrieshin, line 184: what happened here?

ios/FluentUI/People Picker/InitialsView.swift
harrieshin: and this is going to be true for all sizes? or does this bound calculation should only happen for extra small case?

ios/FluentUI/Resources/Localization/he.lproj/Localizable.strings
sophialee0416: There was a merge conflict and the AvatarView localized string populated again. I just removed it so it matched the other files

ios/FluentUI/People Picker/InitialsView.swift
sophialee0416: Ok! And yes this applies to all sizes if the case ever arises where the text is the same/exceeds the size of its containing bounds. However currently the only case that ends up getting affected by the text size adjustment is the extra small case (since the other font sizes are already well within the containing bounds)

ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -753,13 +772,11 @@ class OverflowAvatarView: AvatarView {
         super.init(avatarSize: avatarSize, withBorder: false, style: .circle, preferredFallbackImageStyle: .outlined)
 
         var overflowCountString = NumberFormatter.localizedString(from: NSNumber(value: overflowCount), number: .none)
-        overflowCountString = String(format: "AvatarView.OverflowFormat".localized, overflowCountString)
+        overflowCountString = String(overflowCountString)
```
rdeassis, line 775: If we are intentionally removing the "+" from the overflow, we shouldn't need line 775, right?
I think we can also make overflowCountString let instead of var.

rdeassis (approved):

PR #301, opened 2020-11-05
Added convenience methods to TableViewHeaderFooterView

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added obj-c convenience methods to TableViewHeaderFooterView setups.

### Verification

Manual testing through sample data.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/301)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -280,6 +280,20 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
         NotificationCenter.default.addObserver(self, selector: #selector(handleContentSizeCategoryDidChange), name: UIContentSizeCategory.didChangeNotification, object: nil)
     }
 
+    // MARK: Setup
+
+    /// Sets up the titleView based off of the following parameters:
+    ///
+    /// - Parameters:
+    ///   - style: The `TableViewHeaderFooterView.Style` used to set up the view.
+    ///   - title: The title string.
+    ///   - attributedTitle: Title as an NSAttributedString for additional attributes.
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -280,6 +280,10 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
         NotificationCenter.default.addObserver(self, selector: #selector(handleContentSizeCategoryDidChange), name: UIContentSizeCategory.didChangeNotification, object: nil)
     }
 
+    @objc open func setup(style: Style, title: String, accessoryButtonTitle: String = "") {
```
rdeassis, line 283: Could you add Xcode style documentations to these setup method overloads that are exposed to objc?

rdeassis (commented):
The changes look good. Could you just add the Xcode documentation to those existing setup methods as well as the overloads you're adding?

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
harrieshin: also the simpliest variation we care about might be just style and title ? what is most used variation in office?

rdeassis: It seems like all the usage is indeed only style and title. I agree that providing another overload with these 2 params could be very useful.

rdeassis (approved):

PR #307, opened 2020-11-10
update segment control button accessibility hint

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We shouldn't be adding accessibility hints like "tap to go to something" in our library. 
Interacting with a button with Voiceover is technically double tap. However, we really shouldn't be customizing string to tell users how to interact with our components because depending on the assistive technology that customer is using, how he or she activates a button might be different.
Instead, lets be consistent with our pillbuttonbar control which says the index of the button out of count of all buttons (similar to our tab bar works). 

### Verification

| "Tap to go to Foo" | "1 of 4" |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/307)

### DIFF
--- a/ios/FluentUI/Controls/SegmentedControl.swift
+++ b/ios/FluentUI/Controls/SegmentedControl.swift
@@ -220,6 +220,8 @@ open class SegmentedControl: UIControl {
         if index <= selectedSegmentIndex {
             _selectedSegmentIndex += 1
         }
+
+        updateAccessibilityHints()
     }
 
     /// Remove the segment at the appropriate index. If there are only 2 segments in the control, or if no segment exists at the index, this method is ignored. If the segment is currently selected, we change the selection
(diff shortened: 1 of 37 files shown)


ios/FluentUI/Controls/SegmentedControl.swift
```diff
@@ -428,6 +431,12 @@ open class SegmentedControl: UIControl {
         }
     }
 
+    private func updateAccessibilityHints() {
+        for (index, button) in buttons.enumerated() {
+            button.accessibilityHint = String(format: "Accessibility.MSPillButtonBar.Hint".localized, index + 1, items.count)
```
kubalani, line 436: do we need to use `localizedStringWithFormat` here instead?

Ref: [here](https://developer.apple.com/documentation/foundation/nsstring/1497301-localizedstringwithformat)

kubalani (approved):

rdeassis (approved):

PR #308, opened 2020-11-10
Navigation bar subtitle and title/subtitle interactions

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Included in this change:
- Subtitle support in NavigationBar. The subtitle in positioned under the title.
- Callbacks to handle tap events on the title and the subtitle.
- Chevron configuration for the title and subtitle.
- A property to disable the behavior that expands the Navigation

Before this change, we only used LargeTitleView in NavigationBar when usesLargeTitle is set to true.
Additionally, LargeTitleView only supports aligning the content along the leading edge.
In this change, LargeTitleView was customized to support centered content and subtitles.
The layout of LargeTitleView was also modified to allow for the presentation of chevrons next to the title and subtitle.

Screen Shot 2020-09-24 at 7 14 51 PM
Screen Shot 2020-09-24 at 7 14 49 PM
Screen Shot 2020-09-24 at 7 14 42 PM
Screen Shot 2020-09-24 at 7 14 40 PM
Screen Shot 2020-09-24 at 10 56 07 PM
Screen Shot 2020-09-24 at 7 14 33 PM
Screen Shot 2020-09-24 at 10 55 57 PM
Screen Shot 2020-09-24 at 7 14 27 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/308)


ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -114,12 +181,156 @@ class LargeTitleView: UIView {
...
+        titleButton.contentHorizontalAlignment = .left
+        titleButton.titleLabel?.adjustsFontSizeToFitWidth = true
+        titleButton.addTarget(self, action: #selector(LargeTitleView.titleButtonTapped(sender:)), for: .touchUpInside)
+        titleButton.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
+
+        return titleButton
+    }()
+
+    // button used to display the subtitle of the current navigation item
+    private lazy var subtitleButton: UIButton = {
```
rdeassis, line 213: Can't we leverage from the TwoLineTitleView instead of recreating this layout on the LargeTitleView?

We can follow up with design but the individual handlers for the title/subtitle can't be used simultaneously because of the minimum tappable area of 44ptx44pt (https://developer.apple.com/design/tips/).

MathieuKavalec: It's proving very difficult and expensive to switch to TwoLineTitleView.
TwoLineTitleView doesn't support avatar view and updating based on navigationItem properties. Also, we need to refactor a lot of NavigationBar logic to swap between LargeTitleView and TwoLineTitleView.
I'm going to close this PR as a result and, perhaps, at a later time, revisit the whole thing.

Issue #256, opened 2020-10-09
 Voice over is conveying irrelevant information while navigating to the textfield in the People Picker


tdesroch:
### Environment Information

- **Platform**:
  - [X] iOS
  - [ ] macOS

### Please provide a reproduction of the bug:
Enabled Voice Over and navigate to the PeoplePicker within the FluentUI test app.
VO swipe until your focus is on the empty BadgeField
Double-tap to select

#### Actual behavior:
VO will read out "zero width space"

#### Expected behavior:
VO shouldn't read out "zero width space" as it could be confusing to the user.

### Priorities and help requested:
Looking at the code emptyTextFieldString is being defined to the formal "0 width space" character "\u{200B}" so one could say that it's technically by design.  However, this seems to be an implementation detail of shouldChangeCharactersIn within the BadgeField inadvertently being exposed to VO users.

Requested priority: Normal

Products/applications affected: All consumers of BadgeField

kubalani:
> ### Environment Information
> * **Platform**:
>   
>   * [x]  iOS
>   * [ ]  macOS
> 
> ### Please provide a reproduction of the bug:
> Enabled Voice Over and navigate to the PeoplePicker within the FluentUI test app.
> VO swipe until your focus is on the empty BadgeField
> Double-tap to select
> 
> #### Actual behavior:
> VO will read out "zero width space"
> 
> #### Expected behavior:
> VO shouldn't read out "zero width space" as it could be confusing to the user.
> 
> ### Priorities and help requested:
> Looking at the code emptyTextFieldString is being defined to the formal "0 width space" character "\u{200B}" so one could say that it's technically by design. However, this seems to be an implementation detail of shouldChangeCharactersIn within the BadgeField inadvertently being exposed to VO users.
> 
> Requested priority: Normal
> 
> Products/applications affected: All consumers of BadgeField

@tdesroch thanks for reporting your findings. I was able to replicate the problem on device but I am not sure if I understand the correct expected behavior. A simple replacement to empty string `""` instead of  `"\u{200B}"` does address the voice over issue. 

I am not sure if I understand implementation problems in `shouldChangeCharactersIn` of `BadgeView` related the problems in `PeoplePicker`.

PR #309, opened 2020-11-10
Replacing zero width space to empty string for VO readability

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Fix for #256 . The Voice over reads the default value in the textfield (initialized to [zero space width](https://github.com/microsoft/fluentui-apple/blob/69c7fb74eadc33bcd13539d50bfecf1bcb5d65af/ios/FluentUI/Badge%20Field/BadgeField.swift#L89)) 

```swift
 @objc open func resetTextFieldContent() {
        textField.text = Constants.emptyTextFieldString
    }
```

Since the default is set to non empty string the voice over reads the character description instead of ignoring it. 

#### Fix

Replace unicode character with empty string.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/309)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -78,7 +78,7 @@ open class BadgeField: UIView {
         static let badgeHeight: CGFloat = 26
         static let badgeMarginHorizontal: CGFloat = 5
         static let badgeMarginVertical: CGFloat = 5
-        static let emptyTextFieldString: String = emptyString
+        static let emptyTextFieldString: String = ""
         static let dragAndDropMinimumPressDuration: TimeInterval = 0.2
         static let dragAndDropScaleAnimationDuration: TimeInterval = 0.3
         static let dragAndDropScaleFactor: CGFloat = 1.10
@@ -86,7 +86,6 @@ open class BadgeField: UIView {
         static let labelMarginRight: CGFloat = 5
         static let textStyleFont: UIFont = TextStyle.subhead.font
         static let textFieldMinWidth: CGFloat = 100
-        static let emptyString: String = ""
     }
 
     @objc open var label: String = "" {
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -78,7 +78,7 @@ open class BadgeField: UIView {
         static let badgeHeight: CGFloat = 26
         static let badgeMarginHorizontal: CGFloat = 5
         static let badgeMarginVertical: CGFloat = 5
-        static let emptyTextFieldString: String = zeroWidthSpace
+        static let emptyTextFieldString: String = emptyString
         static let dragAndDropMinimumPressDuration: TimeInterval = 0.2
         static let dragAndDropScaleAnimationDuration: TimeInterval = 0.3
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -78,15 +78,15 @@ open class BadgeField: UIView {
...
+        static let emptyTextFieldString: String = emptyString
         static let dragAndDropMinimumPressDuration: TimeInterval = 0.2
         static let dragAndDropScaleAnimationDuration: TimeInterval = 0.3
         static let dragAndDropScaleFactor: CGFloat = 1.10
         static let dragAndDropPositioningAnimationDuration: TimeInterval = 0.2
         static let labelMarginRight: CGFloat = 5
         static let textStyleFont: UIFont = TextStyle.subhead.font
         static let textFieldMinWidth: CGFloat = 100
-        static let zeroWidthSpace: String = "\u{200B}"
+        static let emptyString: String = ""
```
rdeassis, line 89: This can be removed and you can use "" as the default value for emptyTextFieldString.

harrieshin: why did the previous author use unicode for zerowidthspace? there must have been a reason?

kubalani: I couldn't find any references from blame logs to find any concrete answers. The information I was able to find online was that this character are typically used for [line wrapping](https://www.ptiglobal.com/2018/04/26/the-beauty-of-unicode-zero-width-characters/). 

In addition we use them in [AvatarView](https://github.com/microsoft/fluentui-apple/blob/4cfde746f2dd7d9b337829d6119e797063ce72ab/macos/FluentUI/AvatarView/AvatarView.swift#L349) as character separators. It seemed out of context in this case. I was hoping either of you would have more context :) 

If not, I would probably reach out to broader team to double check on this.

harrieshin: the previous author aren't the the team anymore. removing this doesn't have any affect on badge field layout?

kubalani: Nope, I have attached the before/after changes pics below.

### Zero space characters

peoplepicker_before 
badge_before

### Empty string
peoplepicker_after
badge_after

kubalani: I posted this question to broader team across FX to check if someone else have more knowledge about the use of this context and below is the response. 

> it looks like it was there since BadgeField came in so probably just got overlooked in the initial PR. It's maybe possible removing may expose other code that is incorrectly assuming that a string length of 0 was a deletion rather than just initial state? But most likely I'd guess using "" as a drop-in replacement should be sufficient.

harrieshin: ifthat's the case, i am not sure why we need zerowithspace static let at all?  just directly set emptyTextFieldString with ""?

kubalani: I have updated this.

harrieshin (approved):

rdeassis (approved):

PR #312, opened 2020-11-14
Add primary colors for Mac

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes
Added primary color properties to Colors. I also added the demo colors from iOS to the Mac test app.

### Verification
light
dark

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/312)

### DIFF
--- a/macos/FluentUI/Core/Colors.swift
+++ b/macos/FluentUI/Core/Colors.swift
@@ -247,6 +247,40 @@ public final class Colors: NSObject {
 			}
 		}
 	}
+	
+	// MARK: Primary
+	
+	@objc public var primary: NSColor = Colors.Palette.communicationBlue.color {
+		didSet { }
+	}
+	
+	@objc public var primaryTint10: NSColor = Colors.Palette.communicationBlueTint10.color {
+		didSet { }
+	}
+	
+	@objc public var primaryTint20: NSColor = Colors.Palette.communicationBlueTint20.color {
+		didSet { }
+	}
+	
+	@objc public var primaryTint30: NSColor = Colors.Palette.communicationBlueTint30.color {
+		didSet { }
+	}
+	
+	@objc public var primaryTint40: NSColor = Colors.Palette.communicationBlueTint40.color {
+		didSet { }
+	}
+	
+	@objc public var primaryShade10: NSColor = Colors.Palette.communicationBlueShade10.color {
+		didSet { }
+	}
+	
+	@objc public var primaryShade20: NSColor = Colors.Palette.communicationBlueShade20.color {
+		didSet { }
+	}
+	
+	@objc public var primaryShade30: NSColor = Colors.Palette.communicationBlueShade30.color {
+		didSet { }
+	}
 }
 
 /// Make palette enum CaseIterable for unit testing purposes
--- a/macos/FluentUI/Core/Colors.swift
+++ b/macos/FluentUI/Core/Colors.swift
@@ -250,37 +250,21 @@ public final class Colors: NSObject {
 	
 	// MARK: Primary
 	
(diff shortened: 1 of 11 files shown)


macos/FluentUI/Core/Colors.swift
```diff
@@ -247,6 +247,40 @@ public final class Colors: NSObject {
 			}
 		}
 	}
+	
+	// MARK: Primary
+	
+	@objc public var primary: NSColor = Colors.Palette.communicationBlue.color {
+		didSet { }
```
harrieshin, line 254: no need for this

macos/FluentUITestApp/Assets.xcassets/Colors/DemoPrimaryShade10Color.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "1.000",
```
harrieshin, line 7: where is this demo primary colo coming from? also, json file isn't set as hex it is floating number which won't be consistent with design figma.

macos/FluentUI/Core/Colors.swift
```diff
@@ -247,6 +247,40 @@ public final class Colors: NSObject {
 			}
 		}
 	}
+	
+	// MARK: Primary
+	
+	@objc public var primary: NSColor = Colors.Palette.communicationBlue.color {
```
harrieshin, line 253: similar to iOS, i think it should be static class variable so that you don't need to create instance of MSFColors

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -20,17 +20,28 @@ class TestColorViewController: NSViewController {
...
-			let rowStackView = NSStackView()
-			rowStackView.orientation = .horizontal
-			rowStackView.spacing = 20.0
-			rowStackView.addArrangedSubview(colorView)
-			rowStackView.addArrangedSubview(textView)
-			colorsStackView.addArrangedSubview(rowStackView)
+			colorsStackView.addArrangedSubview(createColorRowStackView(name: color.name, color: color.color))
 		}
 		
+		/// Excel primary colors
```
harrieshin, line 26: if we are showcasing excel colors, it should be similar to ios which you can switch between communication blue to excel green.

macos/FluentUITestApp/Assets.xcassets/Colors/DemoPrimaryShade10Color.colorset/Contents.json
chiuam: @harrieshin This is coming from the iOS demo app. I copied them from the resources folder, don't know why they got redefined in floating point. I will change them to hex.

macos/FluentUITestViewControllers/TestColorViewController.swift
chiuam: iOS uses a table view to switch between themes, we might have to do the same to achieve that.

harrieshin: well instead of creating a table view, couldn't you listen to when the toggle value get changes, unload the entire colorsstackview and re-create it?

chiuam: I see. Maybe we can toggle themes between default and primary.

chripear (approved):

harrieshin (approved):

PR #317, opened 2020-11-18
Rename iOS Static Library from "FluentUILib" to "FluentUI"

### LINKED ISSUE #306: FluentUI framework and static library have different symbols
### Environment Information

- **Platform**:
  - [x] iOS
  - [x] macOS
- **Package version(s)**: 1.16
- **Xcode and OS versions**: 12.2_beta_2

### Please provide a reproduction of the bug:

Internally in our office codebase, we use FluentUI Apple on iOS as a static library, and not a framework. This means we link "libFluentUILib.a" which contains symbols that looks like this:
```
# From libFluentUILib.a
$ swift demangle s11FluentUILib11ShimmerViewCMa
$s11FluentUILib11ShimmerViewCMa ---> type metadata accessor for FluentUILib.ShimmerView
```

On FluentUI React Native, we consume FluentUI as a framework via cocoapods, and wrap some FluentUI Apple controls to create FluentUI React Native controls. 

The corresponding FluentUI React Native static libraries reference symbols from FluentUI Apple that look like this:
```
# From libFluentUI-React-Native-Shimmer.a
$ swift demangle s8FluentUI11ShimmerViewCMa
$s8FluentUI11ShimmerViewCMa ---> type metadata accessor for FluentUI.ShimmerView
```
We import both `libFluentUILib.a` and `libFluentUI-React-Native-Shimmer.a` into our office codebase via NuGet packages, and attempt to link the libraries.

#### Actual behavior:

We run into a linker error, and cannot use our FluentUI React Native libraries.
This is because  
```
s11FluentUILib11ShimmerViewCMa != s8FluentUI11ShimmerViewCMa
```

#### Expected behavior:

Both libraries should link fine, allowing us to use FluentUI React Native internally in our office codebase

#### Proposed Solution

At the current moment, we *must * use FluentUI as a framework, because importing FluentUI as a static library fails because we can't find the header "FluentUI-Swift.h". This is a separate bug that I should probably file. 
Additionally, our Office codebase has made the universal choice to not use frameworks on iOS because of boot performance, and only use static libraries. So switching the office codebase is not an option.

To fix this, it seems we either need to:

- have the framework and static library have the same name and therefore produce the same symbols (preferred solution)

or

- Fix the bug with importing static ibraries via cocoapods, and import FluentUI Apple as a static library into FluentUI React 
Native (less preferred, requires work across both repos)

### Priorities and help requested:

Are you willing to submit a PR to fix? Yes

Requested priority: Blocking

Products/applications affected: (if applicable) Office

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This is to close #306 

Our iOS static library and framework had different names, which meant the compiled products had different symbols, which prevented interop if another library (like FluentUI React Native) was expecting symbols the framework provided, but got symbols the static library provided.

Renaming the Product in xcode to "FluentUI" explicitly solves this.
Let's also update our CI pipeline to still properly find the compiled libraries and pack appropriately.

Note: This is a breaking change, as every `import FluentUILib` must now be a `import FluentUI`

### Verification

Locally created the static libraries and copied them into my office enlistment. Made sure that office compiled and ran for the scenario I was testing

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/317)

### DIFF
--- a/ios/FluentUI.xcodeproj/project.pbxproj
+++ b/ios/FluentUI.xcodeproj/project.pbxproj
@@ -321,7 +321,7 @@
 		8AF03E1B24B6BDBD00E6E2A2 /* ContactCollectionViewCell.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContactCollectionViewCell.swift; sourceTree = "<group>"; };
 		8AF03E1D24B6BDFB00E6E2A2 /* ContactCollectionViewLayout.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ContactCollectionViewLayout.swift; sourceTree = "<group>"; };
 		8FA3CB5A246B19EA0049E431 /* ColorTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = ColorTests.swift; sourceTree = "<group>"; };
-		8FD01166228A820600D25925 /* libFluentUILib.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = libFluentUILib.a; sourceTree = BUILT_PRODUCTS_DIR; };
+		8FD01166228A820600D25925 /* libFluentUI.a */ = {isa = PBXFileReference; explicitFileType = archive.ar; includeInIndex = 0; path = libFluentUI.a; sourceTree = BUILT_PRODUCTS_DIR; };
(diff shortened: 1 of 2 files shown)


Saadnajmi:
I probably need to update our YAML too...

Saadnajmi:
> I probably need to update our YAML too...

Nevermind. The script I updated **is** what our YAML refers to 😁

harrieshin (approved):

rdeassis (approved):

Saadnajmi:
@rdeassis @harrieshin no objections to hardcoding the name as I did? As opposed to using some other environment variable?

PR #320, opened 2020-11-20
Moving xcodeproj customized configs to xcconfig files.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Moving all the configurations for all targets that were customized in the xcodeproj file into specific xcconfig files.

### Verification

Verified all settings remain unchanged comparing Xcode settings(customized tab) before and after they were moved to xcconfig files.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/320)


ios/xcode/FluentUITests.xcconfig
```diff
@@ -0,0 +1,14 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+ALWAYS_EMBED_SWIFT_STANDARD_LIBRARIES = YES
+CODE_SIGN_STYLE = Automatic
+DEVELOPMENT_TEAM = UBF8T346G9
```
kubalani, line 8: Do we need this?

rdeassis: We don't necessarily need it to run the tests. But it'll avoid the warnings in Xcode:

Screen Shot 2020-11-19 at 8 48 03 PM

ios/xcode/FluentUILib.xcconfig
```diff
@@ -0,0 +1,11 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+CODE_SIGN_STYLE = Automatic
+MTL_FAST_MATH = YES
+OTHER_LDFLAGS = -ObjC
+PRODUCT_NAME = FluentUI
```
Saadnajmi, line 9: +1 for not having this hardcoded in the pbxproj now :D

Saadnajmi (approved):

ios/xcode/FluentUITests.xcconfig
harrieshin: i don't think i like developement team tag in public?

harrieshin: also that is specific to office team afaik

rdeassis: This was already public in the project file. These are just being moved to xcconfig files.

That is the Apple Team ID for Microsoft.
It's part of a lot of App Bundles distributed publicly by Microsoft.
An example is Teams App which has this in the embedded.provisionprofile file inside its application Bundle.

```
	<key>TeamIdentifier</key>
	<array>
		<string>UBF8T346G9</string>
	</array>
```

harrieshin (approved):

PR #321, opened 2020-11-20
Expose Button properties

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes
Although Mac design toolkit is still work in progress, let's expose some of the fundamental properties that FluentUI React Native will need to customize. 

Let's expose:
contentTintColor - button content tint color
restBackgroundColor - button background color
cornerRadius - button corner radius

### Verification
Added customized button test cases to demo app:
image
Tested exposed props in JS
image

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/321)

### DIFF
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -25,7 +25,7 @@ open class Button: NSButton {
 	@objc public init(title: String, style: ButtonStyle = .primaryFilled) {
 		super.init(frame: .zero)
 		initialize(title: title, image: nil, style: style)
-    }
+	}
 
 	/// Initializes a Fluent UI Button with an image, setting the imagePosition to imageOnly
 	/// - Parameters:
@@ -35,7 +35,7 @@ open class Button: NSButton {
 		super.init(frame: .zero)
 		imagePosition = .imageOnly
 		initialize(title: nil, image: image, style: style)
-    }
+	}
 	
 	
 	/// Initializes a Fluent UI Button with a title,  image,  imagePosition, and style
@@ -48,14 +48,14 @@ open class Button: NSButton {
 		super.init(frame: .zero)
 		self.imagePosition = imagePosition
 		initialize(title: title, image: image, style: style)
-    }
+	}
 
 	/// Initializes a Fluent UI Button
 	/// Set style to primaryFilled as default
 	@objc public init() {
 		super.init(frame: .zero)
 		initialize(title: nil, image: nil, style: .primaryFilled)
-    }
+	}
 
 	@available(*, unavailable)
 	required public init?(coder decoder: NSCoder) {
@@ -152,7 +152,7 @@ open class Button: NSButton {
 	open override func updateLayer() {
 		if let layer = layer {
 			layer.borderWidth = Button.borderWidth
-			layer.cornerRadius = Button.cornerRadius
+			layer.cornerRadius = self.cornerRadius
 			layer.backgroundColor = layerBackgroundColor.cgColor
 			layer.borderColor = outlineColor.cgColor
 		}
@@ -201,6 +201,11 @@ open class Button: NSButton {
 	@objc public var style: ButtonStyle = .primaryFilled {
 		didSet {
 			if style != oldValue {
+				if style == .primaryFilled {
+					contentTintColor = .white
+				} else {
+					contentTintColor = primaryColor
+				}
 				needsDisplay = true
 			}
 		}
@@ -228,16 +233,15 @@ open class Button: NSButton {
 		}
 	}
 
-	private var restBackgroundColor: NSColor {
-		return fillColor
-	}
-
+	@objc public var restBackgroundColor: NSColor? = nil
+	@objc public var cornerRadius: CGFloat = 3
+	
 	private var hoverBackgroundColor: NSColor {
-		return fillColor.withSystemEffect(.rollover)
+		return restBackgroundColor?.withSystemEffect(.rollover) ?? fillColor.withSystemEffect(.rollover)
 	}
 
 	private var pressedBackgroundColor: NSColor {
-		return fillColor.withSystemEffect(.pressed)
+		return restBackgroundColor?.withSystemEffect(.pressed) ?? fillColor.withSystemEffect(.pressed)
 	}
 
 	private var layerBackgroundColor: NSColor {
@@ -247,10 +251,10 @@ open class Button: NSButton {
 			} else if mouseEntered {
 				return hoverBackgroundColor
 			} else {
-				return restBackgroundColor
+				return restBackgroundColor == nil ? fillColor : restBackgroundColor!
 			}
 		} else {
-			return disabledColor(for: fillColor)
+			return disabledColor(for: restBackgroundColor ?? fillColor)
 		}
 	}
 
@@ -454,8 +458,6 @@ open class Button: NSButton {
 
 
 	private static let borderWidth: CGFloat = 1
-
-	private static let cornerRadius: CGFloat = 3
 
 	private static let verticalEdgePadding: CGFloat = 4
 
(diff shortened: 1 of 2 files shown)


macos/FluentUI/Button/Button.swift
```diff
@@ -152,7 +152,7 @@ open class Button: NSButton {
 	open override func updateLayer() {
 		if let layer = layer {
 			layer.borderWidth = Button.borderWidth
-			layer.cornerRadius = Button.cornerRadius
+			layer.cornerRadius = self.cornerRadius
```
Saadnajmi, line 155: could we just do "cornerRadius"? I don't think we need self

macos/FluentUI/Button/Button.swift
```diff
@@ -228,16 +233,15 @@ open class Button: NSButton {
 		}
 	}
 
-	private var restBackgroundColor: NSColor {
-		return fillColor
-	}
-
+	@objc public var restBackgroundColor: NSColor? = nil
+	@objc public var cornerRadius: CGFloat = 3
```
Saadnajmi, line 237: Because these are public, could we add some simple documentation around them? I realize the other variables do not have this.

Saadnajmi, line 237: I think we should also have a `needsDisplay = true` to ensure that the view is re-rendered when these change

macos/FluentUI/Button/Button.swift
```diff
@@ -228,16 +233,15 @@ open class Button: NSButton {
...
+	
 	private var hoverBackgroundColor: NSColor {
-		return fillColor.withSystemEffect(.rollover)
+		return restBackgroundColor?.withSystemEffect(.rollover) ?? fillColor.withSystemEffect(.rollover)
 	}
 
 	private var pressedBackgroundColor: NSColor {
-		return fillColor.withSystemEffect(.pressed)
+		return restBackgroundColor?.withSystemEffect(.pressed) ?? fillColor.withSystemEffect(.pressed)
```
Saadnajmi, line 244: I think you need to sync to the latest master, as these shouldn't be included as part of your change

Saadnajmi (changes requested):

macos/FluentUI/Button/Button.swift
anandrajeswaran, line 237: Why do we now need to expose the cornerRadius? I didn't know modifying it was part of the toolkit

macos/FluentUI/Button/Button.swift
```diff
@@ -228,16 +233,15 @@ open class Button: NSButton {
 		}
 	}
 
-	private var restBackgroundColor: NSColor {
-		return fillColor
-	}
-
+	@objc public var restBackgroundColor: NSColor? = nil
```
Saadnajmi, line 236: Would the default not be `fillColor` instead of `nil`? Also, can we not make this not an optional?

macos/FluentUI/Button/Button.swift
chiuam: You're right, my mistake...

Saadnajmi:
Might be a stretch, but it would be cool if we added  section where there is just the one custom button and you could customize various properties and see it reflected on the button, kinda like what we have for DatePicker. Totally not a blocker though

macos/FluentUI/Button/Button.swift
chiuam: These were intended changes.

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
chiuam: I think the convention we are following is avoid using struct for constants? Anyway, they're gone now :)

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -14,39 +14,77 @@ class TestButtonViewController: NSViewController {
...
 				Button(title: "Back", image: leadingArrowImage, imagePosition: .imageLeading, style: .borderless),
+				customButtonWithTitleAndImage,
 			]
 		}
 		
-		let disabledButtonsWithTitleAndImage = buttonsWithTitleAndImage().map { button -> NSButton in
-			button.isEnabled = false
-			return button
+		let customDisabledButton = Button()
+		customDisabledButton.title = "DisabledWithImg"
```
Saadnajmi, line 67: Could set the title, image, and imagePosition with a different initializer as well

macos/FluentUI/Button/Button.swift
```diff
@@ -174,6 +176,7 @@ open class Button: NSButton {
 		}
 	}
 	
+	/// Tint color of the button content.
 	open override var contentTintColor: NSColor? {
 		get {
 			if style == .primaryFilled {
```
Saadnajmi, line 182: Note: This should probably also be @obj open override, and I probably just forgot to do so

macos/FluentUI/Button/Button.swift
```diff
@@ -201,11 +204,25 @@ open class Button: NSButton {
...
+				} else {
+					contentTintColor = primaryColor
+				}
 				needsDisplay = true
 			}
 		}
 	}
 
+	/// Background color in rest state. Hover and pressed state colors will adjust accordingly using the systemEffect.
+	@objc public lazy var restBackgroundColor: NSColor  = fillColor {
```
Saadnajmi, line 218: One thing I wonder about now is the interplay between contentTintColor, primaryColor, and restBackgroundColor as all 3 of those are public and separately configurable, and could get in each other's way?

macos/FluentUI/Button/Button.swift
```diff
@@ -25,7 +25,7 @@ open class Button: NSButton {
 	@objc public init(title: String, style: ButtonStyle = .primaryFilled) {
 		super.init(frame: .zero)
 		initialize(title: title, image: nil, style: style)
-    }
+	}
```
HeyImChris, line 28: Is this intentional?

HeyImChris: To avoid spamming the review- same question for all the indentation changes :)

macos/FluentUI/Button/Button.swift
HeyImChris: Worth checking! I'd imagine @chiuam can verify those work as expected when toggled separately in the test app?

macos/FluentUI/Button/Button.swift
```diff
@@ -454,7 +467,7 @@ open class Button: NSButton {
 
 
 	private static let borderWidth: CGFloat = 1
-
+
```
HeyImChris, line 470: nit delete

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -58,8 +96,8 @@ class TestButtonViewController: NSViewController {
 		gridView.addColumn(with: buttonsWithTitle())
 		gridView.addColumn(with: buttonsWithImage())
 		gridView.addColumn(with: buttonsWithTitleAndImage())
-		gridView.addColumn(with: disabledButtonsWithTitleAndImage)
-
+		gridView.addColumn(with: disabledButtons())
+
```
HeyImChris, line 100: nit- delete white space

chripear (approved):

macos/FluentUI/Button/Button.swift
chiuam: Turns out @chripear is making major changes to the Button properties. Since my changes are a subset of his, maybe it's more time efficient to have him do the thorough testing on this?

Saadnajmi (approved):

macos/FluentUI/Button/Button.swift
Saadnajmi: This would be the main thing, otherwise LGTM!

chiuam: I made it public since we don't want it to be overridable.

PR #323, opened 2020-11-24
Fix for regression where Badge field does not allow deleting badges.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The PR #309 merged as 1da82131ddbdcc88d9f092c1de465904ec55e401 introduced a regression which prevents the user from deleting badges by hitting backspace.

The  zero width space ("\u{200B}") was used to trigger the delegate [textField(_:shouldChangeCharactersIn:replacementString:) method](https://developer.apple.com/documentation/uikit/uitextfielddelegate/1619599-textfield) when the user hits backspace.

It was a workaround to force that delegate call when the UITextField is empty.

This fix proposes a change in how that case is handled. A subclass of the UITextView being used in the BadgeField scope will trigger the delegate method when the cursor/caret is in the first position of the text field even though it's not deleting any character.

All the logic needed to account for the zero width space approach was also removed.

### Verification

Exercised the following scenarios on the fixed code in the PeoplePicker section of the Demo app:

 1.  Navigating with the cursor back and forth while ensuring the position of the caret is correct.
 2. Deleted selections of text that include the first position of the text in the UIText field ensuring the caret remains in the first position.
 3. Deleted selections starting in the middle of the text ensuring the caret remains in the beginning of the selection.
 4. Hit backspace with content on the text field validating the last badge was selected as a result.
 5. Hit backspace with an empty text field validating the last badge was selected as a result.
 6. Hit backspace on a selected badge ensuring it was removed as a result.
 7. Exercised tests 5 and 6 on red badges.

| people_picker_delete_bug  | people_picker_delete_fix |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/323)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -206,7 +206,6 @@ open class BadgeField: UIView {
         addGestureRecognizer(tapGesture)
 
         textField.addTarget(self, action: #selector(textFieldTextChanged), for: .editingChanged)
-        textField.addObserver(self, forKeyPath: #keyPath(UITextField.selectedTextRange), options: .new, context: nil)
 
         UIDevice.current.beginGeneratingDeviceOrientationNotifications()
         NotificationCenter.default.addObserver(self, selector: #selector(handleOrientationChanged), name: UIDevice.orientationDidChangeNotification, object: nil)
@@ -220,7 +219,6 @@ open class BadgeField: UIView {
     }
 
     deinit {
-        textField.removeObserver(self, forKeyPath: #keyPath(UITextField.selectedTextRange), context: nil)
 
         UIDevice.current.endGeneratingDeviceOrientationNotifications()
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -1234,4 +1214,19 @@ private class PasteControlTextField: UITextField {
...
+    override func deleteBackward() {
+        // Triggers the delegate method even when the cursor (caret) is in the first postion (regardless of text being empty).
+        // Using the zero width space ("\u{200B}") as the emptyTextFieldString instead of this approach will cause Voice Over
+        // to read it out loud as "zero width space", which is not desirable.
+        if let selectionRange = selectedTextRange,
+           selectionRange.isEmpty,
+           offset(from: beginningOfDocument, to: selectionRange.start) == 0 {
+            _ = self.delegate?.textField?(self,
+                                          shouldChangeCharactersIn: NSRange(location: 0, length: 0),
+                                          replacementString: "")
```
kubalani, line 1227: use `emptyTextFieldString` instead of `""` ?

kubalani (approved):
LGTM !! Thanks for taking a quick look on this

ios/FluentUI/Badge Field/BadgeField.swift
rdeassis: It's a private type of BadgeField. Probably not worth exposing it just for that constant.

PR #215, opened 2020-09-02
Customize ShimmerView with individual properties instead of objects

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The Shimmer view has two objects "ShimmerAppearance" and "ShimmerViewAppearance", each of which contains 5 properties that control the shimmer view's appearance.

Let's just make these individual properties, and deprecate the objects

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/215)

PR #327, opened 2020-12-01
Updated Fluent Button designs for macOS

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
- Incorporated more recent (but still evolving) designs for Fluent button styles
- Added semantic colors associated with the styles
- Added properties and init parameters for button size and accent color
- Added properties to customize button foreground/background/border color for any of the rest/pressed/disabled states, for a total of 9 colors
- Added computed properties that will allow setting just one color (the rest color) and will automatically apply system effects to determine colors for the other states
- Fine-tuned text and image alignment for good looking results in all cases
- Added a property allowing a secondary button to de-highlight a linked primary button when the secondary is pressed:
Taking highlight from linked button

### Verification
FluentUITestApp Screenshots
- Added examples of each the four standard styles and one custom style to the test app, in both large and small sizes
- Included examples of each possible imagePosition
- Included examples of different accent colors
- Added controls to preview different background materials (including translucent effects), background colors, and NSColor system effects
- Made the button pane scrollable so that it doesn't expand to arbitrarily large size with added cases

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/327)


macos/FluentUITestViewControllers/TestViewControllers.swift
```diff
@@ -11,6 +11,7 @@ public struct TestViewController: Identifiable {
 	public let type: NSViewController.Type
 }
 
+@available(OSX 10.14, *)
```
Saadnajmi, line 14: What is this tag for?

macos/xcode/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -981,13 +981,15 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 8F69C7B22295FABD009F69C0 /* FluentUI_testapp.xcconfig */;
 			buildSettings = {
+				MACOSX_DEPLOYMENT_TARGET = 10.14;
```
Saadnajmi, line 984: This should be coming from the xcconfig, and need not be specified here I thought? I guess we need to update the minimum OS version of the test app as well? Looks like line 982 specifies the correct xcconfig

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -6,134 +6,340 @@
 import AppKit
 import FluentUI
 
+@available(OSX 10.14, *)
```
Saadnajmi, line 9: If we bump the min os of the test app, can we not use these tags?

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -6,134 +6,340 @@
...
-		gridView.addColumn(with: buttonsWithTitleAndImage())
-		gridView.addColumn(with: disabledButtons())
-
-		let emptyCell = NSGridCell.emptyContentView
-	
-		// Insert the Row Titles as the last step
-		let rowLabels: [NSView] = [
-			emptyCell,
+		let fluentButtonsGrid = NSGridView(frame: .zero)
+		fluentButtonsGrid.translatesAutoresizingMaskIntoConstraints = false
```
Saadnajmi, line 232: ha, of course I forgot to set this before. That might explain some of the weirdness with the grid view before

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
 import AppKit
 
+// MARK: - Button Formatting Parameters
+
+/// Indicates the size of the button
+@objc(MSFButtonSize)
+public enum ButtonSize: Int, CaseIterable {
+	case large
+	case small
+}
```
Saadnajmi, line 15: Somewhat related, we offhand mentioned standardizing the "size" type in FluentUI React Native, which I put in this bug here https://github.com/microsoft/fluentui-react-native/issues/475

Doesn't affect your change, but is great we are using the same names!

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
+		size: ButtonSize = .large,
+		style: ButtonStyle = .primary,
+		accentColor: NSColor? = Colors.primary
+	)
+	{
+		self.size = size
+		self.style = style
+		self.accentColor = accentColor
+	}
+}
```
Saadnajmi, line 55: I might be wrong but I think we were moving away from structs to contain properties? @rdeassis or @anandrajeswaran might know more.

I previously said that for controls we want to wrap in React Native, we need to keep all the properties as a flat list instead of having a struct as a property that contained other properties. I was wrong, and both should work, and I just wanted to call that out.

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
+@objc(MSFButonColor)
+class ButtonColor: NSObject {
+	public static let brandForegroundDisabled = NSColor(named: "ButtonColors/brandForegroundDisabled", bundle: FluentUIResources.resourceBundle)
+	public static let brandBackgroundDisabled = NSColor(named: "ButtonColors/brandBackgroundDisabled", bundle: FluentUIResources.resourceBundle)
+	public static let neutralInverted = NSColor(named: "ButtonColors/neutralInverted", bundle: FluentUIResources.resourceBundle)
+	public static let neutralForeground2 = NSColor(named: "ButtonColors/neutralForeground2", bundle: FluentUIResources.resourceBundle)
+	public static let neutralBackground2 = NSColor(named: "ButtonColors/neutralBackground2", bundle: FluentUIResources.resourceBundle)
+	public static let neutralStroke2 = NSColor(named: "ButtonColors/neutralStroke2", bundle: FluentUIResources.resourceBundle)
+	public static let neutralForeground3 = NSColor(named: "ButtonColors/neutralForeground3", bundle: FluentUIResources.resourceBundle)
+	public static let neutralBackground3 = NSColor(named: "ButtonColors/neutralBackground3", bundle: FluentUIResources.resourceBundle)
```
Saadnajmi, line 68: Just curious, do these names come from design tokens?

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
 		layer?.contentsScale = window?.backingScaleFactor ?? 1.0
 
-		self.style = style
+		if let cell = cell as? ButtonCell {
+			// The Button properties `contentTintColorDisabled`,
+			// `backgroundColorDisabled`, and `borderColorDisabled` fully
+			// specify the look of the disabled button, and may already include
+			// a 'disabled' system effect.  Don't apply the effect again on top
+			// of that.
+			cell.imageDimsWhenDisabled = false
```
Saadnajmi, line 163: Would this be true if we weren't using a template image as well?

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
 			self.title = title
-		}
-
-		if let image = image {
+			if let image = image {
+				self.image = image
+				self.imagePosition = imagePosition
+			} else {
+				self.imagePosition = .noImage
+			}
```
Saadnajmi, line 173: nit: Indentation looks off?

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
-	open override func mouseEntered(with event: NSEvent) {
-		mouseEntered = true
-		needsDisplay = true
-	}
+	/// While the current Button is pressed, its style is temporarily applied to the linkedPrimary button.
+	/// This emulates an effect seen in the default style of Cocoa buttons, where pressing a secondary
+	/// button takes the accent color highlighting from a nearby primary button. For best results, the
+	/// current button should have the `.secondary` style, the linkedPrimary button should have the
+	/// `.primary` style, and both buttons should have the same accentColor.
+	@objc public var linkedPrimary: Button?
```
Saadnajmi, line 214: When we say "Cocoa Button", do we mean CocoaUI? we probably shouldn't mention internal office frameworks in our open source code. Also, I didn't know about this. I'm guessing this is used in like dialogs and alerts?

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
-	open override func mouseExited(with event: NSEvent) {
-		mouseEntered = false
+	private var linkedPrimaryOriginalStyle: ButtonStyle?
+
+	private func setMouse(down: Bool) {
+		mouseDown = down
 		needsDisplay = true
+		if isEnabled {
+			if let linkedPrimary = linkedPrimary
+			{
```
Saadnajmi, line 223: nit: bracket style

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
 	open override func updateLayer() {
 		if let layer = layer {
 			layer.borderWidth = Button.borderWidth
-			layer.cornerRadius = Button.cornerRadius
-			layer.backgroundColor = layerBackgroundColor.cgColor
-			layer.borderColor = outlineColor.cgColor
+			layer.cornerRadius = cornerRadius
+			if let contentTintColor = contentTintColor {
+				super.contentTintColor = contentTintColor
+			}
```
Saadnajmi, line 249: I think we specifically only want to do layer modifications in `updateLayer`. Why do we need to set `super.contentTintColor` here?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
-			}
+			return ButtonFormat(
+				size: self.size,
+				style: self.style,
+				accentColor: self.accentColor
+			)
 		}
 		set {
-			super.contentTintColor = newValue
-			needsDisplay = true
```
Saadnajmi, line 190: Do we not need a `needsDisplay = true` here? I'm guessing whoever sets format calls it themselves?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
-					contentTintColor = .white
-				} else {
-					contentTintColor = primaryColor
-				}
-				needsDisplay = true
-			}
-		}
-	}
+	/// Foreground text/image color when button is pressed.
+	@objc public var contentTintColorPressed: NSColor?
```
Saadnajmi, line 294: not necessary, but maybe we could incorporate the state-> color mapping into a dictionary that can be modified? Not sure if that's cleaner or works with objc though

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
+				return contentTintColorRest
 			}
 		}
-	}
-	
-	private func disabledColor(for color: NSColor) -> NSColor {
-		return color.withSystemEffect(.disabled)
+		set {
+			contentTintColorRest = newValue
+			contentTintColorPressed = newValue == .clear ? newValue : newValue?.withSystemEffect(.pressed)
```
Saadnajmi, line 316: Just curious, what happens if you apply the pressed system effect to .clear?

Saadnajmi, line 316: So if the client specifies their own `contentTintColorPressed`, but then modifies `contentTintColor`, `contentTintColorPressed` is overridden? Could we add some in code documentation that specifies that, or add a guard that explicity doesn't let you set `contentTintColor` and one of the state specific overrides?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
+	/// colors separately, use the `backgroundColorRest`, `backgroundColorPressed`, and/or
+	/// `backgroundColorDisabled` properties.
+	@objc public var backgroundColor: NSColor? {
+		get {
+			if !isEnabled {
+				return backgroundColorDisabled
+			} else if mouseDown {
+				return backgroundColorPressed
+			} else {
+				return backgroundColorRest
```
Saadnajmi, line 343: So we don't need to have a different color for hover?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
-				break
-			case .imageTrailing:
-				x = isRTL ? Button.horizontalEdgePadding : titleSize.width + Button.titleAndImageRectInterspacing + Button.horizontalEdgePadding
-				y = Button.verticalEdgePadding
-				width = imageSize.width
-				height = imageSize.height
+	/// This color is used for the background in the primary style, the pressed background in the secondary
+	/// style, and the content tint (i.e. foreground text/image) color in the borderless style.  It is not used
+	/// for the acrylic style.
+	@objc public var accentColor: NSColor? = Colors.primary
```
Saadnajmi, line 387: What happens if someone sets both this and contentTintColor on a borderless button?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
+			case .acrylic:
+				contentTintColor = ButtonColor.neutralForeground3
+				backgroundColor = ButtonColor.neutralBackground3
+				borderColor = .clear
+			case .borderless:
+				contentTintColor = accentColor
+				contentTintColorDisabled = ButtonColor.brandForegroundDisabled
+				backgroundColor = .clear
+				borderColor = .clear
+			case .none:
```
Saadnajmi, line 417: None of these have breaks at the end, I think that causes some problems with fallthrough?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
 				break
 			@unknown default:
 				break
 			}
+		}
+	}
 
-			return NSMakeRect(x, y, width, height)
+	private var fontSize: CGFloat = 15 {
```
Saadnajmi, line 425: Nit, move magic number to a constant

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
-			case .imageTrailing:
-				x = isRTL ? Button.horizontalEdgePadding + imageSize.width + Button.titleAndImageRectInterspacing : Button.horizontalEdgePadding
-				y = Button.verticalEdgePadding
-				width = titleSize.width
-				height = titleSize.height
-				break
-			@unknown default:
-				break
-			}
+	private var cornerRadius: CGFloat = 6
```
Saadnajmi, line 431: same for this constant

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
 
-			return NSMakeRect(x, y, width, height)
-		}
+	private static let borderWidth: CGFloat = 1
```
Saadnajmi, line 433: and this one

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +271,355 @@ open class Button: NSButton {
...
+	override func drawingRect(forBounds rect: NSRect) -> NSRect {
+		var horizontalInterCellSpacing: CGFloat = 0
+		var verticalInterCellSpacing: CGFloat = 0
+
+		switch imagePosition {
+		case .imageLeft, .imageLeading, .imageRight, .imageTrailing:
+			horizontalInterCellSpacing = titleToImageSpacing
+		case .imageBelow, .imageAbove:
+			verticalInterCellSpacing = titleToImageSpacing
+		case .noImage, .imageOnly, .imageOverlaps:
```
Saadnajmi, line 611: didn't know you could comma separate, nice!

Saadnajmi (changes requested):

macos/FluentUI/Button/Button.swift
anandrajeswaran: I don't know of any broad guidance on this, but I've generally found it easier to just use flat lists of elements since that tends to translate best for our objC consumers and allow selective updates - some property changes invalidate our intrinsicContentSize and some invalidate our display and some do both. This way we don't need to do both for all changes. Those are admittedly small benefits, but I can't think of any practical benefits of grouping together in a struct unless we are suggesting that all of them should logically only be used together

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
+
+	/// Accent color text/image, no fill or outline.
+	case borderless
+}
+
+/// Combination of all formatting parameters.
+public struct ButtonFormat {
+	var size: ButtonSize
+	var style: ButtonStyle
+	var accentColor: NSColor?
```
anandrajeswaran, line 43: Since we give this a default value that is not nil, can this be non-optional?

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
+	/// Initializes a Fluent UI Button with a title only
 	/// - Parameters:
 	///   - title: String displayed in the button
-	///   - style: The ButtonStyle, defaulting to primaryFilled
-	@objc public init(title: String, style: ButtonStyle = .primaryFilled) {
-		super.init(frame: .zero)
-		initialize(title: title, image: nil, style: style)
+	///   - size: The ButtonSize, default .large
+	///   - style: The ButtonStyle, default .primary
+	///   - accentColor: The accent or highlight NSColor, defaulting to the app primary color
```
anandrajeswaran, line 82: The comment "accent or highlight" might be a bit confusing since these are separate concepts when it comes to NSColor

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
 	}
 
-	/// Initializes a Fluent UI Button
-	/// Set style to primaryFilled as default
-	@objc public init() {
-		super.init(frame: .zero)
-		initialize(title: nil, image: nil, style: .primaryFilled)
+	/// Initializes a Fluent UI Button with no title or image, and with default formatting
+	@objc public convenience init() {
+		self.init()
```
anandrajeswaran, line 131: Why would someone want a button with no title or image?

macos/FluentUI/Button/Button.swift
anandrajeswaran: In general we probably don't want to be calling super unless it's to call the superclass implementation of the currently executing method. I've seen cases of that, but usually it means there is a better way to do things.

But yeah, Saad's comment about restricting updateLayer to doing operations on the layer is the more important thing here

macos/FluentUI/Button/Button.swift
```diff
@@ -5,158 +5,254 @@
...
-			layer.cornerRadius = Button.cornerRadius
-			layer.backgroundColor = layerBackgroundColor.cgColor
-			layer.borderColor = outlineColor.cgColor
+			layer.cornerRadius = cornerRadius
+			if let contentTintColor = contentTintColor {
+				super.contentTintColor = contentTintColor
+			}
+			if let backgroundColor = backgroundColor { 
+				layer.backgroundColor = backgroundColor.cgColor
+			}
```
anandrajeswaran, line 252: do we need to clear the layer's backgroundColor (and borderColor below) if these are nil?

macos/FluentUITestViewControllers/TestViewControllers.swift
chripear: The Button view controller has a bunch of 10.14+ features.  This is the topmost-level availability tag that allows us to load up the button view controller class.  Somehow setting the deployment target to 10.14+ was not sufficient to let me get rid of all these availability tags.

macos/xcode/FluentUI.xcodeproj/project.pbxproj
chripear: I still get all kinds of availability warnings despite setting the deployment target for the test app target, so wherever the setting is currently specified it is not correct.  I guessed that it might need to be set project-wide but I do not actually know.  I will gladly remove all this stuff if we can figure out how to set the minimum OS.

macos/FluentUITestViewControllers/TestViewControllers.swift
chripear: Fixed.  I must have picked up a change to the minimum OS version recently and was able to remove this stuff.

macos/xcode/FluentUI.xcodeproj/project.pbxproj
chripear: Fixed. I must have picked up a change to the minimum OS version recently and was able to remove this stuff.

macos/FluentUITestViewControllers/TestButtonViewController.swift
chripear: Fixed. I must have picked up a change to the minimum OS version recently and was able to remove this stuff.

macos/FluentUI/Button/Button.swift
chripear: Yes.

macos/FluentUI/Button/Button.swift
chripear: @anandrajeswaran @Saadnajmi I cannot visualize the "flat list" structure you are describing.  Is there an example you could point me to?

macos/FluentUI/Button/Button.swift
chripear: There is no automatic fallthrough in swift switches.  You need to explicitly use the 'fallthrough' keyword for that to happen.  The 'break' keyword in the empty cases at the bottom of the switch may be misleading: in that situation, it serves the same purpose as a 'pass' in python, in that its only function is to satisfy the compiler, which requires a statement to be there.

macos/FluentUI/Button/Button.swift
chripear: I did not add this initializer, I just made it better.  As for its purpose, someone might want to initialize the button (e.g. in an outer scope) and then add a title or image later.  That's pretty much what the designated NSButton initializer init(frame: .zero) does.

macos/FluentUI/Button/Button.swift
chripear: What is a `static let` if not a constant?

macos/FluentUI/Button/Button.swift
Saadnajmi: I'm fairly sure we need the default initializer so we could use it within office where someone wanted to use the objective c keyword "new"

macos/FluentUI/Button/Button.swift
chripear: If these all need to be separately defined constants then I would like to organize them into a struct.  But I need to understand the comments about avoiding structs first.

macos/FluentUI/Button/Button.swift
chripear: Fixed.

macos/FluentUI/Button/Button.swift
chripear: The word was there to avoid an unhelpful circular statement.  Removed "highlight" to avoid confusion.

macos/FluentUI/Button/Button.swift
chripear: Since they are nullable, sure.

macos/FluentUI/Button/Button.swift
chripear: RIP `medium`, you were so much help figuring out the alignment algorithm or lack thereof, before you were deemed redundant.

macos/FluentUI/Button/Button.swift
chripear: Do the fluent arrows count as not using template images?

Note that the property is not very well named, since the dimming is also applied to the button title.  It's probably updating the contentTintColor.

macos/FluentUI/Button/Button.swift
chripear: It is correct.  There is an extra level compared to the original, because we need to distinguish the case where *both* `title` and `image` are set to determine whether or not to override `imagePosition`.

macos/FluentUI/Button/Button.swift
chripear: No, I mean [Cocoa](https://en.wikipedia.org/wiki/Cocoa_(API)) as in NSButton.

chripear: As distinct from, say, Cocoa Touch and SwiftUI.

macos/FluentUI/Button/Button.swift
chripear: .clear is black with 100% alpha.  Applying the .pressed system effect removes some of the alpha, making it shade the background.  This generally does not make visual sense, because it adds visible borders and backgrounds to buttons that are not supposed to have them.

This is an artifact of .clear being a degenerate point in the RGBA space, and applying the system effect would look different based on the choice of RGB value in .clear, even though those values should be irrelevant.

macos/FluentUI/Button/Button.swift
chripear: Design says no.  For the sake of simplicity/YAGNI I removed the machinery for the hover state.

macos/FluentUI/Button/Button.swift
chripear: Correct.

I don't want to guard against overriding setting contentTintColor and one of the state specific overrides, because it's useful to override specific states (e.g. Disabled) after auto-computing the others.  I am using this extensively in the `style` property's `willSet` below.

I'll add some documentation.

macos/FluentUI/Button/Button.swift
chripear: If the cleanliness factor is related to the accumulation of lots of properties like this one, what I would really like is for the color set to be a struct (possibly of structs, whatever) that replaces the `style` enum.  That would make it simpler for users to make and apply their own styles.  There are a lot of additional format parameters (padding, corner radius and more) that the designers eventually want to expose, and it would be great to aggregate them together somehow so that they can be applied easily on init or by setting a single property later.

Any suggestions on a way to do this that would work with ObjC?

Coming from python I know the pain of trying to do all this with dictionaries (or project plists)... one spends a lot of lines of code validating the data structure before one can do anything useful with it.

macos/FluentUI/Button/Button.swift
chripear: We are setting contentTintColor here so that we pick up state-specific color changes in the pressed and disabled states.  Prior to this change, the color of a button's title or image did not change when hovered or pressed, and we were tacitly relying on the behavior of NSButtonCell's imageDimsWhenDisabled property to dim the image and title when the button was disabled.

Though the contentTintColor is not a property of the layer, it is relevant to the drawing of the button.  Alternatives would include adding this block to the setMouse function and also adding something to detect when the button is disabled/enabled and adjusting the color then too.

Thoughts?

macos/FluentUI/Button/Button.swift
chripear: Are there any conventions for when to set `needsDisplay = true`?  Would someone setting the `accentColor`, `size`, `style` or `format` expect to need to do it themselves, or would they expect the setting of the property to trigger a display update automatically?  Would this also apply to the properties such as `contentTintColorRest`?

macos/FluentUI/Button/Button.swift
Saadnajmi: alignment algorithm? I guess it is redundant so I don't need to know :l

macos/FluentUI/Button/Button.swift
Saadnajmi: Instead of:

- Class
  - prop 1
  - prop 2
  - prop that is struct
    - prop 3
    - prop 4
    - prop 5

Do:

- Class
  - prop 1
  - prop 2
  - prop 3
  - prop 4
  - prop 5

Saadnajmi: Here's a PR where I did that for Shimmer. In hindsight, I didn't *need* to (now that I understand memo caches in React Native better), but at the time I thought it was necessary https://github.com/microsoft/fluentui-apple/pull/215

macos/FluentUI/Button/Button.swift
Saadnajmi: derp, ignore

macos/FluentUI/Button/Button.swift
Saadnajmi: Fair. I personally wouldn't block a review on it, but the current guidance we as APEX has chosen is fileprivate constants at the bottom of the file, as seen here https://microsoft.github.io/swift-guide/FileContentOrdering.html

I don't quite like `fileprivate` but I did keep my constants at the bottom to follow this.

Saadnajmi:
I know you probably already have talked to designers / gotten requirements and such, but I wanted to ask here so it's somewhat public / I wasn't sure: Why do we want to override the rest/hover/press colors? Wouldn't that be going against the FluentUI design language if we can let clients override those? Or is it because those are design tokens that are overridable, so we need to support them?

chripear:
> I know you probably already have talked to designers / gotten requirements and such, but I wanted to ask here so it's somewhat public / I wasn't sure: Why do we want to override the rest/hover/press colors? Wouldn't that be going against the FluentUI design language if we can let clients override those? Or is it because those are design tokens that are overridable, so we need to support them?

Good question.  I brought this up today with Amy from the design team and she seemed to agree that keeping the configurable parameters limited to size, style and accent color would make sense from the perspective of the Fluent design language.  Might need some input from Amanda on how this affects the custom button style PR she recently completed, but I did notice that the "custom" style example in the test app could be represented equally well by a Primary style button with a different accent color, without tweaking any of the specific colors at all.

macos/FluentUI/Button/Button.swift
chripear: OK, I understand.  But in this case, `format` is a synthesized property that just maps to and from its three components without actually storing a struct anywhere in the class.  The ButtonFormat struct only exists because I found it really convenient to have predefined sets of properties that I could reuse and apply to buttons as a unit.  Examples the test app:

`let largePrimary = ButtonFormat(size: .large, style: .primary, accentColor: communicationBlue)`

`formats.map({ TestButton(image: stopImage, format: $0) })`

`TestButton(title: "Back", image: leadingArrowImage, imagePosition: .imageLeading, format: largePrimary)`

`TestButton(title: "Nope", image: stopImage, imagePosition: .imageAbove, format: largePrimary)`

Because the components are only transmitted and not stored as a struct, we aren't subject to the perf impact Anand mentioned: each of the properties will be set independently and can decide on its own what needs to be invalidated.

anandrajeswaran: This is a great discussion and love the care and attention to detail going into API design! I definitely agree that a struct to coalesce coupled properties is a great idea and that certainly might be the right design here. The two things to think of are
1) Will many clients want to only update some of the values in the struct? If most clients will only want to adjust one or some of the properties but then need to make a struct with all the properties to do so we may be adding unnecessary complexity and verbosity to the clients
2) What does the objC interface look like? The subtleties of structs vs. classes don't translate to objC well, so there's a chance doing this could put objC consumers in a bind.

If neither of the above are relevant (and I'm correctly understanding @Saadnajmi 's point that there is no impact of this decision on JS consumers in fluentui-react-native), then I really like the idea of a struct to tie together commonly coupled configs. Otherwise we may be adding more complexity than we're providing benefit.

macos/FluentUI/Button/Button.swift
anandrajeswaran: dictionaries in swift (or objC with generics) are easier than python because you do have type guarantees. in general the way Apple has done this is to define an enum for the state and expose setters that take in an enum. Then the complication of whether a data structure or dictionary or whatever is needed to back it is just an implementation detail that doesn't leak into the API surface

macos/FluentUI/Button/Button.swift
chripear: 1. Yes, certainly clients may want to update only some values of the struct.  But the client never needs to make a struct in order to update one of the values.

2. The ObjC properties are the individual components of the struct:
@objc public var accentColor: NSColor = Colors.primary
@objc open var style: ButtonStyle = .primary
@objc public var size: ButtonSize = .large

And the ObjC initializers use those properties:
@objc public convenience init(title: String, size: ButtonSize = .large, style: ButtonStyle = .primary, accentColor: NSColor = Colors.primary)
@objc public convenience init(image: NSImage, size: ButtonSize = .large, style: ButtonStyle = .primary, accentColor: NSColor = Colors.primary)
@objc public convenience init(title: String, image: NSImage, imagePosition: NSControl.ImagePosition = .imageLeading, size: ButtonSize = .large, style: ButtonStyle = .primary, accentColor: NSColor = Colors.primary)
@objc public convenience init()

macos/FluentUI/Button/Button.swift
chripear: I did something like this for size mappings.

macos/FluentUI/Button/Button.swift
chripear: Last one set wins.

macos/FluentUI/Button/Button.swift
Saadnajmi: I've been operating under the assumptions that 

1- there isn't a significant perf impact for calling `needsDisplay = true`  as it just tells the run loop to re-render the control the next time it runs (regardless if you called it 1 or 100 times)
2- Any change to the UI of the button (AKA all of the properties you mentioned) should have a `needsDisplay = true` call, and that is done by us and not the client. 

In practice, I found that things rendered fine without explicitly calling it, but I defaulted on calling it anyway. @anandrajeswaran or @markavitale perhaps one of my two assumptions is wrong and we don't really need to call it everywhere?

macos/FluentUI/Button/Button.swift
Saadnajmi: Interesting. I wonder if that's a peculiarity that's worth being an Apple Bug? Not related to the review though

macos/FluentUI/Button/Button.swift
Saadnajmi: TIL :D

macos/FluentUI/Button/Button.swift
anandrajeswaran: Yeah if the issue is updating when hover or disabled state changes, then the fix should be ensuring we re-compute that color when hover state and disable state change. We shouldn't rely on it implicitly happening to work because something else causes an updateLayer call in those cases

macos/FluentUI/Button/Button.swift
anandrajeswaran: I had the same assumption that you could call it whenever and rely on only one coalesced call to the actual display code happening. Pretty sure you can rely on that.

Regarding when to call it, I think it should only be called when we change something that we use in a display override (drawRect: and friends). If we're just setting a property on the button that NSButton uses internally  and we don't use at all, we should be able to trust that the implementation of NSButton will properly mark itself as needing to be drawn.

For something like size, it seems like that would affect layout rather than display so that would need to invalidate the intrinsicContentSize?

chripear: Added needsDisplay to few more places.  Making most of the colors private helped with this too.

macos/FluentUI/Button/Button.swift
chripear: OK, added some machinery to avoid setting contentTintColor in updateLayer.

macos/FluentUI/Button/Button.swift
harrieshin: you can test locally it by making /fluentui-apple/macos/FluentUITestApp/Assets.xcassets/ic_fluent_chevron_left_16_filled not render as "template image" in Xcode project 
Screen Shot 2020-12-07 at 5 53 45 PM

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +234,402 @@ open class Button: NSButton {
...
+			x = (rect.width - imageSize.width) / 2
+			y = (rect.height - imageSize.height + titleSize.height + titleToImageSpacing - titleToImageVerticalSpacingAdjustment) / 2
+			width = imageSize.width
+			height = imageSize.height
+		case .imageAbove:
+			x = (rect.width - imageSize.width) / 2
+			y = (rect.height - imageSize.height - titleSize.height - titleToImageSpacing + titleToImageVerticalSpacingAdjustment) / 2
+			width = imageSize.width
+			height = imageSize.height
+		@unknown default:
```
harrieshin, line 441: can we delete default?

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +234,402 @@ open class Button: NSButton {
...
+	/// Dark mode: dark gray fill, white text/image.
+	case acrylic
+
+	/// Accent color text/image, no fill or outline.
+	case borderless
+}
 
-	private static let verticalEdgePadding: CGFloat = 4
+/// Combination of all formatting parameters.
+public struct ButtonFormat {
```
harrieshin, line 565: do we need this struct?

macos/FluentUIResources-macos/FluentUI-macos.xcassets/ButtonColors/neutralBackground2.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
...
+      "appearances" : [
+        {
+          "appearance" : "luminosity",
+          "value" : "dark"
+        }
+      ],
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "1.000",
```
harrieshin, line 25: this should be hex rather than floating number which could be inconsistent.

harrieshin: same comments for other color files

macos/FluentUI/Button/Button.swift
chripear: There's already another thread on this subject.  I made the struct because it's convenient to create a handful of format structs and apply them to buttons en masse using the designated initializer that accepts this struct (or using the synthesized `format` property.  I'm using it extensively in the test app.

macos/FluentUIResources-macos/FluentUI-macos.xcassets/ButtonColors/neutralBackground2.colorset/Contents.json
chripear: Any way you know of for Xcode to default to hex values when I'm working with color assets?  Changing this drop-down for every single color is getting really old...

macos/FluentUI/Button/Button.swift
chripear: This default cases are in the original code.  I can remove them.

chripear: JK removing this breaks the build.  Xcode gives me this error and has an auto-fix for it:

Switch covers known cases, but 'NSControl.ImagePosition' may have additional unknown values, possibly added in future versions

Handle unknown values using "@unknown default"

macos/FluentUI/Button/Button.swift
chripear: Update: with the latest changes we're no longer respecting contentTintColor.  If the client attempts to set it on a button, we'll clobber it with contentTintColor{Rest,Pressed,Disabled} the next time the pressed or enabled state changes.

macos/FluentUI/Button/Button.swift
chripear: imageDimsWhenDisabled does apply to non-template images.  (contentTintColor does not.)

macos/FluentUI/Button/Button.swift
chripear: @anandrajeswaran how do we invalidate intrinsicContentSize?

anandrajeswaran: https://developer.apple.com/documentation/appkit/nsview/1526864-invalidateintrinsiccontentsize

macos/FluentUI/Button/Button.swift
Saadnajmi: Yeah, I'm not sure why this switch needs that `@uknown` param but others don't. Probably something about it coming from objective c.

Saadnajmi:
Oh yeah... I forgot about this. We do have local documentation in the form of markdown files for each of the macOS controls, [here](https://github.com/microsoft/fluentui-apple/blob/master/macos/docs/Controls/Button.md). 
Those eventually make it to the [FluentUI Public website](https://developer.microsoft.com/en-us/fluentui#/controls/mac/button) by making a PR [like this one](https://github.com/microsoft/fluentui/pull/13910).

I don't know if that's worth blocking the review, since we (surprise) haven't seemed to keep up with documentation muc, but it's worth noting, even if it's just future work for the next documentation day.

macos/FluentUI/Button/Button.swift
anandrajeswaran: Yeah swift 5 added a feature for this which is a regression imo. See https://developer.apple.com/documentation/swift/objective-c_and_c_code_customization/grouping_related_objective-c_constants for some context, but essentially because this is not marked as NS_CLOSED_ENUM it is not frozen and thus may add values in the future. As a "convenience", the swift compiler forces you to add unknown default for this so your code automatically still compiles even when new enums get added.

I might be missing a subtlety of it, but to me that negates one of the main benefits of using switch/case statements in that you can avoid the default case and get a compiler error if a new enum value is added, requiring each case to go explicitly decide how to handle this new case. I believe adding a new case would now result in a warning rather than an error if an unknown default case exists, but don't get me started on the many flaws of SWIFT_TREAT_WARNINGS_AS_ERRORS :). I personally would vote to use NS_CLOSED_ENUM wherever we have code that may be consumed by swift.

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +233,403 @@ open class Button: NSButton {
...
+			backgroundColorRest = accentColor
+			backgroundColorPressed = accentColor.withSystemEffect(.pressed)
+			backgroundColorDisabled = ButtonColor.brandBackgroundDisabled
+			borderColorRest = .clear
+			borderColorPressed = .clear
+			borderColorDisabled = .clear
+		case .secondary:
+			contentTintColorRest = .textColor
+			contentTintColorPressed = ButtonColor.neutralInverted?.withSystemEffect(.pressed)
+			contentTintColorDisabled = NSColor.textColor.withSystemEffect(.disabled)
```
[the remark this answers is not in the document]
chripear: I tried it.  Doesn't work.  The compiler is not able to guess what .textColor is after .withSystemEffect is applied to it.

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +233,403 @@ open class Button: NSButton {
...
+			contentTintColorDisabled = ButtonColor.neutralForeground3?.withSystemEffect(.disabled)
+			backgroundColorRest = ButtonColor.neutralBackground3
+			backgroundColorPressed = ButtonColor.neutralBackground3?.withSystemEffect(.pressed)
+			backgroundColorDisabled = ButtonColor.neutralBackground3?.withSystemEffect(.disabled)
+			borderColorRest = .clear
+			borderColorPressed = .clear
+			borderColorDisabled = .clear
+		case .borderless:
+			contentTintColorRest = accentColor
+			contentTintColorPressed = accentColor.withSystemEffect(.deepPressed)
```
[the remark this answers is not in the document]
chripear: Yes, .deepPressed is darker than .pressed.  This makes the color change more visible in the smaller area covered by the contentTintColor.  That's also reflected in the design docs: there they have shade 10 for brand background color when pressed, and shade 30 for brand foreground color when pressed.  I am using system effects to approximate that, rather than making the client specify a host of related colors for the accent.

macos/FluentUI/Button/Button.swift
```diff
@@ -175,304 +233,403 @@ open class Button: NSButton {
...
 	}
 
-	/// The primary style of the button
-	@objc public var style: ButtonStyle = .primaryFilled {
+	private func setColorValues(forStyle: ButtonStyle, accentColor: NSColor) {
+		switch forStyle {
+		case .primary:
+			contentTintColorRest = ButtonColor.neutralInverted
+			contentTintColorPressed = ButtonColor.neutralInverted?.withSystemEffect(.pressed)
+			contentTintColorDisabled = ButtonColor.brandForegroundDisabled
```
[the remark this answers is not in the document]
chripear: Yes, I am mostly following the design doc, which is a work in progress.  We will almost certainly revisit and tweak these colors later.

macos/FluentUI/Button/Button.swift
chripear: Added this invalidation to the size property's didSet.

chripear:
Added updated docs to the pull request.

---
In reply to: [741579753](https://github.com/microsoft/fluentui-apple/pull/327#issuecomment-741579753) [](ancestors = 741579753)

macos/docs/Controls/Button.md
```diff
@@ -1,40 +1,80 @@
...
+- Make sure the label conveys a clear purpose of the button to the user.
+- Describe the action the button will perform, ideally with a verb. Use concise, specific, self-explanatory labels, usually a single word.
+- When using an image in a button, use a template image so that it takes on the style and color of the button.
+- For best results when using the `linkedPrimary` property, the current button should have the `.secondary` style, the linked button should have the `.primary` style, and both buttons should have the same `accentColor`.
+
+### Things to avoid
+- Don't place the default focus on a button that performs a destructive action. Place the default focus on a button that performs a "safe" action instead (e.g "Cancel").
 - Overriding the underlying NSButtonCell will result in undefined behavior. While this may fit your use case, take caution while doing so.
-- Displaying both a title and an image currently does not render properly, and is not recommended.
+- Setting the contentTintColor property will result in undefined behavior.  Use the style and accentColor properties to determine the look of the button instead.
```
Saadnajmi, line 72: If we felt strongly about it, we could override the setter for `contentTintColor` to cause a `preconditionFailure()`, which we did in other parts of the button. But I think this being in documentation is good enough.

Saadnajmi (approved):

harrieshin (approved):

PR #328, opened 2020-12-02
Add Objective-C support to Drawer background colors

cy255:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add Objective-C support to Drawer background color and popover background color

### Verification

Verified that the generated Obj-C header FluentUI-Swift.h contains definitions for the drawer background colors.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/328)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -80,6 +80,10 @@ public extension Colors {
         public static var background = UIColor(light: surfacePrimary, dark: surfaceSecondary)
         public static var popoverBackground = UIColor(light: surfacePrimary, dark: surfaceQuaternary)
     }
+
+    // Objective-C support
+    @objc static var drawerBackground: UIColor { return Drawer.background }
+    @objc static var popoverBackground: UIColor { return Drawer.popoverBackground }
 }
 
 // MARK: - DrawerControllerDelegate


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -80,6 +80,10 @@ public extension Colors {
         public static var background = UIColor(light: surfacePrimary, dark: surfaceSecondary)
         public static var popoverBackground = UIColor(light: surfacePrimary, dark: surfaceQuaternary)
     }
+
+    // Objective-C support
+    @objc static var drawerBackground: UIColor { return Drawer.background }
+    @objc static var popoverBackground: UIColor { return Drawer.popoverBackground }
```
anandrajeswaran, line 86: Thoughts on just moving this out of a Colors extension (where it has to be in a Drawer struct) and into Drawer so we don't need this dance to make it consumable by objC?

anandrajeswaran: Also do we need to specify public? Or does @objc do that implicitly?

anandrajeswaran (approved):

rdeassis (approved):

ios/FluentUI/Drawer/DrawerController.swift
Saadnajmi: > Thoughts on just moving this out of a Colors extension (where it has to be in a Drawer struct) and into Drawer so we don't need this dance to make it consumable by objC?

Yeah, I thought w were going for that for future changes, but I guess we can do that next time 🙂

PR #330, opened 2020-12-04
Adding the SwiftUI implementation for the Button.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change adds the SwiftUI tokenized implementation of the Button with icon support.

It consists of the following itemized changes:

1. Because we're using the SwiftUI framework which defines types as Button and ButtonStyle, we need to rename the current button implementation to avoid clashes.

2. We're adding the SwiftUI implementation (ButtonVnext.swift) with a hosting controller wrapper to enable it to be added in UIKit view hierarchies.

3. The demo app was updated to use the new button on the following demo controllers:
- **Button (vnext):** added to showcase the new button with its styles (see verification screenshots in light and dark modes below)

- **Notification View:** the new button delivers the new functionality with no issues replacing the previous implementation
NotificationView

- **HUD:** because there are issues with constraints to make it use the entire width of the screen, the stack view was changed to align the buttons to the left

| Screen Shot 2020-12-03 at 1 28 35 PM | HUD |

- **Tab Bar View:** the "+" and "-" buttons were added to showcase state changes (enabled/disabled) being automatically refreshed/rendered as the property is updated (there are some issues with constraints as far as the button's view width goes.

| Screen Shot 2020-12-04 at 11 55 34 AM | TabBarView |

- **Objective C Demos:** Added the new button sample with other buttons to enable/disable, set title and image of the button to showcase the state changes coming from the objc code (selectors triggered by UIButtons)
ObjcDemos

### Verification

There are some issues with constraint support that we still need to discuss before committing to the replacement of the previous implementation of the button (subclass of UIButton) to the new SwiftUI button.

A conversation will be started to reach a decision on that matter.

| Light mode                                       | Dark mode                                      |
| buttonVnextLight | buttonVnextDark
|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/330)


Saadnajmi:
Just curious, does this also render on macOS? If so, could we see a screenshot?

Saadnajmi:
I recall @chripear adding some new styles to the macOS button like "Acrylic". Are the styles standardized across platforms / have we agreed on the new styles? It would be nice if that was consistent across the new converged vnext controls we have :)

ios/FluentUI/Generated/FluentUIStyle.generated.swift
```diff
@@ -0,0 +1,2958 @@
...
+	}
+}
+
+/// Your view should conform to 'AppearaceProxyComponent'.
+public protocol AppearaceProxyComponent: class {
+    associatedtype ApperanceProxyType
+    var appearanceProxy: ApperanceProxyType { get }
+    var themeAware: Bool { get set }
+    func didChangeAppearanceProxy()
+}
```
Saadnajmi, line 18: type, appearance?

ios/FluentUI/Generated/FluentUIStyle.generated.swift
```diff
@@ -0,0 +1,2958 @@
+/// Autogenerated file
```
Saadnajmi, line 1: If this is generated, should it be checked into source?

sophialee0416 (approved):
Let's continue the discussion regarding Constraints and other underlying issues as this will be integral towards our efforts in SwiftUI.

kubalani (approved):
Approving changes to unblock merging `Drawer` changes. 
Three incremental changes (may be) needed post merging this: 
  1. Investigate collision of naming convension between swiftui & uikit.
  2. Optimize flow of observed objects using environment objects.
  3. Investigate layout constraints resolution for switui objects embedded within uikit

ios/FluentUI/Generated/FluentUIStyle.generated.swift
rdeassis: This is because the file is not auto-generated as part of the FluentUI Build process.
This file won't be changing frequently unless a new control is added or a change is made on the tokens from the design team.

But any time it's changed, it should be re-generated from the YAML token files and checked in here since all the controls depend on it.

ios/FluentUI/Generated/FluentUIStyle.generated.swift
rdeassis: I didn't get your point here. Can you elaborate?

rdeassis:
> I recall @chripear adding some new styles to the macOS button like "Acrylic". Are the styles standardized across platforms / have we agreed on the new styles? It would be nice if that was consistent across the new converged vnext controls we have :)

So, this ties to your other question about rendering the button on the Mac.
Technically we can render them on both platforms with no issues. 
From the FluentUI side, we have different tokens and styles for Mac and iOS. But I believe it's still possible to entertain the idea of a single SwiftUI button that could serve both platforms with their specific tokens.

PR #331, opened 2020-12-07
Add Obj-C support to PillButton color constants

cy255:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Add Obj-C support to PillButton color constants
- Add Obj-C support to PillButton corner radius

### Verification
-Verified that PillButton colors and corner radius are included in the generated Obj-C header FluentUI-Swift.h

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/331)

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -5,20 +5,20 @@
 
 import UIKit
 
-// MARK: PillButton Colors
-
-public extension Colors {
-    struct PillButton {
-        public struct Outline {
-            public static var background = UIColor(light: surfaceTertiary, dark: surfaceSecondary)
-            public static var title = UIColor(light: textSecondary, dark: textPrimary)
-            public static var titleSelected = UIColor(light: textOnAccent, dark: textDominant)
-        }
-        public struct Filled {
-            public static var title = UIColor(light: textOnAccent, dark: Outline.title)
-        }
-    }
-}
+//// MARK: PillButton Colors
+//
+//public extension Colors {
+//    struct PillButton {
+//        public struct Outline {
+//            public static var background = UIColor(light: surfaceTertiary, dark: surfaceSecondary)
+//            public static var title = UIColor(light: textSecondary, dark: textPrimary)
+//            public static var titleSelected = UIColor(light: textOnAccent, dark: textDominant)
+//        }
+//        public struct Filled {
+//            public static var title = UIColor(light: textOnAccent, dark: Outline.title)
+//        }
+//    }
+//}
 
 // MARK: - PillButtonStyle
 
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -111,13 +96,13 @@ public typealias MSPillButton = PillButton
...
-        static let horizontalInset: CGFloat = 16.0
-        static let topInset: CGFloat = 6.0
-    }
+
+    @objc public static let outlineStyleBackgroundColor = UIColor(light: Colors.surfaceTertiary, dark: Colors.surfaceSecondary)
+    @objc public static let outlineStyleTitleColor = UIColor(light: Colors.textSecondary, dark: Colors.textPrimary)
+    @objc public static let outlineStyleTitleSelectedColor = UIColor(light: Colors.textOnAccent, dark: Colors.textDominant)
+    @objc public static let filledStyleTitleColor = UIColor(light: Colors.textOnAccent, dark: outlineStyleTitleColor)
+
+    @objc public static let cornerRadius: CGFloat = 16.0
```
anandrajeswaran, line 105: Do we actually need all of these? Is there an simpler approach of modifications needed so we can just use PillButton instead of needing all these constants to recreate it?

harrieshin:
can you update your PR description with compared screenshots please?

ios/FluentUI/Pill Button Bar/PillButton.swift
harrieshin: not sure I think cornerRadius should be exposed

harrieshin:
what's the intended use here? if it is PillButtonBar, there is a way to customize PillButtonBar background color via "pillbuttonbackgroundcolor." also, changing the outlinestyle color itself seems not following the fluent guideline

cy255:
> what's the intended use here? if it is PillButtonBar, there is a way to customize PillButtonBar background color via "pillbuttonbackgroundcolor." also, changing the outlinestyle color itself seems not following the fluent guideline
@harrieshin
For some reason I can't reply to the comment directly, so answering here: in certain scenarios, PillButtonBar is not being used to contain the PillButtons (for example in the case of multiple sections of grouped PillButtons), so some other custom view is used. In order for those custom views (in Obj-C) to adopt similar appearances (e.g. background color, corner radius) these constants need to be publicly accessible to Obj-C.

harrieshin (approved):

ios/FluentUI/Pill Button Bar/PillButton.swift
anandrajeswaran: Since these are public they should probably have some documentation, which in this case seems pretty useful to explain why these are exposed (to allow clients to build custom PillButtonBars with similar appearance)

anandrajeswaran (approved):

PR #334, opened 2020-12-09
Adding SwiftUI implementation of Menu List

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR contains the first draft of List (counterpart of UITableView from UIKit), created in SwiftUI. Currently, it supports leading icons, title, and subtitle properties. 

### Verification

- The TableViewCellVnext Demo Controller uses the same sample data as the non-vNext version. All issues mentioned in [PR#330](https://github.com/microsoft/fluentui-apple/pull/330) also pertain to this component.

| | |
|--------|------|
|Simulator Screen Shot - iPhone 12 Pro - 2020-12-30 at 15 53 26 | Simulator Screen Shot - iPhone 12 - 2020-12-21 at 17 54 03 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/334)


harrieshin:
looking at the screenshot this doesn't seem right. 
-the singleline cell hight seeems incorrect. 
- the title isn't veritcally centered. 
- separator should be indented and more aligned with text
- double cell height seem not right too?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+  public class MSFListTokens: ObservableObject {
+      @Published public var backgroundColor: UIColor!
+      @Published public var borderColor: UIColor!
+      @Published public var leadingIconColor: UIColor!
+      @Published public var trailingIconColor: UIColor!
+      @Published public var leadingTextColor: UIColor!
+      @Published public var trailingTextColor: UIColor!
+
+      @Published public var borderSize: CGFloat!
+      @Published public var padding: CGFloat!
```
harrieshin, line 44: padding of what? lets be more specific?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+      @Published public var textFont: UIFont!
+      @Published public var subtitleFont: UIFont!
+      @Published public var leadingIconSize: CGFloat!
+      @Published public var trailingIconSize: CGFloat!
+      @Published public var interspace: CGFloat!
```
harrieshin, line 49: this needs to be more specific or better documentation

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+          case .button:
+              interspace = appearanceProxy.interspace.button
+          }
+      }
+  }
+
+  public struct MSFListView: View {
+      @ObservedObject var state: MSFListItem
+      @ObservedObject var tokens: MSFListTokens
+      var numCells: [MSFListItem]
```
rdeassis, line 103: why numCells? isn't this the entire collection of cells and not only the number of cells?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+          List(numCells) { cell in
+              HStack {
+                  Image(cell.leadingView ?? "")
+                      .resizable()
+                      .frame(width: tokens.leadingIconSize, height: tokens.leadingIconSize)
+                  VStack(alignment: .leading) {
+                      Text(cell.title)
+                          .font(Font(tokens.textFont))
+                          .foregroundColor(Color(tokens.leadingTextColor))
+                      Text(cell.subtitle ?? "")
```
rdeassis, line 121: Enclosing this with an if statement will help you work on the alignment issues when there's not subtitle.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+      @objc @Published public var subtitle: String?
+  }
+
+  @objc(MSFListAccessoryVnextStyle)
+  /// Pre-defined styles of icons
+  public enum MSFListAccessoryVnextStyle: Int, CaseIterable {
+      case iconOnly
+      case withLabel
+      case disclosure
+  }
```
rdeassis, line 24: Can you make sure these style variations are demoed in the app?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+
+  @objc(MSFListAccessoryVnextSize)
+  /// Pre-defined sizes of icons
+  public enum MSFListAccessoryVnextSize: Int, CaseIterable {
+      case icon
+      case persona
+      case disclosure
+      case button
+  }
```
rdeassis, line 33: Why is this named size?
And what is the difference between disclosure here compared to the disclosure in MSFListAccessoryVnextStyle for example?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -161,3 +162,14 @@ SecondaryButtonTokens extends MSFButtonTokens:
...
   borderColor: [ rest: $Colors.Neutral.clear, hover: $Colors.Neutral.clear, pressed: $Colors.Neutral.clear, selected: $Colors.Brand.shade10, disabled: $Colors.Neutral.clear ]
 
+AP_MSFListTokens:
+  backgroundColor: $Colors.Background.neutral1
+  borderColor: $Colors.Stroke.neutral1
+  borderSize: $Border.size.thin
+  padding: $Spacing.medium
+  textFont: $Typography.body
+  subtitleFont: $Typography.subheadline
+  iconColor: [ iconOnly: $Colors.Foreground.neutral3, withLabel: $Colors.Foreground.neutral4, disclosure: $Colors.Foreground.neutral4 ]
```
rdeassis, line 172: I think the variations in the figma should be appearance proxies that extend MSFListTokens.
Once we get the state colors like pressed, hover, etc... this will be a matrix of styles vs states for this color.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+      @objc open var state: MSFListItem {
+          return self.hostingController.rootView.state
+      }
+
+      @objc public init(cells: [MSFListItem],
+                        style: MSFListAccessoryVnextStyle = .iconOnly,
+                        size: MSFListAccessoryVnextSize = .icon) {
+          self.hostingController = UIHostingController(rootView: MSFListView(cells: cells,
+                                                                                  style: style,
+                                                                                  size: size))
```
rdeassis, line 151: nit: params misaligned

rdeassis (commented):
This file should not be part of the change:
tools/sgen/Pods/Target Support Files/Pods-sgen/Pods-sgen-frameworks.sh

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
+  //
+  //  Copyright (c) Microsoft Corporation. All rights reserved.
+  //  Licensed under the MIT License.
+  //
+
+  import UIKit
+  import SwiftUI
+
+  ///Properties that make up cell content
+  @objc(MSFListVnextItem)
```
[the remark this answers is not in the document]
sophialee0416: I plan on using NavigationView for the cells, it has not been implemented yet

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,169 @@
...
+  public class MSFListItem: NSObject, ObservableObject, Identifiable {
```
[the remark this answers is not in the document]
sophialee0416: I originally had regular list but then decided to follow naming patterns with the rest of vNext components. Open to going back to List though

kubalani (approved):

kubalani (approved):

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+        List {
+            ForEach(cells, id: \.self) { item in
+                MSFListCellView(cell: item, tokens: tokens)
+                    .onTapGesture(perform: item.handler ?? {})
+            }
+            .border(state.hasBorder ? Color(tokens.borderColor) : Color.clear, width: state.hasBorder ? tokens.borderSize : 0)
+            .frame(minHeight: tokens.layoutType.height)
+            .listRowInsets(EdgeInsets())
+        }
+        .frame(width: UIScreen.main.bounds.width, height: 300)
```
kubalani, line 142: Can we use Geometry reader instead of `UIScreen` doesn't have to be part of this PR?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+        .environment(\.defaultMinListRowHeight, 0)
```
rdeassis, line 143: What is using this environment value on the List hierarchy?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -22,23 +23,57 @@ class ListVnextDemoController: DemoController {
...
+
     private func updateLayout(subtitle: String?) -> MSFListCellVnextLayoutType {
         if subtitle != "" {
             return MSFListCellVnextLayoutType.twoLines
         } else {
             return MSFListCellVnextLayoutType.oneLine
         }
     }
+
+    func showAlertForCellTapped() {
```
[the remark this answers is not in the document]
rdeassis: Think of the delegate pattern where you have the sender parameter as the object that triggered the action.
If the calling code wants to access the cell instance that triggered the handler to change a property or to remove the cell for example.

If you're not providing the instance of the cell, you have other options such as:
 - the index path (section and cell indexes as an ordered pair)
 - the cell data object id property to identify the one tapped (public var id = UUID())

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+///Properties that make up cell content
+@objc(MSFListVnextCell)
+public class MSFListVnextCell: NSObject, ObservableObject, Identifiable {
```
rdeassis, line 11: Shouldn't the name of this class be something more like MSFListVnextCellData?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+    public var id = UUID()
+    @objc @Published public var leadingView: UIImage?
```
rdeassis, line 13: Since this is an instance of UIImage, I think it makes more sense to call it iconImage.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+    @objc @Published public var title: String = ""
+    @objc @Published public var subtitle: String?
+    @objc @Published public var trailingView: TableViewCellAccessoryType = .none
```
rdeassis, line 16: Maybe name it trailingViewStyle or even accessoryType?
I think we should define our own enumeration values here instead of reusing the TableViewCellAccessoryType.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -161,3 +163,21 @@ SecondaryButtonTokens extends MSFButtonTokens:
...
+  backgroundColor: [ rest: $Colors.Background.neutral1, pressed: $Colors.Background.neutral5 ]
+  borderColor: $Colors.Stroke.neutral1
+  borderSize: $Border.size.thin
+  disclosureIconForegroundColor: $Colors.Foreground.neutral3
+  disclosureInterspace: $Spacing.xxSmall
+  disclosureSize: $Icon.size.small
+  horizontalCellPadding: $Spacing.medium
+  iconColor: $Colors.Foreground.neutral4
+  iconInterspace: $Spacing.medium
+  iconSize: [ icon: $Icon.size.medium, largeIcon: $Icon.size.xxlarge ]
```
rdeassis, line 176: Since the property is already named "iconSize", can we call we call the properties just with their predefined size name (small, medium, large...)?

Also, it seems like the largeIcon is not being used. What is the scenario you're planning to use it?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -161,3 +163,21 @@ SecondaryButtonTokens extends MSFButtonTokens:
...
+  labelColor: $Colors.Foreground.neutral1
+  sublabelColor: $Colors.Foreground.neutral3
+  textFont: [ label: $Typography.body, subLabel: $Typography.caption1 ]
```
rdeassis, line 179: I think this can be split into 2 tokens:
labelTextFont: $Typography.body
subLabelTextFont: $Typography.caption1

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+    @Published public var trailingItemForegroundColor: UIColor!
+
+    @Published public var highlightedBackgroundColor: UIColor!
+
+    @Published public var borderSize: CGFloat!
+    @Published public var disclosureInterspace: CGFloat!
+    @Published public var disclosureSize: CGFloat!
+    @Published public var horizontalCellPadding: CGFloat!
+    @Published public var iconInterspace: CGFloat!
+    @Published public var iconSize: CGFloat!
```
rdeassis, line 71: If iconSize and largeIconSize are not going to be used at the same time, we can make this only one var (iconSize) and assign the correct token value accordingly.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+                    }
+                }
+                Spacer()
+                ZStack(alignment: .trailing) {
+                    if let trailingView = cell.trailingView {
+                        if trailingView != .none {
+                            Image(uiImage: trailingView.icon!)
+                                .resizable()
+                                .foregroundColor(Color(trailingView == .disclosureIndicator ? tokens.disclosureIconForegroundColor : tokens.trailingItemForegroundColor))
+                                .frame(width: trailingView == .disclosureIndicator ? tokens.disclosureSize : tokens.iconSize,
```
rdeassis, line 189: It's good practice to avoid repeating the property calls.
You can do this on the top of the method for the disclosureIndicator for example:

`let disclosureIndicatorSize = tokens.disclosureSize`

and use it throughout the implementation here.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,225 @@
...
+
+        backgroundColor = appearanceProxy.backgroundColor.rest
+        borderColor = appearanceProxy.borderColor
+        disclosureIconForegroundColor = appearanceProxy.disclosureIconForegroundColor
+        iconColor = appearanceProxy.iconColor
+        leadingTextColor = appearanceProxy.labelColor
+        subtitleColor = appearanceProxy.sublabelColor
+        trailingItemForegroundColor = appearanceProxy.trailingItemForegroundColor
+
+        highlightedBackgroundColor = appearanceProxy.backgroundColor.pressed
```
rdeassis, line 104: I think the press effect (using the highlightedBackgroundColor while the cell is being pressed) is one of the important things we need to have for our MVP implementation.

rdeassis (changes requested):

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: SwiftUI's List currently has built-in default padding in between the rows, and there is no modifier to remove them without hardcoding negative padding. This helps with eliminating the extra padding so that the tokens can be reflected more accurately

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Double-lined cells have the option of using large icons (as per figma)

ios/FluentUI/Table View/ListVnext.swift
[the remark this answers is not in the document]
sophialee0416: For referencing: ended up not using NavigationView, instead passed in a closure

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,296 @@
...
+        var tokens: MSFListTokens
+        @State var isTapped: Bool = false
+
+        init(cell: MSFListVnextCellData, tokens: MSFListTokens) {
+            self.cell = cell
+            self.tokens = tokens
+        }
+
+        var body: some View {
+            HStack(spacing: 0) {
```
rdeassis, line 193: I have just tried enclosing all this content in a Button and got the TableView behavior (including the background color change):

https://user-images.githubusercontent.com/68076145/103319927-6e7f2a00-49e8-11eb-9eab-d575e5e1a827.mov

...```

        var body: some View {
            Button(action: cell.onTapAction ?? {}, label: {
                HStack(spacing: 0) { ...
```

File: ios/FluentUI/Table View/ListVnext.swift [hunk 23]
```diff
@@ -0,0 +1,296 @@
...
+                        }
+                    }
+                }
+            }
+            .contentShape(Rectangle())
+            .padding(.leading, tokens.horizontalCellPadding)
+            .padding(.trailing, tokens.horizontalCellPadding)
+            .frame(minHeight: tokens.layoutType.height)
+            .listRowInsets(EdgeInsets())
+            .background(isTapped ? Color(tokens.highlightedBackgroundColor) : Color.clear)
```
rdeassis, line 238: The only thing I still could not figure out is how to customize the background color for when the button is pressed.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,296 @@
...
+            .onTapGesture(perform: cell.onTapAction ?? {})
+            .gesture(DragGesture(minimumDistance: 0, coordinateSpace: .local)
```
rdeassis, line 240: With the suggestion of using a button as the content (line 193 above), these gestures are probably not going to be needed anymore.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,305 @@
...
+                }
+            }
+            ForEach(cells, id: \.self) { item in
+                MSFListCellView(cell: item, tokens: tokens)
+                    .border(state.hasBorder ? Color(tokens.borderColor) : Color.clear, width: state.hasBorder ? tokens.borderSize : 0)
+                    .frame(maxWidth: .infinity)
+            }
+        }
+        .environment(\.defaultMinListRowHeight, 0)
+        .frame(width: .infinity)
```
rdeassis, line 183: This should be maxWidth, right?
It should also have **maxHeight: .infinity**

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -0,0 +1,85 @@
...
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class ListVnextDemoController: DemoController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        container.alignment = .leading
```
rdeassis, line 12: remove this line so you get the view to stretch and snap on its super view.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -0,0 +1,85 @@
...
+                    indexPath.row = row
+                    indexPath.section = sectionCount
+                    self.showAlertForCellTapped(indexPath: indexPath)
+                }
+                listItems.append(listItem)
+            }
+
+            list = MSFListVnext(cells: listItems, layoutType: updateLayout(subtitle: listItems[0].subtitle), iconStyle: iconStyle)
+            list.state.sectionTitle = section.title
+            addRow(items: [list.view])
```
rdeassis, line 46: If you remove line 12 and use **container.addArrangedSubview(list.view)** instead, you'll get the constraint behavior we want (taking up the entire width and height:


https://user-images.githubusercontent.com/68076145/103374696-14847000-4a8d-11eb-998a-af9b15341239.mov

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -0,0 +1,97 @@
...
+            return .checkmark
+        case 4:
+            return .none
+        default:
+            return .none
+        }
+    }
+
+    private func updateLayout(subtitle: String?) -> MSFListCellVnextLayoutType {
+        if subtitle != "" {
```
rdeassis, line 81: Are we covering the three line scenario?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -63,6 +63,30 @@ - (void)viewDidLoad {
...
+    listCell3.leadingIcon = image;
+    listCell3.trailingIcon = MSFListAccessoryTypeDisclosure;
+    
+    MSFListVnextSectionData *section = [[MSFListVnextSectionData alloc] init];
+    section.layoutType = MSFListCellVnextHeightTwoLines;
+    section.cells = @[listCell1, listCell2, listCell3];
+    NSArray *sections = @[section];
+
+    MSFListVnext *list = [[MSFListVnext alloc] initWithSections:sections iconStyle:MSFListIconVnextStyleNone];
+    [self.container addArrangedSubview:[list view]];
```
rdeassis, line 89: Can you fix the constraints here in the same way they were done on the List Demo Controller?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -63,6 +63,30 @@ - (void)viewDidLoad {
 
     MSFButton *resetButton = [self createButtonWithTitle:@"Reset Vnext Button" action:@selector(resetVnextButton)];
     [self.container addArrangedSubview:resetButton];
+    
+    UILabel *listVnextLabel = [[UILabel alloc] init];
+    [listVnextLabel setText:@"List (vNext)"];
+    [self.container addArrangedSubview:listVnextLabel];
+    
+    MSFListVnextCellData *listCell1 = [[MSFListVnextCellData alloc] init];
```
rdeassis, line 71: Can you add an example of an action in one of those cells?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+        }
+        return icon
+    }
+}
+
+///Properties that make up cell content
+@objc(MSFListVnextCellData)
+public class MSFListVnextCellData: NSObject, ObservableObject, Identifiable {
+    public var id = UUID()
+    @objc @Published public var leadingIcon: UIImage?
```
rdeassis, line 45: We'll soon need to add a view as an option to the image (Avatar cell scenario).
Can you create a GitHub issue to track that work?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+                            }
+                        }
+                    }
+                }
+            })
+            .buttonStyle(ListCellButton(tokens: tokens, layoutType: layoutType))
+        }
+    }
+
+    struct ListCellButton: ButtonStyle {
```
rdeassis, line 254: ListCellButtonStyle instead of ListCellButton?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+        let title: String
+        var tokens: MSFListTokens
+
+        init(title: String, tokens: MSFListTokens) {
+            self.title = title
+            self.tokens = tokens
+        }
+
+        var body: some View {
+            GeometryReader { _ in
```
rdeassis, line 279: Why do you need a GeometryReader here?

sophialee0416: Headers are kinda wonky right now where they're not meant to be very customizable. Adding GeometryReader forces the view to expand (kind of like a Spacer) so that the default background color does not show. Below is a screenshot of what it looks like with the GeometryReader removed. I've also tried resetting list row insets, padding, frame and the header view did not seem to budge.

Screen Shot 2020-12-30 at 5 52 30 PM

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
sophialee0416: As of now, no. We do not have the tokens defined yet, but @Victa should be following up with them later when we need them

rdeassis: Sounds good! I think it's worth creating a GitHub issue to track that work item.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+
+    @objc open var view: UIView {
+        return hostingController.view
+    }
+
+    @objc open var state: MSFListVnextState {
+        return hostingController.rootView.state
+    }
+
+    @objc open var sections: [MSFListVnextSectionData] {
```
rdeassis, line 304: Is there a specific reason why this is not part of the state object?
I think it makes sense to have the sections as part of the state so we get the list updated when rows/sections get added or removed.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+
+/// Pre-defined styles of icons
+@objc(MSFListIconVnextStyle)
+public enum MSFListIconVnextStyle: Int, CaseIterable {
+    case none
+    case iconOnly
+    case large
+}
+
+/// Pre-defined types of icons
```
rdeassis, line 17: nit: Pre-defined accessory types

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+        case .detailButton:
+            icon = UIImage.staticImageNamed("more-24x24")
+        case .checkmark:
+            icon = UIImage.staticImageNamed("checkmark-24x24")
+        }
+        return icon
+    }
+}
+
+///Properties that make up cell content
```
rdeassis, line 41: nit: space after ///

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+    @objc @Published public var leadingIcon: UIImage?
+    @objc @Published public var title: String = ""
+    @objc @Published public var subtitle: String?
+    @objc @Published public var trailingIcon: MSFListAccessoryType = .none
+    @objc @Published public var titleLineLimit: Int = 1
+    @objc @Published public var subtitleLineLimit: Int = 1
+    @objc public var onTapAction: (() -> Void)?
+}
+
+///Properties that make up section content
```
rdeassis, line 54: nit: space after ///

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+@objc(MSFListVnextSectionData)
+public class MSFListVnextSectionData: NSObject, ObservableObject, Identifiable {
+    public var id = UUID()
+    @objc @Published public var cells: [MSFListVnextCellData] = []
+    @objc @Published public var title: String?
+    @objc @Published public var layoutType: MSFListCellVnextLayoutType = .oneLine
```
rdeassis, line 60: Wouldn't it be nicer to allow the calling code to specify the layout type per cell (in the MSFListVnextCellData)?
We can discuss this after your first merge (it's worth a GitHub issue to track as well).

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+}
+
+///Properties that make up list content
```
rdeassis, line 63: nit: space after ///

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+}
+
+@objc(MSFListCellVnextHeight)
+/// Pre-defined layout heights of cells
+public enum MSFListCellVnextLayoutType: Int, CaseIterable {
+    case oneLine
+    case twoLines
+    case threeLines
+
+    var height: CGFloat {
```
rdeassis, line 76: Another thing that can be changed later (GitHub issue):
These values should be tokenized in the YAML file.
Something like:

```
AP_MSFListTokens:
  cellHeight: [ oneLine: 48, twoLines: 64, threeLines: 84 ]
  ...
```

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+                        .frame(maxWidth: .infinity)
+                }
+            }
+        }
+        .environment(\.defaultMinListRowHeight, 0)
+        .frame(maxWidth: .infinity, maxHeight: .infinity)
+    }
+}
+
+extension MSFListView {
```
rdeassis, line 194: What's the benefit of having these types in the MSFListView extension?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+                    }
+                    VStack(alignment: .leading) {
+                        if let title = cell.title {
+                            Text(title)
+                                .font(Font(tokens.textFont))
+                                .foregroundColor(Color(tokens.leadingTextColor))
+                                .lineLimit(cell.titleLineLimit)
+                        }
+                        if let subtitle = cell.subtitle {
+                            if subtitle != "" {
```
rdeassis, line 224: if !subtitle.isEmpty {

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,313 @@
...
+                            .padding(.trailing, tokens.iconInterspace)
+                    }
+                    VStack(alignment: .leading) {
+                        if let title = cell.title {
+                            Text(title)
+                                .font(Font(tokens.textFont))
+                                .foregroundColor(Color(tokens.leadingTextColor))
+                                .lineLimit(cell.titleLineLimit)
+                        }
+                        if let subtitle = cell.subtitle {
```
rdeassis, line 223: if let subtitle = cell.subtitle, !subtitle.isEmpty {

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -63,6 +63,39 @@ - (void)viewDidLoad {
...
+    section.cells = @[listCell1, listCell2, listCell3];
+    NSArray *sections = @[section];
+
+    MSFListVnext *list = [[MSFListVnext alloc] initWithSections:sections iconStyle:MSFListIconVnextStyleNone];
+
+    UIView *listView = [list view];
+    listView.translatesAutoresizingMaskIntoConstraints = false;
+
+    [self.container addArrangedSubview:[list view]];
+
```
rdeassis, line 97: super nit: remove line

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: It was for organization purposes, but I'm thinking of splitting the file (which will be tracked in a GH issue)

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: Updated to include inside state

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: I wanted to consider both cell design + consistency... should we allow sections to contain cells of different heights? I agree we should continue this in a discussion, I will create an issue

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -120,6 +120,9 @@ Typography:
 Border:
   radius: [ none: 0pt, small: 2pt, medium: 4pt, large: 8pt, xlarge: 12pt ]
   size: [ none: 0pt, thin: 1pt, thick: 2pt, thicker: 4pt, thickest: 6pt ]
+  
+Cell:
+  height: [ oneLine: 48pt, twoLines: 64pt, threeLines: 84pt ]
```
rdeassis, line 125: This doesn't seem necessary since it's exclusive to the List cells.
I'd just put the values directly in the AP_MSFListTokens:

AP_MSFListTokens:
  cellHeight: [ oneLine: 48pt, twoLines: 64pt, threeLines: 84pt ]

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/364

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/367

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/365

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/366

tools/sgen/input/FluentUIStyle.yml
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/362

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,308 @@
...
+}
+
+/// Properties that make up cell content
+@objc(MSFListVnextCellData)
+public class MSFListVnextCellData: NSObject, ObservableObject, Identifiable {
+    public var id = UUID()
+    @objc @Published public var leadingIcon: UIImage?
+    @objc @Published public var title: String = ""
+    @objc @Published public var subtitle: String?
+    @objc @Published public var trailingIcon: MSFListAccessoryType = .none
```
rdeassis, line 48: **accessoryType** instead of **trailingIcon**?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,308 @@
...
+                        if let subtitle = cell.subtitle, !subtitle.isEmpty {
+                                Text(subtitle)
+                                    .font(Font(tokens.subtitleFont))
+                                    .foregroundColor(Color(tokens.subtitleColor))
+                                    .lineLimit(cell.titleLineLimit)
+                        }
+                    }
+                    Spacer()
+                    HStack {
+                        if let trailingIcon = cell.trailingIcon {
```
rdeassis, line 233: Instead of the 2 consecutive if statements:

```
                        if let accessoryType = cell.accessoryType,
                           accessoryType != .none,
                           let accessoryIcon = accessoryType.icon {
```

(provided you took my suggestion of renaming **trailingIcon** to **accessoryType** above)

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,308 @@
...
+                            if trailingIcon != .none {
+                                let isDisclosure = trailingIcon == .disclosure
+                                let disclosureSize = tokens.disclosureSize
+                                let iconSize = tokens.iconSize
+                                Image(uiImage: trailingIcon.icon!)
```
rdeassis, line 238: Then you can do just **accessoryIcon** here:

`Image(uiImage: accessoryIcon)`

rdeassis (commented):
Just some last nit comments, but overall looking good! :)

Also, can you remove Pods-sgen-frameworks.sh from the PR?
This is just about the execution permission that won't make any difference if checked in.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -0,0 +1,95 @@
...
+        }
+        list = MSFListVnext(sections: listData, iconStyle: iconStyle)
+
+        let listView = list.view
+        listView.translatesAutoresizingMaskIntoConstraints = false
+
+        let demoControllerView: UIView = self.view
+        demoControllerView.addSubview(listView)
+
+        NSLayoutConstraint.activate([demoControllerView.safeAreaLayoutGuide.topAnchor.constraint(equalTo: listView.topAnchor),
```
kubalani, line 55: If you are using a custom constraint it would a good idea to not inherit from `DemoController` and just use `UIViewController`

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -0,0 +1,95 @@
...
+                listCell.accessoryType = accessoryType(for: rowIndex)
+                listCell.onTapAction = {
+                    indexPath.row = rowIndex
+                    indexPath.section = sectionIndex
+                    self.showAlertForCellTapped(indexPath: indexPath)
+                }
+                listSection.cells.append(listCell)
+            }
+            listSection.layoutType = updateLayout(subtitle: listSection.cells[0].subtitle)
+             listData.append(listSection)
```
kubalani, line 45: How would the list rendering work with an async call?

sophialee0416: List will still get updated with changes post-creation

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -0,0 +1,311 @@
...
+                            .textCase(.none)
+                            .listRowInsets(EdgeInsets())
+                            .padding(.top, tokens.horizontalCellPadding / 2)
+                            .padding(.leading, tokens.horizontalCellPadding)
+                            .padding(.trailing, tokens.horizontalCellPadding)
+                            .background(Color(tokens.backgroundColor))
+                    } else {
+                        Text(sectionTitle)
+                            .listRowInsets(EdgeInsets(
+                                            top: tokens.horizontalCellPadding / 2,
```
rdeassis, line 170: nit: can you move line 170 up and indent the following ones accordingly?

rdeassis (approved):

rdeassis:
Thanks for addressing all the comments! :)

PR #335, opened 2020-12-09
Adding the Avatar Vnext implemented in SwiftUI.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

Adding the SwiftUI implementation of the Avatar.

### Verification

Screen Shot 2020-12-11 at 12 01 40 PM
Screen Shot 2020-12-11 at 12 01 52 PM
Screen Shot 2020-12-11 at 12 01 57 PM
Screen Shot 2020-12-11 at 12 25 53 PM
Screen Shot 2020-12-09 at 12 16 24 PM
Screen Shot 2020-12-09 at 12 16 34 PM
Screen Shot 2020-12-09 at 12 17 08 PM
Screen Shot 2020-12-09 at 12 17 13 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/335)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
@@ -62,6 +62,8 @@ class AvatarVnextDemoController: DemoController {
             let unauthenticatedAvatar = createAvatarView(size: size,
                                                          name: nil,
                                                          style: .unauthenticated)
+
+            avatarViewsUnauthenticated.append(unauthenticatedAvatar.1)
             addRow(text: size.description, items: [unauthenticatedAvatar.1.view], textStyle: .footnote, textWidth: 100)
         }
 
@@ -104,6 +106,10 @@ class AvatarVnextDemoController: DemoController {
                 for avatarView in avatarViewsWithInitials {
                     avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
                 }
+
+                for avatarView in avatarViewsUnauthenticated {
+                    avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
+                }
             }
         }
     }
@@ -143,6 +149,7 @@ class AvatarVnextDemoController: DemoController {
 
     private var avatarViewsWithImages: [AvatarVnext] = []
     private var avatarViewsWithInitials: [AvatarVnext] = []
+    private var avatarViewsUnauthenticated: [AvatarVnext] = []
 
     private func createSection(withTitle title: String,
                                name: String,
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
@@ -24,8 +24,13 @@ class AvatarVnextDemoController: DemoController {
                                                             switchAction: #selector(toggleAlternateBackground(switchView:)),
                                                             isOn: isUsingAlternateBackgroundColor)
 
+        let opaqueBorderSettingView = createLabelAndSwitchRow(labelText: "Use opaque borders",
+                                                            switchAction: #selector(toggleOpaqueBorders(switchView:)),
+                                                            isOn: isUsingOpaqueBorders)
+
         addRow(items: [backgroundSettingView])
         addRow(items: [showPresenceSettingView])
+        addRow(items: [opaqueBorderSettingView])
 
         createSection(withTitle: "Circle style for person",
                       name: "Kat Larrson",
@@ -89,6 +94,20 @@ class AvatarVnextDemoController: DemoController {
         }
     }
 
+    private var isUsingOpaqueBorders: Bool = false {
+        didSet {
+            if oldValue != isUsingOpaqueBorders {
+                for avatarView in avatarViewsWithImages {
+                    avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
+                }
+
+                for avatarView in avatarViewsWithInitials {
+                    avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
+                }
+            }
+        }
+    }
+
     private lazy var presenceIterator = AvatarVnextPresence.allCases.makeIterator()
 
     private func nextPresence() -> AvatarVnextPresence {
@@ -112,6 +131,10 @@ class AvatarVnextDemoController: DemoController {
 
     @objc private func toggleAlternateBackground(switchView: UISwitch) {
         isUsingAlternateBackgroundColor = switchView.isOn
+    }
+
+    @objc private func toggleOpaqueBorders(switchView: UISwitch) {
+        isUsingOpaqueBorders = switchView.isOn
     }
 
     private func updateBackgroundColor() {
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
@@ -1,0 +1,179 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class AvatarVnextDemoController: DemoController {
+    enum BorderStyle: Int {
+    case noBorder
+    case defaultBorder
+    case colorfulBorder
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        let showPresenceSettingView = createLabelAndSwitchRow(labelText: "Show presence",
+                                                              switchAction: #selector(toggleShowPresence(switchView:)),
+                                                              isOn: isShowingPresence)
+
+        let backgroundSettingView = createLabelAndSwitchRow(labelText: "Use alternate background color",
+                                                            switchAction: #selector(toggleAlternateBackground(switchView:)),
+                                                            isOn: isUsingAlternateBackgroundColor)
+
+        addRow(items: [backgroundSettingView])
+        addRow(items: [showPresenceSettingView])
+
+        createSection(withTitle: "Circle style for person",
+                      name: "Kat Larrson",
+                      image: UIImage(named: "avatar_kat_larsson")!,
+                      style: .noring)
+
+        createSection(withTitle: "Square style for group",
+                      name: "NorthWind Traders",
+                      image: UIImage(named: "site")!,
+                      style: .group)
+
+        createSection(withTitle: "Circle style with border",
+                      name: "Kat Larrson",
+                      image: UIImage(named: "avatar_kat_larsson")!,
+                      style: .default,
+                      borderStyle: .defaultBorder)
+
+        addTitle(text: "Fallback Images")
+        for size in AvatarVnextSize.allCases.reversed() {
+            let phoneNumber = "+1 (425) 123 4567"
+            let fallbackImageAvatar = createAvatarView(size: size,
+                                                       name: phoneNumber,
+                                                       style: .noring)
+            addRow(text: size.description, items: [fallbackImageAvatar.1.view], textStyle: .footnote, textWidth: 100)
+        }
+
+        addTitle(text: "Unauthenticated")
+        for size in AvatarVnextSize.allCases.reversed() {
+            let unauthenticatedAvatar = createAvatarView(size: size,
+                                                         name: nil,
+                                                         style: .unauthenticated)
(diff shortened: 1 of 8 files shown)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -137,6 +137,46 @@ Opacity:
...
+  size: [ xSmall: 16pt, small: 24pt, medium: 32pt, large: 40pt, xlarge: 52pt, xxlarge: 72pt ]
+  textFont: [ xSmall: "Font(9px, regular)", small: "Font(12px, regular)", medium: "Font(13px, regular)", large: "Font(15px, regular)", xlarge: "Font(20px, regular)", xxlarge: "Font(28px, medium)" ]
+
+NoRingAvatarTokens extends AvatarTokens:
+  ringInnerGap: [ xSmall: $Border.size.none, small: $Border.size.none, medium: $Border.size.none, large: $Border.size.none, xlarge: $Border.size.none, xxlarge: $Border.size.none ]
+  ringThickness: [ xSmall: $Border.size.none, small: $Border.size.none, medium: $Border.size.none, large: $Border.size.none, xlarge: $Border.size.none, xxlarge: $Border.size.none ]
+  ringOuterGap: [ xSmall: $Border.size.none, small: $Border.size.none, medium: $Border.size.none, large: $Border.size.none, xlarge: $Border.size.none, xxlarge: $Border.size.none ]
+  ringDefaultColor: $Colors.Neutral.clear
+
+UnauthenticatedAvatarTokens extends AvatarTokens:
```
harrieshin, line 161: can we not say unauthenticated?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -137,6 +137,46 @@ Opacity:
   opacity96: 0.96f
   none: 1f
 
+AP_AvatarTokens:
+  backgroundDefaultColor: "#EDACB1"
```
harrieshin, line 141: this can't be right..... default color should be app accent color?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+        let presenceFrameHypotenuse: CGFloat = ringInnerGapHypotenuse + ringInnerGapRadius + presenceIconHypotenuse
+        let presenceIconFrameSide: CGFloat = sqrt(pow(presenceFrameHypotenuse, 2) / 2)
+
+        let ringColor = state.ringColor ?? tokens.ringDefaultColor!
+        let backgroundColor = state.backgroundColor ?? tokens.backgroundDefaultColor!
+        let foregroundColor = state.foregroundColor ?? tokens.foregroundDefaultColor!
+
+        let presence = state.presence
+        let style = tokens.style
+        let initialsString: String = ((style == .overflow) ? state.primaryText ?? "" : InitialsView.initialsText(fromPrimaryText: state.primaryText,
```
harrieshin, line 259: shouldn't the initial's view 21 different background color be tokenized?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+    case unknown
+    case blocked
+
+    public func color() -> UIColor? {
+        var color: UIColor?
+
+        switch self {
+        case .none:
+            break
+        case .available:
```
harrieshin, line 48: i guess this color part could be tokenize?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+    @objc @Published public var backgroundColor: UIColor?
+    @objc @Published public var foregroundColor: UIColor?
+    @objc @Published public var presence: AvatarVnextPresence = .none
+}
+
+/// Representation of design tokens to buttons at runtime which interfaces with the Design Token System auto-generated code.
+/// Updating these properties causes the SwiftUI button to update its view automatically.
+public class AvatarTokens: ObservableObject {
+    @Published public var avatarSize: CGFloat!
+    @Published public var borderRadius: CGFloat!
```
harrieshin, line 118: borderradius seems weird naming?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+            color = Colors.Palette.presenceUnknown.color
+        case .blocked:
+            color = Colors.Palette.presenceBlocked.color
+        }
+
+        return color
+    }
+
+    public func image() -> UIImage? {
+        var imageName: String?
```
sophialee0416, line 70: Is it possible to use SwiftUI's Image instead of UIImage?

Saadnajmi (changes requested):

ios/FluentUI/People Picker/AvatarViewVnext.swift
chiuam: I think that's the naming design uses in the spec?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+public enum AvatarVnextPresence: Int, CaseIterable {
+    case none
+    case available
+    case away
+    case busy
+    case doNotDisturb
+    case outOfOffice
+    case offline
+    case unknown
+    case blocked
```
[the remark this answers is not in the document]
rdeassis: What's the reason or scenario for making it a separate boolean instead of just another presence enum?
What would it change in the way the Avatar looks or behaves?

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,340 @@
...
+
+@objc(AvatarVnextStyle)
+/// Pre-defined styles of the avatar
+public enum AvatarVnextStyle: Int, CaseIterable {
+    case `default`
+    case noring
+    case unauthenticated
+    case overflow
+    case group
+}
```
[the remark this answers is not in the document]
rdeassis: Yes. Those names come from the variations defined in the figma file.
And the group is indeed the square avatar.

ios/FluentUI/People Picker/AvatarViewVnext.swift
rdeassis: Correct, @chiuam! This is the naming being used on the figma files the design team for the Vnext specs.

rdeassis: I've added a comment to follow up on that and ensure the naming convention being used is intentional and consistent across all platforms.

tools/sgen/input/FluentUIStyle.yml
rdeassis: Following up with Victor there. The figma file indicates that we'll have new colors for standard fallback image and background colors.

tools/sgen/input/FluentUIStyle.yml
rdeassis: This is the variation naming convention defined by the design team. I'll follow up with them on changing this convention.

rdeassis: @harrieshin and @Saadnajmi. Any suggestions on how this should be named?

ios/FluentUI/People Picker/AvatarViewVnext.swift
rdeassis: Seems like we're going to have 30 default colors now. I'm following up with Victor to get this updated here.

ios/FluentUI/People Picker/AvatarViewVnext.swift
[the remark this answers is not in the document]
rdeassis: Just started a follow up conversation with the design team on whether we should support "Out Of Office" variations for each of the presence options.

tools/sgen/input/FluentUIStyle.yml
kubalani: How about `Anonymous` ?

harrieshin: it isn't really inactive or anonymous either. it can be used with phone number only accounts, signed out state etc. I think fallback seems like a better naming?

rdeassis: I'll pass the suggestions to the design team!

ios/FluentUI/People Picker/AvatarViewVnext.swift
```diff
@@ -0,0 +1,336 @@
...
+        case .unauthenticated:
+            appearanceProxy = StylesheetManager.S.UnauthenticatedAvatarTokens
+        case .overflow:
+            appearanceProxy = StylesheetManager.S.OverflowAvatarTokens
+        case .group:
+            appearanceProxy = StylesheetManager.S.GroupAvatarTokens
+        }
+
+        ringDefaultColor = appearanceProxy.ringDefaultColor
+        ringGapColor = appearanceProxy.ringGapColor
```
sophialee0416, line 159: I noticed that the ring gap color does not follow the background color in the alternate background toggle, was this intentional?

tools/sgen/input/FluentUIStyle.yml
rdeassis: @harrieshin I think "signed out" and phone only accounts should look different in the UI because a user identified only by his phone number is different from the absence of user info. 

I believe this variation is meant to be used for the signed out state (or some semantic similar to that where there is no user being represented).

Thoughts?

ios/FluentUI/People Picker/AvatarViewVnext.swift
rdeassis: On the figma file it's specified that both inner and outer rings should be Neutral Background. So I believe this is intentional. Started a thread with design to confirm anyways.
Thanks!

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
```diff
@@ -0,0 +1,179 @@
...
+        let avatarView = AvatarVnext(style: style,
+                                     size: size)
+        avatarView.state.primaryText = name
+        avatarView.state.image = image
+
+        return (avatarView.view, avatarView)
+    }
+}
+
+extension AvatarVnextSize {
```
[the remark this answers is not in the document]
rdeassis: Technically we could, but the AvatarSize (current one) has some functions related to the current Avatar that we don't need due to the tokens specification.
I believe the 2 enums could/should be "merged" into one if we decide to keep both the current and the vnext swiftUI implementations. 

I'll keep that on the notes. Thank you for bringing this point!

kubalani (approved):

sophialee0416 (approved):

ios/FluentUI/People Picker/AvatarViewVnext.swift
rdeassis: Create #338 to track this issue.

ios/FluentUI/People Picker/AvatarViewVnext.swift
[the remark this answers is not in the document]
rdeassis: Hey @Saadnajmi. I have created #339 to track this issue.

ios/FluentUI/People Picker/AvatarViewVnext.swift
rdeassis: @sophialee0416 I added code to support the "transparent" background of the presence icons and the outer ring gap (the inner gap is supposed to be the ringGapColor).
Please check the screenshots.

tools/sgen/input/FluentUIStyle.yml
rdeassis: Tracking this with #340

tools/sgen/input/FluentUIStyle.yml
harrieshin: @rdeassis office uses the "fallback" cases rather than "unauthenticated" style because the gray doesn't work for certain scenarios. Then, it causes more confusion why UI naming is different from usage. from previous commit of the fluent avatar, the "unauthenticated" naming was purposely not used in the control because it had too much implication on the identity state which should be the factor of UI component.

harrieshin: u can easily imagine that the grey style could be used for "disabled" state for example

rdeassis: Just issued PR #346 to implement the solution we discussed offline:

 - Presence of the ring is not a style. It's a state property that can be set.
 - Defined styles for the Fallback scenarios supported by the current AvatarView implementation.
 - Added the ability to set styles and sizes after the Avatar has been instantiated.
 
Issue #340 is tracking this and should be resolved by the PR. 
@Victa FYI.

PR #341, opened 2020-12-14
Replacing MSFButton with MSFButtonVnext in the Demo app.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change replaces the usage of the MSFButton (subclass of UIButton) with the new MSFButtonVnext (SwiftUI implementation).

As outlined in #330, there were some issues with the MSFButtonVnext view (UIView from its hosting controller) when layout constraints were applied to it.

1. The layout constraints issue has been resolved and the new button view can also be used with layout constraints.

2. The action closure implementation of the button now provides the sender parameter which allows the calling code to access the object instance that triggered the action.

3. The demo app usage of the button was updated to capture a weak reference of self on the closure an avoid reference cycles with the pattern below:

```
action: { [weak self] sender in
                return
            }
            ...
}
```

### Verification

|                                        |                                      |
| Screen Shot 2020-12-14 at 9 17 31 AM | Screen Shot 2020-12-14 at 9 17 54 AM |
| Screen Shot 2020-12-14 at 9 18 27 AM | Screen Shot 2020-12-14 at 9 18 56 AM |
| Screen Shot 2020-12-14 at 9 19 35 AM | Screen Shot 2020-12-14 at 9 20 00 AM |
| Screen Shot 2020-12-14 at 9 20 18 AM | Screen Shot 2020-12-14 at 9 20 37 AM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/341)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoController.swift
@@ -38,12 +38,9 @@ class DemoController: UIViewController {
 
     var allowsContentToScroll: Bool { return true }
 
-    func createButton(title: String, action: Selector) -> MSFButton {
-        let button = MSFButton()
-        button.titleLabel?.textAlignment = .center
-        button.titleLabel?.numberOfLines = 0
-        button.setTitle(title, for: .normal)
-        button.addTarget(self, action: action, for: .touchUpInside)
+    func createButton(title: String, action: (( MSFButtonVnext) -> Void)?) -> MSFButtonVnext {
+        let button = MSFButtonVnext(style: .secondary, size: .small, action: action)
+        button.state.text = title
         return button
     }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
@@ -113,7 +113,10 @@ class AvatarGroupViewDemoController: DemoController {
         backgroundColorSwitch.addTarget(self, action: #selector(toggleAlternateBackground(switchView:)), for: .valueChanged)
 
         addRow(text: "Use alternate background color", items: [backgroundColorSwitch], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
+
+        maxAvatarButton.view.widthAnchor.constraint(equalToConstant: 50).isActive = true
         addRow(text: "Max displayed avatars", items: [UIStackView(arrangedSubviews: [maxAvatarsTextField, maxAvatarButton.view])], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
+        overflowCountButton.view.widthAnchor.constraint(equalToConstant: 50).isActive = true
         addRow(text: "Overflow count", items: [UIStackView(arrangedSubviews: [overflowCountTextField, overflowCountButton.view])], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
 
         insertLabel(text: "Avatar stack without borders")
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
@@ -15,17 +15,44 @@ class AvatarGroupViewDemoController: DemoController {
         reloadAvatarViews()
     }
 
-    private lazy var incrementBadgeButton: MSFButton = {
-        return createButton(title: "+", action: #selector(incrementBadgeNumbers))
+    private lazy var incrementBadgeButton: MSFButtonVnext = {
+        return createButton(title: "+", action: { [weak self] _ in
+            guard let strongSelf = self else {
+                return
+            }
+
+            if strongSelf.avatarCount < samplePersonas.count {
+                strongSelf.avatarCount += 1
+            }
+        })
     }()
 
-    private lazy var decrementBadgeButton: MSFButton = {
-        return createButton(title: "-", action: #selector(decrementBadgeNumbers))
+    private lazy var decrementBadgeButton: MSFButtonVnext = {
+        return createButton(title: "-", action: { [weak self] _ in
+            guard let strongSelf = self else {
+                return
+            }
+
+            if strongSelf.avatarCount > 1 {
+                strongSelf.avatarCount -= 1
+            }
+        })
     }()
 
-    private lazy var maxAvatarButton: MSFButton = {
-        let button = createButton(title: "Set", action: #selector(maxAvatarButtonWasPressed))
-        button.isEnabled = false
+    private lazy var maxAvatarButton: MSFButtonVnext = {
+        let button = createButton(title: "Set", action: { [weak self] _ in
+            guard let strongSelf = self else {
+                return
+            }
+
+            if let text = strongSelf.maxAvatarsTextField.text, let count = UInt(text) {
+                strongSelf.maxDisplayedAvatars = count
+                strongSelf.maxAvatarButton.state.isDisabled = true
+            }
+
+            strongSelf.maxAvatarsTextField.resignFirstResponder()
+        })
+        button.state.isDisabled = true
 
         return button
     }()
@@ -39,9 +66,20 @@ class AvatarGroupViewDemoController: DemoController {
         return textField
     }()
 
-    private lazy var overflowCountButton: MSFButton = {
-        let button = createButton(title: "Set", action: #selector(overflowCountButtonWasPressed))
-        button.isEnabled = false
+    private lazy var overflowCountButton: MSFButtonVnext = {
+        let button = createButton(title: "Set", action: { [weak self] _ in
+            guard let strongSelf = self else {
+                return
+            }
+
+            if let text = strongSelf.overflowCountTextField.text, let count = UInt(text) {
+                strongSelf.overflowCount = count
+                strongSelf.overflowCountButton.state.isDisabled = true
+            }
+
+            strongSelf.overflowCountTextField.resignFirstResponder()
+        })
+        button.state.isDisabled = true
 
         return button
     }()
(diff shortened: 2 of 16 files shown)


ios/FluentUI.Demo/FluentUI.Demo/DemoController.swift
```diff
@@ -38,12 +38,9 @@ class DemoController: UIViewController {
 
     var allowsContentToScroll: Bool { return true }
 
-    func createButton(title: String, action: Selector) -> MSFButton {
-        let button = MSFButton()
-        button.titleLabel?.textAlignment = .center
-        button.titleLabel?.numberOfLines = 0
-        button.setTitle(title, for: .normal)
-        button.addTarget(self, action: action, for: .touchUpInside)
+    func createButton(title: String, action: (( MSFButtonVnext) -> Void)?) -> MSFButtonVnext {
```
kubalani, line 41: Do you think it would be better to use an extension? This base class `DemoController` fixes problem in one place and creates in another

rdeassis: To be honest I was about to just delete this func and create the buttons directly since it only provides the convenience of the default style/size and setting the text.

Usually we prefer standalone funcs (not instance funcs like this) instead of extensions.
But the main reason I opted to just replace the implementation of this one is that we have other convenience "create" funcs here.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
```diff
@@ -66,15 +104,20 @@ class AvatarGroupViewDemoController: DemoController {
         settingsTitle.text = "Settings"
         container.addArrangedSubview(settingsTitle)
 
-        addRow(text: "Avatar count", items: [incrementBadgeButton, decrementBadgeButton], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
+        let avatarCountButtonsStackView = UIStackView(arrangedSubviews: [incrementBadgeButton.view, decrementBadgeButton.view])
```
kubalani, line 107: Any specific reason for not using container of `DemoController`?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
```diff
@@ -66,15 +104,20 @@ class AvatarGroupViewDemoController: DemoController {
...
 
         let backgroundColorSwitch = UISwitch(frame: .zero)
         backgroundColorSwitch.isOn = isUsingAlternateBackgroundColor
         backgroundColorSwitch.addTarget(self, action: #selector(toggleAlternateBackground(switchView:)), for: .valueChanged)
 
         addRow(text: "Use alternate background color", items: [backgroundColorSwitch], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
-        addRow(text: "Max displayed avatars", items: [maxAvatarsTextField, maxAvatarButton], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
-        addRow(text: "Overflow count", items: [overflowCountTextField, overflowCountButton], textStyle: .footnote, textWidth: Constants.settingsTextWidth)
+
+        maxAvatarButton.view.widthAnchor.constraint(equalToConstant: 50).isActive = true
```
kubalani, line 117: consolidate constraints calls in `viewDidLayoutSubviews` ?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
```diff
@@ -36,16 +36,81 @@ class DateTimePickerDemoController: DemoController {
...
         container.addArrangedSubview(dateLabel)
-        container.addArrangedSubview(createButton(title: "Show date picker", action: #selector(presentDatePicker)))
-        container.addArrangedSubview(createButton(title: "Show date time picker", action: #selector(presentDateTimePicker)))
-        container.addArrangedSubview(createButton(title: "Show date range picker (paged)", action: #selector(presentDateRangePicker)))
-        container.addArrangedSubview(createButton(title: "Show date range picker (tabbed)", action: #selector(presentTabbedDateRangePicker)))
-        container.addArrangedSubview(createButton(title: "Show date time range picker", action: #selector(presentDateTimeRangePicker)))
-        container.addArrangedSubview(createButton(title: "Show picker with custom subtitles or tabs", action: #selector(presentCustomSubtitlePicker)))
+        container.addArrangedSubview(createButton(title: "Show date picker", action: { [weak self] _ in
+            guard let strongSelf = self else {
+                return
```
kubalani, line 41: nit: extra new line

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
```diff
@@ -15,17 +15,44 @@ class AvatarGroupViewDemoController: DemoController {
         reloadAvatarViews()
     }
 
-    private lazy var incrementBadgeButton: MSFButton = {
-        return createButton(title: "+", action: #selector(incrementBadgeNumbers))
+    private lazy var incrementBadgeButton: MSFButtonVnext = {
+        return createButton(title: "+", action: { [weak self] _ in
```
kubalani, line 19: Can we follow a pattern of 

```swift
let button = createButton(title: "Set", action: { [weak self] _ in
                return
            }
           strongSelf. incrementBadgeNumbers() // replace with corresponding selectors
  })
```
This help us preserve GitHub history along with code preserving code modules. 
I see this is done is some cases and not the others. Any specific reason?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
kubalani: I am still in favor of `UIStackview` extension doesn't have to be part of this PR :)

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
```diff
@@ -235,12 +248,12 @@ extension AvatarGroupViewDemoController: UITextFieldDelegate {
         let button = textField == maxAvatarsTextField ? maxAvatarButton : overflowCountButton
         if let count = UInt(text) {
             if textField == maxAvatarsTextField {
-                button.isEnabled = count > 0 && count != maxDisplayedAvatars
+                button.state.isDisabled = count <= 0 || count == maxDisplayedAvatars
```
kubalani, line 251: nit: Can we add wrappers in `UIHostingVontroller` it becomes hard to read the negative connotation while trying to deduce the meaning of the condition?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonVnextDemoController.swift
```diff
@@ -16,27 +16,36 @@ class ButtonVnextDemoController: DemoController {
...
+                    strongSelf.didPressButton()
+                })
                 disabledButton.state.text = "Button"
                 disabledButton.state.isDisabled = true
 
                 addRow(items: [button.view, disabledButton.view], itemSpacing: 20)
 
                 if let image = style.image {
-                    let iconButton = MSFButtonVnext(style: style, size: size) {
+                    let iconButton = MSFButtonVnext(style: style, size: size) {_ in
```
kubalani, line 42: A optional convenience callback without sender would be good addition to reduce the unnecessary parameters

sophialee0416 (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
rdeassis: I added the UIStackView to have control over the size of the buttons and the space between them.
Just passing them on the addRow was changing the buttons size in a way that didn't look good.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
rdeassis: Wouldn't it potentially cause a new layout of the views?
There are other constraints being activated in other funcs like insertAvatarViews, so I believe if anything can be improved, it should be a separate change because not all constraints here are related to the button.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
rdeassis: You mean a wrapper function like "button.isDisabled = true" that calls into the state?

If so, we have explicitly decided to expose the state object as is to make it easier (less code written) when we add or update state properties. (Anand made this comment in the original button prototype PR).

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
rdeassis: You mean add a new line?
The convention we use is not to have extra lines on the else of a guard statement and no extra lines before the first statement of a closure or after the last statement of it.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
rdeassis: So, these methods were not being used anywhere else and they were selectors just because the old button requires it.
I think it's worth cleaning up if there's no good reason to have them for the sake of simpler code and readability.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonVnextDemoController.swift
rdeassis: Closures (or blocks in objC) have been preferred over selectors. I believe we should not support selectors.
Let's talk about the callback with no parameters offline.

kubalani (approved):
Discussed with Rafael offline, making transient changes in hosting controller in favor of backward compatibility is not recommended as it makes sense to make changes in client interface to support switfui functionality.

PR #344, opened 2020-12-16
Adding Drawer controller 

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Adding Drawer Controller SwitUI component. Added design tokens for shadow modification.
Currently the drawer only supports slide out cases for left and right presentation style.  

Edit: 

Updated presentations style to override modal screen size. 
Screen Shot 2020-12-21 at 6 33 11 PM

#### Left Presentation Style
Screen Shot 2020-12-15 at 7 22 34 PM

#### Right Presentation Style
Screen Shot 2020-12-15 at 7 22 47 PM

#### Presentation with Dimmed background 
Screen Shot 2020-12-15 at 7 22 59 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/344)


harrieshin:
why is the drawer not shown on top of the nav bar?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,181 @@
...
+
+        let presentationController = DrawerPresentingController()
+        presentationController.delegate = self
+
+        let controller = VXTDrawerController(contentView: DrawerContentController().containerForActionViews(),
+                                             presentationController: presentationController)
+        container.addArrangedSubview(controller.view)
+        self.drawerController = controller
+
+        controller.view.translatesAutoresizingMaskIntoConstraints = false
```
harrieshin, line 148: isn't it weird that client app need to do this?

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+
+public class DrawerState: NSObject, ObservableObject {
+    @objc @Published public var presentationDirection: DrawerDirection = .left
+    @objc @Published public var isExpanded: Bool = false
+    @objc @Published public var backgroundDimmed: Bool = false
+}
+
+public class DrawerTokens: ObservableObject {
+
+    @Published public var shadowColor: Color! = Color.black
```
harrieshin, line 19: avoid using "!" in property declarations

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+@objc(DrawerState)
+
+public class DrawerState: NSObject, ObservableObject {
+    @objc @Published public var presentationDirection: DrawerDirection = .left
```
harrieshin, line 12: all of these properties should be properly documented

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+                    self.draggedOffsetWidth = nil
+                }
+            }
+    }
+}
+
+@objc(VXTDrawerController)
+
+/// UIKit wrapper that exposes the SwiftUI Button implementation
+open class VXTDrawerController: NSObject {
```
rdeassis, line 150: How about naming this like this?

@objc(**MSFDrawerVnext**)
open class **DrawerVnext**: NSObject

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+            .onChanged { value in
+
+                let withinDragBounds = state.presentationDirection == .left ? value.translation.width < 0 : value.translation.width > 0
+                if withinDragBounds {
+                    draggedOffsetWidth = value.translation.width
+                }
+            }
+            .onEnded { _ in
+                if let draggedOffsetWidth = draggedOffsetWidth {
+                    if abs(draggedOffsetWidth) < 270 / 4 {
```
rdeassis, line 136: Where is 270 and 4 coming from?
Shouldn't this be relative to the width?

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+                            x: tokens.shadowDepth[0],
+                            y: tokens.shadowDepth[1])
+                    .offset(x: computedOffset(screenSize: reader.size))
+                    .animation(backgroundLayerAnimation)
+
+                if state.presentationDirection == .left {
+                    Spacer()
+                }
+
+            }.onTapGesture {
```
rdeassis, line 118: nit: put the modifiers on the next line?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
rdeassis: I think the the init() of the wrapper class is the best place for it.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+    @objc @Published public var isExpanded: Bool = false
+    @objc @Published public var backgroundDimmed: Bool = false
+}
+
+public class DrawerTokens: ObservableObject {
+
+    @Published public var shadowColor: Color! = Color.black
+    @Published public var shadowOpacity: Double! = 0.24
+    @Published public var shadowBlur: CGFloat! = 28
+    @Published public var shadowDepth: [CGFloat]! =  [0, 14.0] // x,y axis
```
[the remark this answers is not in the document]
kubalani: These defaults are coming from the figma file. I will update them to optionals

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
kubalani: This is mostly required to override the default behavior of `UIStackView`. 
Ideally I would expect the client to directly present or transition to `VXTDrawerController` without the need of `DemoController` which is how I think Teams should consume it.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -147,6 +147,13 @@ Opacity:
   opacity96: 0.96f
   none: 1f
 
+Shadow:
+    Blur:
+    - light12: 28pt
+
+    Depth:
+    - shadow28: [ 0pt, 14pt ]
```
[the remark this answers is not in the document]
kubalani: I did this initially seems like Sgen doesn't like nesting beyond 3 levels, it crashed on me.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
kubalani: I updated this to use `MSFDrawerVnext` extension instead of using view controller API's

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: Thanks for catching this, I have updated this to use geometry reader instead

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,224 @@
...
+    @objc open var drawerState: DrawerState {
+        return self.hostingController.rootView.drawerState
+    }
+
+    @objc public init(contentView: UIView, presentationController: UIViewController) {
+        self.hostingController = UIHostingController(rootView: DrawerShimView(contentView: contentView, controller: presentationController))
+        super.init()
+    }
+
+    struct DrawerContentView: UIViewRepresentable {
```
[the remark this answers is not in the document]
kubalani: I am not sure if I entirely understand your proposal, I agree with a convenience UIView API but that would be in the wrapper class as well

kubalani:
> why is the drawer not shown on top of the nav bar?

The API of DemoController is now matched to show full screen.

ios/FluentUI/DrawerVnext/Drawer.swift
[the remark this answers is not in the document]
kubalani: I have updated the API to remove presentation controller. The ShimView is still required to bridge the gap with modal presentation style. I also tried content view with `UIViewRepresentable` but it seems like in that case the controller doesn't receive touch events. 

I have also updated the API to match interface with FluentUI `DrawerController`. I think this is the minimal hosting code to feature complete teams requirement. We can discuss offline if there is a need for additional API

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
rdeassis: You can do it like this:

Shadow:
  shadow28: [ blur: 28pt, color: $Colors.Neutral.black, opacity: $Opacity.opacity24, x: 0pt, y: 14pt ]

AP_DrawerTokens:
  shadowColor: $Shadow.shadow28.color
  shadowOpacity: $Shadow.shadow28.opacity
  shadowBlur:   $Shadow.shadow28.blur
  shadowX: $Shadow.shadow28.x
  shadowY: $Shadow.shadow28.y

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+        dismiss(animated: true)
+    }
+}
+
+extension DrawerVnext: UIViewControllerTransitioningDelegate, UIViewControllerAnimatedTransitioning {
+    public func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
+        return 0
+    }
+
+    public func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
```
rdeassis, line 230: I found a bug where if you drag from the right it will display the drawer from the left side anyways:

https://user-images.githubusercontent.com/68076145/102934669-71fb3a00-4459-11eb-8691-1c2660cddf17.mov

rdeassis: The other bug also shown in this video is the one you mentioned before about displaying it from the left and then from the right.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+    }
+
+    public var body: some View {
+        GeometryReader { reader in
+            HStack {
+                if state.presentationDirection == .right {
+                    Spacer()
+                }
+                content
+                    .frame(width: reader.size.width * percentWidthOfContent)
```
rdeassis, line 116: The current drawer does not change its width in horizontal orientation.
This seems to be the right behavior for the left nav at least.

drawer_horizontal
drawer_vnext_horizontal

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+                            x: tokens.shadowDepth[0],
+                            y: tokens.shadowDepth[1])
+                    .offset(x: computedOffset(screenSize: reader.size))
+                    .animation(backgroundLayerAnimation)
+
+                if state.presentationDirection == .left {
+                    Spacer()
+                }
+            }
+            .background(state.isExpanded ? backgroundLayerColor.opacity(backgroundLayerOpacity) : Color.clear)
```
rdeassis, line 128: You can create a GitHub issue to address this after your first merge:
If you tap out of the drawer (dimmed background or just out of its contents) you should be able to dismiss it:


https://user-images.githubusercontent.com/68076145/102935346-d4086f00-445a-11eb-8a45-e641139c81d9.mov

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+            .gesture(dragGesture(snapWidth: reader.size.width * percentSnapWidthOfScreen))
+        }
+    }
+
+    private func dragGesture(snapWidth: CGFloat) -> some Gesture {
```
rdeassis, line 133: Another issue you can file as a GitHub issue to be addressed after your first merge:

The current drawer presents the contents as you drag from the side instead of detecting the drag and presenting it entirely:


https://user-images.githubusercontent.com/68076145/102935613-585af200-445b-11eb-8a9c-99b48f8a66a7.mov

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+
+        let isPresentingDrawer = toView == view
+
+        let drawerView = isPresentingDrawer ? toView : fromView
+
+        // delegate animation to swiftui by changing state
+        if isPresentingDrawer {
+            transitionContext.containerView.addSubview(drawerView)
+            drawerView.frame = UIScreen.main.bounds
+            // Added a static interval as currently swiftui doesn't have an animation completion callback
```
rdeassis, line 243: Do you think that the withAnimation function could be useful to avoid this?

https://developer.apple.com/documentation/swiftui/withanimation(_:_:)

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,266 @@
...
+            } else {
+                delegate?.drawerDidCollapsed?()
+            }
+        }
+    }
+    @objc @Published public var backgroundDimmed: Bool = false
+}
+
+@objc(DrawerStateDelegate)
+
```
rdeassis, line 27: nit: remove the extra line (27) here and in other types.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,145 @@
...
+
+    public func actionViews() -> [UIView] {
+        let spacer = UIView()
+        spacer.backgroundColor = .orange
+        spacer.layer.borderWidth = 1
+        spacer.heightAnchor.constraint(greaterThanOrEqualToConstant: 20).isActive = true
+
+        var views = [UIView]()
+        views.append(createButton(title: "Dismiss", action: #selector(dismissButtonTapped)))
+        views.append(createButton(title: "Dismiss (no animation)", action: #selector(dismissNotAnimatedButtonTapped)))
```
rdeassis, line 22: FYI: When you pull the latest changes to your branch, you'll need to change the usage of the button to use the ButtonVnext closure instead of the selector.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,145 @@
...
+            present(drawerController, animated: true, completion: nil)
+        }
+    }
+
+    @objc private func handleScreenEdgePan(gesture: UIScreenEdgePanGestureRecognizer) {
+        guard gesture.state == .began else {
+            return
+        }
+
+        let leadingEdge: UIRectEdge = view.effectiveUserInterfaceLayoutDirection == .leftToRight ? .left : .right
```
rdeassis, line 138: hmmmm... I think the client app shouldn't need to worry about this.
Isn't there a way to just attach the drawerController to a specific view/viewcontroller and let it read the gestures and take the appropriate actions?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,145 @@
...
+            drawerController.drawerState.presentationDirection = .right
+            present(drawerController, animated: true, completion: nil)
+        }
+    }
+
+    @objc private func showRightDrawerDimmedBackgroundButtonTapped() {
+        if let drawerController = drawerController {
+            drawerController.drawerState.backgroundDimmed = true
+            drawerController.drawerState.presentationDirection = .right
+            present(drawerController, animated: true, completion: nil)
```
rdeassis, line 129: How would a SwiftUI application use/call the drawer in order to be displayed?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -211,3 +218,8 @@ SecondaryButtonTokens extends MSFButtonTokens:
 GhostButtonTokens extends MSFButtonTokens:
   borderColor: [ rest: $Colors.Neutral.clear, hover: $Colors.Neutral.clear, pressed: $Colors.Neutral.clear, selected: $Colors.Brand.shade10, disabled: $Colors.Neutral.clear ]
 
+AP_DrawerTokens:
```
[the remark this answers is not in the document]
kubalani: The drawer inherits the background color from content view controller. The default color is `.clear`. Overriding it won't have any effect when content color is set.

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: Thanks for catching this, I have updated to match the default functionality. [I've added a small hack to override the default behavior] 
I believe the approach should be to either use `preferredContentSize` from content view controller or stretch to full screen because if any app needs the functionality other way around then currently there is no way to do so. 

We can revisit this when adding vertical support for drawer controller.

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
rdeassis: Makes sense!

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
kubalani: I have added SwiftUI drawer preview code for an application using SwiftUI

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
kubalani: I have created a new issue to track this 

https://github.com/microsoft/fluentui-apple/issues/356

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: I have created a new issue to track this. 
https://github.com/microsoft/fluentui-apple/issues/357

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: The left to right pan gesture is correct, as far the animation when switching side is concerned would be addressed as part of this [issue](https://github.com/microsoft/fluentui-apple/issues/357)

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
kubalani: I have reached out to victor to double check on this. I have created a new issue to track this item. 
https://github.com/microsoft/fluentui-apple/issues/358

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,331 @@
...
+            NavigationView {
+                EmptyView()
+                    .navigationBarTitle(Text("Drawer Background"))
+                    .navigationBarItems(leading: Button(action: {
+                        drawer.state.isExpanded.toggle()
+                    }, label: {
+                        Image(systemName: "sidebar.left")
+                    })).background(Color.blue)
+            }
+            drawer
```
rdeassis, line 204: drawer.edgesIgnoringSafeArea(.all)

(so we can display the drawer in the whole screen?)

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
rdeassis: I have added a comment referencing the code of the current DrawerController to clarify that this logic should live in the control instead of the client app.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,331 @@
...
+
+// Mark - Drawer View
+
+@objc public enum DrawerDirection: Int, CaseIterable {
+    case left
+    case right
+}
+
+/**
+ `Drawer` is used to present a content that is currently in hideout mode when collapsed and when expanded is used to present content view partially either on the left or right.
```
rdeassis, line 66: nit: Xcode style comment

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: Thanks for fixing this!

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: By nesting animation modifiers I believe you can eventually disable the animation of repositioning the drawer while animating just its presentation and avoiding the bug on the video below:


https://user-images.githubusercontent.com/68076145/103310566-c314ac00-49cc-11eb-9a00-b1c2af5c6643.mov

rdeassis (commented):
I think the issue on the video below is the main blocker.
Figuring out the animation comment might solve that comment along with this bug:


https://user-images.githubusercontent.com/68076145/103310694-10911900-49cd-11eb-891d-dc8d79172f17.mov

ios/FluentUI/DrawerVnext/Drawer.swift
sophialee0416: ^this means using `///` instead of the other commenting methods for clarification

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,331 @@
...
+    private let backgroundLayerColor: Color = .black
+    private var backgroundLayerOpacity: Double {
+        return state.backgroundDimmed ? 0.5 : 0
+    }
+    private let backgroundLayerAnimation: Animation = .default
+
+    private let percentWidthOfContent: CGFloat = 0.9
+    private let percentSnapWidthOfScreen: CGFloat = 0.225
+
+    // content view on top of
```
sophialee0416, line 86: Is this comment incomplete?

kubalani: Thanks for catching this, I have updated it

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,331 @@
...
+                    }, label: {
+                        Image(systemName: "sidebar.left")
+                    })).background(Color.blue)
+            }
+            drawer
+        }
+    }
+}
+
+#if DEBUG
```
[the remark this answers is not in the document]
kubalani: For the meantime, I have added a preview with example for SwiftUI in a separate file

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,331 @@
...
+    }
+
+    @objc required dynamic public init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+    }
+}
+
+extension DrawerVnext: DrawerStateDelegate {
+    // two way binding if state is changed
+    func drawerDidCollapsed() {
```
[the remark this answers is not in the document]
kubalani: this method no longer exists

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: This should be now working as expected

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: This is already part of `DrawerShimView`.

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
```diff
@@ -0,0 +1,31 @@
...
+                        state.isExpanded = true
+                    } else {
+                        state.isExpanded = false
+                    }
+                }
+            }
+    }
+
+    private func gestureSnapWidthRatio() -> CGFloat {
+        return 0.225
```
sophialee0416, line 29: Should this be tokenized?

ios/FluentUI/DrawerVnext/Drawer+Modifiers.swift
```diff
@@ -0,0 +1,21 @@
...
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+public extension Drawer {
+    /// Custom modifier for adding a callback placeholder when drawer's state is changed
+    /// - Parameter `didChangedState`: closure executed with drawer is expanded or collapsed
+    /// - Returns: `Drawer`
+    func didChangedState(_ didChangedState: @escaping DrawerStateChangedCompletionBlock) -> Drawer {
```
sophialee0416, line 12: nit: didChangeState

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+        let isPortraitMode = self.size.width < self.size.height
+        if isPortraitMode {
+            return CGSize(width: self.size.width, height: self.size.height)
+        } else {
+            return CGSize(width: self.size.height, height: self.size.width)
+        }
+    }
+}
+
+// MARK: - Previews
```
sophialee0416, line 133: Are previews something we want to check in? I think they're a good idea if someone wants to see the live previews within Xcode, but they could also test the demo app.. Thoughts?

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -0,0 +1,174 @@
...
+    public var body: some View {
+        ZStack {
+            Color.black.opacity(0.001)
+        }
+    }
+}
+
+// MARK: - Preview
+
+struct SlideOverPanelLeft_Previews: PreviewProvider {
```
sophialee0416, line 110: Re: preview comment from before

ios/FluentUI/DrawerVnext/Drawer+Modifiers.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+public extension Drawer {
```
rdeassis, line 8: Same thing here regarding the need for a separate file and an extension: I think this should go to the Drawer file as a func and the extension should be deleted unless we have more modifiers to justify it.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+public class DrawerState: NSObject, ObservableObject {
+
+    /// A callback executed when the drawer is expanded/collapsed
+    public var onStateChange: (DrawerStateChangedCompletionBlock)?
+
+    /// Set `isExpanded` to `true` to maximize the drawer's width to fill the device screen horizontally minus the safe areas.
+    /// Set to `false` to restore it to the normal size.
+    @objc @Published public var isExpanded: Bool = false {
+        didSet {
+            onStateChange?(isExpanded)
```
rdeassis, line 28: Do we need to pass isExpanded as a parameter? The state object property could just be queried instead.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+        }
+    }
+    @objc @Published public var presentationDirection: DrawerDirection = .left
+
+    /// Set `backgroundDimmed` to `true` to dim the spacer area between drawer and base view.
+    /// If set to `false` it restores to `clear` color
+    @objc @Published public var backgroundDimmed: Bool = false
+}
+
+public typealias DrawerStateChangedCompletionBlock = (_ isExpanded: Bool) -> Void
```
rdeassis, line 38: Calling this a completion block seems very "objC"ish...
I vote for just deleting this typealias and declaring line 22 as follows:

` public var onStateChange: (() -> Void)?`

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+    public var content: Content
+
+    // configure the behavior of drawer
+    @ObservedObject public var state = DrawerState()
+
+    // configure the apperance of drawer
+    @ObservedObject public var tokens = DrawerTokens()
+
+    // keep track of dragged offset
+    @State internal var draggedOffsetWidth: CGFloat?
```
rdeassis, line 83: rename to dragOffset?

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+
+    // internal drawer state
+    @State internal var isDrawerOpen: Bool = false
```
rdeassis, line 86: How about renaming to **isContentPresented**?

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+                        self.draggedOffsetWidth = nil // drag ends
+                    }
+                })
+                .gesture(dragGesture(screenWidth: proxy.portraitOrientationAgnosticSize().width))
+        }
+        .edgesIgnoringSafeArea(.all)
+    }
+
+    private var backgroundLayerOpacity: Double {
+        return state.backgroundDimmed ? 0.5 : 0
```
rdeassis, line 112: This opacity value should be tokenized. Can you start a conversation with Victor about it? (and open a GitHub issue for it)

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+
+    private var slideOutDirection: SlideOverDirection {
+        return state.presentationDirection == .left ? .left : .right
+    }
+}
+
+// MARK: - Utility + Helpers
+
+extension GeometryProxy {
+    func portraitOrientationAgnosticSize() -> CGSize {
```
rdeassis, line 123: Can you move this into the implementation of the drawer as a private func that takes the GeometryProxy as a parameter? This function does not make sense for the GeometryProxy out of the Drawer context and that's why I don't think it should be an extension.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+        let isPortraitMode = self.size.width < self.size.height
+        if isPortraitMode {
```
rdeassis, line 125: nit: put the condition directly in the if statement?

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis, line 123: How about calling the method **sizeForCurrentOrientation**?

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: I agree we should question ourselves about it and it's why I created #360.

My first gut reaction is that this should be part of the demo app code to prevent increasing the library size for consumers out of the cocoapods world.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+            return CGSize(width: self.size.width, height: self.size.height)
+        } else {
+            return CGSize(width: self.size.height, height: self.size.width)
+        }
+    }
+}
+
+// MARK: - Previews
+
+import SwiftUI
```
rdeassis, line 135: remove duplicate import

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -0,0 +1,174 @@
...
+            if direction == .left {
+                InteractiveSpacer().onTapGesture (count: 1, perform: actionOnBackgroundTap ?? {})
+            }
+        }
+        .background(isOpen ? backgroundLayerColor.opacity(backgroundLayerOpacity) : Color.clear)
+    }
+
+    private func resolvedContentOffset() -> CGFloat {
+        // override offset if required
+        if let preferredContentWidth = preferredContentOffset {  // parent view wants to take over primarily to conform to user drag gesture
```
rdeassis, line 73: nit: comment on the next line

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -0,0 +1,174 @@
...
+        return slideOutPanelWidth * contentWidthSizeRatio
+    }
+}
+
+// MARK: - Composite View
+
+public struct InteractiveSpacer: View {
+    public var body: some View {
+        ZStack {
+            Color.black.opacity(0.001)
```
rdeassis, line 103: I think the same applies here for a tokenized opacity that can be tracked in the same issue I mentioned before.
Could you talk to Victor about it?

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,170 @@
...
+        GeometryReader { proxy in
+            SlideOverPanel(
+                content: content,
+                isOpen: $isDrawerOpen,
+                preferredContentOffset: $draggedOffsetWidth)
+                .backgroundOpactiy(backgroundLayerOpacity)
+                .direction(slideOutDirection)
+                .width(proxy.portraitOrientationAgnosticSize().width)
+                .performOnBackgroundTap {
+                    state.isExpanded.toggle()
```
rdeassis, line 98: Is there a scenario where tapping on the background will expand the drawer?
I believe there's not and this should be just **state.isExpanded = false**

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
kubalani: @sophialee0416 tokens are only applied to appearances and not behavioral attributes

@rdeassis  we can't add const static property in Drawer extension because of it's generic type. 
I am also not a big fan of redundant documentation. 

this is from FluentUI [drawer](https://github.com/microsoft/fluentui-apple/blob/bb4d1ac255b858755f25ea427ad2305a66c90b0f/ios/FluentUI/Drawer/DrawerController.swift#L262) 
```swift
// If the drawer is already loaded, update the resizing handle ...
 if isViewLoaded {
       updateResizingHandleView()
}
```

Redundant documentation creates a lot of problems than in solves. documentation also goes out of sync more often that code does.

I am like to add comments when they are [necessary](https://blog.cleancoder.com/uncle-bob/2017/02/23/NecessaryComments.html). 

I intentionally create small method where it's name eliminates the need for documentation and I believe `gestureSnapWidthRatio` in Drag extension has sufficient context to interpret the meaning of the value. I am open to other suggestions to improve nomenclature or restructure code to eliminate the need of documentation.

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: I would like to keep previews for testing till we have a solution.. We can move them to appropriate place when we address #360

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: I will address them as part of #360

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,174 @@
...
+        dismiss(animated: false)
+    }
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        view = containerForActionViews()
+    }
+}
+
+extension UIStackView {
```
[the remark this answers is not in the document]
kubalani: The scope of extension is internal by default. If this was intended to be used as a multiple times I would have used a broader modifier (`public` or `open`) 

The only reason an extension is used here is for code separation. It makes code readability much easier
```swift
 addBackground(stackview: container, color: Colors.surfacePrimary)
```

vs 

```swift
 container.addBackground(color: Colors.surfacePrimary)
```

I would be interested to know if there are any specific reason we are trying to limit extension or having excessive extensions is bad for the code?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,174 @@
...
+        views.append(spacer)
+        return views
+    }
+
+    public func containerForActionViews() -> UIView {
+        let container = DemoController.createVerticalContainer()
+        for view in actionViews() {
+            container.addArrangedSubview(view)
+        }
+        container.addBackground(color: Colors.surfacePrimary)
```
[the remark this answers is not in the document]
kubalani: This is demo code not `FluentUI`. Since we are sharing components like images, Democontroller with rest of the Demo code the Colors library is part of it. Adding `StyleSheetManager` to demo code would give an indication that it's ok for client to use the internal style or theme sheet for their own content. 

We don't want client to use `StyleSheetManager` directly. At some point we will replace `Color` wrapper to use `Stylesheet` and clients will continue using their current API. At least that is what I think is the expected functionality.

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: Victor is OOF in the meanwhile I have added a transparent color to stylesheet. I will discuss this with him as a part of #358

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
```diff
@@ -0,0 +1,174 @@
...
+
+        let leadingEdge: UIRectEdge = view.effectiveUserInterfaceLayoutDirection == .leftToRight ? .left : .right
+        if leadingEdge == .left {
+            self.showRightDrawerDimmedBackgroundButtonTapped()
+        } else {
+            self.showLeftDrawerClearBackgroundButtonTapped()
+        }
+    }
+
+    @objc func drawerDidChangeState(state: DrawerState, controller: UIViewController) {
```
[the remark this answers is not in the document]
kubalani: This is intended for the client to show how does the delegates work. It would encourage them to use this rather than manipulating state. 

Is there any thing wrong with keeping it?

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
```diff
@@ -0,0 +1,31 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+internal extension Drawer {
```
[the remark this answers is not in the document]
kubalani: This will add a significant code when while addressing #356 . SwiftUI encourage modular structure as opposed monolithic views.

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
```diff
@@ -0,0 +1,31 @@
...
+    func dragGesture(screenWidth: CGFloat) -> some Gesture {
+        DragGesture()
+            .onChanged { value in
+                let withinDragBounds = state.presentationDirection == .left ? value.translation.width < 0 : value.translation.width > 0
+                if withinDragBounds {
+                    draggedOffsetWidth = value.translation.width
```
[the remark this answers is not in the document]
kubalani: dragOffset is a float and can be across horizontal axis or vertical axis. May be `dragOffsetX`?

ios/FluentUI/DrawerVnext/Drawer+Modifiers.swift
kubalani: I have used extension when there is a logical grouping of functionalities, modifiers are going to be extended as we add vertical drawer. The extensions in Drawer are split based on UIKit/SwiftUI native interfaces. As the drawer will grow there would be necessity to clarify the distinction. A client might end up using the other interface if there is not sufficient clear distinction between them. we can discuss this more offline if required

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: how about `dragOffsetX`? I need some context to indicate this is along x-axis only as it's a float and not a size

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: I like this name, much better.

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: that would make it hard for readability. Which one of this you would prefer?

```swift
let isPortraitMode = self.size.width < self.size.height
        if isPortraitMode {
            return CGSize(width: self.size.width, height: self.size.height)
        } else {
            return CGSize(width: self.size.height, height: self.size.width)
        }
```

```swift
        // check if user is in portrait mode
        if self.size.width < self.size.height {
            return CGSize(width: self.size.width, height: self.size.height)
        } else {
            return CGSize(width: self.size.height, height: self.size.width)
        }
```

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: Extension methods are only a (pretty nice) syntactic sugar for a static helper method. I prefer extensions methods because they are much easier to read as compare to private helper methods. We talk about this more offline.

kubalani: `sizeForCurrentOritentation` seems like to expect a parameter i.e current orientation. The function needs to return portrait orientation size regardless of current orientation. How about `portraitOrientationSize`?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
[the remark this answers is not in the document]
rdeassis: The helper function is specific for this scenario and not for all uses of the UIStackView.
You don't want to have this addBackground available for UIStackView instances everywhere if it only makes sense here. It makes the code more complicated and hard to maintain if this gets used somewhere else and you also need to change this here.

You could argue that making this extension private could solve that, but this is being used only once for a very specific scenario... which makes the extension unnecessary.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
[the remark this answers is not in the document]
rdeassis: The client app will use StylesheetManager directly to override Themes in our current design. That is where it can access all the appearance proxies (including colors).

Although it's not a big deal to have these calls here now, it'll be more work when we discontinue the Colors class.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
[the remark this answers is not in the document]
rdeassis: I don't think it's necessarily wrong to keep it, but this is an optional method (which wouldn't help us catch func signature changes, right?).
We usually avoid doing just print() since it's not necessarily visible in the demo app.
You can keep it if you're very passionate about it.

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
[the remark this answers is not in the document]
rdeassis: horizontalDragOffiset?

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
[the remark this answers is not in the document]
rdeassis: You're not making it modular if your extension is specific to just the Drawer, are you?

Like I said before: if this will add significant amounts of code when you implement the gesture presentation, I can see the value for it.

To me this seems not needed for now.

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: how about horizontaldragOffset?

ios/FluentUI/DrawerVnext/Drawer+Modifiers.swift
rdeassis: I can see the value of creating separate files for modifiers related to a certain functionality.
My understanding of your idea in this case would be something like Drawer+Delegate.swift or Drawer+State.swift.
Drawer+Modifiers.swift seems like a file that will contain all modifiers for it irrespective of their functionality group.
We can indeed discuss offline.

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: We can discuss it offline. I believe the same argument applies here: we need to make sure that the extension makes sense out of the Drawer context, or at least make it private so it's not visible everywhere in the library where it doesn't make sense.

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: I guess the second option is what I tried to suggest (putting the condition you're storing in isPortraitMode directly in the if statement).

ios/FluentUI/DrawerVnext/Drawer.swift
rdeassis: portraitOrientationSize sounds like it does not return a value if it's in landscape.
How about **sizeInCurrentOrientation** or **sizeInActualOrientation**?

ios/FluentUI.Demo/FluentUI.Demo/Demos/DrawerVnextDemoController.swift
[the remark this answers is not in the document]
kubalani: I have removed it for now, since empty function might look like a bug. We can add this as an example later while adding a mock teams demo controller

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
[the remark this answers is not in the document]
kubalani: I have converted this to a private func, I will defer the need for gesture extension and add when required.

ios/FluentUI/DrawerVnext/Drawer+Drag.swift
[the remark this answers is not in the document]
kubalani: yup, this is a much better name. I have updated it !!

ios/FluentUI/DrawerVnext/Drawer+Modifiers.swift
kubalani: I will remove the drawer modifier for now as it contains only single API. We can discuss this later as we extend the functionality and more API's in drawer

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: I have removed the extension and converted to a private func, we can revisit this when we add more functionality to drawer

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -0,0 +1,174 @@
...
+
+// MARK: - Composite View
+
+public struct InteractiveSpacer: View {
+    @Binding public var defaultBackgroundColor: Color
+
+    public var body: some View {
+        ZStack {
+            defaultBackgroundColor
+        }
```
rdeassis, line 104: .contentShape(Rectangle())

Adding this modifier here will enable hit testing even if the color is transparent.
This should fix the "tap out of the drawer dismiss bug".

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -0,0 +1,174 @@
...
+    // configure the apperance of drawer
+    @ObservedObject public var tokens = DrawerTokens()
+
+    private let contentWidthSizeRatio: CGFloat = 0.9
+
+    public var body: some View {
+        HStack {
+            if direction == .right {
+                InteractiveSpacer(defaultBackgroundColor: $tokens.backgroundClearColor)
+                    .onTapGesture (count: 1, perform: actionOnBackgroundTap ?? {})
```
rdeassis, line 49: you don't need the count:1 parameter here, do you? (same in line 62)

rdeassis (approved):
Thanks for addressing all the comments!
I left a few more (including a bug on the "tap outside to dismiss"), but the .contentShape(Rectangle()) modifier will fix it.

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -0,0 +1,209 @@
...
+    public var onStateChange: (() -> Void)?
+
+    /// Set `isExpanded` to `true` to maximize the drawer's width to fill the device screen horizontally minus the safe areas.
+    /// Set to `false` to restore it to the normal size.
+    @objc @Published public var isExpanded: Bool = false {
+        didSet {
+            onStateChange?()
+        }
+    }
+    @objc @Published public var presentationDirection: DrawerDirection = .left
```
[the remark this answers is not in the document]
kubalani: Discuss this offline with Rafael, published property is not required for direction and animation duration

ios/FluentUI/DrawerVnext/SideOverPanel+Modifiers.swift
```diff
@@ -0,0 +1,78 @@
...
+                              direction: direction,
+                              isOpen: $isOpen,
+                              preferredContentOffset: $preferredContentOffset,
+                              tokens: tokens)
+    }
+
+    /// Add action or callback to be executed when background view is Tapped
+    /// - Parameter `performOnBackgroundTap`:  defaults to no-op
+    /// - Returns: `SlideOverPanel`
+    func performOnBackgroundTap(_ performOnBackgroundTap: (() -> Void)?) -> SlideOverPanel {
```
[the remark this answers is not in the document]
kubalani: this is now fixed. thanks for catching this. I will add a new issues to add integration tests

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: this worked, updated !!

kubalani:
> Thanks for addressing all the comments!
> I left a few more (including a bug on the "tap outside to dismiss"), but the .contentShape(Rectangle()) modifier will fix it.

Thanks for reviews

PR #345, opened 2020-12-16
Contextual Command Bar

zhuorantan:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Implement Contextual Command Bar

### Verification

| Light Mode | Dark Mode |
| ----------- | ----------- |
| ccb_light | ccb_dark |

ccb

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/345)


ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -0,0 +1,59 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class CommandBarButtonGroupView: UIView {
```
harrieshin, line 8: are there 2 CommandBarButtonGroupView swift files and class?

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/CommandBar/Delete_24_Regular.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images" : [
+    {
+      "filename" : "Delete_24_Regular.pdf",
```
harrieshin, line 4: if possible we try to keep the fluent-system-icon file name as is so we can easily replace them later if there is a new icon for the same purpose

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,306 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+public class CommandBar: UIView {
```
harrieshin, line 8: should be more documentation on this class for other clients

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,306 @@
...
+
+    // MARK: Public Properties
+
+    public weak var delegate: CommandBarDelegate?
+
+    // MARK: Private Properties
+
+    private let barAppearance: CommandBarAppearance
+    private let itemGroups: [CommandBarItemGroup]
+    private let leadingItem: CommandBarItem?
```
harrieshin, line 28: what's the importance of leading item and trailing item, leading button and trailing button to be stored separately?

ios/FluentUI/Command Bar/CommandBarAppearance.swift
```diff
@@ -0,0 +1,61 @@
...
+        self.tintColor = tintColor
+        self.backgroundColor = backgroundColor
+        self.highlightedTintColor = highlightedTintColor
+        self.highlightedBackgroundColor = highlightedBackgroundColor
+    }
+
+    public enum Defaults {
+        public static let tintColor: UIColor = Colors.gray400
+        public static let backgroundColor: UIColor = Colors.surfaceSecondary
+        public static let highlightedTintColor = UIColor(light: Colors.communicationBlue, dark: .black)
```
harrieshin, line 29: this probably needs to be MSFColors.primary because not all apps will be communication blue but should their their branding color

harrieshin: same for other references

ios/FluentUI/Command Bar/CommandBarAppearance.swift
```diff
@@ -0,0 +1,61 @@
...
+        public static let highlightedBackgroundColor = UIColor(light: UIColor(red: 224 / 255, green: 238 / 255, blue: 249 / 255, alpha: 1), dark: Colors.communicationBlue)
```
harrieshin, line 30: what's this color?

harrieshin: I think this should be msfcolors.primarytint30

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -0,0 +1,59 @@
...
+
+    init(buttons: [CommandBarButton]) {
+        self.buttons = buttons
+
+        super.init(frame: .zero)
+        translatesAutoresizingMaskIntoConstraints = false
+
+        clipsToBounds = true
+        layer.cornerRadius = Constants.cornerRadius
+        if #available(iOS 13, *) {
```
harrieshin, line 28: for iPad, we should also add pointer support?

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zhuorantan: Yeah, must be an error during renaming. Deleted one.

ios/FluentUI/Command Bar/CommandBar.swift
zhuorantan: Removed.

ios/FluentUI/Command Bar/CommandBarAppearance.swift
zhuorantan: These colors are default colors. They can be configured via `CommandBarAppearance`. Should we use a fixed `MSFColor` or just replace the defaults?

zhuorantan: > what's this color?

This is the background color for selected buttons.

Screen Shot 2020-12-17 at 13 20 57

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,287 @@
...
+    // |  |--scrollView (fill containerView)
+    // |  |  |--subviews
+    // |  |  |  |--stackView
+    // |  |  |  |  |--buttons (fill scrollView content)
+
+    // MARK: Public Properties
+
+    @objc public weak var delegate: CommandBarDelegate?
+
+    // MARK: Private Properties
```
Saadnajmi, line 25: Nit: Follow the ordering described at aka.ms/swift-guide 
https://microsoft.github.io/swift-guide/FileContentOrdering.html

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,287 @@
...
+
+// MARK: - Private methods
+
+private extension CommandBar {
+    struct Constants {
+        static let fadeViewWidth: CGFloat = 24
+        static let horizontalButtonInset: CGFloat = 4
+        static let buttonSpacing: CGFloat = 16.0
+        static let insets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+    }
```
Saadnajmi, line 181: I know we have a bunch of this in our code, but we're moving away from structs that are just constants in favor of just having constants at the bottom of our files

Saadnajmi:
It would aldo be great to add a screenshot/gif/video of the intended control to the PR, so we have a visual :)

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zhuorantan: I'm not familiar with iPad pointer. Can other contributors add this feature?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+    public override func layoutSubviews() {
+        super.layoutSubviews()
+
+        containerMaskLayer.frame = containerView.bounds
+        updateShadow()
+    }
+
+    // MARK: - Private properties
+
+    private let itemGroups: [CommandBarItemGroup]
```
harrieshin, line 78: what about dynamically adding and removing the items based on your selection?

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    override var isEnabled: Bool {
+        didSet {
+            updateStyle()
+        }
+    }
+
+    init(item: CommandBarItem, isFixed: Bool = false) {
```
harrieshin, line 29: "isFixed" the right property name? from commandbarbutton point of view it doesn't make much sense.  ithink it just translated to whether or not this button isSelectable?

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+            return Colors.Palette.communicationBlueTint30.color
+        }
+
+        return Colors.primaryTint30(for: window)
+    }
+
+    func updateStyle() {
+        tintColor = isSelected ? selectedTintColor : CommandBarButton.normalTintColor
+
+        if isFixed {
```
harrieshin, line 82: if it is a "fixed" button, then there isn't any visual feedback if user long press and hold? no color changes at all?

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
rdeassis: Here's some reference PRs on Pointer Interaction support:

[https://github.com/microsoft/fluentui-apple/pulls?q=is%3Apr+pointer+in%3Atitle+](https://github.com/microsoft/fluentui-apple/pulls?q=is%3Apr+pointer+in%3Atitle+)

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -0,0 +1,192 @@
...
+    }
+}
+
+extension CommandBarDemoController: CommandBarDelegate {
+    func commandBar(_ commandBar: CommandBar, didSelectItem item: CommandBarItem) {
+        guard let item = item as? Item else {
+            fatalError("Invalid item type")
+        }
+
+        print("Did select command \(item.command)")
```
kubalani, line 183: remove print statement?

kubalani:
The voice over accessibility doesn't honor left to right order. for the scroll view. A person using accessibility tools won't be able to discover that tools scroll horizontally or discover additional items. 

Expected behavior: 
Voice over control should scroll to include all items or assert that there are additional items not visible on the screen. 

https://user-images.githubusercontent.com/63682282/103813844-94728300-5015-11eb-9448-9f6187313144.mov

ios/FluentUI/Command Bar/CommandBarDelegate.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFCommandBarDelegate)
```
kubalani, line 8: Why is there a need of explicit file for delegates?

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -0,0 +1,61 @@
...
+            trailingAnchor.constraint(equalTo: stackView.trailingAnchor)
+        ])
+    }
+
+    func applyInsets() {
+        buttons.first?.contentEdgeInsets.left += CommandBarButtonGroupView.leftRightBuffer
+        buttons.last?.contentEdgeInsets.right += CommandBarButtonGroupView.leftRightBuffer
+    }
+
+    static let cornerRadius: CGFloat = 8
```
rdeassis, line 58: make those private too?

rdeassis (changes requested):

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -0,0 +1,61 @@
...
+        let stackView = UIStackView(arrangedSubviews: buttons)
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .horizontal
+        stackView.spacing = CommandBarButtonGroupView.buttonPadding
+
+        return stackView
+    }()
+}
+
+private extension CommandBarButtonGroupView {
```
kubalani, line 42: why use an additional extension?

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -0,0 +1,61 @@
...
+    func configureHierarchy() {
+        addSubview(stackView)
+        NSLayoutConstraint.activate([
```
kubalani, line 45: you can avoid all the complex layout logic by subclassing this from `UIStackview` instead of `UIView`. 
Everything including padding and adding inset is supported by stackview.

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -0,0 +1,70 @@
...
+                isSelected = false
+            }
+        }
+    }
+
+    /// Whether the selection state is persisted. If this is set to `false`, the button would be deselected immediately after selection.
+    public var isPersistSelection: Bool
+
+    @available(iOS 14.0, *)
+    public lazy var menu: UIMenu? = nil // Only lazy property can be used with @available
```
kubalani, line 66: I am having a hard time figuring out the use case for this? I don't see any menu in iOS 14. Does anything specific action triggers this?

ios/FluentUI/Command Bar/CommandBarDelegate.swift
```diff
@@ -0,0 +1,14 @@
...
+public protocol CommandBarDelegate: AnyObject {
+    @objc optional func commandBar(_ commandBar: CommandBar, shouldSelectItem item: CommandBarItem) -> Bool
+    @objc optional func commandBar(_ commandBar: CommandBar, didSelectItem item: CommandBarItem)
+    @objc optional func commandBar(_ commandBar: CommandBar, shouldDeselectItem item: CommandBarItem) -> Bool
```
kubalani, line 12: How do you expect client to use `shouldSelectItem`? There is already an attribute `isSelected`.  why isn't this as composition as opposed to a `delegate`? 

Also if you need to accept value from client it should be used as a `datasource` as opposed to `delegate`. `Delegate` are intended for callback purposes only.

ios/FluentUI/Command Bar/CommandBarDelegate.swift
```diff
@@ -0,0 +1,14 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFCommandBarDelegate)
+public protocol CommandBarDelegate: AnyObject {
```
kubalani, line 9: Consider using this instead

```swift
public protocol CommandBarDelegate: AnyObject {
    @objc optional func commandBar(_ commandBar: CommandBar,  item: CommandBarItem, isSelected: Bool)
}

public protocol CommandBarDatasource: AnyObject {
    @objc optional func commandBarItem(_ commandBar: CommandBar, atIndex: IndexPath) -> CommandBarItem
}
```

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+
+        return Dictionary(uniqueKeysWithValues: allButtons.map { ($0.item, $0) })
+    }()
+
+    // MARK: Views and Layers
+
+    private lazy var containerView: UIView = {
+        let containerView = UIView()
+        containerView.translatesAutoresizingMaskIntoConstraints = false
+        containerView.layer.mask = containerMaskLayer
```
kubalani, line 92: why is there a need for container view? why can't this be applies directly to the view?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+/**
+ `CommandBar` is a horizontal scrollable list of icon buttons divided by groups.
+ Provide `itemGroups` in `init` to set the buttons in the scrollable area. Optional `leadingItem` and `trailingItem` add fixed buttons in leading and trailing positions. Each `CommandBarItem` will be represented as a button.
+ Set the `delegate` property to determine whether a button can be selected and deselected, and listen to selection changes.
+ */
+@objc(MSFCommandBar)
+open class CommandBar: UIView {
+    // Hierarchy:
+    //
+    // leadingButton
```
kubalani, line 17: does the view hierarchy needs to be this complex? The leading and trailing buttons can be a part of a group with single items?

|--scrollview (commandbar)
|    |--stackview (group)
|    |    |--stackview (items)

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+
+            containerLeadingConstraint,
+            containerTrailingConstraint
+        ])
+
+        stackView.layoutIfNeeded()
+
+        if UIView.userInterfaceLayoutDirection(for: semanticContentAttribute) == .rightToLeft {
+            // Flip the scroll view to invert scrolling direction. Flip its content back because it's already in RTL.
+            let flipTransform = CGAffineTransform(scaleX: -1, y: 1)
```
kubalani, line 231: it RTL layout does the scroll direction change and the ordering of the items itself change?

ios/FluentUI/Command Bar/CommandBar.swift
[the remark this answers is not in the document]
harrieshin: I think trailing button is for that "keyboard" button that always stays on the accessory bar and doesn't scroll like others. the author removed the left and right items.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/**
+ `CommandBar` is a horizontal scrollable list of icon buttons divided by groups.
```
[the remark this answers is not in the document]
chripear: `/** */` is the block comment equivalent to `///`.

Presumably the goal is good Quick Help (option-click).  There are some markup elements that should go in the comments to improve the help formatting.  The keyboard shortcut Rafael mentioned will get you started.

Quick Primer: [Swift Documentation - NSHipster](https://nshipster.com/swift-documentation/)
Apple Reference: [Markup Formatting Reference](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/index.html)

ios/FluentUI/Command Bar/CommandBarDelegate.swift
zetasq: Maybe `UITableViewDelegate` is a good example to follow. In that protocol there are several methods like `func tableView(UITableView, shouldBeginMultipleSelectionInteractionAt: IndexPath) -> Bool` that is used to accept values from client.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zetasq: If `CommandBarButtonGroupView` is exposed as `UIStackView` subclass, callers may have the chance of calling `UIStackView` API (e.g. `addArrangedSubview`), which is not expected by the designer of `CommandBarButtonGroupView` (all the `arrangedSubviews` should be managed by `CommandBarButtonGroupView` internally).

Furthermore,  If we later want to make `CommandBarButtonGroupView` inherit from another `UIStackView` substitution, there will be an API breaking change (users may already depend on the inherited `UIStackView` interface). So it seems appropriate to me that using a wrapper view to hide the internal implementation with `UIStackView`.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+    // |--subviews
+    // |  |--scrollView (fill containerView)
+    // |  |  |--subviews
+    // |  |  |  |--stackView
+    // |  |  |  |  |--buttons (fill scrollView content)
+    // trailingButton
+
+    // MARK: - Public methods
+
+    public init(itemGroups: [CommandBarItemGroup], leadingItem: CommandBarItem? = nil, trailingItem: CommandBarItem? = nil) {
```
owenconnolly, line 29: I don't see this public initializer with ObjC usage, looks like something is missing to make it available? (I cloned and built this PR branch locally to test out the UI)

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -0,0 +1,192 @@
...
+            switch self {
+            case .add, .mention, .calendar, .arrowUndo, .arrowRedo, .copy, .delete, .link, .keyboard:
+                return false
+            case .textBold, .textItalic, .textUnderline, .textStrikethrough, .checklist, .bulletList, .numberList:
+                return true
+            }
+        }
+    }
+
+    class Item: CommandBarItem {
```
owenconnolly, line 80: Is it expected that the consumer should have to sub-class, or that is optional (there seems to be some problem with the ObjC headers which are being exposed so it is hard to see the intended usage)

ios/FluentUI/Command Bar/CommandBar.swift
owenconnolly: Ah, so this is missing @objc, seems some other places are also. I will comment where I find these missing, thanks!

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -0,0 +1,70 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+public typealias CommandBarItemGroup = [CommandBarItem]
+
+open class CommandBarItem: NSObject {
```
owenconnolly, line 10: Missing @objc(MSFCommandBarItem)

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -0,0 +1,70 @@
...
+    public init(
```
owenconnolly, line 11: Missing @objc

ios/FluentUI/Command Bar/CommandBar.swift
zhuorantan: Is it a requirement for the initial version of CCB? It would be quite complicated to implement and currently there is no animation specs. For now, only the state of the buttons can be updated via `public func updateButtonsState()`.

ios/FluentUI/Command Bar/CommandBarButton.swift
zhuorantan: Yeah. That's an issue. Do we have the design spec for this scenario?

ios/FluentUI/Command Bar/CommandBarButton.swift
zhuorantan: Renamed to `isPersistSelection`.

ios/FluentUI/Command Bar/CommandBar.swift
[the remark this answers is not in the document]
zhuorantan: Yeah. Only the items are removed because they can be retrieved from the leading and trailing buttons.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+        sender.updateState()
+
+        if newSelected {
+            delegate?.commandBar?(self, didSelectItem: sender.item)
+        } else {
+            delegate?.commandBar?(self, didDeselectItem: sender.item)
+        }
+    }
+
+    static let fadeViewWidth: CGFloat = 16.0
```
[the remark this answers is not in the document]
zhuorantan: There are private because they are in a `private extension`.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+        updateButtonsState()
+    }
+
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Apply `isEnabled` and `isSelected` state from `CommandBarItem` to the buttons
+    @objc public func updateButtonsState() {
```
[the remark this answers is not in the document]
zhuorantan: It is for programmatically changing the state of the buttons.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+        if UIView.userInterfaceLayoutDirection(for: semanticContentAttribute) == .rightToLeft {
+            // Flip the scroll view to invert scrolling direction. Flip its content back because it's already in RTL.
+            let flipTransform = CGAffineTransform(scaleX: -1, y: 1)
+            scrollView.transform = flipTransform
+            stackView.transform = flipTransform
+            containerMaskLayer.setAffineTransform(flipTransform)
+        }
+    }
+
+    func button(forItem item: CommandBarItem, isFixed: Bool = false) -> CommandBarButton {
```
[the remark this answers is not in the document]
zhuorantan: It is private because it is in a `private extension`.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+        let button = CommandBarButton(item: item, isFixed: isFixed)
+        button.addTarget(self, action: #selector(handleCommandButtonTapped(_:)), for: .touchUpInside)
+
+        return button
+    }
+
+    func updateShadow() {
```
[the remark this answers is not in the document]
zhuorantan: Same above.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+        if trailingButton != nil {
+            let trailingOffset = max(0, stackView.frame.width - scrollView.frame.width - scrollView.contentOffset.x)
+            let percentage = min(1, trailingOffset / scrollView.contentInset.right)
+            locations[2] = 1 - CommandBar.fadeViewWidth / containerView.frame.width * percentage
+        }
+
+        containerMaskLayer.locations = locations.map { NSNumber(value: Float($0)) }
+    }
+
+    @objc func handleCommandButtonTapped(_ sender: CommandBarButton) {
```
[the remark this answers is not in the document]
zhuorantan: Same above.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,290 @@
...
+            if let trailingButton = trailingButton {
+                return trailingButton.leadingAnchor.constraint(equalTo: containerView.trailingAnchor, constant: CommandBar.fixedButtonSpacing)
+            } else {
+                return trailingAnchor.constraint(equalTo: containerView.trailingAnchor)
+            }
+        }()
+
+        NSLayoutConstraint.activate([
+            containerView.topAnchor.constraint(equalTo: topAnchor, constant: CommandBar.insets.top),
+            bottomAnchor.constraint(equalTo: containerView.bottomAnchor, constant: CommandBar.insets.bottom),
```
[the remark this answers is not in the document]
zhuorantan: The constant is applied from the first anchor to the second anchor. If `containerView.bottomAnchor` is the first anchor, the bottom of `containerView` would be under the bottom of `CommandBar`.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zhuorantan: They are private.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zhuorantan: To group private members together.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
zhuorantan: Before iOS 14, `UIStackView` is non rendering view, making it unable to have rounded corners.

ios/FluentUI/Command Bar/CommandBarItem.swift
zhuorantan: Long press to show the menu. Or set `showsMenuAsPrimaryAction` to `true`, then the button would show the menu when tapped.

ios/FluentUI/Command Bar/CommandBar.swift
zhuorantan: leading and trailing buttons are fixed and can't be placed inside the scroll view.

ios/FluentUI/Command Bar/CommandBar.swift
zhuorantan: RTL on iOS doesn't change the scroll direction but `UIStackView` would change the ordering of its items.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
zhuorantan: It is optional. But it would be hard to identify which button is tapped without subclassing and adding a property specifically for this purpose.

zhuorantan:
> The voice over accessibility doesn't honor left to right order. for the scroll view. A person using accessibility tools won't be able to discover that tools scroll horizontally or discover additional items.
> 
> Expected behavior:
> Voice over control should scroll to include all items or assert that there are additional items not visible on the screen.
> 
>  Screen.Recording.2021-01-06.at.11.48.26.AM.mov

I think you chose the wrong target in accessibility inspector. It should be the simulator rather than your Mac. The accessibility support of scroll view is provided by iOS itself and doesn't require additional code from developers.

Screen Shot 2021-01-21 at 16 54 16

ios/FluentUI/Command Bar/CommandBarItem.swift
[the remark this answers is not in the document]
zhuorantan: To prevent exposing `UIButton` properties and methods.

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    override var isEnabled: Bool {
+        didSet {
+            updateStyle()
+        }
+    }
+
+    init(item: CommandBarItem, isFixed: Bool = false) {
+        self.item = item
```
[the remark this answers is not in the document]
zhuorantan: Currently, to change the state of the buttons programmatically, the developer changes the properties of `CommandBarItem`, then call `CommandBar.updateButtonsState()`. It would be not much different from implementing a delegate pattern with batch update support.

ios/FluentUI/Command Bar/CommandBar.swift
zhuorantan: The mask layer is to apply the gradient effect on the sides of the scroll view. Applying directly to `CommandBar` would corp the sides of the view. And also it can't be applied to the scroll view since the coordinate is different.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,288 @@
...
+
+    private var leadingButton: CommandBarButton?
+    private var trailingButton: CommandBarButton?
+
+    private let containerMaskLayer: CAGradientLayer = {
+        // A mask layer using alpha color channel.
+        let layer = CAGradientLayer()
+        layer.colors = [UIColor.clear, UIColor.white, UIColor.white, UIColor.clear].map { $0.cgColor }
+        layer.startPoint = CGPoint(x: 0, y: 0.5)
+        layer.endPoint = CGPoint(x: 1, y: 0.5)
```
owenconnolly, line 170: worth using constants for values like these?

ios/FluentUI/Command Bar/CommandBarButton.swift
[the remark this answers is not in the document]
rdeassis: +1 on removing the call to updateButtonsState

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
kubalani: extensions in `FluentUI` are preserved for grouping functionality as opposed to access levels

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
kubalani: @zhuorantan Thanks for brining this to my attention  if we can document the need for additional wrapper as a requirement for iOS 13 then we can remove them later. 

I understand the need for not exposing `UIStackView` api's. I could argue the same thing that exposing `UIView` could have unexpected consequences in other scenarios. A final subclass of `UIStackView` that suppress the unintended consequences could be used a drop in replacement instead. 

My primary concerns by using too many wrapper views are accessibility bugs, boilerplate code and forward compatibility issues with UIKit. This comment is **not** a blocking issue.

kubalani:
> > The voice over accessibility doesn't honor left to right order. for the scroll view. A person using accessibility tools won't be able to discover that tools scroll horizontally or discover additional items.
> > Expected behavior:
> > Voice over control should scroll to include all items or assert that there are additional items not visible on the screen.
> > Screen.Recording.2021-01-06.at.11.48.26.AM.mov
> 
> I think you chose the wrong target in accessibility inspector. It should be the simulator rather than your Mac. The accessibility support of scroll view is provided by iOS itself and doesn't require additional code from developers.
> 
> Screen Shot 2021-01-21 at 16 54 16

Thanks for addressing these comments, I verified the accessibility voice over pass in device and it works as expected !!

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
harrieshin: I don't think it is a must. we should follow up with the feature crew?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
zhuorantan: Using a string as identifier is not a friendly API for Swift. I originally intended to use generic feature of Swift, to enable providing a custom enum type to identify buttons. But that would make CCB completely not compatible with ObjC.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -0,0 +1,192 @@
...
+        let label = Label(style: .subhead, colorStyle: .regular)
+        label.text = text
+        label.textAlignment = .center
+        return label
+    }
+}
+
+extension CommandBarDemoController: CommandBarDelegate {
+    func commandBar(_ commandBar: CommandBar, didSelectItem item: CommandBarItem) {
+        guard let item = item as? Item else {
```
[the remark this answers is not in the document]
zhuorantan: A solution would be adding a block property to `CommandBarItem` instead of a delegate method to handle item selection, but this would not be aligned with designs in other Fluent UI views. What do you think?

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/CommandBar/textItalic24Regular.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
```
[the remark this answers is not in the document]
zhuorantan: I exported them from the design spec file. I think they are the same icons as the fluent icons but binary may not be the same. Should I replace them with the ones from fluent icons repo?

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -0,0 +1,71 @@
...
+//
+
+import UIKit
+
+public typealias CommandBarItemGroup = [CommandBarItem]
+
+@objc(MSFCommandBarItem)
+open class CommandBarItem: NSObject {
+    @objc public init(
+        iconImage: UIImage?,
```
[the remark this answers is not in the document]
zhuorantan: This is answered in above convos. Also, if we migrate the delegate to block based, we may not need to identify each button.

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -0,0 +1,71 @@
...
+        isEnabled: Bool = true,
+        isSelected: Bool = false,
+        isPersistSelection: Bool = true,
+        accessbilityLabel: String? = nil
+    ) {
+        self.iconImage = iconImage
```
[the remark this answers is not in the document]
zhuorantan: Yes, but I'm worried about the appearance. Since the design assumes all buttons have the same size icon, using labels might look odd.

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+        }
+    }
+
+    override var isEnabled: Bool {
+        didSet {
+            updateStyle()
+        }
+    }
+
+    init(item: CommandBarItem, isPersistSelection: Bool = true) {
```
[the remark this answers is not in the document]
zhuorantan: I will add some documentation.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -0,0 +1,206 @@
...
+        }
+
+        container.addArrangedSubview(createLabelWithText("Default"))
+
+        let defaultCommandBar = CommandBar(itemGroups: itemGroups)
+        defaultCommandBar.backgroundColor = Colors.Navigation.System.background
+        defaultCommandBar.delegate = self
+        container.addArrangedSubview(defaultCommandBar)
+
+        container.addArrangedSubview(createLabelWithText("With Fixed Button"))
```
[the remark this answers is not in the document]
zhuorantan: I copied from other demos. Looks like it doesn't.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,288 @@
...
+        configureHierarchy()
+        updateButtonsState()
+    }
+
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    /// Apply `isEnabled` and `isSelected` state from `CommandBarItem` to the buttons
```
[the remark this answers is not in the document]
zhuorantan: I'm not sure. Maybe we can provide properties to set icons for different states (disabled, selected). If there is such need in other scenarios, we should implement reloading of the entire command bar.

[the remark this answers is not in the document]
owenconnolly: So there are scenarios where the command state needs to be updated, but the icon should remain the same, which should be fine. We do also have a scenario where the command button just has a label (no icon) and the label can be updated after the command is executed. Can we add that capability as it should be relatively straightforward?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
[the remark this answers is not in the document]
owenconnolly: @rdeassis what is the Fluent preference here? Completion blocks seem the right design in terms of Apple platforms, but it may not be preferred for Fluent?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
owenconnolly: Okay, thanks, for my purposes I can look at using a Swift subclass, but the general design is still an issue if you are determined to use ObjC. The block option may be better, depending on what the Fluent team say

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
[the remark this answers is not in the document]
rdeassis: Although there are no official standards defined in FluentUI, I believe the order of preference to be:

1. Completion blocks/closures
2. Protocol/delegate pattern
3. Selectors

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -0,0 +1,288 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFCommandBarDelegate)
+public protocol CommandBarDelegate: AnyObject {
+    @objc optional func commandBar(_ commandBar: CommandBar, shouldSelectItem item: CommandBarItem) -> Bool
```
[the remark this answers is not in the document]
rdeassis: Is documentation for the protocol methods going to be added?

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class CommandBarButton: UIButton {
```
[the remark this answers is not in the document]
rdeassis: Yes! Everything that's public should have /// doc comments.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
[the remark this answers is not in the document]
owenconnolly: That seems like the right option then, @zhuorantan would you be able to incorporate a completion block for the item?

ios/FluentUI/Command Bar/CommandBarItem.swift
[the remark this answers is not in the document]
owenconnolly: Thanks, yes we don't need button identifiers now with the block, the design is more flexible now in that sense.

ios/FluentUI/Command Bar/CommandBarItem.swift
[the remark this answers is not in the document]
owenconnolly: Thanks for adding the label support, it looks good in the demo app!

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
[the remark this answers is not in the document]
owenconnolly: Okay, thanks, something for FluentUI team to follow up on at some point @rdeassis ?

ios/FluentUI/Command Bar/CommandBar.swift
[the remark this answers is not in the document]
owenconnolly: Thanks, I see that in the current iteration and demo app that command button labels can be used and also updated!

ios/FluentUI/Command Bar/CommandBar.swift
owenconnolly: @rdeassis is this something that Fluent wants to track as a potential future enhancement?

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+    }
+
+    init(item: CommandBarItem, isPersistSelection: Bool = true) {
+        self.item = item
+        self.isPersistSelection = isPersistSelection
+
+        super.init(frame: .zero)
+
+        translatesAutoresizingMaskIntoConstraints = false
+        setImage(item.iconImage, for: .normal)
```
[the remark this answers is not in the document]
owenconnolly: Thanks, confirmed this now works in the demo app!

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -0,0 +1,99 @@
...
+        updateState()
+    }
+
+    @available(*, unavailable)
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func updateState() {
+        isEnabled = item.isEnabled
```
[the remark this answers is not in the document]
owenconnolly: We can leave this for a future enhancement if it adds flexibility

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
owenconnolly: @rdeassis where can we track this for future work?

ios/FluentUI/Command Bar/CommandBarButton.swift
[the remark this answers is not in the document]
owenconnolly: This likely could be a future enhancement tied to changes for more dynamic updates to the command items, can we track it with that future work @rdeassis ?

owenconnolly (approved):

rdeassis (approved):

Issue #340, opened 2020-12-11
[AvatarVnextView] Define fallback / unauthenticated scenarios.


rdeassis:
#### Platform
- [x] iOS
- [ ] macOS

#### Describe the feature that you would like added
[AvatarVnextView] Define fallback / unauthenticated scenarios.
The current implementation of the AvatarView has 4 variations of fallback image styles:

**primaryFilled** and **onAccentFilled** respectively depicted below:
![Screen Shot 2020-12-11 at 3 16 14 PM](https://user-images.githubusercontent.com/68076145/101963729-45773080-3bc4-11eb-8601-98e8e3eb91f9.png)

These variations are used to represent an existing user that does not have associated information (primary and secondary text) enough to generate background colors and initials. But it still a valid user.

The different background colors are meant to work well with different background (white vs primary color) variations like the ones below:

| primaryFilled                                       | onAccentFilled                                      |
|----------------------------------------------|--------------------------------------------|
| ![IMG_0514](https://user-images.githubusercontent.com/68076145/101964201-7146e600-3bc5-11eb-8072-2be076ece7a2.jpg) | ![IMG_0513](https://user-images.githubusercontent.com/68076145/101964197-6f7d2280-3bc5-11eb-86e8-a1e0850022d5.jpg) |

**outlined** and **primaryOutlined** respectively depicted below:
![Screen Shot 2020-12-11 at 3 16 23 PM](https://user-images.githubusercontent.com/68076145/101963751-53c54c80-3bc4-11eb-97ab-b5608104a5b7.png)

These variations are used to indicate no user is associated with the avatar (unauthenticated / sign out scenarios).

We need to make a decision on what are the variations that will be supported by the new AvatarVnextView.
Victor mentioned we can make this simpler by defining fewer colors and variations that could work for all the apps.

#### What component or utility would this be added to

AvatarVnextView

#### Have you discussed this feature with our team, and if so, who

Currently discussing this with Victor Coulon.

Issue #338, opened 2020-12-11
[AvatarVnextView] Add support for background/foreground/ring colors based on the primaryText and secondaryText


rdeassis:
#### Platform
- [x] iOS
- [ ] macOS

The current AvatarView relies on the logic of the InitialsView to get the background color of the Avatar.
The design team is currently proposing 30 different color variations to be used based on the hash of the primaryText and secondaryText properties.

These 30 colors should be tokenized in the YAML file (light, dark and HC variations) and used to define the AvatarVnext default colors for:
 1. background
 2. foreground (used for the text color or icon image tint color)
 3. avatar ring

#### What component or utility would this be added to

AvatarVnextView

#### Have you discussed this feature with our team, and if so, who

Discussion currently happening with Victor.

#### Additional context/screenshots

![avatar_colors](https://user-images.githubusercontent.com/68076145/101949274-31bdd100-3ba8-11eb-9e39-a059adcf8659.png)

PR #346, opened 2020-12-16
[AvatarVnextView] Revisiting Avatar state, styles, covering fallback scenarios and adding tokenized calculated background colors.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

1. These changes revisit the Avatar styles and resolves issue #340.
We now have parity with the current Avatar implementation in terms of fallback scenarios (where the user information is not provided or the initials can't be calculated such as in phone number only scenario).

2. NoRing is not a style anymore. It's part of the state of the avatar where a flag can be set to display the ring.

3. The styles and sizes can now be set/updated after the Avatar has been initialized. Updating any of those properties will trigger SwiftUI to refresh the views.

4. The foundation is also laid for issue #338. We are currently using the same calculated color options as the current Avatar, but they are tokenized and only need to be updated to the new 30 variations of colors once the design team has them ready.

### Verification

Ran variations of the Demo app as well as setting states of show/hide ring, custom ring and background colors.
Light and dark mode variations.

|                                       |                                      |
| Screen Shot 2020-12-16 at 10 17 28 AM | Screen Shot 2020-12-16 at 10 18 02 AM |
| Screen Shot 2020-12-16 at 10 12 13 AM | Screen Shot 2020-12-16 at 10 12 22 AM |
| Screen Shot 2020-12-16 at 3 29 10 PM | Screen Shot 2020-12-16 at 3 29 25 PM |
| Screen Shot 2020-12-16 at 3 29 16 PM | Screen Shot 2020-12-16 at 3 29 32 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/346)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
@@ -51,8 +51,8 @@ class AvatarVnextDemoController: DemoController {
         for size in AvatarVnextSize.allCases.reversed() {
             let phoneNumber = "+1 (425) 123 4567"
             let defaultAvatar = createAvatarView(size: size,
-                                                       name: phoneNumber,
-                                                       style: .default)
+                                                 name: phoneNumber,
+                                                 style: .default)
             avatarViews.append(defaultAvatar)
 
             let accentAvatar = createAvatarView(size: size,
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
@@ -7,72 +7,103 @@ import FluentUI
 import UIKit
 
 class AvatarVnextDemoController: DemoController {
-    enum BorderStyle: Int {
-    case noBorder
-    case defaultBorder
-    case colorfulBorder
-    }
-
     override func viewDidLoad() {
         super.viewDidLoad()
-
-        let showPresenceSettingView = createLabelAndSwitchRow(labelText: "Show presence",
-                                                              switchAction: #selector(toggleShowPresence(switchView:)),
-                                                              isOn: isShowingPresence)
 
         let backgroundSettingView = createLabelAndSwitchRow(labelText: "Use alternate background color",
                                                             switchAction: #selector(toggleAlternateBackground(switchView:)),
                                                             isOn: isUsingAlternateBackgroundColor)
-
-        let opaqueBorderSettingView = createLabelAndSwitchRow(labelText: "Use opaque borders",
-                                                            switchAction: #selector(toggleOpaqueBorders(switchView:)),
-                                                            isOn: isUsingOpaqueBorders)
-
         addRow(items: [backgroundSettingView])
+
+        let transparencySettingView = createLabelAndSwitchRow(labelText: "Do not use transparency",
+                                                              switchAction: #selector(toggleOpaqueBorders(switchView:)),
+                                                              isOn: !isTransparent)
+        addRow(items: [transparencySettingView])
+
+        let showPresenceSettingView = createLabelAndSwitchRow(labelText: "Show presence",
+                                                              switchAction: #selector(toggleShowPresence(switchView:)),
+                                                              isOn: isShowingPresence)
         addRow(items: [showPresenceSettingView])
-        addRow(items: [opaqueBorderSettingView])
 
-        createSection(withTitle: "Circle style for person",
-                      name: "Kat Larrson",
-                      image: UIImage(named: "avatar_kat_larsson")!,
-                      style: .noring)
+        let showRingsSettingView = createLabelAndSwitchRow(labelText: "Show rings",
+                                                           switchAction: #selector(toggleShowRings(switchView:)),
+                                                           isOn: isShowingRings)
+        addRow(items: [showRingsSettingView])
 
-        createSection(withTitle: "Square style for group",
-                      name: "NorthWind Traders",
-                      image: UIImage(named: "site")!,
-                      style: .group)
+        addTitle(text: "Default style")
+        for size in AvatarVnextSize.allCases.reversed() {
+            let name = "Kat Larrson"
+            let imageAvatar = createAvatarView(size: size,
+                                               name: name,
+                                               image: UIImage(named: "avatar_kat_larsson")!,
+                                               style: .default)
+            avatarViews.append(imageAvatar)
 
-        createSection(withTitle: "Circle style with border",
-                      name: "Kat Larrson",
-                      image: UIImage(named: "avatar_kat_larsson")!,
-                      style: .default,
-                      borderStyle: .defaultBorder)
+            let initialsAvatar = createAvatarView(size: size,
+                                                  name: name,
+                                                  style: .default)
+            avatarViews.append(initialsAvatar)
 
-        addTitle(text: "Fallback Images")
+            addRow(text: size.description, items: [imageAvatar.view, initialsAvatar.view], textStyle: .footnote, textWidth: 100)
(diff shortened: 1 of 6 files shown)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -148,44 +148,61 @@ Opacity:
...
+  foregroundDefaultColor: $Icon.accentColor
 
-UnauthenticatedAvatarTokens extends AvatarTokens:
-  backgroundDefaultColor: $Colors.Background.neutral4
-  foregroundDefaultColor: $Colors.Foreground.neutral3
-  ringDefaultColor: $Colors.Background.neutralDisabled
+GroupAvatarTokens extends AvatarTokens:
+  borderRadius: [ xSmall: $Border.radius.small, small: $Border.radius.medium, medium: $Border.radius.medium, large: $Border.radius.large, xlarge: $Border.radius.large, xxlarge: $Border.radius.xlarge ]
+
+OutlinedAvatarTokens extends AvatarTokens:
```
rdeassis, line 194: Hey @Victa,

Just FYI regarding the colors for fallback styles: I have used the closest colors in the gray pallet to replicate the current behavior of the Avatarview. Here is the conversion from the current palette to the vNext palette:
	 - Gray50 (#F1F1F1) -> grey94 (#F0F0F0)
	 - Gray600 (#404040) -> grey26 (#424242)
	 - Gray500 (#6E6E6E) -> grey42 (#6B6B6B)
	 - Gray200 (#C8C8C8) -> grey78 (#C7C7C7)

Victa: Oh yeah, that's right. We don't have the latest colors yet. Thanks!

ios/FluentUI/People Picker/InitialsView.swift
```diff
@@ -12,7 +12,7 @@ import UIKit
  The initials are generated from a provided primary text (e.g. a name) or secondary text (e.g. an email address) and placed as a label above a colored background.
  */
 class InitialsView: UIView {
-    private static func initialsBackgroundColor(fromPrimaryText primaryText: String?, secondaryText: String?) -> UIColor {
+    static func initialsBackgroundColor(fromPrimaryText primaryText: String?, secondaryText: String?, colorOptions: [UIColor]? = nil) -> UIColor {
```
kubalani, line 15: this won't update the initials view if the token is changed?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -148,44 +148,61 @@ Opacity:
...
   presenceIconOutlineColor: $Colors.Background.neutral1
   presenceIconOutlineThickness: [ xSmall: $Border.size.none, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thick ]
-  ringDefaultColor: "#C50F1F"
+  ringDefaultColor: $Colors.Background.neutralDisabled
   ringGapColor: $Colors.Background.neutral1
   ringInnerGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
   ringThickness: [ xSmall: $Border.size.thin, small: $Border.size.thin, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
   ringOuterGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
   size: [ xSmall: 16pt, small: 24pt, medium: 32pt, large: 40pt, xlarge: 52pt, xxlarge: 72pt ]
+  textCalculatedBackgroundColors: [ "NamedColor(cyanBlue10)",
```
kubalani, line 163: these colors are specific to Avatar's only, is that right? I am not sure if I understand the significance of making them in tokens? These are just random bucket of colors used to hash out a background view?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
```diff
@@ -85,30 +116,28 @@ class AvatarVnextDemoController: DemoController {
...
+    private var isShowingRings: Bool = false {
         didSet {
-            if oldValue != isUsingOpaqueBorders {
-                for avatarView in avatarViewsWithImages {
-                    avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
-                }
-
-                for avatarView in avatarViewsWithInitials {
-                    avatarView.state.shouldUseOpaqueBorders = isUsingOpaqueBorders
+            if oldValue != isShowingRings {
```
kubalani, line 128: might be worth it to add a `toggle` extension Bool.

```swift
extension Bool {
   mutating func toggle() {
       self = !self
   }
}
```

kubalani (approved):
LGTM !! I would recommend to use `Style` as an `observedobject` decouples from `state` and `tokens` as state is more functional entity (at-least to my understanding). Might need a breakdown if state has more attributes.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
```diff
@@ -7,72 +7,103 @@ import FluentUI
...
+            addRow(text: size.description, items: [imageAvatar.view, initialsAvatar.view], textStyle: .footnote, textWidth: 100)
+        }
 
-        addTitle(text: "Fallback Images")
+        addTitle(text: "Fallback (default style and accent style)")
         for size in AvatarVnextSize.allCases.reversed() {
             let phoneNumber = "+1 (425) 123 4567"
-            let fallbackImageAvatar = createAvatarView(size: size,
+            let defaultAvatar = createAvatarView(size: size,
                                                        name: phoneNumber,
```
sophialee0416, line 54: Spacing seems off

sophialee0416 (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarVnextDemoController.swift
rdeassis: Fixed!

ios/FluentUI/People Picker/InitialsView.swift
rdeassis: Nope. The AvatarVnext does not use the InitialsView.
We are just borrowing the functionality of the Initials calculation and the hash to determine the background color (which  will be used to calculate the ring and foreground colors shortly.

tools/sgen/input/FluentUIStyle.yml
rdeassis: Correct. These are all the colors that can be used depending on the hash of the primary and secondary texts.
In the vNext tokenized world it's currently just used here, but I definitely anticipate that we'll move this to a more semantically generic place once more tokenized controls need these colors.

But the colors are not random. They are a pre-defined group of options that should be consistent. These colors are the same used by the current implementation of the AvatarView. But the design team will shortly release a new group of 30 colors to replace these. They will also include ringColors and text/foreground colors.

rdeassis (commented):
> LGTM !! I would recommend to use `Style` as an `observedobject` decouples from `state` and `tokens` as state is more functional entity (at-least to my understanding). Might need a breakdown if state has more attributes.

I agree that it makes sense to break it down into more observableObjects if it grows larger (like we spoke offline today).
We'll keep an eye on this to assess if it's needed in the future.

Issue #339, opened 2020-12-11
[AvatarVnextView] Add support "Out Of Office aware" presence icons


rdeassis:
#### Platform
- [x] iOS
- [ ] macOS

#### Describe the feature that you would like added

As part of the discussion raised on #335, we need to add support for presence icons that vary based on the state of the avatar being "Out Of Office".

The screenshot below shows how the presence status icons should vary based on the "Out Of Office" state of the Avatar.

#### What component or utility would this be added to

AvatarVnextView

#### Have you discussed this feature with our team, and if so, who

Currently discussing the issue with Victor Coulon on the design team.

#### Additional context/screenshots

![Screen Shot 2020-09-17 at 12 03 18 PM](https://user-images.githubusercontent.com/68076145/101954299-ac8aea00-3bb0-11eb-8379-b1a80b43d1c9.png)

PR #347, opened 2020-12-17
[AvatarVnextView] Adding support for OOF state.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The Avatar is not supposed to have "Out Of Office" as a presence enumeration.
The OOF is instead a state of the Avatar that determine how the presence is represented in the avatar.

This change adds support for the isOutOfOffice property and renders the presence icons accordingly and solves issue #339.

### Verification

| Regular                                       | Out Of Office                                      |
| Screen Shot 2020-12-16 at 6 05 49 PM | Screen Shot 2020-12-16 at 6 06 00 PM |
| Screen Shot 2020-12-16 at 6 06 14 PM | Screen Shot 2020-12-16 at 6 06 17 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/347)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -92,11 +92,12 @@ Colors:
...
     - available: "NamedColor(presenceAvailable)"
     - away: "NamedColor(presenceAway)"
+    - awayOutOfOffice: "NamedColor(presenceOof)"
     - blocked: "NamedColor(presenceBlocked)"
     - busy: "NamedColor(presenceBusy)"
     - doNotDisturb: "NamedColor(presenceDnd)"
     - offline: "NamedColor(presenceOffline)"
-    - outOfOffice: "NamedColor(presenceOof)"
+    - offlineOutOfOffice: "NamedColor(presenceOof)"
     - unknown: "NamedColor(presenceUnknown)"
```
Saadnajmi, line 101: Only away and offline need a separate "OOF" entry? and why don't these have to line up exactly with the Avatar presence enum?

Saadnajmi (approved):

tools/sgen/input/FluentUIStyle.yml
rdeassis: As far as colors go, Away and Offline are the only ones that change.
The icons have a different mapping though.

What Teams passed as a requirement is that each item of the presence menu should be rendered in a certain way for both "Regular" or "OOF" modes.

Away and Offline should render as the "Out Of Office" icon.
The other ones are just a regular icon as opposed to a filled one (and it's not for all of the status).

PR #354, opened 2020-12-23
Support button action

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes
In the current button implementation, we override the mouse events to customize the mouse down behavior. Since we are not calling the super class, we need to manually trigger the target action.

### Verification
In the test app, ensured the button action is executed (mouse & keyboard), and the mouse events are fired. 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/354)

### DIFF
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -88,6 +88,7 @@ open class Button: NSButton {
 		isBordered = false
 		wantsLayer = true
 		layer?.contentsScale = window?.backingScaleFactor ?? 1.0
+		setButtonType(.momentaryChange)
 
 		if let cell = cell as? ButtonCell {
 			// The Button properties `contentTintColorDisabled`,
@@ -183,47 +184,10 @@ open class Button: NSButton {
 		}
 	}
 
-	private var trackingArea: NSTrackingArea?
-	
-	override public func updateTrackingAreas() {
-		super.updateTrackingAreas()
-
-		if let trackingArea = trackingArea {
-			removeTrackingArea(trackingArea)
-			self.trackingArea = nil
-		}
-
-		let opts: NSTrackingArea.Options = [.mouseEnteredAndExited, .activeAlways]
-		let trackingArea = NSTrackingArea(rect: bounds, options: opts, owner: self, userInfo: nil)
-		addTrackingArea(trackingArea)
-		self.trackingArea = trackingArea
-	}
-
 	open override func mouseDown(with event: NSEvent) {
-		var pressed = true
-		var nextEvent:NSEvent
-		var mouseInRect = true
-
 		mouseDown = true
-		while (pressed) {
-			nextEvent = (self.window?.nextEvent(matching: NSEvent.EventTypeMask(rawValue: NSEvent.EventTypeMask.RawValue(NX_LMOUSEUPMASK | NX_MOUSEEXITEDMASK | NX_MOUSEENTEREDMASK))))!
-			
-			switch nextEvent.type {
-			case NSEvent.EventType.leftMouseUp:
-				/// send action if cursor is inside the button
-				if mouseInRect {
-					self.sendAction(action, to: target)
-				}
-				pressed = false
-				mouseDown = false
-			case NSEvent.EventType.mouseExited:
-				mouseInRect = false
-			case NSEvent.EventType.mouseEntered:
-				mouseInRect = true
-			default: // ignore other events
-				break
-			}
-		}
+		super.mouseDown(with: event)
+		mouseDown = false
 	}
 
 	open override var isEnabled: Bool {
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -183,16 +183,47 @@ open class Button: NSButton {
 		}
 	}
 
-	open override func mouseDown(with event: NSEvent) {
-		mouseDown = true
-		super.mouseDown(with: event)
-		/// mouseUp doesn't register when super.mouseDown is called
-		self.mouseUp(with: event)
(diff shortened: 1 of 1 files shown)


macos/FluentUI/Button/Button.swift
```diff
@@ -188,6 +188,7 @@ open class Button: NSButton {
 	}
 
 	open override func mouseUp(with event: NSEvent) {
+		_ = target?.perform(action, with: event)
```
anandrajeswaran, line 191: Should we be doing this by calling super in these overrides instead so we inherit superclass behavior instead of recreating it?

Saadnajmi, line 191: Cann you add a comment explaining why we specifically want to trigger the action here, as opposed to `mouseDown` or somewhere else? And why you have to assign to `_`

anandrajeswaran (changes requested):
Has this just always been broken? Or was this broken recently?

chripear:
> Has this just always been broken? Or was this broken recently?

It has always been broken.

macos/FluentUI/Button/Button.swift
chiuam: I think so. As I'm doing more testing, calling perform action manually in mouseUp() introduces bugs like button action fires on click-drag-release gesture.

chiuam: For some reason when I call super.mouseDown, contentTintColor doesn't render as expected.
fluent_button
(investigating)

macos/FluentUI/Button/Button.swift
```diff
@@ -185,11 +185,14 @@ open class Button: NSButton {
 
 	open override func mouseDown(with event: NSEvent) {
 		mouseDown = true
+		super.mouseDown(with: event)
+		/// mouseUp doesn't register when super.mouseDown is called
```
chripear, line 189: I don't understand this comment.  Surely we don't a mouseDown event to trigger a mouseUp event?  Or is this saying that somehow calling super.mouseDown causes subsequent mouseUp events to not register?

macos/FluentUI/Button/Button.swift
chripear: Reversing the order might help:

`super.mouseDown(with: event)`
`mouseDown = true`

it's the assignment `mouseDown = true` that sets the contentTintColor, so it would override anything that the `super` call did.

chiuam: super.mouseDown blocks until the button is released. So if we call updateContentTintColor() after mouseDown, contentTintColor will not update while button is in pressed state.

macos/FluentUI/Button/Button.swift
chiuam: The latter. I'm not sure why though...

anandrajeswaran: https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/EventOverview/HandlingMouseEvents/HandlingMouseEvents.html#//apple_ref/doc/uid/10000060i-CH6-SW4 explains what's going on under the hood here. Can we just put mouseDown = false after the call to super.mouseDown so we're toggling our FluentUI.Button behavior based on how the super class works instead of sending the mouseUp message to super and potentially modifying the behavior we get from the super class?

anandrajeswaran (changes requested):

macos/FluentUI/Button/Button.swift
chiuam: @anandrajeswaran Thanks for the leads. I couldn't get the UI to work as expected with super.mouseDown. I think NSButton uses the mouse-tracking loop approach to handle mouse events. So, if we want to customize the behavior of any mouse events, the proper way to do it may be to recreate the loop and handle the events ourselves.

chripear (approved):

macos/FluentUI/Button/Button.swift
anandrajeswaran: But we don't want to customize the behavior of mouse events, right? We want existing mouse event behavior with but just to change some things when that status changes. Could we instead just have our mouseDown implementation be:
mouseDown = true
super.mouseDown
mouseDown = false

chiuam: I tried the approach you proposed and other alternatives with `super` but none of them worked for me :/ 

We do want to customize the mouse down and mouse up behaviors for updating the contentTintColor. According to the Apple documentation, the mouse-tracking loop filters and processes all mouse events, which means mouse Up gets processed inside the loop. I agree that we should leverage the super class whenever possible, but unfortunately in this case it cannot suffice the design requirements.

macos/FluentUI/Button/Button.swift
chiuam: Obsolete, but I added comments to the updated version.

macos/FluentUI/Button/Button.swift
anandrajeswaran: What didn't work about this approach? It works locally for me
```
	open override func mouseDown(with event: NSEvent) {
		mouseDown = true
		super.mouseDown(with: event)
		mouseDown = false
	}
```

chiuam: The only issue I had with it was if I hold down the button, the button title disappears but reappears when you drag the cursor outside the button. 
ezgif com-gif-maker (1)

chiuam: Setting the button typed worked, thanks!

harrieshin (approved):

anandrajeswaran (approved):

PR #361, opened 2020-12-30
Improve rect calculations for constrained buttons

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

#### Problem:
Previous frame calculations assumed that the button frame size was equal to the natural content sizes plus known paddings and spacings.  For example, for a button with both image and title laid out horizontally, we assumed the following equality:

`rect.width == imageSize.width + titleSize.width + titleToImageSpacing + 2*horizontalPadding`

However, constraining the button to a greater or lesser width changes the value of the left-hand side without changing anything on the right-hand side, resuling in an inequality.  If we constrain this button to a width greater than the natural content width, the effect is that the button image and title appear left-aligned in a larger button frame.

Before 😭

#### Fix:
Adjusted coordinate calculations to ensure image and title remain centered in the button frame when subject to such constraints.

After 😁

Refactored the math to make consistency between cases clear.

### Verification
Added controls to the test app to constrain button widths and heights to larger and smaller sizes to verify the fix and ensure image and title are properly centered and aligned for all permutations.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/361)


macos/FluentUI/Button/Button.swift
```diff
@@ -388,125 +388,77 @@ class ButtonCell: NSButtonCell {
...
+		let xOffsetFromTitle = (titleSize.width + titleToImageSpacing) / 2
+		let yOffsetFromTitle = (titleSize.height + titleToImageSpacing - titleToImageVerticalSpacingAdjustment) / 2
 
 		switch imagePosition {
 		case .noImage:
-			x = 0
-			y = 0
-			width = 0
-			height = 0
+			return NSZeroRect
```
Saadnajmi, line 403: nit: can we avoid the early return in this one case? That would require introducing two vars `width` and `height` that default to `imageSize.width` and `imageSize.height` which is arguably less clean, so I don't feel too strongly. 

The preference for no early returns comes from our [obj-s guide](https://microsoft.github.io/objc-guide/MethodsAndImplementations/EarlyReturns.html) and our [swift guide](https://microsoft.github.io/swift-guide/EarlyReturn.html).

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -89,10 +105,24 @@ class TestButtonViewController: NSViewController, NSMenuDelegate {
 		buttonStatesPopup.target = self
 		buttonStatesPopup.action = #selector(TestButtonViewController.stateChanged)
 
+		widthPopup.addItems(withTitles: widths.keys.sorted())
+		widthPopup.selectItem(withTitle: "natural")
```
Saadnajmi, line 109: nit: I would factor out the string to a constant

Saadnajmi (approved):

macos/FluentUI/Button/Button.swift
```diff
@@ -388,125 +388,77 @@ class ButtonCell: NSButtonCell {
...
 		case .imageOnly, .imageOverlaps:
```
rdeassis, line 404: Are the cases circled in red the imageOverlaps case? What is the scenario for that use case?

Screen Shot 2020-12-30 at 11 57 55 PM

macos/FluentUI/Button/Button.swift
```diff
@@ -388,7 +388,7 @@ class ButtonCell: NSButtonCell {
 			return NSZeroRect
 		}
 
-		let isRTL = NSApp.userInterfaceLayoutDirection == .rightToLeft
+		let layoutDirectionSign : CGFloat = NSApp.userInterfaceLayoutDirection == .rightToLeft ? -1 : 1
```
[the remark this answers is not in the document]
rdeassis: for reference: 

https://developer.apple.com/documentation/appkit/nsview/1483254-userinterfacelayoutdirection

anandrajeswaran (approved):

macos/FluentUI/Button/Button.swift
Saadnajmi: I don't think there's an explicit use case, but it's an enum value and we want to handle all the enum values.

macos/FluentUI/Button/Button.swift
chripear: What's bad about early return?  Would it be better if it was even earlier?

chripear: Literally no other case uses zero for any of the four values: this case is a total outlier in terms of the values.

chripear: The intermediate variables improve readability and reduce the chance of mistakes since I am not writing multiple long lines of math for each case.  I am not too worried about the performance implications of a couple of CGFloat calculations done upfront in the service of a much simpler switch block.

chripear: > choosing a reasonable default for all values that need to be set

Out of 9 cases, 8 have the same width and height values.  4 have the same x values, and 4 of the other 5 are those same values plus or minus a constant.  6 have the same y values, and 2 of the other 3 are those same values plus or minus a constant.  It seems to me the choice of initial values is pretty optimal...

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
chripear: The type specifier is helpful because Swift infers `int` based on the literals, and complains later when I try to use this value in CGFloat calculations.  Sometimes I think language designers need to go back to math class and forget the bad habits they learned from C...

macos/FluentUI/Button/Button.swift
chripear: I covered this case and included an example to ensure the layout is as intended (image and text properly centered together).  I didn't really care that this specific example looks ugly in the test app.  If you are trying to imagine a use case where it wouldn't look ugly... it would probably be a semitransparent image or an elaborate framing/border for the title.

macos/FluentUI/Button/Button.swift
anandrajeswaran: Then break that case out into a separate logical block or break handling from the other cases into a separate func? As mentioned in the links Saad shared, early returns are often used as a crutch where different factoring would improve clarity - imagine I'm investigating a bug in the no image case and I'm debugging through and trying to understand why we're doing calculations for things that aren't used and trying to understand whether that's by design or a code bug.

In ObjC/Swift I've not seen a case (other than guard lets for input validation) where adding early returns actually improves code clarity relative to more explicit factoring. I've seen many cases where it causes bugs due to control flow being more difficult to parse with stuff like added cleanup surprisingly not applying to all code paths.

chripear: Moved the early return to the `guard` clause at top.  Not sure if the preferred style is to use one or many of these, but they all return the same thing so I used one.

chripear: Introduced a couple of new 'sign' vars to track +/-/0 until the end of the switch block where I can determine whether to compute the actual offset values.  I still left these as mutating x/y to keep line lengths reasonable and math semi-intelligible.

macos/FluentUI/Button/Button.swift
```diff
@@ -384,129 +384,101 @@ class ButtonCell: NSButtonCell {
...
-			height = titleSize.height
+			yOffsetSign = -1
 		case .imageAbove:
-			x = (rect.width - titleSize.width) / 2
-			y = (rect.height - titleSize.height + imageSize.height + titleToImageSpacing) / 2 + titleVerticalPositionAdjustment
-			width = titleSize.width
-			height = titleSize.height
+			yOffsetSign = 1
+		case .noImage, .imageOnly, .imageOverlaps:
+			// .imageOnly case is covered by guard return at top
```
anandrajeswaran, line 466: Can be preconditionFailure to enforce?

anandrajeswaran: Though does it really apply to the .noImage or imageOverlaps case?

anandrajeswaran: Similar question for imageRect

anandrajeswaran (approved):

macos/FluentUITestViewControllers/TestButtonViewController.swift
chripear: I was always arranging these popups so that the initial selection is also the first menu item, so these 'selectItem' calls were redundant.

macos/FluentUI/Button/Button.swift
chripear: Put cases back where they were previously and added preconditionFailure.

PR #368, opened 2021-01-05
Fix deprecation warning when using primary colors as fallbacks

zetasq:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

Use private values (`FallbackThemeColor.primary`) instead of deprecated public API (`Colors.primary`) in `Colors.primary(for:)` color API.

### Verification

There are no more compiler warnings regard using the deprecated `Colors.primary` API.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/368)

### DIFF
--- a/ios/FluentUI/Core/Colors.swift
+++ b/ios/FluentUI/Core/Colors.swift
@@ -281,64 +281,139 @@ public final class Colors: NSObject {
 
     /// Use these funcs to grab a color customized by a ColorProviding object for a specific window.. If no colorProvider exists for the window, falls back to deprecated singleton theme color
     @objc public static func primary(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryColor(for: window) ?? primary
+        return colorProvidersMap.object(forKey: window)?.primaryColor(for: window) ?? FallbackThemeColor.primary
     }
 
     @objc public static func primaryTint10(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryTint10Color(for: window) ?? primaryTint10
+        return colorProvidersMap.object(forKey: window)?.primaryTint10Color(for: window) ?? FallbackThemeColor.primaryTint10
     }
 
     @objc public static func primaryTint20(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryTint20Color(for: window) ?? primaryTint20
+        return colorProvidersMap.object(forKey: window)?.primaryTint20Color(for: window) ?? FallbackThemeColor.primaryTint20
     }
 
     @objc public static func primaryTint30(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryTint30Color(for: window) ?? primaryTint30
+        return colorProvidersMap.object(forKey: window)?.primaryTint30Color(for: window) ?? FallbackThemeColor.primaryTint30
     }
 
     @objc public static func primaryTint40(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryTint40Color(for: window) ?? primaryTint40
+        return colorProvidersMap.object(forKey: window)?.primaryTint40Color(for: window) ?? FallbackThemeColor.primaryTint40
     }
 
     @objc public static func primaryShade10(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryShade10Color(for: window) ?? primaryShade10
+        return colorProvidersMap.object(forKey: window)?.primaryShade10Color(for: window) ?? FallbackThemeColor.primaryShade10
     }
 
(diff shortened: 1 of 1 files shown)


harrieshin:
this is the right fix. There are clients who are still using MSFColors.primary etc for setting their branding colors. the intention of this code to make clients move away from the primary property and use the MSFColors primary for window instead. the fallback color to be communicationBlue is incorrect because different apps will have different app branding color

zetasq:
There is not any modification to the existing behavior. I just created move the usage of primary colors to the private scope so the warnings of using deprecated public API is gone.

zetasq:
Clients using `MSFColors.primary` will still get warnings as before. With this fix the compilation of `Colors.primary(for:)` (MSFColors primary for window) will get no more warnings (it now use private scoped colors now).

zetasq:
As you can see, `FallbackThemeColor.primary` is the same as the existing `Colors.primary`. The only difference is that `FallbackThemeColor.primary` is of `fileprivate` access scope.

harrieshin:
fallbacktheme colors won't necessarily have the same color as colors.primary. if client code set the primary color, there are places within sdk that calls colorsforwindow instead of getting the app branding color it will now fallback to communicationblue.

zetasq:
Thanks Harrie. Now all the `Colors.primary` colors will act as proxy and get/set colors from `FallbackThemeColor.primary` as single source.

ios/FluentUI/Core/Colors.swift
```diff
@@ -281,67 +281,150 @@ public final class Colors: NSObject {
...
     }
 
     @objc public static func primaryShade30(for window: UIWindow) -> UIColor {
-        return colorProvidersMap.object(forKey: window)?.primaryShade30Color(for: window) ?? primaryShade30
+        return colorProvidersMap.object(forKey: window)?.primaryShade30Color(for: window) ?? FallbackThemeColor.primaryShade30
     }
 
     private static var colorProvidersMap = NSMapTable<UIWindow, ColorProviding>(keyOptions: .weakMemory, valueOptions: .weakMemory)
 
+    fileprivate enum FallbackThemeColor {
```
rdeassis, line 317: Avoid using fileprivate. If you have Swiftlint installed, it should be giving you a warning for that.

Private over fileprivate Violation: Prefer `private` over `fileprivate` declarations. (private_over_fileprivate)

rdeassis, line 317: Why creating an enum these properties?
It's confusing to have a enum with no enumeration values.

zetasq: If I change `fileprivate` to `private` there will be compiler error like "'primaryShade30' is inaccessible due to 'private' protection level". Should we add special comment to silence Swiftlint warnings, or should we change `private` into `internal` instead (though no other parts of FluentUI should access these values)?

zetasq: It's used as an [uninhabitable type](https://nshipster.com/never/), which means this type act as only a namespace. Actually Apple uses this technique a lot in Swift repo (https://github.com/apple/swift/blob/a4230ab2ad37e37edc9ed86cd1510b7c016a769d/stdlib/public/core/Unicode.swift#L918)

zetasq: Just added some comment to make my intention clear :)

ios/FluentUI/Core/Colors.swift
```diff
@@ -281,67 +281,151 @@ public final class Colors: NSObject {
...
+    @objc public static var primaryShade30: UIColor {
+        get {
+            return FallbackThemeColor.primaryShade30
+        }
+        set {
+            FallbackThemeColor.primaryShade30 = newValue
+        }
+    }
 
     @available(*, deprecated, renamed: "textOnAccent")
```
harrieshin, line 420: why does this have to change?

zetasq: Since you mentioned earlier that clients may change `Colors.primary` which also affects fallback colors in `Colors.primary(for:)`, I just move the actual storage to `FallbackThemeColor.primay`. Now both `Colors.primary(for:)` and `Colors.primary` refer to the same fallback color.

harrieshin: github selected the wrong thing. I meant for foregroundprimary. it doesn't have to do with branding color

zetasq: Okay my intention is to keep the logic consistent (as with other colors). I can revert to the original logic for foregroundprimary.

ios/FluentUI/Core/Colors.swift
harrieshin: is it because you didn't remove the "private" on the variable itself?

zetasq: Yes! Preserve `private` only in type declaration solved this problem.

harrieshin (approved):

PR #371, opened 2021-01-06
Link custom color

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
FluentUI links have a fixed color value that is part of the `attributedTitle`, which gets clobbered by `mouseEntered` and `mouseExited` events (and therefore shouldn't be set directly).  This removes the color attribute and instead sets the `contentTintColor`, which can be overridden.  Combined with the existing `font` property, this allows us to customize the link text a bit more.

Also did some refactoring and a bugfix.

### Verification
Added a button to the test app with color and font customizations:
Screen Shot 2021-01-05 at 14 54 57

I also noticed an issue with the "Link with overridden Target/Action" disappearing when clicked.  This is fixed by setting the button type to `.momentaryChange` (thanks Anand):
| Disappearing Link Before | Disappearing Link Fixed |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/371)

### DIFF
--- a/macos/FluentUI/Link/Link.swift
+++ b/macos/FluentUI/Link/Link.swift
@@ -113,7 +113,7 @@ open class Link: NSButton {
 	}
 	
 	private func updateTitle() {
-		let titleAttributes = (showsUnderlineWhileMouseInside && mouseEntered) ? underlinedlinkAttributes: linkAttributes
+		let titleAttributes = (showsUnderlineWhileMouseInside && mouseEntered) ? underlinedLinkAttributes: linkAttributes
 		self.attributedTitle = NSAttributedString(string: title, attributes: titleAttributes)
 	}
 	
@@ -128,7 +128,7 @@ fileprivate let linkAttributes: [NSAttributedString.Key: Any] = [
 	.foregroundColor: NSColor.linkColor
 ]
 
-fileprivate let underlinedlinkAttributes: [NSAttributedString.Key: Any] = [
+fileprivate let underlinedLinkAttributes: [NSAttributedString.Key: Any] = [
 	.foregroundColor: NSColor.linkColor,
 	.underlineStyle: NSUnderlineStyle.single.rawValue
 ]
--- a/macos/FluentUI/Link/Link.swift
+++ b/macos/FluentUI/Link/Link.swift
@@ -47,6 +47,7 @@ open class Link: NSButton {
 		alignment = .natural
 		isBordered = false
 		contentTintColor = .linkColor
+		setButtonType(.momentaryChange)
 		target = self
 		action = #selector(linkClicked)
 		updateTitle()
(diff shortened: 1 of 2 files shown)


macos/FluentUI/Link/Link.swift
```diff
@@ -122,10 +131,9 @@ open class Link: NSButton {
 }
 
 fileprivate let linkAttributes: [NSAttributedString.Key: Any] = [
-	.foregroundColor: NSColor.linkColor
+	.underlineStyle: NSUnderlineStyle()
```
Saadnajmi, line 134: What is this? It just initializes the enum to 0 AKA UnderlineStyleNone? https://developer.apple.com/documentation/uikit/nsunderlinestyle/nsunderlinestylenone?language=objc

Saadnajmi (approved):

Saadnajmi:
How did you fix the link disappearing bug? Also nice catch on that bug

macos/FluentUI/Link/Link.swift
chripear: Right.  I didn't find a corresponding Swift `.none` or anything similar, but this did the trick.

chripear:
> How did you fix the link disappearing bug? Also nice catch on that bug

It's only partly fixed: the link still disappears while the modal dialog is up, but reappars as soon as it's dismissed.

The relevant commit has some more notes, but this improvement was made by adding the `mouseDown` override with `updateTitle()`, combined with setting the underline style back to none in `linkAttributes`.

macos/FluentUI/Link/Link.swift
```diff
@@ -93,24 +97,29 @@ open class Link: NSButton {
 		self.trackingArea = trackingArea
 	}
 	
+	open override func mouseDown(with event: NSEvent) {
+		super.mouseDown(with: event)
+		updateTitle()
```
anandrajeswaran, line 102: I may be misunderstanding something, but why do we need a call to updateTitle here? We're not changing anything in this mouseDown implementation nor does the implementation of updateTitle seem to be using anything that would be affected by super.mouseDown?

chripear: "Link with overridden Target/Action" (in the test app) disappears when clicked.  I wasn't able to figure out why that happened, but I can improve the behvior somewhat by fixing up the title as soon as the modal dialog is dismissed.  That's what this is for.

anandrajeswaran: What's the buttonType for this? Maybe needs to be .momentaryChange instead of the default?

chripear: Behavior is unchanged if I `setButtonType(.momentaryPushIn)`, so that is probably the default.  `.momentaryChange` is a better fix: link doesn't disappear at all now.

anandrajeswaran (approved):

PR #373, opened 2021-01-06
Mac button improvements

chiuam:
### Platforms Impacted
- [ ] iOS
- [X] macOS

A few fixes:

1. When button gets initialized with `Button()`, the app goes into an infinite loop. This is because it's calling itself instead of the  designated initializer. The potential fix is to call the designated initializer explicitly.

2. Previously we recalculate the image position based on the init value of title and image. While this works for buttons that have image and title initialized, it breaks the case when the button is init with no image/title. The fix is to recalculate the image position in the button cell.

3. After a discussion with design, we decided to have the button title and icon tint color to be consistent. Make image overridable so that isTemplate flag can be set. 

### Verification
Added a non template image icon.

|image|image|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/373)


macos/FluentUI/Button/Button.swift
```diff
@@ -133,6 +131,10 @@ open class Button: NSButton {
 
 	/// Image to display in the button.
 	override public var image: NSImage? {
+		didSet{
+			/// force image to render as a template image
```
harrieshin, line 135: i am not sure we should force it. shoudn't client decide?

chiuam: Following the Fluent design, the icon tint color should be the same as the title tint color.

harrieshin: but if the button has an avatar image for example, we shouldn't tint it? shouldn't the client try to pass in templated image if they want to be colored?

chiuam: Good point. I think the right fix is to expose the isTemplateImage flag to react native?

harrieshin: i do think react native should handle it because if eng want to put random stuff in button, it should jsut respect that. rest of the changes of this PR seems fine.

harrieshin: the other thing we could do something like  image?.renderingMode == .automatic ? image?.withRenderingMode(.alwaysTemplate) : image

macos/FluentUI/Button/Button.swift
```diff
@@ -64,7 +64,7 @@ open class Button: NSButton {
 
 	/// Initializes a Fluent UI Button with no title or image, and with default formatting
 	@objc public convenience init() {
-		self.init()
+		self.init(title: nil, image: nil, imagePosition: .imageLeading, format: ButtonFormat())
```
anandrajeswaran, line 67: Maybe set title to "" instead of nil here? Then you wouldn't need the default override below?

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -159,9 +159,15 @@ class TestButtonViewController: NSViewController, NSMenuDelegate {
...
+		let customButton = Button()
+		customButton.title = "Custom Button"
+		customButton.image = NSImage(named: TestButtonViewController.nonTemplateImage)!
+		customButton.accentColor = communicationBlue
+		customButton.size = .large
+		customButton.style = .primary
 		let buttonsWithTitleAndImageHorizontal: () -> [Button] = {
 			return [
-				Button(title: "Back", image: leadingArrowImage, imagePosition: .imageLeading, format: largePrimary),
+				customButton,
```
anandrajeswaran, line 170: Should we have a customButton for all configurations (image only, title only, title and image)?

macos/FluentUI/Button/Button.swift
Saadnajmi, line 67: I just wanna say lol

macos/FluentUI/Button/Button.swift
```diff
@@ -100,17 +100,15 @@ open class Button: NSButton {
...
 			self.title = title
-			if let image = image {
-				self.image = image
-				self.imagePosition = imagePosition
-			} else {
-				self.imagePosition = .noImage
-			}
-		} else if let image = image {
+		} else {
+			/// NSButton defaults title to "Button"
```
Saadnajmi, line 104: Nit. Three slashes is meant to be public documentation, while 2 is meant to be internal comments. I think this should be 2 slashes

macos/FluentUI/Button/Button.swift
chripear: `nil` makes an important distinction between a "button with a title", a "button with an image", and "a button with both a title and an image".

macos/FluentUITestViewControllers/TestButtonViewController.swift
chripear: The unique thing about this button is the non-template image (and the way it's constructed using the empty `Button()` init).  The other configurations are already covered in the test app.

macos/FluentUI/Button/Button.swift
```diff
@@ -398,6 +400,16 @@ class ButtonCell: NSButtonCell {
 		var width = imageRect.size.width
 		var height = imageRect.size.height
 
+		if title.count == 0 {
+			if imageRect.size != NSZeroRect.size {
+				imagePosition = .imageOnly
+			}
+		} else {
+			if imageRect.size == NSZeroRect.size {
```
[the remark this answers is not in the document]
chripear: then this section is simply
```
if title.count == 0 {
    imagePosition = .imageOnly
}
```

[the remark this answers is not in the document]
chripear: The `titleRect` method should have the converse: the `guard` clause should early-return `NSZeroRect` unless `title.count != 0`, and around this point in that method we should have
```
if imageSize == NSZeroSize {
    imagePosition = .noImage
}
```

macos/FluentUI/Button/Button.swift
chripear: I learned about this craziness today and had to work around it for Link.  A convenience initializer must call a designated initializer, right?  Why would it default to calling itself?

macos/FluentUITestViewControllers/TestButtonViewController.swift
chripear: i.e. instead of modifying one button in the top row, make a new row where all the buttons are constructed in this way

chripear (changes requested):

macos/FluentUITestViewControllers/TestButtonViewController.swift
chripear: I have some updates ready for the test app that will incorporate this suggestion.  Feel free to leave it as-is here and I will submit a follow-up PR.

macos/FluentUI/Button/Button.swift
chiuam: I reverted the template image changes and will let react native handle the custom tinting.

macos/FluentUI/Button/Button.swift
Saadnajmi: Oh I thought this was originally from me, my bad.  I'm not sure what the "convenience" keyword does, I jsut saw it on other code

chripear: [Designated Initializers and Convenience Initializers - The Swift Programming Language](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html#ID216)

macos/FluentUITestViewControllers/TestButtonViewController.swift
chiuam: So I should revert my changes in the test controller then?

chripear: You don't need to.  It's fine to leave them as they are.

chripear (approved):

macos/FluentUI/Button/Button.swift
chiuam: noted.

macos/FluentUI/Button/Button.swift
chiuam: Quoting from the swift doc:
Rule 1
A designated initializer must call a designated initializer from its immediate superclass.
Rule 2
A convenience initializer must call another initializer from the same class.
Rule 3
A convenience initializer must ultimately call a designated initializer.

I think the key part is Rule 2, where self.init() was allowed to call another convenience initializer (itself), but it never reached to the designated initializer.

Issue #365, opened 2020-12-31
[ListVnext] Replace leading view input type for broader customization


sophialee0416:
#### Platform
- [x] iOS
- [ ] macOS

#### Describe the feature that you would like added

LeadingView in ListVnext takes inputs as UIImage only. However, we will need to allow support for other components as well, ie. AvatarView. We should change this logic so that LeadingView can be more customizable.

#### What component or utility would this be added to

ListVnext

PR #378, opened 2021-01-07
[ListVnext] Replace custom LeadingView input type to UIView

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

List's custom LeadingView only accepted Images, but with this new change should be able to accept any UIView as its input. (https://github.com/microsoft/fluentui-apple/issues/365)

### Verification

Sample section was added in main List.

| Current                                       | Vnext                                      |
| Simulator Screen Shot - iPhone 12 Pro - 2021-01-06 at 20 55 18 | Simulator Screen Shot - iPhone 12 Pro - 2021-01-06 at 20 54 40 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/378)

### DIFF
--- a/ios/FluentUI/Table View/ListVnext.swift
+++ b/ios/FluentUI/Table View/ListVnext.swift
@@ -42,7 +42,7 @@ public enum MSFListAccessoryType: Int, CaseIterable {
 @objc(MSFListVnextCellData)
 public class MSFListVnextCellData: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
-    @objc @Published public var leadingIcon: UIView?
+    @objc @Published public var leadingView: UIView?
     @objc @Published public var title: String = ""
     @objc @Published public var subtitle: String?
     @objc @Published public var accessoryType: MSFListAccessoryType = .none
@@ -203,8 +203,8 @@ extension MSFListView {
         var body: some View {
             Button(action: cell.onTapAction ?? {}, label: {
                 HStack(spacing: 0) {
-                    if let leadingIcon = cell.leadingIcon {
-                        UIViewWrapper(leadingIcon)
+                    if let leadingView = cell.leadingView {
+                        UIViewWrapper(leadingView)
                             .frame(width: tokens.iconSize, height: tokens.iconSize)
                             .padding(.trailing, tokens.iconInterspace)
                     }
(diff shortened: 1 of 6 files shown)


ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -42,7 +42,7 @@ public enum MSFListAccessoryType: Int, CaseIterable {
 @objc(MSFListVnextCellData)
 public class MSFListVnextCellData: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
-    @objc @Published public var leadingIcon: UIImage?
+    @objc @Published public var leadingIcon: UIView?
```
rdeassis, line 45: I think it's worth calling it leadingView now that it's not only an icon.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -289,6 +288,25 @@ extension MSFListView {
     }
 }
 
+/// Generic UIView wrapper
+struct UIViewWrapper: UIViewRepresentable {
```
rdeassis, line 292: This seems generic enough to be in a separate file to be leveraged by other controls.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -204,8 +204,7 @@ extension MSFListView {
             Button(action: cell.onTapAction ?? {}, label: {
                 HStack(spacing: 0) {
                     if let leadingIcon = cell.leadingIcon {
-                        Image(uiImage: leadingIcon)
-                            .resizable()
+                        UIViewWrapper(leadingIcon)
```
rdeassis, line 207: What would be the story for the app that is using SwiftUI fully?
I thought initially about having a modifier for the MSFListCellView where you'd pass "some View" (the AvatarVnext) in this case and it would replace the icon there.

And I believe that doesn't dismiss the current approach of supporting an instance of UIView.
I'm thinking about a path for both consumers (UIKit and SwiftUI) to use it.

sophialee0416: How would this be translated when declaring the input type? I understand that we can assign/return a `some View` type in SwiftUI but I do not think it can be represented in ObjC?

sophialee0416: https://github.com/microsoft/fluentui-apple/issues/380

rdeassis (approved):

PR #384, opened 2021-01-09
Button Test Pane Improvements

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
Added an `imagePosition` pop-up menu to the button pane.  This makes the pane more compact, allows us to more clearly see the effect of each `imagePosition` without having to consult the test code, and exercises the cases where we override `imagePosition` in the `titleRect` and `imageRect` methods for buttons that are missing one or the other.

| Before | After |

### Verification
Testing exposed an issue where title-only buttons acquired an invisible pixel image whenever their `imagePosition` was set to `.imageOnly` or `.imageOverlaps`, which remains if the `imagePosition` is changed back to something else.  Added a check to detect and remove this pixel to avoid offsetting the title.

| Bug Version                                       | Fixed Version                                      |
| Bug | Fix |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/384)


macos/FluentUI/Button/Button.swift
```diff
@@ -607,3 +613,5 @@ fileprivate struct ButtonSizeParameters {
 		}
 	}
 }
+
+fileprivate let OnePointSize = NSSize(width: 1, height: 1)
```
rdeassis, line 617: Can you use private instead of fileprivate? swiftlint will complain about it with a warning.
Or even better: just create it directly in the scope that you're using it.

Saadnajmi: We should probably update [aka.ms/swiftguide](https://microsoft.github.io/swift-guide/FileContentOrdering.html) to reflect we don't like fileprivate anymore

Saadnajmi: Annnnnnd PR sent https://github.com/microsoft/swift-guide/pull/9

rdeassis: And I figured that the mac folder does not have swiftlint validation:
#390

Saadnajmi: Yeah, I was supposed to add that a while back for an FHLand I think I got hung up on sharing the yml / choosing SwfitLint over SwiftFormat. Thanks for adding it now!

chripear: Moved this line into a narrower scope and removed all other 'fileprivate' usage in Button.swift.

macos/FluentUI/Button/Button.swift
```diff
@@ -156,14 +156,14 @@ open class Button: NSButton {
 
 	private var linkedPrimaryOriginalStyle: ButtonStyle?
 
-	private var mouseDown: Bool = false {
+	public var isPressed: Bool = false {
```
anandrajeswaran, line 159: So a user of a button could change the isPressed value? Is that really a valid use case? Shouldn't that just be an internal detail?

macos/FluentUI/Button/Button.swift
```diff
@@ -366,20 +366,34 @@ open class Button: NSButton {
 // MARK: - ButtonCell
 
 class ButtonCell: NSButtonCell {
-	fileprivate var verticalPadding: CGFloat = 0
-	fileprivate var horizontalPadding: CGFloat = 0
-	fileprivate var titleVerticalPositionAdjustment: CGFloat = 0
-	fileprivate var titleToImageSpacing: CGFloat = 0
-	fileprivate var titleToImageVerticalSpacingAdjustment: CGFloat = 0
+	var verticalPadding: CGFloat = 0
```
anandrajeswaran, line 369: Should be private for these rather than inheriting value from ButtonCell (internal?)

macos/FluentUI/Button/Button.swift
```diff
@@ -366,20 +366,34 @@ open class Button: NSButton {
...
+	var horizontalPadding: CGFloat = 0
+	var titleVerticalPositionAdjustment: CGFloat = 0
+	var titleToImageSpacing: CGFloat = 0
+	var titleToImageVerticalSpacingAdjustment: CGFloat = 0
+
+	let OnePointSize = NSSize(width: 1, height: 1)
```
anandrajeswaran, line 375: minor: lowercase? `onePointSize`

macos/FluentUI/Button/Button.swift
```diff
@@ -366,20 +366,34 @@ open class Button: NSButton {
...
 		else {
-			return NSZeroRect
+			return .zero
+		}
+
+		// The test app exposed cases where a button with no image has its
+		// imagePosition set to .imageOnly or .imageOverlaps, then back to
+		// something else.  An invisible 1-pixel image gets assigned to the
+		// button, having an NSCustomImageRep with no drawing delegate.  Detect
+		// and remove this image so it doesn't offset the title.
```
anandrajeswaran, line 391: This feels like something we should file an Apple bug on? I think long term this seems like something that would break often as OS implementation changes so we should push to actually get the behavior fixed and maybe consider just not supporting that scenario if we don't expect it to be a common case so all usages of it don't get the weight/complexity they may not need

macos/FluentUI/Button/Button.swift
chripear: `Button` needs to set these values on its cell, so `private` is not an option.  Rafael said `fileprivate` is bad.  `internal` is the next least permissive option.

macos/FluentUI/Button/Button.swift
chripear: Moved this into the test app code.  Can follow up with a minimal test project and Apple bug.

macos/FluentUI/Button/Button.swift
Saadnajmi: I don't know if it's bad, I was just saying we should be consistent with our documentation if we decided not to like `fileprivate` anymore. Ironically, the comment in my PR to update documentation was "So then, what is a good example of `fileprivate`? Aaaannnnnnnnddd I think your scenario (A Button and a ButtonCell in the same file) is a perfect example 🙃

macos/FluentUI/Button/Button.swift
chripear: I want the test app to be able to put the buttons in a rest/pressed/disabled state so we can easily see each color set applied to all styles of button.  Previously I achieved this by sending mouseDown and mouseUp events to the button, but since the Action overhaul this no longer works, and I need to set the isPressed var directly.  The `internal` access level is not sufficient for this, and I couldn't figure out how to expose the property in a testable subclass.  Do you know of a better way?

It's not intended to be part of the Fluent button API, but it's also not the end of the world if a user sets this value (similar to setting isHighlighted).

macos/FluentUI/Button/Button.swift
chripear: @rdeassis what is the specific rule that prompted the original `fileprivate` commant?  Is it to prefer `private` over `fileprivate` in the outermost scope?  Is `fileprivate` fine in narrower scopes, such as the uses I had here?

macos/FluentUI/Button/Button.swift
Saadnajmi: Just curious, what happens if we manually set the `state` property? https://developer.apple.com/documentation/appkit/nsbutton/1528907-state

chripear: If I try to use `state` in place of `isPressed` then the highlight-stealing effect stops working for some reason.  Probably the system is doing something complicated. The NSButton `isHighlighted` property is similarly glitchy.

macos/FluentUI/Button/Button.swift
anandrajeswaran: That was my understanding - avoid using fileprivate where private is identical (outermost scope). I agree that this is a good case for fileprivate

macos/FluentUI/Button/Button.swift
anandrajeswaran: Probably worth checking the buttonType after using state. My guess is that implicitly changes its type to a type that supports on/off states? 

But I guess the question is whether the test app really needs to be able to toggle it. Can we instead count on someone changing it to actually go press the buttons for verification and avoid unnecessarily increasing the size of our production component (even if a tiny bit)? That would also ensure that our test app code actually set patterns for how we want this used by consumers. You'd be surprised how often I find our hacky test app code copied by well-meaning consumers

macos/FluentUI/Button/Button.swift
rdeassis: @chripear, below is the warning with the reference of samples of code that trigger it :

[Private over fileprivate Violation: Prefer `private` over `fileprivate` declarations. (private_over_fileprivate)](https://realm.github.io/SwiftLint/private_over_fileprivate.html)

https://realm.github.io/SwiftLint/private_over_fileprivate.html

macos/FluentUI/Button/Button.swift
chripear: It is a huge help for the test app to be able to toggle all of the buttons together to see the effects of color scheme changes on all the button variants, and capture these states in a screenshot.

PR #389, opened 2021-01-11
SideTabBar account button AX label

MathieuKavalec:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The SideTabBar's account button AX label is set to: "<account name> image".
In the FluentUI navigation bar, the account button's AX label is: "account profile button".
Let's be consistent and set the SideTabBar's account button label to the same value as the nav bar's account button.
Additionally, the AX trait for the button should be set to "button" instead of "image". It is fair to assume that anyone that will consume the SideTabBar will want the avatar view to be an interactive button.

### Verification

Run on device and make sure that VoiceOver announcements match expectations.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/389)

### DIFF
--- a/ios/FluentUI/People Picker/AvatarView.swift
+++ b/ios/FluentUI/People Picker/AvatarView.swift
@@ -345,6 +345,8 @@ open class AvatarView: UIView {
 
         super.init(frame: containerView.frame)
 
+        self.accessibilityTraits = .image
+
         addSubview(containerView)
     }
 
@@ -744,7 +746,6 @@ open class AvatarView: UIView {
     // MARK: Accessibility
 
     open override var isAccessibilityElement: Bool { get { return true } set { } }
-    open override var accessibilityTraits: UIAccessibilityTraits { get { return .image } set { } }
 
     open override var accessibilityLabel: String? {
         get {
--- a/ios/FluentUI/People Picker/AvatarView.swift
+++ b/ios/FluentUI/People Picker/AvatarView.swift
@@ -345,7 +345,7 @@ open class AvatarView: UIView {
 
         super.init(frame: containerView.frame)
 
-        self.accessibilityTraits = .image
+        accessibilityTraits = .image
 
         addSubview(containerView)
     }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/People Picker/AvatarView.swift
```diff
@@ -345,6 +345,8 @@ open class AvatarView: UIView {
 
         super.init(frame: containerView.frame)
 
+        self.accessibilityTraits = .image
```
harrieshin, line 348: do we need self here?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -39,6 +39,8 @@ open class SideTabBar: UIView {
             if let avatarView = avatarView {
                 avatarView.avatarSize = .medium
                 avatarView.translatesAutoresizingMaskIntoConstraints = false
+                avatarView.overrideAccessibilityLabel = "Accessibility.LargeTitle.ProfileView".localized
+                avatarView.accessibilityTraits = .button
```
harrieshin, line 43: should it be button trait only if the client is actually going to handle the tap on avatarview?

MathieuKavalec: I mentioned in the description that there's an assumption that the avatar view will always be used as a button. I wonder if a client would ever want a different behavior. 
Here are 3 options:
- Always assume that it will be a button. From a design perspective, this might make sense since we don't expect anyone to show the avatar view if it can't be interacted with.
- Only set the trait as button if the client provides a handler for the tap gesture recognizer. Otherwise, default to image.
- Never set the trait. Since the AvatarView is a public property, the clients can set the AX trait themselves.

What do you prefer?

rdeassis: I vote for option #2

"Only set the trait as button if the client provides a handler for the tap gesture recognizer. Otherwise, default to image."

MathieuKavalec: There is no way, in swift, to check if a specific optional method is implemented. The best I can do here is check if the delegate is nil or not which isn't great.

ios/FluentUI/People Picker/AvatarView.swift
MathieuKavalec: removed

harrieshin (approved):

rdeassis (approved):

PR #390, opened 2021-01-11
Adding swiftlint validation to Mac.

rdeassis:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

This changes enables swiftlint validation to the Mac Xcode project by adding the .swiftlint.yml file to the repo root folder and referencing it from the Xcode projects for iOS and Mac.

The path to lint is set through the **LINTPATH** environment variable set on the RunScript phase before the swiftlint is called.

Also updated all the files to resolve the warnings.

### Verification

Rebuilt the Mac project/demo app and ensured all swiftlint warnings were solved.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/390)


Saadnajmi (approved):

macos/FluentUITestApp/TestControlsViewController.swift
```diff
@@ -6,8 +6,8 @@
 import AppKit
 import FluentUITestViewControllers
 
-fileprivate struct Constants {
-	static let previousSelectionIndexUserDefaultsKey = "FluentUITestApp.previousSelectionIndexUserDefaultsKey"
+private struct Constants {
+	static let previousSelectionIndexUserDefaultsKey: String = "FluentUITestApp.previousSelectionIndexUserDefaultsKey"
```
Saadnajmi, line 10: Do we have a rule that says we prefer explicit types? I thought we preffered them being inferred

macos/xcode/.swiftlint.yml
```diff
@@ -0,0 +1,193 @@
+included: # paths to include during linting. `--path` is ignored if present.
```
Saadnajmi, line 1: Do we plan to share the same ios and macos yml file eventually?

macos/FluentUITestApp/TestControlsViewController.swift
rdeassis: Yes, this is the rule:
**Explicit Type Interface Violation: Properties should have a type interface (explicit_type_interface)**

I think the inferred ones you're referring to are the variables:
**Redundant Type Annotation Violation: Variables should not have redundant type annotation (redundant_type_annotation)**

Saadnajmi: I remember @markavitale preferred implicit to explicit when we started macOS to begin with, but obviously prefer following a common guide overall. Tagging anyway in case he has comments

macos/xcode/.swiftlint.yml
rdeassis: Yes. That makes sense. I moved it to the root of the folder and now the lint path is being passed through an environment variable.

Saadnajmi: Nice! :D

anandrajeswaran:
Wouldn't we prefer to use swift-format instead and use the first-party supported dependency?

rdeassis:
> Wouldn't we prefer to use swift-format instead and use the first-party supported dependency?

I took a bit of time to research the potential benefits of swift-format over swiftlint for the linting purposes that we need.

1. It seems like swiftlint is more mature [providing nearly 200 rules](https://realm.github.io/SwiftLint/rule-directory.html) that can be applied and configured to the swift code while [swift-format provides around 40](https://github.com/apple/swift-format/tree/main/Sources/SwiftFormatRules).

2. The way they are integrated into XCode is the same (by using run script build phases):

**SwiftLint:**
```
if which swiftlint >/dev/null; then
    swiftlint
else
    echo "warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint"
fi
```

**swift-format:**
```
if which swift-format >/dev/null; then
    swift-format -m format -i -r ${PROJECT_DIR}
else
    echo "warning: swift-format not installed"
fi
```

3. [Installing swiftlint](https://realm.github.io/SwiftLint/) is more convenient for developers/contributors with package managers while swift-format requires the developer [to clone the repo and build the tool locally](https://github.com/apple/swift-format#matching-swift-format-to-your-swift-version).

4. This may be a minor issue: Since we're already using swiftlint for the iOS projects, the current consumers of FluentUI would have a small disruption in their process to ensure their linting of swift files keeps working if we change it from swiftlint to swift-format.

I personally think it makes sense to remain with swiftlint for now (just adding support to cover the Mac files) and eventually entertain the possibility of a migration to swift-format when we have more time for a more thorough investigation on the benefits of adopting it.

Saadnajmi:
> > Wouldn't we prefer to use swift-format instead and use the first-party supported dependency?
> 
> I took a bit of time to research the potential benefits of swift-format over swiftlint for the linting purposes that we need.
> 
> 1. It seems like swiftlint is more mature [providing nearly 200 rules](https://realm.github.io/SwiftLint/rule-directory.html) that can be applied and configured to the swift code while [swift-format provides around 40](https://github.com/apple/swift-format/tree/main/Sources/SwiftFormatRules).
> 2. The way they are integrated into XCode is the same (by using run script build phases):
> 
> **SwiftLint:**
> 
> ```
> if which swiftlint >/dev/null; then
>     swiftlint
> else
>     echo "warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint"
> fi
> ```
> 
> **swift-format:**
> 
> ```
> if which swift-format >/dev/null; then
>     swift-format -m format -i -r ${PROJECT_DIR}
> else
>     echo "warning: swift-format not installed"
> fi
> ```
> 
> 1. [Installing swiftlint](https://realm.github.io/SwiftLint/) is more convenient for developers/contributors with package managers while swift-format requires the developer [to clone the repo and build the tool locally](https://github.com/apple/swift-format#matching-swift-format-to-your-swift-version).
> 2. This may be a minor issue: Since we're already using swiftlint for the iOS projects, the current consumers of FluentUI would have a small disruption in their process to ensure their linting of swift files keeps working if we change it from swiftlint to swift-format.
> 
> I personally think it makes sense to remain with swiftlint for now (just adding support to cover the Mac files) and eventually entertain the possibility of a migration to swift-format when we have more time for a more thorough investigation on the benefits of adopting it.
> 
> Thoughts?

I guess the initial question I would have is, how has SwiftFormat been going inside devmain? I know that @cy255 added support for that in devmain and it seems to have been going well?

I think switching from SwiftLint to SwiftFormat is a different task than just making sure both iOS and macOS have the same linter (which is what this PR does). Perhaps just file an issue to track separately?

rdeassis:
> > > Wouldn't we prefer to use swift-format instead and use the first-party supported dependency?
> > 
> > 
> > I took a bit of time to research the potential benefits of swift-format over swiftlint for the linting purposes that we need.
> > 
> > 1. It seems like swiftlint is more mature [providing nearly 200 rules](https://realm.github.io/SwiftLint/rule-directory.html) that can be applied and configured to the swift code while [swift-format provides around 40](https://github.com/apple/swift-format/tree/main/Sources/SwiftFormatRules).
> > 2. The way they are integrated into XCode is the same (by using run script build phases):
> > 
> > **SwiftLint:**
> > ```
> > if which swiftlint >/dev/null; then
> >     swiftlint
> > else
> >     echo "warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint"
> > fi
> > ```
> > 
> > 
> > **swift-format:**
> > ```
> > if which swift-format >/dev/null; then
> >     swift-format -m format -i -r ${PROJECT_DIR}
> > else
> >     echo "warning: swift-format not installed"
> > fi
> > ```
> > 
> > 
> > 
> > 1. [Installing swiftlint](https://realm.github.io/SwiftLint/) is more convenient for developers/contributors with package managers while swift-format requires the developer [to clone the repo and build the tool locally](https://github.com/apple/swift-format#matching-swift-format-to-your-swift-version).
> > 2. This may be a minor issue: Since we're already using swiftlint for the iOS projects, the current consumers of FluentUI would have a small disruption in their process to ensure their linting of swift files keeps working if we change it from swiftlint to swift-format.
> > 
> > I personally think it makes sense to remain with swiftlint for now (just adding support to cover the Mac files) and eventually entertain the possibility of a migration to swift-format when we have more time for a more thorough investigation on the benefits of adopting it.
> > Thoughts?
> 
> I guess the initial question I would have is, how has SwiftFormat been going inside devmain? I know that @cy255 added support for that in devmain and it seems to have been going well?
> 
> I think switching from SwiftLint to SwiftFormat is a different task than just making sure both iOS and macOS have the same linter (which is what this PR does). Perhaps just file an issue to track separately?

I have synced offline with @cy255  before posting my findings and it learned that swift-format in devmain is being used for formatting (not linting swift code integrated to Xcode). From what I heard from her, it seems swift-format is doing well to format code as an automated process.

And I agree with your point of eventually creating a separate issue to investigate the migration as I pointed out before.

cy255:
In the specific case of Fluent UI Mac, I think SwiftLint is a better option because it provides consistency within the repo. However, I do think we should migrate to swift-format eventually when swift-format becomes more mature.

harrieshin (approved):

Issue #356, opened 2020-12-29
Add gesture convenience API for drawer view


kubalani:
#### Platform
- [X] iOS
- [ ] macOS

#### Describe the feature that you would like added

hmmmm... I think the client app shouldn't need to worry about this.
Isn't there a way to just attach the drawerController to a specific view/viewcontroller and let it read the gestures and take the appropriate actions?

_Originally posted by @rdeassis in https://github.com/microsoft/fluentui-apple/pull/344#r547524809_

rdeassis:
The current DrawerController has a way to present itself through a gesture recognizer that is passed as a parameter from the host of the Drawer.

We should have a similar simple API that does not rely on the client app to have all the logic of presenting it.

Reference:

[@objc private func handlePresentingPan(gesture: UIPanGestureRecognizer)](https://github.com/microsoft/fluentui-apple/blob/7ec4cacc489f013b329952a616122a940da10583/ios/FluentUI/Drawer/DrawerController.swift#L642)

Issue #357, opened 2020-12-29
Add custom animation for DrawervNext


kubalani:
#### Platform
- [X] iOS
- [ ] macOS

#### Describe the feature that you would like added

Another issue you can file as a GitHub issue to be addressed after your first merge:

The current drawer presents the contents as you drag from the side instead of detecting the drag and presenting it entirely:

https://user-images.githubusercontent.com/68076145/102935613-585af200-445b-11eb-8a9c-99b48f8a66a7.mov

_Originally posted by @rdeassis in https://github.com/microsoft/fluentui-apple/pull/344#r547520295_

rdeassis:
This is closely related to #356.
Once we can support a gesture recognizer for the presentation, the feature requested in this issue will also be accomplished.

PR #395, opened 2021-01-13
[DrawerVnext] Adding support for custom pan gesture to show/hide drawer horizontal access

kubalani:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR is fix issue #356  & #357. It accepts pan gesture input from the client and monitors it via publish object to mimic FluentUI native component of DrawerController. 

https://user-images.githubusercontent.com/63682282/104423454-fa3d9e00-5532-11eb-990d-c5a446643499.mov

Updated Video with dimmed background

https://user-images.githubusercontent.com/63682282/104764393-32ec9b80-571c-11eb-811b-3aa12b7f38db.mov

Since the `UIPercentDrivenInteractiveTransition` is not supported in swiftUI. The current PR mimics percent driven transition as a drop in replacement for UIKIt functionality. 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/395)


ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -157,20 +207,36 @@ public struct Drawer<Content: View>: View {
     private func dragGesture(screenWidth: CGFloat) -> some Gesture {
         DragGesture()
             .onChanged { value in
-                let withinDragBounds = state.presentationDirection == .left ? value.translation.width < 0 : value.translation.width > 0
-                if withinDragBounds {
-                    horizontalDragOffset = value.translation.width
-                }
+                let velocity = value.translation.width
```
rdeassis, line 210: I found this bug on the video below.
If you add logic like the snippet below to skip the updateTransition it'll fix it:

```
                if (slideOutDirection == .right && value.startLocation.x > value.location.x) ||
                    (slideOutDirection == .left && value.startLocation.x < value.location.x) {
                    return
                }
```


https://user-images.githubusercontent.com/68076145/104497752-add47b80-558f-11eb-9e09-bc7b4f1ffe2a.mov

ios/FluentUI/DrawerVnext/Drawer.swift
```diff
@@ -146,6 +188,14 @@ public struct Drawer<Content: View>: View {
         }
     }
 
+    public func isPresentationGestureActive() -> Bool {
+        return state.presentingGesture != nil && (state.presentingGesture?.state == .none || state.presentingGesture?.state == .began)
+    }
+
+    private func defaultAnimation() -> Animation {
```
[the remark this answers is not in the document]
kubalani: In addition, I will upgrade this to a computed property as well

ios/FluentUI/DrawerVnext/Drawer.swift
kubalani: thanks for catching this, I have update the code with the fix

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -62,33 +72,62 @@ struct SlideOverPanel<Content: View>: View {
...
         }
-        .background(isOpen ? tokens.backgroundDimmedColor.opacity(backgroundLayerOpacity) : tokens.backgroundClearColor)
+        .background(transitionState == .collapsed ? tokens.backgroundClearColor : tokens.backgroundDimmedColor.opacity(backgroundLayerOpacity))
+    }
+
+    public func contentWidth() -> CGFloat {
+        return slideOutPanelWidth * contentWidthSizeRatio
+    }
+
+    private func resolvedShadowOpacity() -> Double {
```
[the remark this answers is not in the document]
kubalani: thanks for catching this, completely slipped off my mind. This is now fixed

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -62,33 +72,62 @@ struct SlideOverPanel<Content: View>: View {
                     .onTapGesture (perform: actionOnBackgroundTap ?? {})
             }
         }
-        .background(isOpen ? tokens.backgroundDimmedColor.opacity(backgroundLayerOpacity) : tokens.backgroundClearColor)
+        .background(transitionState == .collapsed ? tokens.backgroundClearColor : tokens.backgroundDimmedColor.opacity(backgroundLayerOpacity))
+    }
+
+    public func contentWidth() -> CGFloat {
```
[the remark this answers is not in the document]
kubalani: I am promoting these to read-only properties, the only reason have used private func to discourage adding setter in future and might be source of multiple truths that can create additional bugs

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -62,33 +72,62 @@ struct SlideOverPanel<Content: View>: View {
...
         if direction == .left {
             return -offset
         } else {
             return offset
         }
     }
 
-    private func expandedConentOffset() -> CGFloat {
-        return CGSize.zero.width
+    private func percentTransistionOffset() -> CGFloat {
```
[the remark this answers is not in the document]
kubalani: yes, I will promote private func to read only computed properties across the module

ios/FluentUI/Core/SwiftUIAdapter.swift
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+public struct SwiftUIAdapter {
```
rdeassis, line 9: The goal here is to create a namespace, right? Shouldn't this be just an enum instead?

ios/FluentUI/DrawerVnext/Drawer+UIKit.swift
```diff
@@ -77,15 +79,19 @@ open class DrawerVnext: UIHostingController<AnyView>, FluentUIWindowProvider {
...
+    /// Represents the drawer's state with properties to configure its behavior.
+    /// @see `DrawerState`
     @objc open var state: DrawerState {
         return self.drawer.state
     }
 
     @objc public init(contentViewController: UIViewController,
                       theme: FluentUIStyle? = nil) {
-        let drawer = Drawer(content: DrawerContentViewController(contentViewController: contentViewController))
+        let drawer = Drawer(content: SwiftUIAdapter.UIViewControllerAdapter(contentViewController))
```
rdeassis, line 94: Lots of "Adapter" in this call. How about removing the namespace and have the swift file called UIKitAdapter**s**.swift (plural)? (It's probably better than SwiftUIAdapters since we have adapters for UIKit types)

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -50,46 +103,38 @@ struct SlideOverPanel<Content: View>: View {
...
     }
 
-    private func resolvedContentOffset() -> CGFloat {
-        // override offset if required
-        if let preferredContentWidth = preferredContentOffset {
-            // parent view wants to take over primarily to conform user drag gesture
-            return preferredContentWidth
+    private func resolveTransitionOpacity(_ maxValue: Double) -> Double {
+        let clearBackgroundOpacity = 0.0
+        guard backgroundDimmed else {
```
rdeassis, line 123: This is always setting the shadow opacity to 0.
This check shouldn't be in this func because it only applies to the background.

How about creating a computed property to return the background color? You can check the backgroundDimmed and transition state in there and return the color with the opacity returned by this func.

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -112,24 +76,66 @@ struct SlideOverPanel<Content: View>: View {
...
+    private var expandedContentOffset: CGFloat {
         return CGSize.zero.width
     }
 
-    private func collapsedContentOffset() -> CGFloat {
+    private var collapsedContentOffset: CGFloat {
         return slideOutPanelWidth * contentWidthSizeRatio
     }
+
+    public var body: some View {
```
rdeassis, line 98: The file content order is public -> internal -> private according to the [swift-guide](https://github.com/microsoft/swift-guide/blob/gh-pages/FileContentOrdering.md).

ios/FluentUI/Core/SwiftUIAdapter.swift
kubalani: Yes, the basic idea was to use namespace, I don't have any specific preference with enum vs struct for namespace.

ios/FluentUI/DrawerVnext/Drawer+UIKit.swift
kubalani: I assume the intent of using a separate file is to create a module. File with orphan class or loosely based modules usually encourage generic programming or doesn't provide clear directions for extensions.

Naming convention could be adjusted We can have any of the following

- `SwiftUIAdapter.UIView` 
-  'SwiftUIAdapter.UIViewController` 
-  'UIKitAdapter.UIViewAdapter` 
-  'UIKitAdapter.UIViewControllerAdapter` 

I like something that clearly articulates for a UIKit client that this is to be used when integrating with SwiftUI only. 
I am open to any of the above options, thoughts??

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: thanks for catching this, seems like I am almost numb to recurring issues, I will add some snapshot tests right after blocking issues are resolved

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: marked as internal to conform to `View` protocol

ios/FluentUI/DrawerVnext/Drawer+UIKit.swift
rdeassis: The idea behind moving the code to a separate file is organizing the code.

One of the benefits of organizing them like this is: when we're publishing subspecs on cocoapods we are preventing another control that uses these adapters from depending on the Drawer or List.

Since these files contain code that can be used by multiple controls, it's a good idea to move them to their own file.

Maybe the file could be named **UIKit+SwiftUI_interoperability.swift** and we have the times named as they are with no namespace? (**UIViewAdapter** and **UIViewControllerAdapter**)?

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -115,26 +149,7 @@ struct SlideOverPanel<Content: View>: View {
...
-            return clearBackgroundOpacity
-        case .expanded:
-            return maxValue
-        case .inTransisiton:
-            if let percent = percentTransition {
-                return maxValue * percent
-            }
-        }
-        return 0
+        .background(transitionState == .collapsed ? tokens.backgroundClearColor : tokens.backgroundDimmedColor.opacity(resolvedBackgroundOpacity))
```
rdeassis, line 152: This logic could be moved to the **backgroundColor** computed property:

`transitionState == .collapsed ? tokens.backgroundClearColor : tokens.backgroundDimmedColor.opacity(resolvedBackgroundOpacity)`

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -76,6 +76,40 @@ struct SlideOverPanel<Content: View>: View {
         }
     }
 
+    private var resolvedBackgroundOpacity: Double {
```
rdeassis, line 79: You can return the Colors directly in those computed properties:

- backgroundColor
- shadowColor

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -95,7 +129,7 @@ struct SlideOverPanel<Content: View>: View {
         return slideOutPanelWidth * contentWidthSizeRatio
     }
 
-    public var body: some View {
+    var body: some View {
```
rdeassis, line 132: nit: this is internal by default if you don't have a specifier, then it should be moved up so the private props are the last thing in the file? (Public -> Internal -> Private)

harrieshin:
is the video outdated? because when you slowly drag the drawercontroller to show the animation should be in sync with shadow to slowly get "darker"?

kubalani:
> is the video outdated? because when you slowly drag the drawercontroller to show the animation should be in sync with shadow to slowly get "darker"?

I have updated the video in the description section

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
```diff
@@ -76,6 +76,40 @@ struct SlideOverPanel<Content: View>: View {
         }
     }
 
+    private var resolvedBackgroundOpacity: Double {
+        let clearBackgroundOpacity = 0.0
```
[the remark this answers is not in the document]
kubalani: thanks for catching this, yes it seems to be redundant code, I have added a new property for `resolvedBackgroundColor` that returns clear color

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: I have added two separate properties instead of parametrizing them because they have a slightly different behavior.

ios/FluentUI/DrawerVnext/SlideOverPanel.swift
kubalani: thanks for catching this, I have updated the ordering across the module

ios/FluentUI/Vnext/Drawer/SideOverPanel+Modifiers.swift
```diff
@@ -6,73 +6,73 @@
...
-        return MSFSlideOverPanel(slideOutPanelWidth: width,
-                                 actionOnBackgroundTap: actionOnBackgroundTap,
-                                 content: content,
-                                 backgroundLayerOpacity: backgroundLayerOpacity,
-                                 direction: direction,
-                                 isOpen: $isOpen,
-                                 preferredContentOffset: $preferredContentOffset,
-                                 tokens: tokens)
+        return MSFSlideOverPanel(percentTransition: $percentTransition,
+                              tokens: tokens,
```
rdeassis, line 14: I think the merge messed up with the indentation of the other params.

ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
```diff
@@ -12,84 +12,137 @@ import SwiftUI
...
             if direction == .right {
                 MSFInteractiveSpacer(defaultBackgroundColor: $tokens.backgroundClearColor)
                     .onTapGesture (perform: actionOnBackgroundTap ?? {})
             }
 
             content
-                .frame(width: contentWidth())
-                .shadow(color: tokens.shadowColor.opacity(isOpen ? tokens.shadowOpacity : 0),
+                .frame(width: contentWidth)
+                .shadow(color: resolvedShadowColor,
```
rdeassis, line 62: nit: can we call these **shadowColor** and **backgroundColor**(line73)?

ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
```diff
@@ -12,84 +12,137 @@ import SwiftUI
...
     }
 
-    private func collapsedContentOffset() -> CGFloat {
+    private var collapsedContentOffset: CGFloat {
         return slideOutPanelWidth * contentWidthSizeRatio
     }
 
-    private func contentWidth() -> CGFloat {
-        return slideOutPanelWidth * contentWidthSizeRatio
+    private func resolvedPanelStateValue(_ clear: Double, _ opaque: Double) -> Double {
```
rdeassis, line 134: I like the idea of having a generic function to calculate values based on the percentage, but the parameters should be consistent on the generic name. So, can you rename the parameters from clear to **minValue** and opaque to **maxValue**?

ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
```diff
@@ -111,14 +164,15 @@ struct SlideOverPanelLeft_Previews: PreviewProvider {
...
-            MSFSlideOverPanel<MockContent>(slideOutPanelWidth: UIScreen.main.bounds.width,
-                                           actionOnBackgroundTap: nil,
-                                           content: MockContent(),
-                                           backgroundLayerOpacity: 0.5,
-                                           direction: .left,
-                                           isOpen: Binding.constant(true),
-                                           preferredContentOffset: Binding.constant(nil),
-                                           tokens: MSFDrawerTokens())
+            MSFSlideOverPanel<MockContent>(
+                percentTransition: Binding.constant(0),
```
rdeassis, line 168: nit: move the first parameter to the line above and indent accordingly? (here and below too)

rdeassis (approved):
Just some nit comments. Thanks for addressing all the previous comments!

PR #398, opened 2021-01-15
[ListVnext] Adding Expand/Collapse feature

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

List can now take in children cells assigned to a parent cell. This nested list structure can be iterated over multiple levels. Clicking on a cell will expand its children cell contents if defined.

### Verification

Added sample data:

| Swift | Obj-C Demo |
|----|-----|
| https://user-images.githubusercontent.com/22566866/105097453-352d5d80-5a6e-11eb-95dc-f7e7adea6821.mov | https://user-images.githubusercontent.com/22566866/104665732-bc0dbf00-5697-11eb-842c-4cefb590204b.mov |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/398)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ListVnextDemoController.swift
```diff
@@ -29,6 +29,21 @@ class ListVnextDemoController: DemoController {
...
+        for index in 0...2 {
+            listCell = MSFListVnextCellData()
+            listCell.title = "Child #\(index)"
+            children.append(listCell)
+        }
+        children[0].children = subchildren
+
         var avatar: AvatarVnext
 
         /// AvatarView section
```
rdeassis, line 49: How about naming the section "Custom leading view and collapsible sub cells"?

Then you add 2 personas as top level items with a 2 line cell (name on the title and "Tap to show reports" as the subtitle)

Then you can organize them like this (maybe even using more of the personas samples):

- Kat Larrson
  - Ashley McCarthy
  - Allan Munger
    - Amanda Brady
    - Kristin Patterson
    - Persona 1
     
- Persona 2
  - Persona 3
    - Persona 4
    - Persona 5
    - Persona 6
    - Persona 7

And if you implement the section as a button (other comment below), we can probably make the entire section collapsible too (initially appearing expanded).

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -223,7 +225,13 @@ extension MSFListView {
         }
 
         var body: some View {
-            Button(action: cell.onTapAction ?? {}, label: {
+            Button(action: cell.onTapAction ?? {
```
rdeassis, line 228: I think the functionality of having a cell collapse or expand its subcells is great. But what about the header?
I was thinking the header could be a button with a different style (that looks like a header) and the section can be configured as collapsible?

What did you have in mind for those scenarios?

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -214,6 +215,7 @@ extension MSFListView {
     struct MSFListCellView: View {
         var cell: MSFListVnextCellData
         var layoutType: MSFListCellVnextLayoutType
+        @State private var isExpanded: Bool = false
```
rdeassis, line 218: Is the calling code able to configure the initial value of a cell's sub items?
For example, the Left Nav wants to have all the users listed when it loads, but still with the capability of collapsing them if the header is tapped.

ios/FluentUI/Table View/ListVnext.swift
```diff
@@ -261,6 +269,15 @@ extension MSFListView {
                 }
             })
             .buttonStyle(ListCellButtonStyle(tokens: tokens, layoutType: layoutType))
+            if let children = cell.children, isExpanded == true {
+                ForEach(children, id: \.self) { child in
```
rdeassis, line 273: I noticed these dividers for the sub items and I think this is probably not the way it should look.
Can you check with Victor on that and hide them while you don't have clear design guidelines? (creating a GitHub issue if the final look won't be merged as part of this PR as well)

Screen Shot 2021-01-14 at 6 43 53 PM

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: I was thinking anything related to the Header/Footer would be worked on as a separate component entirely (treated similarly to current arch). I envision there being different types of Headers/Footers, including an expand/collapse type. I think that could be treated as a separate issue, and still coexist with individual expand/collapse cells. Thoughts?

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: Lack of modifiers + Apple fixing UITableView accessor -- created an iOS 14 check and changed to a ScrollView, otherwise uses a List with the UITableView modifier.

ios/FluentUI/Table View/ListVnext.swift
rdeassis: Yes! As we agreed offline, the Header/Footer provide the functionality of sticking at the top while the contents are being scrolled, so we can keep it as a separate issue.
The cells being collapsed and expanded look great!
Is there a GitHub issue for the Header expand/collapse functionality?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -106,16 +143,24 @@ extension MSFListView {
...
             self.tokens = tokens
+            _isExpanded = State(initialValue: cell.isExpanded)
         }
 
         var body: some View {
-            Button(action: cell.onTapAction ?? {}, label: {
+            Button(action: cell.onTapAction ?? {
+                if cell.children != nil {
+                    withAnimation {
+                        isExpanded.toggle()
```
rdeassis, line 160: This should change the underlying state of the cell to make sure the expanded/collapsed states remain intact.

I the scenario below, the "Subchild #" nodes should remain expanded when I expand Kat Larrson again.


https://user-images.githubusercontent.com/68076145/105077382-4e6be500-5a41-11eb-9d41-a46568d9ca46.mov

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -153,6 +198,15 @@ extension MSFListView {
                 }
             })
             .buttonStyle(ListCellButtonStyle(tokens: tokens, layoutType: layoutType))
+            if let children = cell.children, isExpanded == true {
+                ForEach(children, id: \.self) { child in
+                    MSFListCellView(cell: child,
```
rdeassis, line 203: I think the MSFListCellView should have a reference to its object data (MSFListVnextCellData) to be able to update its layout when any of its properties change.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -153,6 +198,15 @@ extension MSFListView {
                 }
             })
             .buttonStyle(ListCellButtonStyle(tokens: tokens, layoutType: layoutType))
+            if let children = cell.children, isExpanded == true {
```
rdeassis, line 201: If you use the isExpanded property from the MSFListVnextCellData, you probably don't need the @Sate isExpanded var in this View.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -106,16 +143,24 @@ extension MSFListView {
     struct MSFListCellView: View {
         var cell: MSFListVnextCellData
         var layoutType: MSFListCellVnextLayoutType
+        @State var isExpanded: Bool
```
rdeassis, line 146: I think this should be @ObservedObjcect var state: MSFListVnextCellData as you need all of the cell behavior in sync with the object data.

ios/FluentUI/Table View/ListVnext.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/issues/402

rdeassis (commented):
Since we discussed the other changes offline (divider, removing the list scenario used for iOS 13, etc...), I can take another look once you update your branch.
Thanks for addressing those comments!

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -53,66 +55,64 @@ public struct MSFListView: View {
...
             }
-        }
     }
 }
 
 extension MSFListView {
     /// View for List Cells
     struct MSFListCellView: View {
-        var cell: MSFListCellData
+        var cell: MSFListCellState
```
rdeassis, line 98: `@ObservedObject var state: MSFListCellState`

The collapse/expand seems to be broken as of the last iteration because this property is not using the @observedobject wrapper.

ios/FluentUI/Controls/ScrollView.swift
```diff
@@ -12,12 +12,12 @@ protocol ScrollableContainerView: AnyObject {
 }
 
 @available(*, deprecated, renamed: "ScrollView")
-public typealias MSScrollView = ScrollView
+public typealias MSScrollView = MSFScrollView
```
rdeassis, line 15: Did you change this because it was conflicting with the SwiftUI ScrollView?

I think we can remove lines 14 and 15 since the client code won't be referring to type "MSFScrollView".

ios/FluentUI/Controls/ScrollView.swift
```diff
@@ -12,12 +12,12 @@ protocol ScrollableContainerView: AnyObject {
...
 
 /// `UIScrollView` subclass that automatically adjusts content insets when keyboard is shown/hidden to make sure scrollable area is always visible. Also provides methods to scroll any subview (or first responder if it's a subview) to visible area.
 @available(iOSApplicationExtension, unavailable)
 @objc(MSFScrollView)
-open class ScrollView: UIScrollView, ScrollableContainerView {
+open class MSFScrollView: UIScrollView, ScrollableContainerView {
```
rdeassis, line 20: `@objc open class MSFScrollView: UIScrollView, ScrollableContainerView {`

Use just the **@objc** in front of the declaration since the name of the class and the objC are the same?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -53,66 +55,64 @@ public struct MSFListView: View {
...
-                        .frame(maxWidth: .infinity)
+                        ForEach(section.cells.indices, id: \.self) { index in
+                            let cell = section.cells[index]
+                            let hasDividers = (index < section.cells.count - 1 && section.hasDividers) || (section.hasDividers && (cell.children != nil))
+                            MSFListCellView(cell: cell,
+                                            tokens: tokens,
+                                            hasDividers: hasDividers)
+                                .frame(maxWidth: .infinity)
+                        }
+                        Divider()
```
rdeassis, line 73: Should this also respect the hasDividers property?

Screen Shot 2021-01-25 at 11 44 28 AM

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -53,66 +55,64 @@ public struct MSFListView: View {
...
     }
 }
 
 extension MSFListView {
     /// View for List Cells
     struct MSFListCellView: View {
-        var cell: MSFListCellData
+        var cell: MSFListCellState
         @ObservedObject var tokens: MSFListTokens
+        var hasDividers: Bool
```
rdeassis, line 100: I think this should be part of the cellState, shouldn't it?

ios/FluentUI/Controls/ScrollView.swift
sophialee0416: I think The ScrollView is used in the main DemoController, would it still be okay to remove?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -150,6 +150,19 @@ extension MSFListView {
                 }
             })
             .buttonStyle(ListCellButtonStyle(tokens: tokens, layoutType: cell.layoutType))
+            if hasDividers {
+                Divider()
+                    .padding(.leading, (tokens.horizontalCellPadding + tokens.iconSize + tokens.iconInterspace))
```
rdeassis, line 155: This should consider the icon width and interspace only if it's actually there, right?
Otherwise this will happen with the dividers.

Screen Shot 2021-01-25 at 11 57 37 AM

ios/FluentUI/Vnext/List/List.swift
sophialee0416: I think this is a section property because it should automatically applied per section (on top of including the section dividers). Thoughts?

ios/FluentUI/Vnext/List/List.swift
sophialee0416: Adjusted

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -95,7 +97,7 @@ public struct MSFListView: View {
 extension MSFListView {
     /// View for List Cells
     struct MSFListCellView: View {
-        var cell: MSFListCellState
+        @ObservedObject var cell: MSFListCellState
```
rdeassis, line 100: change the variable name to state?

ios/FluentUI/Vnext/List/List.swift
rdeassis: I agree that it's very likely that the divider will be set on the section level rather than on the cell level.
But I think it'd be nice to have the level of granularity on the cell.
We can defer this decision and check with design whether we'll have that kind of scenario (specific cells not having dividers while the section is set otherwise).

rdeassis (approved):

Issue #77, opened 2020-05-27
UIScreen.main issue


li-bei:
I saw a lot of `UIScreen.main` in the codebase. As a UI framework, I don't think this is a good implementation.

iOS devices can use external screens, it means your view will not alway in the main screen, we should get screen form window:

```swift
extension UIView {
   var screen: UIScreen { window?.screen ?? .main }
}
```

And screen should be accessed from methods like `layoutSubviews`, `sizeThatFits`. For example, `Separator` should access screen in `sizeThatFits`:

```swift
class Separator: UIView {
  ...

  override func sizeThatFits(_ size: CGSize) -> CGSize {
    switch orientation {
      case .horizontal:
        return CGSize(width: size.width, height: screen.devicePixel)
      case .vertical:
        return CGSize(width: screen.devicePixel, height: size.height)
    }
  }
```

### Environment Information

- **Platform**:
  - [x] iOS
  - [ ] macOS

### Priorities and help requested:

Are you willing to submit a PR to fix? Neutral

Requested priority: Normal

markavitale:
@harrieshin can you take a look at this and help triage the issue?

PR #400, opened 2021-01-15
Make several internal extensions public and bring back UIFontDescriptor's extensions

zetasq:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

1. Make `UIFont` extension methods public again.
2. Make `UITableViewCell` extension method (`hideSystemSeparator()`) public again.
3. Bring back the `UIFontDescriptor` extension methods.

With these changes it would be much easier for Outlook Mobile iOS to migrate from `OfficeUIFabric`(1.1.2) to `FluentUI`. And these functionality exposed could also be used by other consumers.

### Verification

N/A

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/400)

### DIFF
--- a/ios/FluentUI/Extensions/UIFontDescriptor+Extension.swift
+++ b/ios/FluentUI/Extensions/UIFontDescriptor+Extension.swift
@@ -1,0 +1,27 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+public extension UIFontDescriptor {
+    var traits: [UIFontDescriptor.TraitKey: Any]? {
+        return object(forKey: .traits) as? [UIFontDescriptor.TraitKey: Any]
+    }
+
+    var weight: UIFont.Weight {
+        if let weight = traits?[.weight] as? NSNumber {
+            return UIFont.Weight(CGFloat(weight.floatValue))
+        }
+        return .regular
+    }
+
+    func withWeight(_ weight: UIFont.Weight) -> UIFontDescriptor {
+        var attributes = fontAttributes
+        var traits = self.traits ?? [:]
+        traits[.weight] = NSNumber(value: Float(weight.rawValue))
+        attributes[.traits] = traits
+        return UIFontDescriptor(fontAttributes: attributes)
+    }
+}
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Extensions/UITableViewCell+Extension.swift
```diff
@@ -5,7 +5,7 @@
 
 import UIKit
 
-extension UITableViewCell {
+public extension UITableViewCell {
```
rdeassis, line 8: In what objects this extension will be used? Plain UITableViewCell instances?
 The functionality should be offered through the MSFTableViewCell (TableViewCell) internally.

ios/FluentUI/Extensions/UIFontDescriptor+Extension.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+public extension UIFontDescriptor {
```
rdeassis, line 8: These seem to be just helper functions that don't necessarily align with any FluentUI standard.
Why should FluentUI be the best place to host this code?

rdeassis (changes requested):
Hi @zetasq.
I'd like to understand more about the context/reasons to add this code into FluentUI.
We can sync offline about details.

ios/FluentUI/Extensions/UITableViewCell+Extension.swift
```diff
@@ -5,7 +5,7 @@
...
     func hideSystemSeparator() {
         separatorInset = UIEdgeInsets(top: 0, left: max(UIScreen.main.bounds.width, UIScreen.main.bounds.height), bottom: 0, right: 0)
```
sophialee0416, line 10: UIScreen usage is not encouraged (https://github.com/microsoft/fluentui-apple/issues/77)

zetasq:
After offline sync it seems better to keep these helper functions from public exposure (client could depend more on controls instead of helper function logic).

PR #403, opened 2021-01-19
Set imagePosition for title-only and image-only init methods

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
Updated convenience initializers for title-only and image-only buttons to pass appropriate imagePosition values to the designated initializer, so that we don't have to rely on the automatic content-based detection in the `titleRect` and `imageRect` methods of `ButtonCell`.

### Verification
Used nuget publish script to bring changes into setupui, and tested via SetupUITestApp.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/403)

### DIFF
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -23,7 +23,7 @@ open class Button: NSButton {
 		accentColor: NSColor = Colors.primary
 	) {
 		let format = ButtonFormat(size: size, style: style, accentColor: accentColor)
-		self.init(title: title, format: format)
+		self.init(title: title, imagePosition: .noImage, format: format)
 	}
 
 	/// Initializes a Fluent UI Button with an image only
@@ -39,7 +39,7 @@ open class Button: NSButton {
 		accentColor: NSColor = Colors.primary
 	) {
 		let format = ButtonFormat(size: size, style: style, accentColor: accentColor)
-		self.init(image: image, format: format)
+		self.init(image: image, imagePosition: .imageOnly, format: format)
 	}
 
 	/// Initializes a Fluent UI Button with a title and image


macos/FluentUI/Button/Button.swift
```diff
@@ -23,7 +23,7 @@ open class Button: NSButton {
 		accentColor: NSColor = Colors.primary
 	) {
 		let format = ButtonFormat(size: size, style: style, accentColor: accentColor)
-		self.init(title: title, format: format)
+		self.init(title: title, imagePosition: .noImage, format: format)
```
harrieshin, line 26:
```
	public init(
		title: String = "",				title: String = "",
		image: NSImage? = nil,				image: NSImage? = nil,
		imagePosition: NSControl.ImagePosition = .imageLeading,				imagePosition: NSControl.ImagePosition = .imageLeading,
		format: ButtonFormat = ButtonFormat()				format: ButtonFormat = ButtonFormat()
	) {			) {
```
Since the default param for this function image is nil should the impagePostion default be .noImage?

chripear:
Correction: The bug I was seeing in Office where Fluent buttons had offset titles was actually resolved by an earlier PR (#384).  I confirmed this by running the publish script on the git master with no changes and testing the effects in SetupUITestApp.  So this change is not fixing a bug, but simply setting some more appropriate defaults for title-only and image-only buttons.

chripear:
Correction: The bug I was seeing in Office where Fluent buttons had offset titles was actually resolved by an earlier PR (#384). I confirmed this by running the publish script on the git master with no changes and testing the effects in SetupUITestApp.

PR #406, opened 2021-01-22
Removing the 'Vnext' suffix from MSFButtonVnext and associated types.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Before making the Vnext types available for client apps, we need to remove the 'Vnext' suffix from the type names.
This change renames the Button Vnext and associated types to remove that suffix.
It also adds the 'Legacy' suffix to the current Button to avoid conflicts.

### Verification

- Built and ran the demo app to ensure both the AvatarVnext and the legacy controls are still working.
- Ran pod lib lint to ensure the Vnext podspec remains valid.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](http://wpcp.azurewebsites.net/CodeFlowProtocolProxy2.php?pullrequest=https://github.com/microsoft/fluentui-apple/pull/406)


ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
```diff
@@ -7,24 +7,24 @@ import UIKit
 
 // Register your control demos here
 let demos: [(title: String, controllerClass: UIViewController.Type)] = [
+    ("Avatar (Vnext)", AvatarDemoController.self),
```
kubalani, line 10: Need a formal name instead of Vnext?  "Avatar 2.0" ?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift
```diff
@@ -189,7 +189,7 @@ class AvatarGroupViewDemoController: DemoController {
     private func insertAvatarViews(style: AvatarGroupViewStyle, showBorders: Bool) {
         var constraints: [NSLayoutConstraint] = []
 
-        for size in AvatarSize.allCases.reversed() {
+        for size in AvatarLegacySize.allCases.reversed() {
```
kubalani: Legacy would show an intent of being deprecated, do we want to discourage users from using existing Avatar view?

rdeassis: Yes! Once we merge this to master, we do want to remove the Legacy one and encourage the client apps to update to the SwiftUI version of it (Vnext).

kubalani (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
rdeassis: The problem with "2.0" is that we're starting with version 0.0.1 for the private drop....
Maybe once we merge it to master this could be a good strategy: "Avatar 0.3" or something like that.

Issue #358, opened 2020-12-29
Generate stylesheet for Shadow


kubalani:
#### Platform
- [X] iOS
- [ ] macOS

#### Task
Update Stylesheet to include shadow definitions

Shadow:
  shadow28: [ blur: 28pt, color: $Colors.Neutral.black, opacity: $Opacity.opacity24, x: 0pt, y: 14pt ]

AP_DrawerTokens:
  shadowColor: $Shadow.shadow28.color
  shadowOpacity: $Shadow.shadow28.opacity
  shadowBlur:   $Shadow.shadow28.blur
  shadowX: $Shadow.shadow28.x
  shadowY: $Shadow.shadow28.y

_Originally posted by @rdeassis in https://github.com/microsoft/fluentui-apple/pull/344#r547508741_

PR #414, opened 2021-01-27
[DrawerVnext] Conforming to box-shadow variant by including dual shadows from Figma

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Bugfix to issue #358 

Adding design token for Shadow28 variant. Figma files conform to CSS [box-shadow](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow) method that allows to add multiple shadows to a component. 

The current changes apply dual shadows as opposed to one (marked in red). The illustrate the effect of combined shadow effect I have added a different color. 

**First Shadow**
Screen Shot 2021-01-28 at 11 06 45 AM
**Second Shadow**
Screen Shot 2021-01-28 at 11 08 56 AM

**Combined Shadow**
Screen Shot 2021-01-28 at 11 06 05 AM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/414)

### DIFF
--- a/ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
+++ b/ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
@@ -60,11 +60,11 @@ struct MSFSlideOverPanel<Content: View>: View {
             content
                 .frame(width: contentWidth)
                 .offset(x: resolvedContentOffset)
-                .shadow(color: shadowColor(true),
+                .shadow(color: tokens.shadow1Color,
                         radius: tokens.shadow1Blur,
                         x: tokens.shadow1DepthX,
                         y: tokens.shadow1DepthY)
-                .shadow(color: shadowColor(false),
+                .shadow(color: tokens.shadow2Color,
                         radius: tokens.shadow2Blur,
                         x: tokens.shadow2DepthX,
                         y: tokens.shadow2DepthY)
@@ -153,10 +153,6 @@ struct MSFSlideOverPanel<Content: View>: View {
             }
         }
         return minValue
-    }
-
-    private func shadowColor(_ isPrimary: Bool = true) -> Color {
-        return isPrimary ? tokens.shadow1Color : tokens.shadow2Color
     }
 }
 
--- a/ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
+++ b/ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
@@ -107,7 +107,19 @@ struct MSFSlideOverPanel<Content: View>: View {
         guard backgroundDimmed else {
             return tokens.backgroundClearColor
         }
-        return tokens.backgroundDimmedColor
+
+        // animate opacity
+        var opacity: Double = 0.0
+        switch transitionState {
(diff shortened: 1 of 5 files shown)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -150,8 +151,9 @@ Opacity:
   none: 1f
   clear: 0.0f
 
-Shadow:
-    shadow28: [ blur: 28pt, color: $Colors.Neutral.black, opacity: $Opacity.opacity24, x: 0pt, y: 14pt ]
+Shadow28:
```
rdeassis, line 154: I've sent a follow-up question to design on this.
I think we should have all Shadow values under the same Appearance Proxy.

Think of the scenario when we create Shadow32 for example. It will be a different Appearance Proxy instead of another variation of Shadow.

If we get the confirmation that the shadows are always composed of 2 fixed values, we can make those named properties and avoid the [0] and [1] use throughout the code.

rdeassis (commented):
Is the difference between the shadows before and after the change visible to the naked eye?

Can you also add a video or screenshot to the description of the validation?

Another question I have is about dark mode? Do all these shadows apply equally on light and dark modes?

tools/sgen/input/FluentUIStyle.yml
kubalani: Updated as per our discussion

ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
```diff
@@ -144,6 +154,10 @@ struct MSFSlideOverPanel<Content: View>: View {
         }
         return minValue
     }
+
+    private func shadowColor(_ isPrimary: Bool = true) -> Color {
+        return isPrimary ? tokens.shadow1Color : tokens.shadow2Color
```
rdeassis, line 159: I think this is not needed. Can you just use them directly on the .shadow modifier?

ios/FluentUI/Vnext/Drawer/SlideOverPanel.swift
```diff
@@ -103,13 +107,19 @@ struct MSFSlideOverPanel<Content: View>: View {
...
+        switch transitionState {
+        case .collapsed:
+            opacity = 0
+        case .expanded:
+            opacity = 1
+        case .inTransisiton:
+            opacity = percentTransition ?? 0
+        }
+
+        return tokens.backgroundDimmedColor.opacity(opacity)
```
rdeassis, line 122: Why aren't we animating from clear color to the background dimmed instead of using the opacity?
Is it because of the gesture scenario that calculates the percentage that reflects the opacity?

kubalani: yes, also swiftui only animates properties that are modified within `withAnimation` block

rdeassis (approved):

PR #418, opened 2021-01-30
unwrap optionals safely

IvanBl:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
This PR changes the way of optionals unwrapping and unwraps them safely. Force unwrapping is not safe and might lead to crashes.

**_Question_**
In this code base `unowned self` in closures is used in most places.  Even though `unowned` has better performance than `weak`, it should be only used when we are sure that the reference always refers to an instance that has not been deallocated. I understand that if `self` is `nil` inside closure, something has gone wrong, however wouldn't it be safer to use `weak`? What do you think? I am happy to improve this if needed 😄 

**Testing**
This PR only improves optional unwrapping and doesn't change UI or any other logic. So the demo app as well as all components should work as before.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/418)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -37,8 +37,8 @@ class DemoListViewController: UITableViewController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
-        
-        guard let title = FluentUIFramework.bundle.object(forInfoDictionaryKey: "CFBundleExecutable") as? String else { 
+
+        guard let title = FluentUIFramework.bundle.object(forInfoDictionaryKey: "CFBundleExecutable") as? String else {
             return assertionFailure("CFBundleExecutable is nil")
         }
         let titleView = TwoLineTitleView()
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -9,11 +9,11 @@ import FluentUI
 class DemoListViewController: UITableViewController {
 
     static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding {
+        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
             Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: colorProvider.primaryColor(for: window)!, whenContainedInInstancesOf: [type(of: window)])
+            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
         } else {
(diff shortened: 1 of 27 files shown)


ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -9,11 +9,11 @@ import FluentUI
...
 
     static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding {
+        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
             Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: colorProvider.primaryColor(for: window)!, whenContainedInInstancesOf: [type(of: window)])
+            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
         } else {
-            FluentUIFramework.initializeAppearance()
+            FluentUIFramework.initializeAppearance(with: Colors.primary(for: window))
```
IvanBl, line 16: This is just a warning fix. `'initializeAppearance()' is deprecated`

harrieshin:
`In this code base unowned self in closures is used in most places. Even though unowned has better performance than weak, it should be only used when we are sure that the reference always refers to an instance that has not been deallocated.`
I agree. I think we should use weak rather than unowned for most of these cases.

harrieshin:
`Force unwrapping is not safe and might lead to crashes.`
Yes! thanks for tackling this. can you just update the PR description with test notes? making sure the existing controls in the demo app work as expected?

harrieshin:
> [CLA assistant check](https://cla.opensource.microsoft.com/microsoft/fluentui-apple?pullRequest=418) 
> Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/fluentui-apple?pullRequest=418) before we can accept your contribution.
> 
> ❌ IvanBl [sign now](https://cla.opensource.microsoft.com/microsoft/fluentui-apple?pullRequest=418)
> You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.opensource.microsoft.com/check/microsoft/fluentui-apple?pullRequest=418) it.

please make sure you go through the CLA link so we can merge your PR

IvanBl:
@harrieshin thanks for the feedback, how does it look now? 😄

harrieshin:
> @harrieshin thanks for the feedback, how does it look now? 😄

one more thing. can you make sure you install SwiftLint and run it with your change? I think some of the whitespace stuff can be fixed https://github.com/microsoft/fluentui-apple#swift-lint

harrieshin (approved):

IvanBl:
swift lint warnings have been fixed, thanks! ✅

PR #419, opened 2021-02-01
[ListVnext] Customizing Leading and Trailing Views

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding in large leading view support and additional trailing view.

### Verification

Manual testing

| Simulator Screen Shot - iPhone 11 Pro - 2021-02-01 at 15 05 07 | Simulator Screen Shot - iPhone 11 Pro - 2021-02-01 at 15 04 24 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/419)


ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -49,53 +51,61 @@ struct MSFListCellView: View {
...
-                VStack(alignment: .leading) {
-                    if let title = state.title {
-                        Text(title)
+                VStack(alignment: .leading, spacing: 0) {
+                    if hasTitle {
+                        Text(state.title)
                             .font(Font(tokens.textFont))
                             .foregroundColor(Color(tokens.leadingTextColor))
-                            .lineLimit(state.titleLineLimit)
+                            .lineLimit(state.titleLineLimit == 0 ? nil : state.titleLineLimit)
```
kubalani, line 67: over reliance on state blocks the functionality or interfaces for native swiftUI clients, how would a swiftUI client consume or set the line limit?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -49,53 +51,61 @@ struct MSFListCellView: View {
             }
         }, label: {
             HStack(spacing: 0) {
+                let defaultIconSize = tokens.iconDefaultSize
```
kubalani, line 54: isn't the icon size in relation with to the size of the list?

sophialee0416: Nope, there are currently 2 icon sizes, default and large.

ios/FluentUI/Vnext/List/ListCell.swift
sophialee0416: If the client chooses to set the line limit then they would set it in the state along with the rest of the cell properties. By default there is no line limit set.

ios/FluentUI/Vnext/Design Token System/ListTokens.swift
```diff
@@ -6,13 +6,6 @@
 import UIKit
 import SwiftUI
 
-/// Pre-defined styles of icons
-@objc public enum MSFListIconStyle: Int, CaseIterable {
```
rdeassis, line 10: Can we incorporate the different icon sizes in this enum and bring it back?
Then we can tackle the 16x16 icon issue too.

ios/FluentUI/Vnext/List/ListCell.swift
rdeassis: And we now need to add at least the 16x16 size in addition to these 2.

ios/FluentUI/Vnext/Design Token System/ListTokens.swift
```diff
@@ -55,15 +48,12 @@ public class MSFListTokens: MSFTokensBase, ObservableObject {
     @Published public var disclosureSize: CGFloat!
     @Published public var horizontalCellPadding: CGFloat!
     @Published public var iconInterspace: CGFloat!
-    @Published public var iconSize: CGFloat!
+    @Published public var iconDefaultSize: CGFloat!
```
rdeassis, line 51: I believe there's value in splitting the cell tokens into a different AP.
If we do that, we'll be able to have just iconSize as a property here and set it on a "per cell" basis.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -117,10 +127,12 @@ struct ListCellButtonStyle: ButtonStyle {
         }
         return configuration.label
             .contentShape(Rectangle())
-            .frame(minHeight: height)
             .listRowInsets(EdgeInsets())
+            .padding(.top, tokens.horizontalCellPadding / 2)
```
rdeassis, line 131: replace these 4 .padding calls with the padding(EdgeInsets(top,bottom,leading,trailing))?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -9,12 +9,14 @@ import SwiftUI
 /// Properties that make up cell content
 @objc public class MSFListCellState: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
-    @objc @Published public var leadingView: UIView?
+    @objc @Published public var leadingView: UIView? // Supporting square views only
+    @objc @Published public var isLargeIcon: Bool = false
     @objc @Published public var title: String = ""
     @objc @Published public var subtitle: String?
+    @objc @Published public var trailingView: UIView? // Supporting square views only
```
rdeassis, line 16: I think the scenario we support in the trailingView is not necessarily just square views, right?
Any non-text-based view should still render correctly, I believe. It's just the label views that get compressed and disappear.

I think it's worth adding /// documentation here explaining more about the bug so that consumers of this code can be aware that it's a known issue.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -9,12 +9,14 @@ import SwiftUI
 /// Properties that make up cell content
 @objc public class MSFListCellState: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
-    @objc @Published public var leadingView: UIView?
+    @objc @Published public var leadingView: UIView? // Supporting square views only
```
rdeassis, line 12: Same thing about the documentation comment as below. The only thing I'd change in the wording is that this is following the "iconSize tokens" instead of "square". (iconSize is currently square, but it could change in the future).

As another nit comment on the naming not specifically related to this PR, how about changing the name from "iconSize" to "leadingViewSize"? I think it makes more sense after we replaced the property from an image to a view.

ios/FluentUI/Vnext/List/ListCell.swift
sophialee0416: Resolved offline: leading/trailing views can only support views with the same width/height regardless of type

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -244,7 +244,7 @@ public struct AvatarView: View {
     }
 }
 
-/// UIKit wrapper that exposes the SwiftUI Button implementation
+/// UIKit wrapper that exposes the SwiftUI Avatar implementation
```
rdeassis, line 247: Thanks for fixing this! :)

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -24,10 +24,9 @@ public struct MSFListView: View {
     @ObservedObject var state: MSFListState
     @ObservedObject var tokens: MSFListTokens
 
-    public init(sections: [MSFListSectionState],
-                iconStyle: MSFListIconStyle) {
+    public init(sections: [MSFListSectionState]) {
         self.state = MSFListState()
-        self.tokens = MSFListTokens(iconStyle: iconStyle)
+        self.tokens = MSFListTokens()
```
rdeassis, line 29: Are you planning to set the windowProvider to this tokens list on the other PR?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -247,17 +255,21 @@ AP_MSFListTokens:
...
   labelColor: $Colors.Foreground.neutral1
   labelFont: $Typography.body
+  leadingViewColor: $Colors.Foreground.neutral4
+  leadingViewSize: $Icon.size.medium
   sublabelColor: $Colors.Foreground.neutral3
   sublabelFont: $Typography.caption1
   trailingItemForegroundColor: $Colors.Foreground.neutral3
+  trailingItemSize: $Icon.size.medium
+
+MSFListCellSmallTokens extends MSFListCellTokens:
```
rdeassis, line 268: I guess what I meant on defining the different sizes for the leading view size was more on the lines of the backgroundColor in line 250, for example:

` trailingItemSize: [ small: $Icon.size.xSmall, medium: $Icon.size.medium, large: $Icon.size.xxlarge]`

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -239,6 +239,14 @@ MSFGhostButtonTokens extends MSFButtonTokens:
   borderColor: [ rest: $Colors.Neutral.clear, hover: $Colors.Neutral.clear, pressed: $Colors.Neutral.clear, selected: $Colors.Brand.shade10, disabled: $Colors.Neutral.clear ]
 
 AP_MSFListTokens:
+  backgroundColor: [ rest: $Colors.Background.neutral1, pressed: $Colors.Background.neutral5 ]
+  borderColor: $Colors.Stroke.neutral1
+  borderSize: $Border.size.thin
+  horizontalCellPadding: $Spacing.medium
+  sublabelColor: $Colors.Foreground.neutral3
```
rdeassis, line 246: What's the goal of having the sublabel font and color on the list since they're already on the cell tokens?
And why only the sublabels and not the labels?

ios/FluentUI/Vnext/Design Token System/ListTokens.swift
```diff
@@ -37,11 +36,51 @@ import SwiftUI
 }
 
 public class MSFListTokens: MSFTokensBase, ObservableObject {
+    @Published public var backgroundColor: UIColor!
```
rdeassis, line 39: What's the order you're using for the order of these variables?
I'd suggest something like this: semantic usage and variable name

```
@Published public var backgroundColor: UIColor!
@Published public var highlightedBackgroundColor: UIColor!

@Published public var borderColor: UIColor!
@Published public var borderSize: CGFloat!

@Published public var horizontalCellPadding: CGFloat!

@Published public var subtitleColor: UIColor!
@Published public var subtitleFont: UIFont!
```

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Oops! I see. Let me check with Victor on these, trailing items should only be 1 size AFAIK

tools/sgen/input/FluentUIStyle.yml
sophialee0416: The sub label tokens are used by header/footer, which will get their own AP from the next PR, but adding that right now would cause overlap... this will get fixed in #424 . For now I am letting the headers get their tokens from the List AP

ios/FluentUI/Vnext/Design Token System/ListTokens.swift
```diff
@@ -55,14 +94,16 @@ public class MSFListTokens: MSFTokensBase, ObservableObject {
...
     @Published public var iconInterspace: CGFloat!
-    @Published public var iconSize: CGFloat!
+    @Published public var leadingViewSize: CGFloat!
+    @Published public var trailingItemSize: CGFloat!
+
     @Published public var subtitleFont: UIFont!
     @Published public var textFont: UIFont!
 
-    var iconStyle: MSFListIconStyle!
+    var cellLeadingViewSize: MSFListCellLeadingViewSize!
```
rdeassis, line 103: This should be a @Published public var too.
If the theme is changed and the Tokens class resets this property, the controls need to be updated automatically.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -31,13 +48,16 @@ import SwiftUI
...
+    @ObservedObject var tokens: MSFListCellTokens
     var hasDividers: Bool
 
-    init(state: MSFListCellState, tokens: MSFListTokens, hasDividers: Bool = false) {
+    init(state: MSFListCellState, hasDividers: Bool = false, windowProvider: FluentUIWindowProvider?) {
         self.state = state
-        self.tokens = tokens
         self.hasDividers = hasDividers
+        self.tokens = MSFListCellTokens(cellLeadingViewSize: state.leadingViewSize)
+        if let windowProvider = windowProvider {
```
rdeassis, line 58: The windowProvider property is **FluentUIWindowProvider?**.
You can just set the value passed as a parameter directly. (without the if let)

rdeassis (approved):
Thanks for addressing the comments!
I left 2 more and will send you a patch so you can test the theming portion.

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Custom trailing items are 1 size excluding specific components (button/toggle), so I'll leave this alone for now.

PR #424, opened 2021-02-03
[ListVnext] Added bolded header style

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A new header style was added. This style is larger and bolded, and is the vNext parity of current FluentUI's HeaderFooter Primary style.

### Verification

Manual test

| Simulator Screen Shot - iPhone 11 Pro - 2021-02-03 at 17 20 07 | Simulator Screen Shot - iPhone 11 Pro - 2021-02-03 at 17 17 30 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/424)


ios/FluentUI/Vnext/Design Token System/ListHeaderFooterTokens.swift
```diff
@@ -0,0 +1,54 @@
...
+/// HeaderFooter styles
+@objc public enum MSFListHeaderFooterStyle: Int, CaseIterable {
+    case header
+    case headerBolded
+}
+
+public class MSFListHeaderFooterTokens: MSFTokensBase, ObservableObject {
+    @Published public var backgroundColor: UIColor!
+    @Published public var textColor: UIColor!
+
```
kubalani, line 18: nit: extra line

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,26 +6,27 @@
...
 struct Header: View {
-    let title: String
-    var tokens: MSFListTokens
+    let state: MSFListSectionState
+    var tokens: MSFListHeaderFooterTokens
 
-    init(title: String, tokens: MSFListTokens) {
-        self.title = title
-        self.tokens = tokens
+    init(state: MSFListSectionState) {
```
kubalani, line 13: any specific reason to pass additional data? The header doesn't seem to use anything in addition to title and tokens

kubalani (approved):

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
sophialee0416: Eventually I will be passing in the rest of the HeaderFooter data, including footer data, accessory data , # of lines, handler etc! :)

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -439,6 +440,12 @@ AP_MSFListTokens:
 
 MSFIconOnlyListTokens extends MSFListTokens:
   iconColor: $Colors.Foreground.neutral3
+  
+AP_MSFListHeaderFooterTokens:
+  backgroundColor: [ rest: $Colors.Background.neutral1, highlighted: $Colors.Background.neutral5 ]
+  textColor: [ default: $Colors.Foreground.neutral3, bolded: $Colors.Foreground.neutral1 ]
```
rdeassis, line 446: I think default and bolded are good candidates for styles. Like the primary/ghost buttons for example.

rdeassis, line 446: nit: regarding the "bolded" naming. Can you double check that with the design team? The reason I bring this up is that it can be misleading if it's implying a bold font and later needs to be changed to something that is not bold.

ios/FluentUI/Vnext/Design Token System/ListHeaderFooterTokens.swift
```diff
@@ -0,0 +1,54 @@
...
+    @objc open func didChangeAppearanceProxy() {
+        updateForCurrentTheme()
+    }
+
+    public override func updateForCurrentTheme() {
+        let currentTheme = theme
+        let appearanceProxy = currentTheme.MSFListHeaderFooterTokens
+        
+        switch style {
+        case .header:
```
rdeassis, line 43: If you change the YAML file to have 2 different style definitions, here is where you would assign the proper appearance proxy.

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,26 +6,27 @@
...
-                .padding(.top, tokens.horizontalCellPadding / 2)
-                .padding(.leading, tokens.horizontalCellPadding)
-                .padding(.trailing, tokens.horizontalCellPadding)
-                .padding(.bottom, tokens.horizontalCellPadding / 2)
+            if let title = state.title {
+                Text(title)
+                    .font(Font(tokens.textFont))
+                    .foregroundColor(Color(tokens.textColor))
+                    .listRowInsets(EdgeInsets())
+                    .padding(.top, tokens.horizontalCellPadding / 2)
```
rdeassis, line 25: You can apply just one padding modifier specifying all the edge insets:

`.padding(EdgeInsets(top: <#T##CGFloat#>, leading: <#T##CGFloat#>, bottom: <#T##CGFloat#>, trailing: <#T##CGFloat#>))`

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,26 +6,27 @@
...
-                .listRowInsets(EdgeInsets())
-                .padding(.top, tokens.horizontalCellPadding / 2)
-                .padding(.leading, tokens.horizontalCellPadding)
-                .padding(.trailing, tokens.horizontalCellPadding)
-                .padding(.bottom, tokens.horizontalCellPadding / 2)
+            if let title = state.title {
+                Text(title)
+                    .font(Font(tokens.textFont))
+                    .foregroundColor(Color(tokens.textColor))
+                    .listRowInsets(EdgeInsets())
```
rdeassis, line 24: What does the EdgeInsets() do with no parameters?

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,26 +6,27 @@
...
 
-    init(title: String, tokens: MSFListTokens) {
-        self.title = title
-        self.tokens = tokens
+    init(state: MSFListSectionState) {
+        self.state = state
+        self.tokens = MSFListHeaderFooterTokens(style: state.style)
     }
 
     var body: some View {
```
rdeassis, line 18: You need the modifier to retrieve the current theme in case there is an override for the window or container control:
This applies for every SwiftUI View that has a tokens property. (meanwhile I'm trying to find a way to avoid this duplication of code and yet affect all of the controls we need)

```
.onAppear {
                // When environment values are available through the view hierarchy:
                //  - If we get a non-default theme through the environment values,
                //    we use to override the theme from this view and its hierarchy.
                //  - Otherwise we just refresh the tokens to reflect the theme
                //    associated with the window that this View belongs to.
                if theme == ThemeKey.defaultValue {
                    self.tokens.updateForCurrentTheme()
                } else {
                    self.tokens.theme = theme
                }
            }
```

ios/FluentUI/Vnext/Design Token System/ListHeaderFooterTokens.swift
sophialee0416: Intentionally leaving spaces between UIColor/CGFloat/UIFont tokens as they will grow later (maintaining similar formatting with ListTokens)

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
sophialee0416: It removes built-in padding that SwiftUI adds to their views

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListDemoController.swift
```diff
@@ -91,6 +92,7 @@ class ListDemoController: DemoController {
             listSection = MSFListSectionState()
             listSection.title = section.title
             listSection.cells = []
+            listSection.style = MSFListHeaderFooterStyle.headerSecondary
```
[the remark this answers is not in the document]
sophialee0416: I'm actually thinking of maintaining the default/secondary header style for the List demo controller so it reflects the TableViewCell demo, and eventually showcase the different header styles in a header/footer demo.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -445,6 +444,19 @@ AP_MSFListCellTokens:
   sublabelFont: $Typography.caption1
   trailingItemForegroundColor: $Colors.Foreground.neutral3
   trailingItemSize: $Icon.size.medium
+  
+AP_MSFHeaderFooterTokens:
+  backgroundColor: [ default: $Colors.Background.neutral1, divider: $Colors.Background.neutral5 ]
+  headerHeight: [ default: $Spacing.xxxlarge, divider: $Spacing.xlarge ]
+  horizontalHeaderPadding: $Spacing.medium
```
rdeassis, line 451: You can replace **horizontalHeaderPadding** and **topHeaderPadding** with a "padding" token that would be like:

`padding: Insets($Spacing.xlarge, $Spacing.medium, $Spacing.medium, $Spacing.medium)`

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -445,6 +444,19 @@ AP_MSFListCellTokens:
...
+  textColor: $Colors.Foreground.neutral3
+  textFont: $Typography.caption1
+  topHeaderPadding: $Spacing.xlarge
+
+MSFPrimaryHeaderTokens extends MSFHeaderFooterTokens:
+  textColor: $Colors.Foreground.neutral1
+  textFont: $Typography.headline
+  topHeaderPadding: $Spacing.medium
```
rdeassis, line 459: Same for the padding here:

`padding: Insets($Spacing.medium, $Spacing.medium, $Spacing.medium, $Spacing.medium)`

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -445,6 +444,19 @@ AP_MSFListCellTokens:
   sublabelFont: $Typography.caption1
   trailingItemForegroundColor: $Colors.Foreground.neutral3
   trailingItemSize: $Icon.size.medium
+  
+AP_MSFHeaderFooterTokens:
+  backgroundColor: [ default: $Colors.Background.neutral1, divider: $Colors.Background.neutral5 ]
+  headerHeight: [ default: $Spacing.xxxlarge, divider: $Spacing.xlarge ]
```
rdeassis, line 450: Where is the **divider: $Spacing.xlarge**  being used? (same for the background color)

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -445,6 +444,19 @@ AP_MSFListCellTokens:
   sublabelFont: $Typography.caption1
   trailingItemForegroundColor: $Colors.Foreground.neutral3
   trailingItemSize: $Icon.size.medium
+  
+AP_MSFHeaderFooterTokens:
+  backgroundColor: [ default: $Colors.Background.neutral1, divider: $Colors.Background.neutral5 ]
```
rdeassis, line 449: Seems like both height and background color only use the default property:


```
backgroundColor = appearanceProxy.backgroundColor.default
headerHeight = appearanceProxy.headerHeight.default
```

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -10,6 +10,7 @@ import SwiftUI
 @objc public class MSFListSectionState: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
     @objc @Published public var cells: [MSFListCellState] = []
+    @objc @Published public var style: MSFHeaderFooterStyle = .headerPrimary
     @objc @Published public var title: String?
     @objc @Published public var hasDividers: Bool = false
```
rdeassis, line 15: The cell view [hasDividers](https://github.com/microsoft/fluentui-apple/blob/5609a48bee3a4e4cdb0741c980f140a31ecec21e/ios/FluentUI/Vnext/List/ListCell.swift#L52) should follow this pattern too.

Can you make these changes?
- name hasDivider (singular for the cell)
- should be a published property of the state object

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
sophialee0416: I was thinking of including both header and footer tokens here to maintain parity

sophialee0416: Not using right now but there is a header style called divider, which will be implemented later

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Re: the divider has different styles so I wanted to add them in ahead of time

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
rdeassis: But then the footer would be a different style (like MSFPrimaryFooterTokens for example) that extends the AP_MSFHeaderFooterTokens that would override the height, right?

rdeassis: As per our offline conversation, we decided not to add stuff that the code is not using right now.

tools/sgen/input/FluentUIStyle.yml
rdeassis: As per our offline conversation, we decided not to add stuff that the code is not using right now.

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,28 +6,44 @@
...
+                    .foregroundColor(Color(tokens.textColor))
+            }
             Spacer()
         }
         .background(Color(tokens.backgroundColor))
+        .padding(EdgeInsets(top: tokens.topPadding,
+                            leading: tokens.leadingPadding,
+                            bottom: tokens.bottomPadding,
+                            trailing: tokens.trailingPadding))
+        .listRowInsets(EdgeInsets())
```
rdeassis: As discussed offline, if this is needed only if we use the Header SwiftUI type in iOS13, we should remove it from code now and document it (out of the code).

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListDemoController.swift
```diff
@@ -104,6 +105,9 @@ class ListDemoController: DemoController {
                 listCell.trailingView = section.hasAccessory ? createCustomView(imageName: cell.image) : nil
                 listCell.accessoryType = accessoryType(for: rowIndex)
                 listCell.layoutType = updateLayout(subtitle: listCell.subtitle)
+                if rowIndex != TableViewCellSampleData.numberOfItemsInSection - 1 {
```
rdeassis, line 108: This logic could be handled inside the component.
It seems weird that the client code needs to remove the divider for the last item.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -35,15 +36,13 @@ public struct MSFListView: View {
...
                     ForEach(sections, id: \.self) { section in
-                        if let sectionTitle = section.title {
-                            Header(title: sectionTitle, tokens: tokens)
+                        if let sectionTitle = section.title, !sectionTitle.isEmpty {
+                            Header(state: section, windowProvider: tokens.windowProvider)
                         }
 
                         ForEach(section.cells.indices, id: \.self) { index in
                             let cellState = section.cells[index]
-                            let hasDividers = (index < section.cells.count - 1 && section.hasDividers) || (section.hasDividers && (cellState.children != nil))
```
rdeassis, line 44: Related to my other comment, setting the cellState.hasDivider seems like a better option compared to leaving that up to the client.
We can sync offline about it.

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
rdeassis: Just noticed your last push. Thanks for fixing this!

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -68,6 +68,21 @@ public struct MSFListView: View {
                 }
             }
     }
+
+    private func dividers(_ sections: [MSFListSectionState]) -> [MSFListSectionState] {
+        var cells: [MSFListCellState]
+        var hasDividers: Bool
+        for sectionIndex in 0...sections.count - 1 {
```
rdeassis, line 75: It's cleaner to do "for section in sections {" instead of using indexes in this case.

Or you can use the ForEach func for convenience:

  
```
sections.forEach { section in
    section.cells.forEach { cell in
        cell.hasDivider = section.hasDividers
    }
}
```

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -68,6 +68,21 @@ public struct MSFListView: View {
                 }
             }
     }
+
+    private func dividers(_ sections: [MSFListSectionState]) -> [MSFListSectionState] {
```
rdeassis, line 72: What would be a more meaningful name for this function? Just "dividers" seems unclear to me.

rdeassis, line 72: Do we need the sections parameter here or can we just use the sections property of the struct?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -69,19 +69,17 @@ public struct MSFListView: View {
...
-            cells = sections[sectionIndex].cells
-            hasDividers = sections[sectionIndex].hasDividers
-            for cellIndex in 0...cells.count - 1 {
-                if cellIndex != cells.count - 1 {
-                    cells[cellIndex].hasDivider = hasDividers
+    private func updateCellDividers() -> [MSFListSectionState] {
+        state.sections.forEach { section in
+            section.cells.forEach { cell in
+                if cell != section.cells.last {
+                    cell.hasDivider = false
```
rdeassis, line 76: You can do this with only one ternary operator:

`cell.hasDivider = section.hasDividers && cell != section.cells.last`

Or to be more efficient and not check the condition all the time you can do:

```
cell.hasDivider = section.hasDividers // (inside the loops)

section.cells.last.hasDivider = false  // (outside of the loops)
```

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -440,12 +436,27 @@ AP_MSFListCellTokens:
...
   labelFont: $Typography.body
   leadingViewColor: $Colors.Foreground.neutral4
-  leadingViewSize: [ small: $Icon.size.xSmall, medium: $Icon.size.medium, large: $Icon.size.xxlarge]
+  leadingViewSize: [ small: $Icon.size.xSmall, medium: $Icon.size.medium, large: $Icon.size.xxlarge ]
   sublabelColor: $Colors.Foreground.neutral3
   sublabelFont: $Typography.caption1
   trailingItemForegroundColor: $Colors.Foreground.neutral3
   trailingItemSize: $Icon.size.medium
 
+AP_MSFHeaderFooterTokens:
```
rdeassis, line 445: FYI: With the new changes I have merged today, there should be a new generated file for MSFHeaderFooterTokens that will contain these tokens and the MSFPrimaryHeaderTokens too.

That's one of the sources of the conflicts of this branch.

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,28 +6,43 @@
...
+            }
             Spacer()
         }
         .background(Color(tokens.backgroundColor))
+        .padding(EdgeInsets(top: tokens.topPadding,
+                            leading: tokens.leadingPadding,
+                            bottom: tokens.bottomPadding,
+                            trailing: tokens.trailingPadding))
+        .frame(minHeight: tokens.headerHeight)
+        .onAppear {
```
rdeassis, line 35: Can you add a screenshot of one of the tokens overridden as part of the theme validation?

rdeassis (commented):
You'll need to resolve the merge conflicts (caused by the generated files and tokens that were moved out of the "Design Token System" folder).

Could you also validate the theme overriding and add a video/screenshots on the PR description? (I think changing the textColor and topPadding to drastically different values could be good visual validations)

PR #425, opened 2021-02-06
[PeoplePicker] Add delegate methods to notify client that PersonaListView is shown or hidden

huanlums:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

- PeoplePicker.swift: Adding two delegate methods peoplePickerDidShowPersonaSuggestions and peoplePickerDidHidePersonaSuggestions to notify client that the PersonaListView is shown/hidden.
- PeoplePickerDemoController.swift: Implement delegate method peoplePickerDidHidePersonaSuggestions to move the VoiceOver focus to the last added persona after the PersonaListView gets dismissed.

### Verification

| Before the change, when voice over is enabled, selecting a contact from PersonaListView would dismiss the list but highlight the voice over focus on the view below wherever user touches.  | After the change, selecting a contact would dismiss the list and VoiceOver focus would land on the last added person |

### Pull request checklist

This PR has considered:
- [ X ] VoiceOver and Keyboard Accessibility

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/425)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
@@ -96,4 +96,8 @@ extension PeoplePickerDemoController: PeoplePickerDelegate {
             completion(personas, true)
         }
     }
+
+    func peoplePickerDidHidePersonaSuggestions(_ peoplePicker: PeoplePicker) {
+        UIAccessibility.post(notification: .screenChanged, argument: peoplePicker.badges.last)
+    }
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
@@ -98,6 +98,6 @@ extension PeoplePickerDemoController: PeoplePickerDelegate {
     }
 
     func peoplePickerDidHidePersonaSuggestions(_ peoplePicker: PeoplePicker) {
-        UIAccessibility.post(notification: .screenChanged, argument: peoplePicker.badges.last)
+        UIAccessibility.post(notification: .layoutChanged, argument: peoplePicker.badges.last)
     }
 }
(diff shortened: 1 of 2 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
```diff
@@ -96,4 +96,8 @@ extension PeoplePickerDemoController: PeoplePickerDelegate {
             completion(personas, true)
         }
     }
+
+    func peoplePickerDidHidePersonaSuggestions(_ peoplePicker: PeoplePicker) {
+        UIAccessibility.post(notification: .screenChanged, argument: peoplePicker.badges.last)
```
harrieshin, line 101: is it a screen change? or more a layout change?

ios/FluentUI/People Picker/PeoplePicker.swift
```diff
@@ -192,13 +198,14 @@ open class PeoplePicker: BadgeField {
         }
 
         personaListView.searchDirectoryState = .idle
-
+        delegate?.peoplePickerDidShowPersonaSuggestions?(self)
```
harrieshin, line 201: at this point the people suggestion is actually shown? has the layout change?

ios/FluentUI.Demo/FluentUI.Demo/Demos/PeoplePickerDemoController.swift
huanlums: Changed to layout change.

ios/FluentUI/People Picker/PeoplePicker.swift
huanlums: Thanks for pointing this out. I never had a deep understanding about update circle/setNeedsLayout/layoutIfNeeded. I just learned it. Added "layoutIfNeeded" after "setNeedsLayout" so that layoutSubviews would be called synchronously.

harrieshin (approved):

PR #426, opened 2021-02-08
Adding button accessibility trait even when the cell is selected.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The PopupMenuController is currently failing to set the button accessibility trait for items that are selected.
Previously, the only scenario where it would be set is when the accessory type was set to disclosure or detail button.

### Verification

Ensured all the PopupMenuController cells have the button accessibility trait even if they're disabled or selected.

| Screen Shot 2021-02-08 at 9 10 00 AM | Screen Shot 2021-02-08 at 9 28 33 AM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/426)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1367,11 +1367,6 @@ open class TableViewCell: UITableViewCell {
     }
 
     private func initAccessibilityForAccessoryType() {
-        if _accessoryType == .disclosureIndicator || _accessoryType == .detailButton {
-            accessibilityTraits.insert(.button)
-        } else {
-            accessibilityTraits.remove(.button)
-        }
         if _accessoryType == .checkmark || isSelected {
             accessibilityTraits.insert(.selected)
         } else {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1380,6 +1375,8 @@ open class TableViewCell: UITableViewCell {
     }
 
     private func updateAccessibility() {
+        accessibilityTraits.insert(.button)
```
harrieshin, line 1378: is that a fair assumption? what if client is using this that isn't actionable tableviewcell? shouldn't client have more say what accessible trait it should be?

rdeassis: Initially I thought of conditioning this trait with the **isUserInteractionEnabled** property which is the scenario where the cell does not represent an action of any sort.

But if you look at the [isEnabled property](https://github.com/microsoft/fluentui-apple/blob/78aa5434e23d4c84333a3c27a9c9a60dc133307f/ios/FluentUI/Table%20View/TableViewCell.swift#L708), it sets isUserInteractionEnabled for a disabled button (that still needs to be treated as a button.

kubalani (approved):

Issue #343, opened 2020-12-15
Rename repository default branch to "main"


markavitale:
#### Describe the work to be done
Microsoft is gradually moving to have its repositories use "main" as the default branch. Fluent UI Apple is no exception here and this issue tracks the work to migrate over to "main".

##### Migration
- Update all the CI/CD/pipelines to operate on both "master" and "main" branches
- Create a new "main" branch based off of the current default branch's latest commit
- Reassign any protected branch settings to the new "main" default branch from the former
- Update any open pull requests against the current default branch for the new "main" branch
- Update the default branch to the new "main" branch

##### Cleanup
- Delete the current default branch
- Update all the CI/CD/pipelines to remove the references to "master"
- Update all links/documentation to point to the main branch instead of the master branch
- Update the links on https://developer.microsoft.com/en-us/fluentui#/controls/ios to point to the main branch instead of the master branch.

#### Potential Impact
Any consumers of this library simply specifying "master" for Carthage consumption may* run into trouble. Instead, we recommend using a tagged commit for better stability.

\* I suspect that because [GitHub now redirects links to deleted branches to the default branch automatically](https://github.blog/changelog/2020-07-17-links-to-deleted-branches-now-redirect-to-the-default-branch/), things may still work. 

#### Additional context/screenshots
See https://github.com/github/renaming for more context.

Saadnajmi:
Also worth updating the internal documentation we have for the macOS controls in the repo

markavitale:
Yep, that’ll happen in the “Update all links/documentation to point to the main branch instead of the master branch” cleanup phase.

markavitale:
The last remaining bit of this work to update the public website is in review at https://github.com/microsoft/fluentui/pull/16978

markavitale:
The website update is merged and now being cherry picked to the v7 (currently live) website. https://github.com/microsoft/fluentui/pull/16987. Going ahead and closing this issue as completed.

PR #429, opened 2021-02-10
Update CI and publishing to work on new default branch named "main

markavitale:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes
Update our CI and CD to support "main" as well as "master". This will allow a seamless transition when renaming the default branch. This is the first step of Issue #343.

Note that the publishing steps for simplicity only reference "main", not "master". This means that there will be a small gap of time where publishing won't work as expected once this PR has been merged but before the rename has been executed. The intention is to keep this gap as small as possible, under an hour if everything goes as planned.

### Verification
CI/CD changes aren't trivial to test in progress, especially since the main branch doesn't exist yet, but part of this work will be making a followup PR to update all the remaining references to the master branch targeting the new main branch. At that time we'll confirm that all the pipelines are working as expected.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/429)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
+      - main
   pull_request:
     branches:
       - master
+      - main
 
 jobs:
   xcodebuild:
--- a/.github/workflows/localize.yml
+++ b/.github/workflows/localize.yml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - master
+      - main
     paths:
       - '**.strings'
       - '**.stringsdict'
--- a/.github/workflows/podPublish.yml
+++ b/.github/workflows/podPublish.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   Pod-Publish:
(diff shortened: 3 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -4,9 +4,11 @@ on:
   push:
     branches:
       - master
```
rdeassis, line 6: I assume there will be another pass where these references to the master will be removed, right?

rdeassis (approved):

.github/workflows/ci.yml
markavitale: Yep! I'll likely rebase #430 on top of this after its committed and then remove the old references. This is just to make sure everything continues working in the interim.

rdeassis: Cool! Thanks for the confirmation!

anandrajeswaran (approved):

PR #431, opened 2021-02-10
Button: Add correct focus ring

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
The keyboard focus ring for Buttons hugs the image/text in the Button, which is within the button frame.  This change adds the drawFocusRingMask method to make the focus ring hug the button frame itself.

### Verification
Turned on full keyboard navigation and navigated through Buttons in the test app.

| Before | After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/431)

### DIFF
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -211,6 +211,13 @@ open class Button: NSButton {
 		return true
 	}
 
+	open override func drawFocusRingMask() {
+		// Ensure we draw the focus ring around the entire button bounds
+		// rather than just around the image or title.
+		let path = NSBezierPath(roundedRect: bounds, xRadius: cornerRadius, yRadius: cornerRadius)
+		path.fill()
+	}
+
 	open override func viewDidChangeBackingProperties() {
 		super.viewDidChangeBackingProperties()
 


rdeassis (approved):

anandrajeswaran (approved):

macos/FluentUI/Button/Button.swift
```diff
@@ -211,6 +211,13 @@ open class Button: NSButton {
 		return true
 	}
 
+	open override func drawFocusRingMask() {
+		// Ensure we draw the focus ring around the entire button bounds
+		// rather than just around the image or title.
+		let path = NSBezierPath(roundedRect: bounds, xRadius: cornerRadius, yRadius: cornerRadius)
+		path.fill()
```
harrieshin, line 218: why is it fill rather than stroke?

chripear: This is the mask for the focus ring.  I'm not drawing the ring itself.

chripear: i.e. how it works is that the actual focus ring hugs the outside edge of this shape.

harrieshin (approved):

PR #439, opened 2021-02-16
Updating FluentUI target iOS version to 13.0

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change bumps the target OS version from iOS 11.0 to iOS 13.0.
A maintenance branch was also created for the continuous support for iOS 11. (https://github.com/microsoft/fluentui-apple/tree/main_0.1).

### Verification

- Successfully built all targets (Framework, Static Lib, Demo App).
- Archived Demo app version and successfully ran on a iOS 13 device while validating it does not install on an iOS 12 device.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/439)


Saadnajmi:
Would I be good to set the new minimum on FURN to iOS 13 as well?

MicrosoftFluentUI.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'MicrosoftFluentUI'
-  s.version          = '0.1.28'
+  s.version          = '0.2.0'
```
Saadnajmi, line 3: :o

Saadnajmi (approved):

harrieshin:
> Would I be good to set the new minimum on FURN to iOS 13 as well?

you should have been able to mark minimum of iOS 13 independently from this PR. right?

ios/FluentUI/Extensions/UIImage+Extensions.swift
```diff
@@ -14,17 +14,6 @@ import UIKit
     }
 
     @objc func image(withPrimaryColor primaryColor: UIColor) -> UIImage {
```
harrieshin, line 16: we should start deprecating this function from this extension I think.

harrieshin (approved):

ios/FluentUI/Extensions/UIImage+Extensions.swift
rdeassis: I'll create a separate work item to address that.

rdeassis:
> > Would I be good to set the new minimum on FURN to iOS 13 as well?
> 
> you should have been able to mark minimum of iOS 13 independently from this PR. right?

And btw, If it's not set to >= iOS 13, you won't be able to update to 0.2.X when it's released (since the podspec deployment_target changed to 13.0).

Saadnajmi:
> > Would I be good to set the new minimum on FURN to iOS 13 as well?
> 
> you should have been able to mark minimum of iOS 13 independently from this PR. right?

Yep. I can make that PR the next time we need  to update the FUA version. 
Also, now that we have a precedent for minor version increments, can we enforce that for static library compatibility? I can talk offline about that.

PR #443, opened 2021-02-19
[ListVnext] Automating cell layout/height

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Cell layout is now automatically defined - presence of only a title will produce a one-line cell height, and a subtitle will produce a double-line cell height. There is still the option to override the default layout as well.

### Verification

https://user-images.githubusercontent.com/22566866/108574570-6606e980-72dd-11eb-91eb-02d795afb2c7.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/443)

### DIFF
--- a/ios/FluentUI/Vnext/List/ListCell.swift
+++ b/ios/FluentUI/Vnext/List/ListCell.swift
@@ -27,19 +27,20 @@ import SwiftUI
     @objc @Published public var leadingView: UIView?
     @objc @Published public var leadingViewSize: MSFListCellLeadingViewSize = .medium
     @objc @Published public var title: String = ""
-    @objc @Published public var subtitle: String?
+    @objc @Published public var subtitle: String = ""
     @objc @Published public var trailingView: UIView?
     @objc @Published public var accessoryType: MSFListAccessoryType = .none
     @objc @Published public var titleLineLimit: Int = 0
     @objc @Published public var subtitleLineLimit: Int = 0
     @objc @Published public var children: [MSFListCellState]?
     @objc @Published public var isExpanded: Bool = false
-    @objc @Published public var layoutType: MSFListCellLayoutType = .oneLine
+    @objc @Published public var layoutType: MSFListCellLayoutType = .none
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -27,19 +27,20 @@ import SwiftUI
...
     @objc @Published public var children: [MSFListCellState]?
     @objc @Published public var isExpanded: Bool = false
-    @objc @Published public var layoutType: MSFListCellLayoutType = .oneLine
+    @objc @Published public var layoutType: MSFListCellLayoutType = .none
     @objc public var onTapAction: (() -> Void)?
 }
 
 /// Pre-defined layout heights of cells
 @objc public enum MSFListCellLayoutType: Int, CaseIterable {
+    case none
```
rdeassis, line 43: It seems like the implementation you've done for this is more like "automatic" than "none" because it considers the subtitle string to define the height.

Thoughts?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -128,11 +129,13 @@ struct MSFListCellView: View {
 
 struct ListCellButtonStyle: ButtonStyle {
     let tokens: MSFListCellTokens
-    let layoutType: MSFListCellLayoutType
+    let state: MSFListCellState
```
rdeassis, line 132: Have you tested changing the state.subtitle property at runtime and see if the cell updates?
I'll be good to test both the cell height and the display of the subtitle itself.

sophialee0416: The cell updates as expected. If the subtitle exists but the title is empty, it should retain a 2-line state. If the subtitle is removed, the cell should go back to a 1-line state :)

rdeassis (approved):

PR #445, opened 2021-02-20
SGen generated code size optimizations, modularization per control.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

These changes simplify the generated code by SGen in order to reduce the binary size footprint of the library.
The highlights included in this change are:

1. Generated files (.generated.swift) are broken down per control into separate files. This will make the FluentUI Vnext library more lightweight for cocoapods consumers using subspecs.

2. @objc exposure of the types generated by SGen were removed. Calls or overrides can still be made on the Swift layer or added later to specific scenarios.

3. All properties of the classes generated are now read only calculated properties, which means that they need to be overridden through inheritance of the types intended instead of just setting the properties as before.

The FluentUI Vnext addition to the library for a fat binary (x86_64 + arm64) of a static library build **initially caused an increase of 74% on its size (27.6MB to 48MB)**.

With these changes, the increase has dropped significantly on that same configuration: **33% (27.6MB to  36.8MB).**

**Reducing the increase by 55% (20.4MB down to 9.2MB).**

Screen Shot 2021-02-19 at 2 17 50 PM

### Verification

1. Exercised all theming scenarios on the Demo App including:
 -- Updating the fixed theme override set for specific controls.
 -- Overriding themes per UIWindow instance

2. pod lib lint on the FluentUIVnext.podspec file to ensure controls can now be consumed individually from the Design Tokens aspect too (an application will use only the tokens needed for the controls it's importing).

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/445)


tools/sgen/input/StylesheetsConfig.yml
```diff
@@ -1,5 +1,6 @@
-objc: true
+objc: false
 generateBaseStyle: true
+generateFilePerAppearanceProxy: true
```
kubalani, line 3: seems like you finally found this in config

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -153,3 +116,77 @@ class ThemingDemoController: DemoController {
         }
     }
 }
+
+open class CustomStyle: FluentUIStyle {
```
kubalani, line 120: This doesn't seem to be autogenerated, do we need any manual updates to this wehn when we add a new design token?

ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -28,7 +28,9 @@ public struct UIViewAdapter: UIViewRepresentable {
             return
         }
 
-        stackView.removeAllSubviews()
+        for view in stackView.arrangedSubviews {
```
kubalani, line 31: is this related change?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -420,7 +420,7 @@ MSFGhostButtonTokens extends MSFButtonTokens:
   borderColor: [ rest: $Colors.Neutral.clear, hover: $Colors.Neutral.clear, pressed: $Colors.Neutral.clear, selected: $Colors.Brand.shade10, disabled: $Colors.Neutral.clear ]
 
 AP_MSFListTokens:
-  backgroundColor: [ rest: $Colors.Background.neutral1, pressed: $Colors.Background.neutral5 ]
+  backgroundColor: $Colors.Background.neutral1
```
kubalani, line 423: is this intended?

ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
rdeassis: That's because we don't want to add the UIView+extensions.swift into the podspec.
I just changed not to use that extension and save that bit of space.

tools/sgen/Sources/SGen/Generator.swift
```diff
@@ -156,11 +156,22 @@ struct Generator  {
             
         }
         if Config.generateFilePerAppearanceProxy {
```
kubalani, line 158: Use an alternative flag instead of overriding

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
rdeassis: No. It's not autogenerated. This is a demonstration on how to override the new classes with their read only properties.

tools/sgen/input/FluentUIStyle.yml
rdeassis: Yes. The pressed color was not being used, so removing it to save that space too.

kubalani (approved):

PR #448, opened 2021-02-24
Make FluentUI allow app extension API only

kkj1996:
### Platforms Impacted
- [✅] iOS
- [❌] macOS

### Description of changes
Make FluentUI allow app extension API only.
#### Why creates this PR?
There are two reasons:
* It is better choice to keep this project app extension API safe.
* We want to use Fluent on our app's framework and the framework is already app extension API safe.
### Verification
This changes has no build error or warning.
### Pull request checklist

This PR has considered:
- [✅]] Light and Dark appearances
- [✅]] VoiceOver and Keyboard Accessibility
- [✅]] Internationalization and Right to Left layouts
- [✅]] Different resolutions (1x, 2x, 3x)
- [✅]] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/448)


ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -1625,6 +1625,7 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 53FC90F825673626008A06FD /* FluentUI_framework.xcconfig */;
 			buildSettings = {
+				APPLICATION_EXTENSION_API_ONLY = YES;
```
anandrajeswaran, line 1628: Can we add to FluentUI_framework.xcconfig instead so we don't have to duplicate for each config in the very difficult-to-read .pbxproj file?

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -111,7 +111,7 @@ open class ContactCollectionView: UICollectionView {
     }
 
     @objc private func updateHeightConstraint() {
-        let height = UIApplication.shared.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
+        let height = UIScreen.main.traitCollection.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
```
rdeassis, line 114: Can you explain in the PR description why this change is needed?

rdeassis (commented):
Thank you for your contribution to FluentUI.

Could you please elaborate on the PR description with more details about the reason for these changes? (the code needs to be compatible with iOS extensions, etc...)

Same thing for the verification: I believe you have an iOS extension using FluentUI and you validated the changes in it. It'll be great to have those details here as well.

ios/FluentUI/People Picker/ContactCollectionView.swift
sophialee0416: Either way we might not want to use UIScreen... #77

ios/FluentUI.xcodeproj/project.pbxproj
kkj1996: Great suggestion. I'll do it.

ios/FluentUI/People Picker/ContactCollectionView.swift
kkj1996: Thanks for your review. I've updated.

ios/FluentUI/Extensions/UIView+Extensions.swift
```diff
@@ -6,6 +6,11 @@
 import UIKit
 
 extension UIView {
+
+    var viewScreen: UIScreen {
+        return window?.screen ?? .main
```
rdeassis, line 11: The UIWindows' screen property is deprecated:
https://developer.apple.com/documentation/uikit/uiwindow/1621597-screen

We shouldn't be using it.
Please remove this computed property.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -111,7 +111,7 @@ open class ContactCollectionView: UICollectionView {
     }
 
     @objc private func updateHeightConstraint() {
-        let height = UIApplication.shared.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
+        let height = viewScreen.traitCollection.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
```
rdeassis, line 114: We should use the UIView's traitCollection instead. (here and in the other changes too)

rdeassis (changes requested):
Can you please elaborate with more details on the reasons for this change so we can benefit from the documentation point of view?

I've also requested some changes in the implementation as well.
We should be using the UIView's trait collection instead of retrieving it from the deprecated UIWindow's screen property.

Thanks.

ios/FluentUI/People Picker/ContactCollectionView.swift
```diff
@@ -111,7 +111,7 @@ open class ContactCollectionView: UICollectionView {
     }
 
     @objc private func updateHeightConstraint() {
-        let height = UIApplication.shared.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
+        let height = self.traitCollection.preferredContentSizeCategory.contactHeight(size: size.contactViewSize) + 2 * Constants.verticalInset
```
rdeassis, line 114: nit: can you remove the redundant "self." from these calls?

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -41,7 +41,7 @@ open class ContactView: UIControl {
 
         var width: CGFloat {
             var width = avatarSize.size.width
-            switch UIApplication.shared.preferredContentSizeCategory {
+            switch UIView().traitCollection.preferredContentSizeCategory {
```
rdeassis, line 44: Using [UITraitCollection.current](https://developer.apple.com/documentation/uikit/uitraitcollection/3238080-current) seems like a better option here.

Your branch seems a bit outdated. We've already bumped the target deployment version to iOS13.

ios/FluentUI/Table View/TableViewCellFileAccessoryView.swift
```diff
@@ -379,7 +381,7 @@ open class TableViewCellFileAccessoryView: UIView {
         let availableColumnSpace = cellSize.width - reservedCellSpace - width
         let columnMinWidth = Constants.columnMinWidth + Constants.columnSpacing
         let canShowFirstColumn = cellSize.width >= Constants.layoutBreakPoints[4] && availableColumnSpace > columnMinWidth
-        let canShowSecondColumn = cellSize.width >= Constants.layoutBreakPoints[5] && availableColumnSpace > 2 * columnMinWidth && UIApplication.shared.preferredContentSizeCategory.canShowSharedStatus
+        let canShowSecondColumn = cellSize.width >= Constants.layoutBreakPoints[5] && availableColumnSpace > 2 * columnMinWidth && self.traitCollection.preferredContentSizeCategory.canShowSharedStatus
```
rdeassis, line 384: nit: remove "self."?

kkj1996: Thanks!

ios/FluentUI/People Picker/ContactView.swift
kkj1996: I've aligned to main branch.

ios/FluentUI/People Picker/ContactCollectionView.swift
kkj1996: Ok.

ios/FluentUI/People Picker/ContactView.swift
```diff
@@ -200,7 +200,7 @@ open class ContactView: UIControl {
     }
 
     private func updateLabelHeight() {
-        let contactHeight = UIApplication.shared.preferredContentSizeCategory.contactHeight(size: size)
+        let contactHeight = self.traitCollection.preferredContentSizeCategory.contactHeight(size: size)
```
rdeassis, line 203: Still missing this one:

nit: remove "self."?

rdeassis (approved):

PR #451, opened 2021-02-26
Adding support for accessibility to the Avatar Vnext.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The Vnext version of the Avatar written in SwiftUI does not have a good accessibility story.
This change fixes that by adding accessibility label and value while consolidating all of its internal components into a single accessibility item.

Screen Shot 2021-02-26 at 11 08 12 AM

### Verification

1. Ran the Accessibility inspector reading through the Avatars and ensured the values correctly announced the Avatar status.
2. Deployed the local build to the real device and navigated through the same items to validate accessibility labels, values and traits.

https://user-images.githubusercontent.com/68076145/109346293-54b26580-7826-11eb-9c7a-edcd4868fddb.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/451)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
@@ -106,6 +106,8 @@ class AvatarDemoController: DemoController {
             let overflowAvatar = createAvatarView(size: size,
                                                   name: "20",
                                                   style: .overflow)
+            overflowAvatar.state.accessibilityLabel = "20 more contacts"
+
             avatarViews.append(overflowAvatar)
 
             addRow(text: size.description, items: [overflowAvatar.view], textStyle: .footnote, textWidth: 100)
--- a/ios/FluentUI/Vnext/Avatar/Avatar.swift
+++ b/ios/FluentUI/Vnext/Avatar/Avatar.swift
@@ -245,7 +245,7 @@ public struct AvatarView: View {
 
             let defaultAccessibilityText = state.primaryText ?? state.secondaryText ?? ""
             return (state.isOutOfOffice ?
-                        String(format: "Accessibility.AvatarView.LabelFormat".localized, defaultAccessibilityText, "Presence.OOF".localized) :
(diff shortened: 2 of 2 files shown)


ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -217,7 +238,22 @@ public struct AvatarView: View {
...
                              alignment: .topLeading))
 
+        let accessibilityLabel: String = {
+            if let overriddenAccessibilityLabel = state.accessibilityLabel {
+                return overriddenAccessibilityLabel
+            }
+
+            let defaultAccessibilityText = state.primaryText ?? state.secondaryText ?? ""
+            return (state.isOutOfOffice ?
+                        String(format: "Accessibility.AvatarView.LabelFormat".localized, defaultAccessibilityText, "Presence.OOF".localized) :
```
kubalani, line 248: use `localizedStringWithFormat` instead?

https://developer.apple.com/documentation/foundation/nsstring/1497301-localizedstringwithformat

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -106,6 +106,8 @@ class AvatarDemoController: DemoController {
             let overflowAvatar = createAvatarView(size: size,
                                                   name: "20",
                                                   style: .overflow)
+            overflowAvatar.state.accessibilityLabel = "20 more contacts"
```
kubalani, line 109: does the client needs to add accessibility label for all the avatars? What's the default value?

rdeassis: This is a customization scenario where the client App could set "Signed Out" as the accessibility label for displaying an avatar with no data.
Or as in this example, instead of showing just "20", it provides a more meaningful description of what it means. ("20 more contacts" for example)

kubalani (approved):

Issue #392, opened 2021-01-12
macOS Avatar backgroundColor algorithm should match iOS


Saadnajmi:
#### Platform
- [ ] iOS
- [ ] macOS

#### Describe the feature that you would like added

There is a hashing algorithm in Avatar to choose a background color given the primaryText/secondaryText of the avatar. 
On iOS this was purposefully chosen to match the FluentUI Android implementation. Let's use the same implementation for macOS, which currently uses a slightly different hashing algorithm resulting in different colors.

#### What component or utility would this be added to

macOS Avatar

#### Have you discussed this feature with our team, and if so, who

Yes. @markavitale and @harrieshin

harrieshin:
@kubalani we can close this now?

Saadnajmi:
The algorithms still don't match, so the same Name/Email won't give the same color. That was a followup task

Saadnajmi:
This is closed now? @kubalani

kubalani:
The scope of previous PR was only restricted to theme changes, I have updated the initials view algorithm to address this issue specifically.

PR #434, opened 2021-02-11
Updating background and text avatar colors.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updating the AvatarView to match the same background/foreground(text) colors as the FluentUI Vnext Avatar which was updated as part of #420 according to #338:

The design team released a new set of 30 colors for background and foreground colors that should be used for the Avatar.
These colors are calculated based on the primary and secondary properties:

Screen Shot 2021-02-01 at 11 18 01 PM

### Verification

| Light mode                                       | Dark mode                                      |
|----------------------------------------------|--------------------------------------------|
| Screen Shot 2021-02-10 at 9 09 24 PM | Screen Shot 2021-02-10 at 9 09 46 PM |
| Screen Shot 2021-02-10 at 9 09 06 PM | Screen Shot 2021-02-10 at 9 10 02 PM |

### List of all 30 colors for visual validation only (not included in the demo app)
| Light mode                                       | Dark mode                                      |
|----------------------------------------------|--------------------------------------------|
| Screen Shot 2021-02-10 at 9 22 35 PM | Screen Shot 2021-02-10 at 9 22 41 PM |
| Screen Shot 2021-02-10 at 9 22 55 PM | Screen Shot 2021-02-10 at 9 23 01 PM |
| Screen Shot 2021-02-10 at 9 23 13 PM | Screen Shot 2021-02-10 at 9 23 19 PM |
| Screen Shot 2021-02-10 at 9 23 26 PM | Screen Shot 2021-02-10 at 9 23 32 PM |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/434)

PR #455, opened 2021-03-01
Updating background and text avatar colors - MacOS

kubalani:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes

(1/2) Fix to issue #392 . The current PR addressed color and theme changes to Avatar view.  This is extension to corresponding [PR for iOS](https://github.com/microsoft/fluentui-apple/pull/434).

This change includes two updates

1. Update primary and secondary colors with new asset catalog from previous PR for iOS
2. Update Avatar color background and foreground color to consume new set of colors

### Verification

Before (Dark Theme):
Screen Shot 2021-03-01 at 2 14 58 PM

After (Dark Theme):

Screen Shot 2021-03-01 at 2 28 10 PM

Before (Light Theme):

Screen Shot 2021-03-01 at 2 16 43 PM

After (Light Theme):

Screen Shot 2021-03-01 at 2 28 51 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/455)

### DIFF
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -472,9 +472,8 @@ fileprivate extension Unicode.Scalar {
 }
 
 extension AvatarView {
-
-	/// the table of background colors for the initials views
-	static let avatarColors: [ColorSet] = [
+	/// Table of background and text colors for the AvatarView
+		static let avatarColors: [ColorSet] = [
 		ColorSet(background: DynamicColor(light: Colors.Palette.darkRedTint40.color, dark: Colors.Palette.darkRedShade30.color),
 				 foreground: DynamicColor(light: Colors.Palette.darkRedShade30.color, dark: Colors.Palette.darkRedTint40.color)),
 		ColorSet(background: DynamicColor(light: Colors.Palette.cranberryTint40.color, dark: Colors.Palette.cranberryShade30.color),
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -473,7 +473,7 @@ fileprivate extension Unicode.Scalar {
 
 extension AvatarView {
 	/// Table of background and text colors for the AvatarView
-		static let avatarColors: [ColorSet] = [
+	static let avatarColors: [ColorSet] = [
 		ColorSet(background: DynamicColor(light: Colors.Palette.darkRedTint40.color, dark: Colors.Palette.darkRedShade30.color),
 				 foreground: DynamicColor(light: Colors.Palette.darkRedShade30.color, dark: Colors.Palette.darkRedTint40.color)),
 		ColorSet(background: DynamicColor(light: Colors.Palette.cranberryTint40.color, dark: Colors.Palette.cranberryShade30.color),
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -312,10 +312,10 @@ open class AvatarView: NSView {
 		updateAppearance()
 	}
 
-	private func updateAppearance(_ appreance: NSAppearance = NSAppearance.current) {
+	private func updateAppearance(_ appearance: NSAppearance = NSAppearance.current) {
 		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
-		avatarBackgroundColor = color.background.resolvedColor(appreance)
-		initialsFontColor = color.foreground.resolvedColor(appreance)
+		avatarBackgroundColor = color.background.resolvedColor(appearance)
+		initialsFontColor = color.foreground.resolvedColor(appearance)
 	}
 
 	/// Get the color associated with a given index
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -309,7 +309,7 @@ open class AvatarView: NSView {
 		}
 
 		initialsTextField.stringValue = AvatarView.initialsWithFallback(name: contactName, email: contactEmail)
-		updateAppearance()
+		updateAppearance(self.window?.effectiveAppearance)
 	}
 
 	private func updateAppearance(_ appearance: NSAppearance? = nil) {
@@ -325,7 +325,7 @@ open class AvatarView: NSView {
 	///
 	/// - note: Internal visibility exists only for unit testing
 	@objc static func getColor(for index: Int) -> ColorSet {
-		let avatarBackgroundColors = NSColor.avatarColors
+		let avatarBackgroundColors = AvatarView.avatarColors
 		return avatarBackgroundColors[index % avatarBackgroundColors.count]
 	}
 
@@ -471,7 +471,7 @@ fileprivate extension Unicode.Scalar {
 	static let zeroWidthSpace = Unicode.Scalar(0x200B)!
 }
 
-extension NSColor {
+extension AvatarView {
 	/// the text color of the text in the initials view
 
 	/// the table of background colors for the initials views
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -30,8 +30,8 @@ open class AvatarView: NSView {
 
 		// Prefer contactEmail to contactName for uniqueness
 		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
-		avatarBackgroundColor = color.resolvedBackgroundColor()
-		initialsFontColor = color.resolvedForegroundColor()
+		avatarBackgroundColor = color.background.resolvedColor()
+		initialsFontColor = color.foreground.resolvedColor()
 		self.contactName = contactName
 		self.contactEmail = contactEmail
 		self.contactImage = contactImage
@@ -314,8 +314,8 @@ open class AvatarView: NSView {
 
 	private func updateAppearance(_ appreance: NSAppearance = NSAppearance.current) {
 		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
-		avatarBackgroundColor = color.resolvedBackgroundColor(appreance)
-		initialsFontColor = color.resolvedForegroundColor(appreance)
+		avatarBackgroundColor = color.background.resolvedColor(appreance)
+		initialsFontColor = color.foreground.resolvedColor(appreance)
 	}
 
 	/// Get the color associated with a given index
@@ -324,7 +324,7 @@ open class AvatarView: NSView {
 	/// - returns: the color table entry for the given index
 	///
 	/// - note: Internal visibility exists only for unit testing
-	@objc public static func getColor(for index: Int) -> DynamicColor {
+	@objc static func getColor(for index: Int) -> ColorSet {
 		let avatarBackgroundColors = NSColor.avatarColors
 		return avatarBackgroundColors[index % avatarBackgroundColors.count]
 	}
@@ -415,17 +415,16 @@ open class AvatarView: NSView {
 
 	open override func viewWillMove(toWindow newWindow: NSWindow?) {
 		super.viewWillMove(toWindow: newWindow)
-		if let window = newWindow {
-			appearanceObserver = window.observe(\.effectiveAppearance) { [weak self] (window, value) in
-				guard let strongSelf = self else {
-					return
-				}
-				print("window \(window), value:\(value)")
 
-				strongSelf.updateAppearance(window.appearance ?? NSAppearance.current)
-			}
-		} else {
+		guard let window = newWindow else {
 			appearanceObserver = nil
+			return
+		}
+		appearanceObserver = window.observe(\.effectiveAppearance) { [weak self] (window, _) in
+			guard let strongSelf = self else {
+				return
+			}
+			strongSelf.updateAppearance(window.appearance ?? NSAppearance.current)
 		}
 	}
 }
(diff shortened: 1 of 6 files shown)


macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -436,34 +452,69 @@ fileprivate extension Unicode.Scalar {
...
-		Colors.Palette.blueMagenta20.color,
-		Colors.Palette.pinkRed10.color,
-		Colors.Palette.gray30.color,
-		Colors.Palette.blueMagenta30.color,
-		Colors.Palette.gray20.color,
-		Colors.Palette.cyan30.color,
-		Colors.Palette.orange30.color
+	static let avatarColors: [ColorSet] = [
+		ColorSet(background: NSColor.getColor(light: Colors.Palette.darkRedTint40.color, dark: Colors.Palette.darkRedShade30.color),
+				 foreground: NSColor.getColor(light: Colors.Palette.darkRedShade30.color, dark: Colors.Palette.darkRedTint40.color)),
```
rdeassis, line 459: nit: is there a line break for the 2 params? They look misaligned on the GitHub web interface for the PR.

macos/FluentUIUnitTest/AvatarViewTests.swift
```diff
@@ -115,13 +115,16 @@ class AvatarViewTests: XCTestCase {
 		XCTAssertNil(noNameNoEmailAvatar.toolTip)
 	}
 
-	func testBackgroundColors () {
+	func testAvatarColors () {
```
rdeassis, line 118: Is there a way to test dynamic colors?
I think it's very valuable to have the colorsets compared in dark and light mode since the background and foreground values are just swapped when going from light to dark mode.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -480,3 +531,26 @@ fileprivate extension CGPath {
 		return CGPath(ellipseIn: NSRect(x: 0, y: 0, width: diameter, height: diameter), transform: nil)
 	}
 }
+
+fileprivate extension NSColor {
```
rdeassis, line 535: Isn't swiftlint complaining about this fileprivate access modifier?

rdeassis, line 535: This eventually can be leveraged from in other parts of the internal library code. Maybe move it to the Colors.swift file?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -436,34 +452,69 @@ fileprivate extension Unicode.Scalar {
...
-	///
-	/// - note: this value doesn't change for dark mode by design as our default background color table
-	/// only provides colors designed to be used with white text, even on dark mode
-    static let initialsViewTextColor: NSColor = .white
 
 	/// the table of background colors for the initials views
-	static let avatarBackgroundColors: [NSColor] = [
-		Colors.Palette.cyanBlue10.color,
-		Colors.Palette.red10.color,
-		Colors.Palette.magenta20.color,
```
harrieshin, line 448: if there are palette colors that aren't being used anymore.... should we delete them from the asset catalog?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -88,6 +90,18 @@ open class AvatarView: NSView {
 			initialsTextField.backgroundColor = avatarBackgroundColor
 		}
 	}
+	
+	/// The initials font color of the avatar view when no image is provided.
+	/// Setting this property will override the default provided color.
+	@objc open var initialsFontColor: NSColor {
```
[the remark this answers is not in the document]
kubalani: we already allow them to override background color, any client who wants to customize background color probably needs to customize font color too for accessibility.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: That seems to be the github interface, I have attached snapshot for XCode. 

Screen Shot 2021-03-01 at 4 50 50 PM

I believe we don't have line character limit in Xcode or Swiftlint so this might look different on your end. However, there is no intentional line break added in the code

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I like the idea, it would save space as well, @rdeassis  any objections?

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: Not sure why isn't this working. Seems like it's enabled [here](https://github.com/microsoft/fluentui-apple/blob/93264f87f66d89075915da14618d934924da6d60/.swiftlint.yml#L84). 

I also see extensive use of file private modifier across the [code](https://github.com/microsoft/fluentui-apple/search?q=fileprivate)

macos/FluentUI/AvatarView/AvatarView.swift
Saadnajmi: I think Github thinks tabs are 8 spaces by default. You can fix that by adding a .editorconfig file

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -480,3 +531,26 @@ fileprivate extension CGPath {
...
+		}
+	}
+
+	static func getColor(light: NSColor, dark: NSColor) -> NSColor {
+		let currentStyle = InterfaceStyle()
+		if currentStyle == .light {
+			return light
+		}
+		return dark
+	}
```
Saadnajmi, line 555: I think we very much don't like things like "isDarkMode" boolean from my memory of when we added Dark Mode to office, but I can't remember specific reasons. Is there any reason we can't just pull the asset catalog color and let the system pull the right color?

macos/FluentUIUnitTest/AvatarViewTests.swift
kubalani: Unfortunately, there isn't a trivial way to test without using a dependency injection framework like [OcMock](https://ocmock.org/swift/). We need a way to override theme values which isn't currently supported API. 

If you have any ideas, I can look into it.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: editorconfig? This is for github editor? If there is an example, I can add it to the repo.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: dynamic traits are not supported by [NSColor](https://developer.apple.com/documentation/appkit/nscolor/color_creation) creation API's. 

Hopefully, Apple will have an addition where we can load them dynamically as opposed to using memory values.

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: No objections at all if they're not being used!

macos/FluentUI/AvatarView/AvatarView.swift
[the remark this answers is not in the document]
rdeassis: For the coauthoring scenario I know that the background color needs to be customized so that each coauthor gets the color matching the canvas presence indicator. I assume the foreground color may need to be changed if they do in the same direction of tint/shade combinations for both text and background colors?

macos/FluentUI/AvatarView/AvatarView.swift
Saadnajmi: @heyimchris do you remember (from the Canvas UX days of Dark Mode integration) whether this might cause any issues? I would ask Mark but he is on parental leave.

macos/FluentUI/AvatarView/AvatarView.swift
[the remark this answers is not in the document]
harrieshin: for coauthor scenario, we should have "ring" color like ios though right? is this something office is currently using?

macos/FluentUIUnitTest/AvatarViewTests.swift
rdeassis: I think based on the issue I just reported on the other comment about light colors being displayed in both light and dark modes, we can follow an approach like:

 1. create a type that holds the 2 values of the color and returns the correct color in a computed property ([resolvedColor](https://developer.apple.com/documentation/uikit/uicolor/3238042-resolvedcolor) to mimic the iOS API)
 
 2. Subscribe to the effective appearance change inside the Avatar code so you can update the visuals of it when it changes from aqua to darkAqua (calling updateAvatarViewContents() whenever it changes)
 
 Something like this example:
 https://developer.apple.com/forums/thread/104515
 ```
private var appearanceObserver: NSKeyValueObservation?

  override func viewWillMove(toWindow newWindow: NSWindow?) {
       super.viewWillMove(toWindow: newWindow)
       if let window = newWindow {
            appearanceObserver = window.observe(\.effectiveAppearance) { [weak self] (window, change) in
                 self?.updateAppearance(forWindow: window)
            }
       } else {
            appearanceObserver = nil
       }
  }
```

File: macos/FluentUI/AvatarView/AvatarView.swift (hunk 5 above)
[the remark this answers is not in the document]
Comment (kubalani) [reply macos/FluentUI/AvatarView/AvatarView.swift:96]: I have created a different task to follow up on parity with iOS avatar view.

File: macos/FluentUI/AvatarView/AvatarView.swift (hunk 1 above)
Comment (kubalani) [reply macos/FluentUI/AvatarView/AvatarView.swift:459]: I have created a new github issue to track the addition of editorconfig file

File: macos/FluentUI/AvatarView/AvatarView.swift (hunk 4 above)
Comment (kubalani) [reply macos/FluentUI/AvatarView/AvatarView.swift:448]: It seems like these are still being references in iOS

File: macos/FluentUIUnitTest/AvatarViewTests.swift (hunk 2 above)
Comment (Saadnajmi) [reply macos/FluentUIUnitTest/AvatarViewTests.swift:118]: I had good luck with something that looked like this:
```
func startObserving() {
	kvoToken = NSApplication.shared.observe(\.effectiveAppearance) { (application, change) in
		guard self.bridge != nil else {
			return
		}
		// Do work
	}
}

func stopObserving() {
	kvoToken?.invalidate()

}
```

**Comment by Saadnajmi** [comment]
Question: Do the hashing algorithms between macOS and iOS match now? AKA, well the same name give the same colors?

File: macos/FluentUI/AvatarView/AvatarView.swift [hunk 7]
```diff
@@ -390,6 +413,21 @@ open class AvatarView: NSView {
...
 	}
 
+	open override func viewWillMove(toWindow newWindow: NSWindow?) {
+		super.viewWillMove(toWindow: newWindow)
+		if let window = newWindow {
+			appearanceObserver = window.observe(\.effectiveAppearance) { [weak self] (window, value) in
+				guard let strongSelf = self else {
+					return
+				}
+				print("window \(window), value:\(value)")
```
rdeassis, line 423: I think we don't need this print statement here.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -390,6 +413,21 @@ open class AvatarView: NSView {
 		}
 	}
 
+	open override func viewWillMove(toWindow newWindow: NSWindow?) {
+		super.viewWillMove(toWindow: newWindow)
+		if let window = newWindow {
```
rdeassis, line 418: Can we use guard let here instead of the if let?

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
```
rdeassis, line 6: Do we need to import Foundation here?

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,65 @@
...
+			}
+		}
+		return false
+	}
+}
+
+@objc(MSFDynamicColor)
+public class DynamicColor: NSObject {
+
+	private let background: NSColor
```
rdeassis, line 34: These properties should be light and dark instead of background and foreground.
And then you have just resolvedColor as a method.

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,65 @@
...
+				if appleInterfaceStyle.lowercased().contains("dark") {
+					return true
+				}
+			}
+		}
+		return false
+	}
+}
+
+@objc(MSFDynamicColor)
```
rdeassis, line 31: I think we can keep this Swift only for now in order to save in the binary size, at least for now.

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,65 @@
...
+public class DynamicColor: NSObject {
```
rdeassis, line 32: let's keep this internal instead of public.

macos/FluentUIUnitTest/AvatarViewTests.swift
```diff
@@ -115,13 +115,12 @@ class AvatarViewTests: XCTestCase {
...
 		// Cherry pick a few known values and test them
-		XCTAssertEqual(AvatarView.backgroundColor(for: 0), Colors.Palette.cyanBlue10.color)
-		XCTAssertEqual(AvatarView.backgroundColor(for: 2), Colors.Palette.magenta20.color)
-		XCTAssertEqual(AvatarView.backgroundColor(for: 3), Colors.Palette.green10.color)
-		XCTAssertEqual(AvatarView.backgroundColor(for: 6), Colors.Palette.orange20.color)
-		XCTAssertEqual(AvatarView.backgroundColor(for: 20), Colors.Palette.orange30.color)
+		XCTAssertEqual(AvatarView.getColor(for: 0),
```
rdeassis, line 120: The test I suggested is to iterate over all color sets and compare like the iOS one does it:

for each colorset:
assert -> Light mode foreground = Dark mode background
assert -> Dark mode foreground = Light mode background

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -436,34 +474,39 @@ fileprivate extension Unicode.Scalar {
...
-		Colors.Palette.gray40.color,
-		Colors.Palette.green20.color,
-		Colors.Palette.blueMagenta20.color,
-		Colors.Palette.pinkRed10.color,
-		Colors.Palette.gray30.color,
-		Colors.Palette.blueMagenta30.color,
-		Colors.Palette.gray20.color,
-		Colors.Palette.cyan30.color,
-		Colors.Palette.orange30.color
+	static let avatarColors: [DynamicColor] = [
```
rdeassis, line 479: I think the ColorSet class is still useful here to encapsulate a DynamicColor
 tuple of background/foreground.

rdeassis (commented):
Thanks @kubalani for addressing the theme changes! That's looking great!
I've added a few comments around the design of the classes. Let me know if you have questions on those.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I replaced this call with reading from `userDefaults`. The notification when `effectiveApperance` is changed doesn't provide information about theme. If there are alternatives suggestion please feel free to create a new GitHub issue or let me know.

rdeassis (commented):
It seems like there's a test failing on the PR checks.

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,70 @@
...
+
+import AppKit
+
+private let kAppleInterfaceStyle = "AppleInterfaceStyle"
+
+public extension NSAppearance {
+
+	/// Pseudo algorithm picked up from https://developer.apple.com/forums/thread/118974
+	var isDarkMode: Bool {
+		// Included for unit testing
```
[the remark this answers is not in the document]
kubalani: yes, this is used for unit test

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -29,7 +29,9 @@ open class AvatarView: NSView {
                       contactImage: NSImage? = nil) {
 
 		// Prefer contactEmail to contactName for uniqueness
-		avatarBackgroundColor = AvatarView.backgroundColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		avatarBackgroundColor = color.background.resolvedColor()
+		initialsFontColor = color.foreground.resolvedColor()
```
Saadnajmi, line 34: Note, this doesn't actually fix #392 because this algorithm differs from iOS. if both primaryText (contactName) and secondaryText (contactEmail) are available, then iOS concatenates the strings. Check out this method: https://github.com/microsoft/fluentui-apple/blob/d3126dd5097de7fd3c551e2e038bec3e5232408a/ios/FluentUI/Avatar/InitialsView.swift#L91

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -292,7 +309,13 @@ open class AvatarView: NSView {
...
 
 		initialsTextField.stringValue = AvatarView.initialsWithFallback(name: contactName, email: contactEmail)
-		avatarBackgroundColor = AvatarView.backgroundColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		updateAppearance()
+	}
+
+	private func updateAppearance(_ appreance: NSAppearance = NSAppearance.current) {
+		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		avatarBackgroundColor = color.background.resolvedColor(appreance)
+		initialsFontColor = color.foreground.resolvedColor(appreance)
```
Saadnajmi, line 318: typo in appearance?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -89,6 +91,18 @@ open class AvatarView: NSView {
...
+	/// The initials font color of the avatar view when no image is provided.
+	/// Setting this property will override the default provided color.
+	@objc open var initialsFontColor: NSColor {
+		didSet {
+			guard oldValue != initialsFontColor else {
+				return
+			}
+			needsDisplay = true
+			initialsTextField.textColor = initialsFontColor
+		}
```
Saadnajmi, line 103: Do we need to update the updateLayer() method in line 76?

That method is also called every time a light/dark switch happens btw, and is the Apple-accepted way of setting dynamic colors to a layer based view

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -161,6 +175,9 @@ open class AvatarView: NSView {
 	/// The layer used to mask the overall AvatarView to a circle
 	private let circleMask = CAShapeLayer()
 
+	/// The apperance observer to update colors when theme is changed
+	private var appearanceObserver: NSKeyValueObservation?
```
Saadnajmi, line 179: I think if we go with this approach we also need to invalidate this token on deinit. But also, could we just do the color calculation inside updateAppearance() in  updateLayer()? (I know I told you to go for this method of KVO, sorry). I'm gonna summon @anandrajeswaran because he had strong opinions about this the last time I spoke with him

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: Thanks Saad for bringing this to my attention, it doesn't seem this is related to `foregroundColor`. This is base view for avatar. I have highlighted the base view in the image below. 

Screen Shot 2021-03-04 at 12 36 24 PM

It does get called during theme switch so does any other view layout methods. Based on my understanding it doesn't need to be updated.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: `viewWillMove` is called anytime there is a change in layout hierarchy. It does set appearance to nil on line 420

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: Let's restrict the scope of this PR to colors and theme, I will add another PR to address the issue.

macos/FluentUI/DatePicker/CalendarDayButton.swift
```diff
@@ -279,11 +279,7 @@ class CalendarDayButton: NSButton {
 						   constant: CalendarDayButton.dualModeMargin)
 
 	private var isDarkMode: Bool {
-		var isInDarkAppearance = false
-		if effectiveAppearance.bestMatch(from: [.darkAqua]) == .darkAqua {
-			isInDarkAppearance = true
-		}
-		return isInDarkAppearance
+		return NSAppearance.current.isDarkMode
```
rdeassis, line 282: Keep using effectiveAppearance here.

macos/FluentUI/Extensions/Color+Apperance.swift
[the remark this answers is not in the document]
rdeassis: The implementation of this method should be just:

`return self.bestMatch(from: [.darkAqua, .aqua]) == .darkAqua`

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -390,6 +413,20 @@ open class AvatarView: NSView {
...
+
+		guard let window = newWindow else {
+			appearanceObserver = nil
+			return
+		}
+		appearanceObserver = window.observe(\.effectiveAppearance) { [weak self] (window, _) in
+			guard let strongSelf = self else {
+				return
+			}
+			strongSelf.updateAppearance(window.appearance ?? NSAppearance.current)
```
rdeassis, line 427: **window.effectiveAppearance** should do the trick for you here:

`strongSelf.updateAppearance(window.effectiveAppearance)`

macos/FluentUI/DatePicker/CalendarDayButton.swift
kubalani: The appearance instance itself isn't used in determining light/dark theme. The underlying instance is only used in test mode. During application run time the theme information is read from `userdefaults` or `effectiveAppearance` based on iOS version. 

so swapping this with another instance won't have any effect.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: This worked, thanks. Saved a lot of complexity for testing

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -434,36 +471,71 @@ fileprivate extension Unicode.Scalar {
 	static let zeroWidthSpace = Unicode.Scalar(0x200B)!
 }
 
-fileprivate extension NSColor {
+extension NSColor {
```
rdeassis, line 474: I think this shouldn't be an extension of NSColor anymore.
Semantically it belongs to the AvatarView class.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -29,7 +29,9 @@ open class AvatarView: NSView {
                       contactImage: NSImage? = nil) {
 
 		// Prefer contactEmail to contactName for uniqueness
-		avatarBackgroundColor = AvatarView.backgroundColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		let color = AvatarView.getColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		avatarBackgroundColor = color.background.resolvedColor()
```
rdeassis, line 33: pass the window.effectiveAppearance to the resolveColor here too? (and in the next line as well)

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -292,7 +309,13 @@ open class AvatarView: NSView {
 		}
 
 		initialsTextField.stringValue = AvatarView.initialsWithFallback(name: contactName, email: contactEmail)
-		avatarBackgroundColor = AvatarView.backgroundColor(for: AvatarView.colorIndex(for: contactEmail ?? contactName ?? ""))
+		updateAppearance()
```
rdeassis, line 312: pass window's effective appearance?

kubalani: not passing it will default to window's effective appearance.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: not passing it will default to window's effective appearance.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -434,36 +471,71 @@ fileprivate extension Unicode.Scalar {
 	static let zeroWidthSpace = Unicode.Scalar(0x200B)!
 }
 
-fileprivate extension NSColor {
+extension NSColor {
 	/// the text color of the text in the initials view
```
rdeassis, line 475: Can you update update the comment here?

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: nevermind this one.. it's in the initializer.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: Updated as per our discussion

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: updated as per our discussion

macos/FluentUI/Extensions/Color+Apperance.swift
```diff
@@ -0,0 +1,50 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+private let kAppleInterfaceStyle = "AppleInterfaceStyle"
```
rdeassis, line 8: remove (not needed anymore).

rdeassis (approved):
Just one last comment.

PR #461, opened 2021-03-03
[List vNext] Adding triple-line cell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added tokens + properties for a triple-lined cell.

### Verification

Made sure that adding footnote (third line) changes subline look.

| Simulator Screen Shot - iPhone 11 Pro - 2021-03-02 at 19 05 50 | Simulator Screen Shot - iPhone 11 Pro - 2021-03-03 at 19 00 52 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/461)

### DIFF
--- a/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
+++ b/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
@@ -10,7 +10,6 @@ import SwiftUI
 struct UIViewAdapter: UIViewRepresentable {
 
     var makeView: () -> UIView
-    private var tintColor: UIColor?
 
     init(_ makeView: @escaping @autoclosure () -> UIView) {
         self.makeView = makeView
@@ -24,9 +23,6 @@ struct UIViewAdapter: UIViewRepresentable {
     }
 
     func updateUIView(_ view: UIView, context: Context) {
-        if let tintColor = tintColor {
-            view.tintColor = tintColor
-        }
         // This logic should be removed wnce the "wrapping in a UIStackView" workaround is removed.
         guard let stackView = view as? UIStackView else {
             return
@@ -36,15 +32,6 @@ struct UIViewAdapter: UIViewRepresentable {
             view.removeFromSuperview()
         }
         stackView.addArrangedSubview(makeView())
-    }
-}
-
-extension UIViewAdapter {
-    func foregroundColor(_ color: UIColor) -> Self {
-        var view = self
-        view.tintColor = color
-        view.makeView().tintColor = color
-        return view
     }
 }
 
--- a/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
+++ b/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
@@ -17,11 +17,10 @@ struct UIViewAdapter: UIViewRepresentable {
     }
 
     func makeUIView(context: Context) -> UIView {
-        let view = makeView()
         // Wrapping the view passed on a StackView is a workaround for a hang that occurs
         // on iOS 13 when the view passed directly comes from a UIHostingController.view
         // property. (e.g. using the MSFAvatar.view property).
-        return UIStackView(arrangedSubviews: [view])
+        return UIStackView(arrangedSubviews: [makeView()])
     }
 
     func updateUIView(_ view: UIView, context: Context) {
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Vnext/List/ListTokens.swift
```diff
@@ -117,7 +118,7 @@ class MSFListCellTokens: MSFTokensBase, ObservableObject {
         disclosureIconForegroundColor = appearanceProxy.disclosureIconForegroundColor
         leadingTextColor = appearanceProxy.labelColor
         leadingViewColor = appearanceProxy.leadingViewColor
-        subtitleColor = appearanceProxy.sublabelColor
+        sublabelColor = appearanceProxy.sublabelColor
```
rdeassis, line 121: Since you're changing names of the properties, can you normalize all properties as "label"? (I see subtitleFont in line 138)

rdeassis: Same for textFont vs labelFont in line 139.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -79,8 +83,7 @@ struct MSFListCellView: View {
                 let labelAccessorySize = tokens.labelAccessorySize
 
                 if let leadingView = state.leadingView {
-                    UIViewAdapter(leadingView)
-                        .foregroundColor(Color(hasTitle ? tokens.backgroundColor : tokens.trailingItemForegroundColor))
+                    UIViewAdapter(leadingView, foregroundColor: hasTitle ? tokens.backgroundColor : tokens.trailingItemForegroundColor)
```
rdeassis, line 86: Why did you stop using the foregroundColor modifier? It seems to be the right thing to do here.

sophialee0416: The foreground modifier appears that it doesn’t work, along with an overlay. I had to modify the uiview’s tintcolor instead for results to show

ios/FluentUI/Vnext/List/ListTokens.swift
sophialee0416: I've updated the property names but left title/subtitle/footnote in the cell state. I figured we should distinguish that sub labels include subtitle + footnote but that they are still separate properties

ios/FluentUI/Vnext/List/ListCell.swift
rdeassis: Is it possible to retrieve the foregroundColor value and apply as the UIView tint color?
If not, creating a custom modifier for the UIViewAdapter seems like a better approach to me (instead of the initializer parameter).

ios/FluentUI/Vnext/List/ListTokens.swift
rdeassis: Okay. Something else I noticed as you renamed leadingTextColor to leadinglabelColor:
Why is it not just labelColor?

sophialee0416: Oh good catch. It was a remnant from when we used to have a trailing label value. Updated, thank you

ios/FluentUI/Vnext/List/ListCell.swift
sophialee0416: I added a custom modifier instead as I wasn't able to find a way to retrieve the foregroundColor value

ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -23,6 +24,9 @@ struct UIViewAdapter: UIViewRepresentable {
     }
 
     func updateUIView(_ view: UIView, context: Context) {
+        if let tintColor = tintColor {
```
kubalani, line 27: I think a much better alternative would be to provide a closure handler to the client where they can update the view themselves. 

```swift
    var updateView: (UIView) -> UIView
```

this way you won't need extension for every additional properties that client wants to modify.

sophialee0416:
Resolved offline: removing modifiers and leaving foreground customization to the client

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -175,7 +196,7 @@ struct ListCellButtonStyle: ButtonStyle {
         let height: CGFloat
         switch state.layoutType {
         case .automatic:
-            height = !state.subtitle.isEmpty ? tokens.cellHeightTwoLines : tokens.cellHeightOneLine
+            height = !state.footnote.isEmpty ? tokens.cellHeightThreeLines : (!state.subtitle.isEmpty ? tokens.cellHeightTwoLines : tokens.cellHeightOneLine)
```
rdeassis, line 199: nit: can you break this into 2 lines for readability?

rdeassis (approved):

PR #468, opened 2021-03-04
update main0.1 branch to have version 0.1.29

harrieshin:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

update main_0.1 to have 0.1.29 version

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/468)

### DIFF
--- a/MicrosoftFluentUI.podspec
+++ b/MicrosoftFluentUI.podspec
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'MicrosoftFluentUI'
-  s.version          = '0.1.28'
+  s.version          = '0.1.29'
   s.summary          = 'Fluent UI is a set of reusable UI controls and tools'
   s.homepage         = "https://www.microsoft.com/design/fluent/#/"
   s.license          = { :type => 'MIT', :file => 'LICENSE' }
--- a/MicrosoftFluentUI.podspec
+++ b/MicrosoftFluentUI.podspec
@@ -5,7 +5,7 @@ Pod::Spec.new do |s|
   s.homepage         = "https://www.microsoft.com/design/fluent/#/"
   s.license          = { :type => 'MIT', :file => 'LICENSE' }
   s.author           = { "Microsoft" => "fluentuinativeowners@microsoft.com"}
-  s.source       = { :git => "https://github.com/microsoft/fluentui-apple.git", :tag => "#{s.version}" }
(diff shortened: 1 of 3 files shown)


Saadnajmi:
What's the difference between the 0.1.x versions and the 0.2.x versions? The latter is both Appkit/UIKit and SwiftUI, or only SwiftUI?

MicrosoftFluentUI.podspec
```diff
@@ -1,6 +1,6 @@
 Pod::Spec.new do |s|
   s.name             = 'MicrosoftFluentUI'
-  s.version          = '0.1.28'
+  s.version          = '0.1.29'
```
rdeassis, line 3: We have to update the source (line 8) too:

s.source       = { :git => "https://github.com/microsoft/fluentui-apple.git", :branch => 'main_0.1', :tag => "#{s.version}_main_0.1" }

rdeassis, line 3: Can you also remove the extra line on line 158?
I took care of it on main only.

rdeassis:
> What's the difference between the 0.1.x versions and the 0.2.x versions? The latter is both Appkit/UIKit and SwiftUI, or only SwiftUI?

0.1.XX has the minimum requirement of iOS 11 (deployment target) while 0.2.XX requires iOS 13.
Vnext (which includes the SwiftUI code will likely be 0.3.XX.

MicrosoftFluentUI.podspec
harrieshin: for my knowledge, can the branch and tag work together in podspec? i thought it was oneway or the other

rdeassis: It can. We just need to make sure the tag is created on the main_0.1 branch.
I suggested the suffix 0.1.29**_main_0.1** just to match the branch name so we don't get confused in the future.

rdeassis (approved):

MicrosoftFluentUI.podspec
rdeassis: > They can. We just need to make sure the tag is created on the main_0.1 branch.
> I suggested the suffix 0.1.29**_main_0.1** just to match the branch name so we don't get confused in the future.

PR #472, opened 2021-03-08
make CI run on main_0.1 and vnext_prototype branches as well

harrieshin:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

make CI run on main_0.1 and vnext_prototype branches as well
add "pod lib lint" in CI to check Podspec for the project

### Verification

not applicable 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/472)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,9 +4,13 @@ on:
   push:
     branches:
       - main
+      - main_0.1
+      - vnext-prototype
   pull_request:
     branches:
       - main
+      - main_0.1
+      - vnext-prototype
 
 jobs:
   xcodebuild:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,5 +39,5 @@ jobs:
       run: scripts/xcode_select_current_version.sh
     - name: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
       run: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
-    - script: pod lib lint
-      displayName: pod lib lint
+    - name: pod lib lint
+      run: scripts/podliblint.sh
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,3 +39,5 @@ jobs:
       run: scripts/xcode_select_current_version.sh
     - name: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
       run: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
+    - script: pod lib lint
+      displayName: pod lib lint
--- a/scripts/ci.sh
+++ b/scripts/ci.sh
@@ -77,4 +77,8 @@ else
 	echo "CI Build Succeeded"
 fi
 
+echo "Pod lib lint"
+POD_LINT_SCRIPT='scripts/podliblint.sh'
+$POD_LINT_SCRIPT
+
 exit $EXIT_CODE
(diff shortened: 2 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -4,9 +4,13 @@ on:
   push:
     branches:
       - main
+      - main_0.1
+      - vnext-prototype
   pull_request:
     branches:
       - main
+      - main_0.1
+      - vnext-prototype
```
kubalani, line 13: Does this run build pipeline on all branches for evert PR? 

I am assuming a CI job should be executed based on branch it's forked from. This way if there is a build break it won't block other branches.

harrieshin: I don't know I understood your question... CI should run any PR that it wants to merge to.

.github/workflows/ci.yml
```diff
@@ -35,3 +39,5 @@ jobs:
       run: scripts/xcode_select_current_version.sh
     - name: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
       run: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
+    - script: pod lib lint
```
rdeassis, line 42: Can we have this in a .sh file in the scripts folder instead of having the command directly here?

rdeassis (commented):
Can you also update the description to mention the changes regarding running "pod lib lint" as part of the CI validation action?

.github/workflows/ci.yml
harrieshin: what's the advantage of having a .sh file for this one line?

rdeassis: It is a good practice in case we want to use it in other workflows. (like **publishCocoaPod.sh**)

One of the scenarios I can think of now is: If we need to update it in any way (add more validation commands to the script or simply changing/adding parameters of **pod lib lint** it'll be a one file change.

kubalani: My understanding about current framework is that the `.sh` script is for local validation and `.yml` script is for cloud validation. In most cases commands are executed in `.sh` script and wrapper are invoked in `.yml` . This way if the cloud build fails then we can use `.sh` script to repo error locally.

harrieshin: the yml script can still do multilines and params. but Kunal argument on .sh for local ci.sh makes sense. I can go create one.

.github/workflows/ci.yml
kubalani: CI executes PR against a branch merged with it's changes. In this case it happens to be `main`. So if I have a feature branch I assume CI is executed with my feature branch merged with main. 

I am trying to understand what's the expected behavior after this change? The current CI run associated with this PR doesn't seems to be using `vnext` or `main_0.1` branch. Is there a way to test `ci.yml` changes before merging to main?

.github/workflows/ci.yml
kubalani, line 42: How does the CI know which `cocoapod` version to use for the script?

harrieshin: I think it is whatever it has one the virtual machine. similar to Cocoapod publishing :/

rdeassis: I understand the YAML has support for params and multi lines.

The point I'm trying to make is about reusability of these commands/parameters when and if they need to be used in other YAML workflows.

rdeassis (approved):

kubalani (approved):

PR #480, opened 2021-03-13
Update segment control examples in demo app

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Now that we have the Fluent Design recommended pill style segment control update all the instance of segment control to use pill style.
- while updating the datetimepicker, I've discovered the bug that pill segment will always try to take the full window width rather than within the contained view. Update the sizeToFit calculation and constraints of the pillcontainerview so that segment control can layout correctly.
- Pill style background color can't be surfaceSecondary in dark mode because surfaceSecondary has different dark elevated color. Rather we should just define as gray950.
- Extend the SegmentControl "style" property to be changeable based on light/dark mode for datetimepicker.
- For Demo app, Color and TableViewHeaderFooterView pages are now updated to use pill style.
- Fix dark mode background color bug in DateTimePickerView.

### Verification

Tested on iPhone and iPad

| Before                                       | After light                                      | After Dark |
|----------------------------------------------|--------------------------------------------|-------------------------------------------|
| before_color | light_color|dark_color|
| before_datepicker | light_datepicker |dark_datepicker|
| before_headerfooter | light_headerfooter|dark_headerfooter|
| before_segment| light_segment |dark_segment|
|before_timepicker| light_timepicker |dark_timepicker|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/480)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
@@ -107,6 +107,7 @@ class ColorDemoController: UIViewController {
         tableView.separatorStyle = .none
 
         let stackView = UIStackView(arrangedSubviews: [segmentedControl, tableView])
+        stackView.setCustomSpacing(16, after: segmentedControl)
         stackView.axis = .vertical
         stackView.translatesAutoresizingMaskIntoConstraints = false
 
@@ -119,7 +120,7 @@ class ColorDemoController: UIViewController {
         NSLayoutConstraint.activate([
             stackView.leadingAnchor.constraint(equalTo: safeAreaLayoutGuide.leadingAnchor),
             stackView.trailingAnchor.constraint(equalTo: safeAreaLayoutGuide.trailingAnchor),
-            stackView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor),
+            stackView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor, constant: 16),
             stackView.bottomAnchor.constraint(equalTo: safeAreaLayoutGuide.bottomAnchor)
         ])
     }
@@ -131,7 +132,7 @@ class ColorDemoController: UIViewController {
         }
     }
     private lazy var segmentedControl: SegmentedControl = {
-        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return $0.name }), style: .tabs)
+        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return $0.name }), style: .primaryPill)
         segmentedControl.addTarget(self, action: #selector(segmentedControlValueChanged(sender:)), for: .valueChanged)
         return segmentedControl
     }()
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
@@ -106,8 +106,9 @@ class ColorDemoController: UIViewController {
         tableView.delegate = self
         tableView.separatorStyle = .none
 
-        let stackView = UIStackView(arrangedSubviews: [segmentedControl, tableView])
-        stackView.setCustomSpacing(16, after: segmentedControl)
+        let separator = Separator(style: .shadow, orientation: .horizontal)
+        let stackView = UIStackView(arrangedSubviews: [segmentedControl, separator, tableView])
+        stackView.setCustomSpacing(8, after: segmentedControl)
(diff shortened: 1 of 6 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -105,33 +105,42 @@ class ColorDemoController: UIViewController {
...
-
-        let safeAreaLayoutGuide = view.safeAreaLayoutGuide
+        view.backgroundColor = Colors.NavigationBar.background
 
         NSLayoutConstraint.activate([
-            stackView.leadingAnchor.constraint(equalTo: safeAreaLayoutGuide.leadingAnchor),
-            stackView.trailingAnchor.constraint(equalTo: safeAreaLayoutGuide.trailingAnchor),
-            stackView.topAnchor.constraint(equalTo: safeAreaLayoutGuide.topAnchor),
-            stackView.bottomAnchor.constraint(equalTo: safeAreaLayoutGuide.bottomAnchor)
+            stackView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
```
huwilkes, line 122: Did you test this on a notched iPhone in landscape?

harrieshin: yeah I did. this entire view controller seems not right to host the tableview inside a stack. because the cell should go all the way edge to edge but the content inset should be within safe ware bound.

huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -105,33 +105,42 @@ class ColorDemoController: UIViewController {
...
     }
+
+    override func viewWillDisappear(_ animated: Bool) {
+        super.viewWillDisappear(animated)
+        navigationController?.navigationBar.shadowImage = nil
+    }
+
     private lazy var segmentedControl: SegmentedControl = {
-        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return $0.name }), style: .tabs)
+        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return $0.name }), style: .primaryPill)
```
huwilkes, line 143: Does the segmented control still switch to green when you select green?

harrieshin: yes it does!

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
```diff
@@ -195,6 +193,14 @@ class DatePickerController: UIViewController, GenericDateTimePicker {
         }
     }
 
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if let segmentedControl = segmentedControl, previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle {
+            segmentedControl.style = (traitCollection.userInterfaceStyle == .dark) ? .onBrandPill : .primaryPill
```
rdeassis, line 200: This is interesting. Shouldn't there be a new style to cover these variations of colors between dark/light modes?
(since there's a valid requirement for these different styles)

rdeassis (approved):

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
huwilkes: The segmented control supports custom colors, so we probably don't need a third style that's onBrand light and primary dark. The onBrand style would still have a colorful background in dark mode.

rdeassis: @huwilkes I understand that it's technically possible either through custom colors or the way it's currently implemented here (setting the style based on the trait collection).
My question is more about the meaning of this variation of "primary for light and onBrand for dark".
If it's a scenario that's part of the design specs (toolkit), there should probably be a standardized way of using it instead of just setting custom colors.

harrieshin: yeah, I went back and forth about actually exposing the style property of the segment control to public. it seemed bit overkill to have another style which is practically the same color as existing style. I think because segment control is mostly focused on large title navigation bar but not really for this "edge" case of date picker which the segment control shouldn't take too much attention compared to other ui.

PR #486, opened 2021-03-22
Adding swiftLint as a build step in CI

kubalani:
### Platforms Impacted
- [X] iOS
- [X] macOS

### Description of changes

- Removed SwiftLint check for `redundant_type_annotation`. 
  Known issue in 0.43 version of SwiftLint.  https://github.com/realm/SwiftLint/issues/3578

- Adding SwiftLint check in CI. Enforced strict check to catch warnings as errors
 ```bash
     swiftlint --config --strict .swiftlint.yml
 ```
- Excluded paths for Sgen and Pods globally via config file

### Verification

Success Lint Job : https://github.com/microsoft/fluentui-apple/runs/2170521865

Screen Shot 2021-03-22 at 4 02 37 PM

Failed Lint Job: https://github.com/microsoft/fluentui-apple/runs/2170396417

Screen Shot 2021-03-22 at 2 29 47 PM

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/486)


.github/workflows/ci.yml
```diff
@@ -40,4 +40,6 @@ jobs:
     - name: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
       run: scripts/xcodebuild_wrapper.sh ${{ matrix.build_command }}
     - name: pod lib lint
-      run: scripts/podliblint.sh
\ No newline at end of file
+      run: scripts/podliblint.sh
```
harrieshin, line 43: this is partially my fault but can we make this a separate job so we only run it once per PR rather than 4 times for different build flavor?

kubalani: I can look into it.

scripts/validation.sh
```diff
@@ -0,0 +1,7 @@
+# A simple script to validate project settings.
+
+if which swiftlint >/dev/null; then
```
rdeassis, line 3: As another improvement to be part of this change, this script could be called from the script phase of the Xcode project (without the --strict param, I believe).

.github/workflows/ci.yml
kubalani: Screen Shot 2021-03-25 at 12 58 12 PM

I have updated this.

scripts/validation.sh
kubalani: that's a good idea, I will update this. The only thing is instead of using `LOCROOT` I will use config file to exclude sgen.

kubalani: I made an attempt to reuse the script. However, the XCode instead of showing warnings in IDE now just shows in phase execution script. I guess it sets some environment variables that doesn't get set correctly with this approach. Any other ideas to reuse this?

rdeassis: Let's follow up offline on this. I thought of passing the root folder as a parameter to the script (which would be resolved from the script build phase. I don't see why the IDE wouldn't show the warnings... but we can experiment with that.

.github/workflows/ci.yml
```diff
@@ -39,5 +51,3 @@ jobs:
       run: scripts/xcode_select_current_version.sh
```
harrieshin, line 51: if we already did xcdoeselect_current_version.sh above why are we running this agin?

kubalani: I have managed to get two successive builds successful. All jobs are executed on same machine. I have removed the duplicate call to xcode selection.

PR #488, opened 2021-03-23
add accessibilityHint for tabbaritemview for iOS 14

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
We would like VoiceOver to read out "Home, 1 of 3, selected" when the VO focus is on a tabbar. However, right now it just says "Home, selected" in iOS 14 device.

This change is a partial revert of commit `1e17860` : it appears to be that .tabbar accessibilityTraits no longer automatically append index of the subview in iOS 14. 
This change further improves sideTabBar experience so that voiceover focus order would be vertical navigation within the sidetabbar before it goes to other views. Also, VoiceOver will announce the index of the tabbaritemview.

### Verification
iPad and iPhone device testing for sideTabBar and tabBar scenarios

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/488)

### DIFF
--- a/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
+++ b/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
@@ -99,8 +99,8 @@
 /* Accessibility label format string for avatar view. Format: "<Name>, <Presence>". Example: "Kat, Available" */
 "Accessibility.AvatarView.LabelFormat" = "%@, %@";
 
-/* Accessibility hint for TabBarItemView to indicate its index relative to total number of items */
-"Accessibility.TabBarItemView.Hint" = "%d of %d";
+/* Accessibility hint for TabBarItemView in TabBarView. %1$d is placeholder for index and %2$d is a placeholder for total number of tab items  */
+"Accessibility.TabBarItemView.Hint" = "%1d of %2d";
 
 /* Accessibility hint for more actions button */
 "Accessibility.TableViewCell.MoreActions.Hint" = "Double tap to view more actions";
--- a/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
+++ b/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
@@ -99,6 +99,9 @@
 /* Accessibility label format string for avatar view. Format: "<Name>, <Presence>". Example: "Kat, Available" */
 "Accessibility.AvatarView.LabelFormat" = "%@, %@";
 
+/* Accessibility hint for TabBarItemView to indicate its index relative to total number of items */
+"Accessibility.TabBarItemView.Hint" = "%d of %d";
+
 /* Accessibility hint for more actions button */
 "Accessibility.TableViewCell.MoreActions.Hint" = "Double tap to view more actions";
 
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -99,6 +99,9 @@
 /* Accessibility label format string for avatar view. Format: "<Name>, <Presence>". Example: "Kat, Available" */
 "Accessibility.AvatarView.LabelFormat" = "%@, %@";
 
+/* Accessibility hint for TabBarItemView to indicate its index relative to total number of items */
+"Accessibility.TabBarItemView.Hint" = "%d of %d";
+
```
Saadnajmi, line 104: Do we need to specify the format of "%d of %d" like we do above?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -123,6 +124,7 @@ open class SideTabBar: UIView {
         addInteraction(UILargeContentViewerInteraction())
 
         accessibilityTraits = .tabBar
+        shouldGroupAccessibilityChildren = true
```
kubalani, line 127: is this being referenced anywhere?

kubalani (approved):

ios/FluentUI/Tab Bar/SideTabBar.swift
harrieshin: what do you mean? it is an iOS api

ios/FluentUI/Tab Bar/TabBarView.swift
```diff
@@ -40,10 +40,15 @@ open class TabBarView: UIView {
...
 
-            for item in items {
+            for (index, item) in items.enumerated() {
                 let tabBarItemView = TabBarItemView(item: item, showsTitle: showsItemTitles)
                 let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleTabBarItemTapped(_:)))
                 tabBarItemView.addGestureRecognizer(tapGesture)
+
+                // seems like iOS 14 `.tabBar` accessibilityTrait doesn't seem to read out the index automatically
+                if #available(iOS 14.0, *) {
+                    tabBarItemView.accessibilityHint = String(format: "Accessibility.TabBarItemView.Hint".localized, index + 1, numberOfItems)
```
rdeassis, line 50: I think it won't have a practical difference in the outcome here since the index is an integer, but we're using [localizedStringWithFormat:](https://developer.apple.com/documentation/foundation/nsstring/1497301-localizedstringwithformat) as a standard for these calls.

For example: https://github.com/microsoft/fluentui-apple/blob/4bd410eef400cb1340bec4b18122cf89b47dfef6/ios/FluentUI/SegmentedControl/SegmentedControl.swift#L624

rdeassis: Same in line 283.

rdeassis (approved):

PR #490, opened 2021-03-24
[FHL] Initial support for multiple images in the Fluent Button

AllieStoyanova:
Initial changes for adding different images for different ButtonStates

### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

1) Added new public enum ButtonState
2) Modified test app for the custom button
3) Set different ButtonState when we update button properties:
  a) when button isEnabled is set to False, set buttonState to .disabled
  b) when button isPressed, we set buttonState to .selected (**TODO**: need to discuss with design/FX if selected will be set until the button is pressed again)
  c) all other states, for now set buttonState to .normal
  d) **TODO**: need to implement .hover state when mouse is over the button

### Verification

 Validated on the FluentUITestApp both ImageOnly and other options for imagePosition
Tested with Light and Dark Mode

Tested keyboard navigation, seems like currently buttons don't support SPACE pressed when keyboard selected

Tested with VO: VO reads button, but didn't perform selection with VO+SPACE

| Screenshot or description before this change | happy to provide video screenshot|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/490)


Saadnajmi:
Just a note: design specifically did not support hover state effects when they designed the button, because the macOS system buttons don't have hover state effects

harrieshin:
^you can add Gifs in the PR description or screenshots

macos/FluentUI/Button/Button.swift
```diff
@@ -546,6 +591,18 @@ public enum ButtonStyle: Int, CaseIterable {
 	case borderless
 }
 
+/// Indicates the state of the button for providing different images
+@objc(MSFButtonState)
+public enum ButtonState: Int, CaseIterable {
```
harrieshin, line 596: is there particular reason you made it iterable?

macos/FluentUI/Button/Button.swift
```diff
@@ -139,6 +146,32 @@ open class Button: NSButton {
...
+	
+	/// Dict of images for different button states
+	private var imagesForButtonState = [ButtonState: NSImage]()
+	
+	public func addImage(forButtonState: ButtonState, image: NSImage) {
+		self.imagesForButtonState[forButtonState] = image;
+	}
+	
+	private func updateImageForButtonState() {
+		if imagesForButtonState[self.buttonState!] != nil {
```
harrieshin, line 168: don't use !, also you should reuse the value right? if let bar = buttonState, let foo = imageforbuttonState[bar] {

macos/FluentUI/Button/Button.swift
```diff
@@ -139,6 +146,32 @@ open class Button: NSButton {
...
+			}
+			updateImageForButtonState()
+		}
+	}
+	
+	/// Dict of images for different button states
+	private var imagesForButtonState = [ButtonState: NSImage]()
+	
+	public func addImage(forButtonState: ButtonState, image: NSImage) {
+		self.imagesForButtonState[forButtonState] = image;
```
harrieshin, line 164: dont use self unless it is needed

macos/FluentUI/Button/Button.swift
```diff
@@ -185,6 +224,12 @@ open class Button: NSButton {
 			guard oldValue != isEnabled else {
 				return
 			}
+			if isEnabled {
+				buttonState = ButtonState.normal;
```
harrieshin, line 228: .normal would be the right convention

macos/FluentUI/Button/Button.swift
```diff
@@ -128,6 +132,9 @@ open class Button: NSButton {
 			guard wantsLayer == true else {
 				preconditionFailure("wantsLayer must be set so that the image is rendered on the layer")
 			}
+			if image != nil {
+				addImage(forButtonState: ButtonState.normal, image: image!)
```
harrieshin, line 136: don't use '!' through out this file

macos/FluentUI/Button/Button.swift
```diff
@@ -546,6 +591,18 @@ public enum ButtonStyle: Int, CaseIterable {
 	case borderless
 }
 
+/// Indicates the state of the button for providing different images
+@objc(MSFButtonState)
+public enum ButtonState: Int, CaseIterable {
+	case normal
+	
+	case selected
```
harrieshin, line 599: how is this different from pressed?

harrieshin (changes requested):

macos/FluentUI/Button/Button.swift
```diff
@@ -139,6 +146,32 @@ open class Button: NSButton {
...
+				return
+			}
+			updateImageForButtonState()
+		}
+	}
+	
+	/// Dict of images for different button states
+	private var imagesForButtonState = [ButtonState: NSImage]()
+	
+	public func addImage(forButtonState: ButtonState, image: NSImage) {
```
Saadnajmi, line 163: Could this be encapsulated in one public variable of type [ButtonState: NSImage] with a getter and setter?

macos/FluentUI/Button/Button.swift
```diff
@@ -139,6 +146,32 @@ open class Button: NSButton {
 			}
 		}
 	}
+	
+	/// the current state of the button
+	private var buttonState: ButtonState? {
```
Saadnajmi, line 151: nit: move private variables to the bottom: https://microsoft.github.io/swift-guide/FileContentOrdering.html

Saadnajmi: But also (this can be followup after this is merged) we could make this public, and then make `isPressed` private? I think @chripear made `isPressed` public just to make things easier to test in the test app, and this seems like the better place to do it

macos/FluentUI/Button/Button.swift
Saadnajmi: Asked @rdeassis, selected is not used but is meant for something like a checkbox. So we should probably name the state pressed here

macos/FluentUI/Button/Button.swift
```diff
@@ -546,6 +591,18 @@ public enum ButtonStyle: Int, CaseIterable {
 	case borderless
 }
 
+/// Indicates the state of the button for providing different images
+@objc(MSFButtonState)
```
rdeassis, line 595: This type name will clash with our Vnext Button State if when we bring it to MacOS:
https://github.com/microsoft/fluentui-apple/blob/b2ba3d28873d4d688d14af59de1fd340d312f11e/ios/FluentUI/Vnext/Button/Button.swift#L10

Can you change it to something like **MSFButtonInteractionState**?

macos/FluentUI/Button/Button.swift
```diff
@@ -185,6 +224,12 @@ open class Button: NSButton {
 			guard oldValue != isEnabled else {
 				return
 			}
+			if isEnabled {
```
rdeassis, line 227: You can also use the ternary operator here:

`buttonState = isEnabled? .normal : .disabled`

macos/FluentUI/Button/Button.swift
```diff
@@ -162,6 +195,12 @@ open class Button: NSButton {
 				return
 			}
 			updateContentTintColor()
+			if isPressed {
```
rdeassis, line 198: You can also use the ternary operator here:

`buttonState = isPressed ? .selected : .normal`

macos/FluentUI/Button/Button.swift
rdeassis: The suggested swift pattern for this case is:

```
if let buttonImage = image {
    addImage(forButtonState: ButtonState.normal, image: buttonImage)
}
```

macos/FluentUI/Button/Button.swift
rdeassis, line 151: Does the variable need to be optional? (**ButtonState?** instead of just **ButtonState**)
You could make it non-optional and assign a default value.

This would solve the explicitly unwrapping issues throughout the file (**buttonsState!**).

macos/FluentUI/Button/Button.swift
AllieStoyanova: no particular reason, just saw we use it in other enums, will remove it

macos/FluentUI/Button/Button.swift
AllieStoyanova: @Saadnajmi thanks for pointing me to the guide, will reorder :)

AllieStoyanova: It doesn't have to be optional. Will be happy to always have it set as .normal as default

Saadnajmi: small nit: I think "rest" is the more aligned name than normal

rdeassis (commented):
Hi @AllieStoyanova,

Do you have swiftlint installed in your machine?
https://realm.github.io/SwiftLint/

Some of the checks are failing due to swiftlint violations:
https://github.com/microsoft/fluentui-apple/pull/490/checks?check_run_id=2241747596#step:5:206
```
Linting 'TestSeparatorViewController.swift' (200/207)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:151:47: error: Trailing Semicolon Violation: Lines should not have trailing semicolons. (trailing_semicolon)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:149:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:153:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:159:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:185:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:209:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:211:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:271:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:281:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:590:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
/Users/runner/work/fluentui-apple/fluentui-apple/macos/FluentUI/Button/Button.swift:592:1: error: Trailing Whitespace Violation: Lines should not have trailing whitespace. (trailing_whitespace)
```

Thanks!

AllieStoyanova:
@rdeassis I just pushed my new changes and all builds are failing. I was able to build locally FluentUI-macOS and FluentUITestApp-macOS. Any idea what the issue might be?

macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -193,6 +193,10 @@ class TestButtonViewController: NSViewController, NSMenuDelegate {
 		let customTitle = "Atom"
 		let customImage = NSImage(named: TestButtonViewController.nonTemplateImage)!
 		customImage.isTemplate = true
+		let imageForSelectedState = NSImage(named: NSImage.actionTemplateName)!
```
harrieshin, line 196: >imageForSelectedState [](start = 6, length = 21)

can we create a new section for this behavior? i am not sure if people would understand why for selected case it will show one image and disabled state another without context and reading the this test code line by line.

macos/FluentUI/Button/Button.swift
```diff
@@ -140,6 +148,18 @@ open class Button: NSButton {
 		}
 	}
 
+	public func addImage(forButtonState: ButtonState, image: NSImage) {
+		imagesForButtonState[forButtonState] = image
+	}
+
```
Saadnajmi, line 154: We should have clear documentation to distinguish the difference between this method and the public property `image`. For example, does setting `image` set it for all states? or just rest? I think I would personally prefer just having one property instead of a property and a method. Something like:

```swift
exampleButton.imageForState = [
    .rest: restImage,
    .pressed: disabledImage
    .disabled: disabledImage
]
```
or just removing the image property altogether, and having the `addImage()` method take an optional parameter for state, that defaults to setting it for all the states

```swift
exampleButton1.addImage(imageForAllStates)

exampleButton2.addImage(restImage, state: .rest)
```

These are both just ideas though! Whatever is most ergonomic / easy to understand on the client side. I might go see how other Fluent buttons / frameworks have solved this as well.

Saadnajmi: Also, to use this method in objective c, I think you need to addd the `@objc` decorator to the method.

harrieshin: I think it should be closer to iOS https://developer.apple.com/documentation/uikit/uibutton/1623997-setimage

Saadnajmi: Ah, the iOS method does exactly what I was about to propose thinking about it more (using the rest/normal state as the default if other states aren't set). Thanks @harrieshin !

rdeassis (commented):
Hey @AllieStoyanova, 

Is the PR ready for another look by the reviewers?
I can only see a merge to update your branch, but didn't see changes to address the comments.

Please let us know when the PR is ready.

Thank you.

macos/FluentUI/Button/Button.swift
AllieStoyanova: So in iOS the ImageView property is readOnly, and user is suppose to use setImage:ForState: and set it for at least   UIControlStateNormal (which can be equivalent of our .rest case). Are you guys (FX) okay with the button being alligned with IOS more than Mac in this case. (I will make the image property be readOnly then)

rdeassis:
Hi @AllieStoyanova,

This PR had no activity in the last 3 months.
Are you still planning to work on this PR going forward?
If not, can you close it and reopen when you get a chance to work on it?

Thank you!

rdeassis:
Hi @AllieStoyanova,

I'll close the PR since it's now stale.
Feel free to reopen and reach out when you can work on it.

Thank you!

PR #491, opened 2021-03-25
Match Avatar Initial's color algorithm on MacOS with iOS/Android

kubalani:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes

Fix to issue #392 . Updating algorithms used to select foreground and background to match on other platforms. 

Screen Shot 2021-03-24 at 7 44 05 PM

### Verification

| Screen Shot 2021-03-24 at 5 39 33 PM | Screen Shot 2021-03-24 at 9 19 25 PM |

| Screen Shot 2021-03-24 at 9 23 12 PM| Screen Shot 2021-03-24 at 9 21 45 PM
 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/491)


macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -409,22 +429,34 @@ open class AvatarView: NSView {
...
+			combined = ""
+		}
+
+		let colors = AvatarView.avatarColors
+		let combinedHashable = combined as NSString
+		let hashCode = Int(abs(javaHashCode(combinedHashable)))
+		return colors[hashCode % colors.count]
+	}
+
+	/// To ensure iOS/MacOS and Android achieve the same result when generating string hash codes (e.g. to determine avatar colors) we've copied Java's String implementation of `hashCode`.
```
harrieshin, line 450: we should take one step further and actually have a .swift file that is shared between ios and mac projects

kubalani: we have a lot of scope to share code across both repos, the handling of pod spec and framework is beyond the scope of this PR. This PR is required to update devmain code for MacOS Avatar view that I intend to merge by Monday to unblock other PR's

macos/FluentUI/DynamicColor/DynamicColor.swift
```diff
@@ -28,7 +28,7 @@ public class DynamicColor: NSObject {
 	}
 
 	/// resolves color based on theme
-	@objc func resolvedColor(_ appearance: NSAppearance? = nil) -> NSColor {
+	@objc public func resolvedColor(_ appearance: NSAppearance? = nil) -> NSColor {
 		guard let appearance = appearance else {
 			return self.light
 		}
```
Saadnajmi, line 34: Completely off topic for this review, but if you wanted to make DynamicColor.swift be an actual NSColor instead of an NSObject, you can look into what we do in react-native-macos (https://github.com/microsoft/react-native-macos/blob/master/React/Base/macOS/RCTDynamicColor.m)
It's not the prettiest though :P

Saadnajmi (approved):

macos/FluentUI/DynamicColor/DynamicColor.swift
```diff
@@ -28,7 +28,7 @@ public class DynamicColor: NSObject {
 	}
 
 	/// resolves color based on theme
-	@objc func resolvedColor(_ appearance: NSAppearance? = nil) -> NSColor {
+	@objc public func resolvedColor(_ appearance: NSAppearance? = nil) -> NSColor {
```
harrieshin, line 31: can you give me context why this is public now?

kubalani: This logic could be reused in devmain to detect theme changes instead of duplicating code in devmain.

macos/FluentUI/DynamicColor/DynamicColor.swift
kubalani: Thanks Saad, I really like the idea. And code is not that bad :)

I initially had this as an extension but then I changed it to `NSObject` to have parity with iOS. At some point I would like to consolidate code across platforms for reusability and keep the interface same helps to accomplish it down the line.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -409,22 +429,34 @@ open class AvatarView: NSView {
...
-	public static func colorIndex(for identifyingString: String) -> Int {
-		return identifyingString.utf16.enumerated().reversed().reduce(0) { (hashCode, enumeratedCodePoints) -> Int in
-			// upcast our UInt16s to standard Ints as that's how the hashing algorithm works in other codebases
-			let integerCodePoint = Int(enumeratedCodePoints.element)
-			let shift = enumeratedCodePoints.offset % 8
-			return hashCode ^ Int((integerCodePoint << shift) + (integerCodePoint >> (8 - shift)))
+	@objc(getInitialsColorSetFromPrimaryText:secondaryText:)
+	public static func getInitialsColorSet(fromPrimaryText primaryText: String?, secondaryText: String?) -> ColorSet {
+		// Set the color based on the primary text and secondary text
+		var combined: String
```
harrieshin, line 435: does this has to be var instead of let?

kubalani: its mutated on line 439 and 441 so this needs to be a var. If you are referring to initializing this one based on condition then yes we can. Since the condition to initialize this is complex I am concerned that would make code readability much harder. Is there any other specific reason you would to make it immutable?

harrieshin: the recommendation is always let unless there is a reason to be var in swift. Isn't combined only set once depending on which if statement ?

kubalani: if `combined` is defined as a `let` variable then it won't be accessible outside of its enclosing if/else. That's the only reason for using `var`

macos/FluentUI/DynamicColor/DynamicColor.swift
Saadnajmi: I think we have similar code in devmain that already does this. Unless there was a specific scenario that needed this, I would not make it public unless necessary.

macos/FluentUI/AvatarView/AvatarView.swift
Saadnajmi:
```swift
let combined: String = {
  //... logic
}()
```

Saadnajmi: ^ I think that works. I feel like I've used it at least once

kubalani: Sync'ed with Saad offline the above conditional initialization didn't work. @Harrie you have any other ideas?

rdeassis: @kubalani if you just change it to let it'll work because you're initializing it in one of the 3 paths of the if statement. And it's not changed afterwards.

```
                let combined: String
		if let secondaryText = secondaryText, let primaryText = primaryText, secondaryText.count > 0 {
			combined = primaryText + secondaryText
		} else if let primaryText = primaryText {
			combined = primaryText
		} else {
			combined = ""
		}
```

Alternatively you can also do the closure initialization (as @Saadnajmi suggested):
Not sure what didn't work there?

```
		let combined: String = {
			if let primaryText = primaryText {
				if let secondaryText = secondaryText, secondaryText.count > 0 {
					return primaryText + secondaryText
				}

				return primaryText
			}

			return ""
		}()
```

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Although I understand that refactoring this code and start sharing more across mac/ios is beyond the scope of this task, we really need to start laying out foundation for that.
Do we have a task in ADO to track that?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -325,11 +325,31 @@ open class AvatarView: NSView {
...
+	/// - parameter identifyingString: the unique identifying string for the user
+	///
+	/// - Returns: a unique index generated by walking backwards through the letters in the \p identifyingString
+	/// creating a cumulative hash
+	///
+	/// - note: the returned index is not limited to any size and should not be considered safe for indexing into an array without
+	/// first checking the size of the array in question
+	@objc(colorIndexForIdentifyingString:)
+	@available(*, deprecated, message: "Use getInitialsColorSetFromPrimaryText:secondaryText: instead")
+	public static func colorIndex(for identifyingString: String) -> Int {
```
rdeassis, line 344: This is being put back for the Outlook scenario only? Or all devmain apps will still use it?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -325,11 +325,31 @@ open class AvatarView: NSView {
 	///
 	/// - note: Internal visibility exists only for unit testing
 	@objc(getColorForIndex:)
+	@available(*, deprecated, message: "Use getInitialsColorSetFromPrimaryText:secondaryText: instead")
```
rdeassis, line 328: Does this one still need to be public? It seems like it's not used in devmain.

rdeassis: Line 360 has a deprecation directive pointing to this one which is also deprecated.

`	@available(*, deprecated, message: "Use getColorForIndex instead")`

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -325,11 +325,31 @@ open class AvatarView: NSView {
...
 	public static func getColor(for index: Int) -> ColorSet {
 		let avatarBackgroundColors = AvatarView.avatarColors
 		return avatarBackgroundColors[index % avatarBackgroundColors.count]
 	}
 
+	/// Returns a color table index for a given display name
```
rdeassis, line 334: Is this a table index or just a hash?
The Int returned by this function would have to divided by the table entry count and the remainder (mod) of it would be the index of the table, correct?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -325,11 +325,31 @@ open class AvatarView: NSView {
...
+	/// - parameter identifyingString: the unique identifying string for the user
+	///
+	/// - Returns: a unique index generated by walking backwards through the letters in the \p identifyingString
```
rdeassis, line 337: Is it unique? If it's a hash being generated from the string, uniqueness is not guaranteed.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: just replacing `let` with `var` did worked. Swift doesn't like the second initialization

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: outlook scenario only

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: thanks for catching this, updated the comment

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: yes, I didn't alter the original comment but I think the author's intent was that.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I don't know for sure, I haven't modified the implementation of original method.

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Based on your code review open in devmain, it seems like all the current uses in devmain will still use this, no?

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Let's make the comment accurate with what the code is doing!

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Let's remove "unique" from this comment. It may be misleading.

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Swift does not allow a second assignment because "let" define a constant that can only be initialized (first assignment).
You can still declare the constant and initialize it later.
Swift will also catch if you try to use a constant before it's initialized.

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: Is there an ADO item to track that?

macos/FluentUI/AvatarView/AvatarView.swift
rdeassis: The documentation comment seems a bit off too:
"Get the **color** associated with a given index"

it's actually returning a ColorSet object. Can you improve the comments there?

Also, what's the internal visibility comment about?
"- note: Internal visibility exists only for unit testing"

macos/FluentUI/DynamicColor/DynamicColor.swift
rdeassis: Yes! Let's not make it public or @objc visible unless it's needed!
This will always have an impact on the binary size.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: yes, after we merge this PR these methods will still be required for outlook but rest of cocoaui will use the new API's

macos/FluentUI/DynamicColor/DynamicColor.swift
kubalani: I've reached out to Saad offline to see if there is any alternative. This method is required in devmain to retrieve theme variant of NSColor color.

kubalani: @Saadnajmi  do you have any objections to exposure to these methods? This method is required for because cocoaui consumes color API for their avatar view rendering

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I have updated the description of the method, the mod usage is already described in note section of the comment.

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I have removed unique from the description.

macos/FluentUI/DynamicColor/DynamicColor.swift
Saadnajmi: If we have to, then we have to 🤷🏾
I would file a bug on cocoaui to make sure they request the color on light/dark changes, even if we provide the right colors for them with this API regardless

macos/FluentUI/AvatarView/AvatarView.swift
kubalani: I have updated the comment. The note is no longer valid because I changed the test case in previous iteration.

Screen Shot 2021-04-01 at 8 44 26 AM
 
https://github.com/microsoft/fluentui-apple/pull/455/files#diff-48e9b1830c83b38133ae54b9dca476f4ab456571c9d162135bd0ac6e1ff65c8f

macos/FluentUI/DynamicColor/DynamicColor.swift
kubalani: I discussed offline with @Saadnajmi and @rdeassis  offline the possibility of using asset catalog to resolve colors. 
https://developer.apple.com/documentation/xcode/supporting_dark_mode_in_your_interface?language=objc

This would increase binary size as we need to duplicate colors. I am resolving this conversation and I have created another ADO task to continue this discussion. Please feel free to reopen if there are other concerns with the use of `DynamicColor`

rdeassis (approved):

PR #495, opened 2021-03-30
Update CommandBarButton.swift

owenconnolly:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This fixes two related issues with updating command button state:
- when a command button with label is disabled, the label font color is incorrect because the isEnabled property of the button's label is not appropriately set; the property should always match the button enabled state
- when the state of a command has changed, it may update the icon to display, or may change between a label and icon; the current code does not cater for this, and the logic should be updated to always show the current icon if there is one, or else show the label

### Verification

Verified in local test client which supports these scenarios:
- label color is correct for both enabled/disabled state
- command button updates to correctly display the current icon or text (where there is no icon)

| Before                                       
|----------------------------------------------|
| Enabled and Disabled State |
enabled-and-disabled

| After                                      
|----------------------------------------------|
| Disabled State (enabled state is unchanged) | 
disabled-state

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/495)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBarButton.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButton.swift
@@ -56,9 +56,9 @@ class CommandBarButton: UIButton {
         isSelected = isPersistSelection && item.isSelected
 
         // always update icon and title as we only display one; we may alterenate between them, and the icon may also change
-        let iconImage: UIImage? = item.iconImage
+        let iconImage = item.iconImage
         setImage(iconImage, for: .normal)
-        setTitle(((iconImage != nil) ? nil : item.title), for: .normal)
+        setTitle(iconImage != nil ? nil : item.title, for: .normal)
         titleLabel?.isEnabled = isEnabled
         titleLabel?.font = item.titleFont
     }
--- a/ios/FluentUI/Command Bar/CommandBarButton.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButton.swift
@@ -55,12 +55,12 @@ class CommandBarButton: UIButton {
         isEnabled = item.isEnabled
         isSelected = isPersistSelection && item.isSelected
 
-        if item.iconImage == nil {
-            setTitle(item.title, for: .normal)
-            if let font = item.titleFont {
-                titleLabel?.font = font
-            }
-        }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -55,12 +55,12 @@ class CommandBarButton: UIButton {
...
         isSelected = isPersistSelection && item.isSelected
 
-        if item.iconImage == nil {
-            setTitle(item.title, for: .normal)
-            if let font = item.titleFont {
-                titleLabel?.font = font
-            }
-        }
+        // always update icon and title as we only display one; we may alterenate between them, and the icon may also change
+        let iconImage: UIImage? = item.iconImage
```
rdeassis, line 59: I think you don't need the explicit type here since item.iconImage is of type **UIImage?** .
You can just do the following:

`let iconImage = item.iconImage`

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -55,12 +55,12 @@ class CommandBarButton: UIButton {
...
+        setImage(iconImage, for: .normal)
+        setTitle(((iconImage != nil) ? nil : item.title), for: .normal)
```
rdeassis, line 61: nit: parenthesis style

`setTitle(iconImage != nil ? nil : item.title, for: .normal)`

ios/FluentUI/Command Bar/CommandBarButton.swift
owenconnolly: Yep, I wanted the code to be clear that it was optional and we should also deliberately set to null, but the comment will cover that, thanks!

ios/FluentUI/Command Bar/CommandBarButton.swift
owenconnolly: Updated (the second brackets were added by Xcode, the first were intended for readability)

rdeassis (approved):

PR #500, opened 2021-03-31
Command Item handle tapped event should include the event source

owenconnolly:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There are two changes here, one in the Command Bar APIs, and the other a test app change:
- for consumers on iOS versions < 14.0 the OS menu items cannot be used; in this case where the consumer wishes to present a submenu they will likely want to do so from the view for the button which sent the event; the handler should include the UIButton which sent the event
- in a previous PR (#495) a fix was added for disabled button labels, so this also adds a sample disabled label command button to the test app

### Verification

- the test app correctly renders the disabled label command button
- in a host integrating with the CCB (and these changes applied locally) we can present a sub-menu from the command button which sent the event

| After                                      |
|--------------------------------------------|
| disabled-text-test-app |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/500)


ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -9,7 +9,7 @@ public typealias CommandBarItemGroup = [CommandBarItem]
 
 @objc(MSFCommandBarItem)
 open class CommandBarItem: NSObject {
-    public typealias ItemTappedHandler = ((CommandBarItem) -> Void)
+    public typealias ItemTappedHandler = ((UIButton, CommandBarItem) -> Void)
```
harrieshin, line 12: curious, does it need to know it is a button?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -68,13 +70,17 @@ class CommandBarDemoController: DemoController {
                 return UIImage(named: "keyboardDock24Regular")
             case .textStyle:
```
harrieshin, line 71: this can be case .textstyle, .disabledText:

ios/FluentUI/Command Bar/CommandBarItem.swift
owenconnolly: No, just a UIView, I will update to reflect that, thanks!

PR #510, opened 2021-04-06
initial prototype for BottomSheetViewController

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Initial prototype for bottomsheetviewcontroller
- There were couple of limitation of DrawerController when interactive view was being used, VoiceOver or Voice Control (iOS assistive technology) won't be able to access the view behind the UIPresentationController. This viewcontroller now directly animates the UIView frame size to handle expand and collapse pan gestures instead of customizing presentationTransition animations.
- When user pans or tap resize handle to show expand the bottomsheet, there is a dimming view that animates in by showing darker shadow.
- When user pans or tap resize handle to collapse the bottomsheet, there is a dimming view that animates out.
- Client can set preferred expanded height and collapse height
- According to Design, when BottomSheetViewController is shown in the iPad regular horizontal size, it shows up as fixed size floating bar and it is not expandable.

### Verification

|iPhone |Simulator Screen Shot - iPhone 11 Pro - 2021-04-05 at 22 45 25|Simulator Screen Shot - iPhone 11 Pro - 2021-04-05 at 22 48 59|
|iPad| Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-05 at 22 43 58|Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-05 at 22 44 16|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/510)


lcapkovic:
Screen Shot 2021-04-06 at 10 42 02 AM
Not sure if this is just a side effect of how the screenshot was made, or the actual view rounded corners, but a bottom sheet should only have the top corners rounded.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,392 @@
...
+                futureState = .collapse
+            }
+        }
+
+        animator?.isReversed = shouldReverse
+        addAnimatorCompletion(to: futureState, completion: nil)
+        animator?.continueAnimation(withTimingParameters: nil, durationFactor: 0)
+    }
+
+    private func panningProgress(in verticalPos: CGFloat) {
```
rdeassis, line 300: 4 topics I wanted to bring up for discussion:

**1.  Content scroll view interaction while the bottomsheet is being resized:**
There is an issue with the content view while the bottomsheet is being resized (during the animation). The icons are changing sizes and the scrollview is also tracking the gesture and being scrolled.

**2. Is the scrollview (content) supposed to support horizontal scrolling too?**
Related to the item above, I ask this question because if we're not passing the gesture to the content view during the scroll gesture, we need to provide a way for the calling code to allow that on a horizontal/vertical basis.

**3. Safe area insets support?**
In the horizontal orientation (video below) the notch covers the rounded corners and potentially one of the options of the sheet. Is there any definition from design on whether we should support both scenarios (respecting or not the safe areas) or provide a setting for that?

**4. Do we intend to provide the resistance bounce when the user tries to dismiss the bottom sheet?**
This is more of an UX consideration, but I think that the resistance effect when the user tries to drag the bottom sheet down to dismiss it is a great clue that the sheet does not allow dismissing.

https://user-images.githubusercontent.com/68076145/113758852-3d8a5180-96c9-11eb-8008-05f9e9f5c82c.mov

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -5,10 +5,10 @@
 
 import UIKit
 
-class TabBarItemView: UIView {
+public class TabBarItemView: UIView {
```
rdeassis, line 8: Is this turning public just because of the Demo app?
If that's the only goal, I think we should find a way around it so it does not add up on the final binary size.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,392 @@
...
+        }
+    }
+}
+
+extension BottomSheetViewController: UIGestureRecognizerDelegate {
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer, shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        if  !shouldShowFloatingViewForPad(), !resizingHandleView.isHidden, let scrollView = (otherGestureRecognizer as? UIPanGestureRecognizer)?.view as? UIScrollView, scrollView.contentOffset.y >= (scrollView.contentSize.height - scrollView.frame.size.height) || (scrollView.contentOffset.y <= 0) {
+            // If scroll view has reached the bottom or top bring the bottom sheet pan in action too.
+            return true
+        } else {
```
rdeassis, line 388: nit: remove the else condition and always return false in the end?

rdeassis:
I found another glitch on the gesture handling on iPad (video below):
When I try to scroll sideways I can see one of the options appearing on the top right corner of the sheet and from that point onwards the scrolling doesn't work.
If I go back and load the demo controller again it's reset, but reproable again.


https://user-images.githubusercontent.com/68076145/113761985-eedeb680-96cc-11eb-994c-717a52d477fc.mov

harrieshin:
> I found another glitch on the gesture handling on iPad (video below):
> When I try to scroll sideways I can see one of the options appearing on the top right corner of the sheet and from that point onwards the scrolling doesn't work.
> If I go back and load the demo controller again it's reset, but reproable again.
> 
>  bottom_sheet_ipad_scroll.mov

right not able to scroll in bottomsheet is by design but the demo controller do needs to lock the scrolling in this case.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
harrieshin: 1. I think it is more of the details of the collectionviewcontroller setup. I think we are still in discussion if the collection view be part of fluent control or not.
2. as far as I know, no horizontal scrolling is not supported and that is by design
3. no, detail design on safe area so I was making my best guess.
4. don't think that was considered by design but could bring it up to discussion.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,392 @@
...
+        if currentState == .expand {
+            resizingHandleView.accessibilityLabel = "Accessibility.Drawer.ResizingHandle.Label.Collapse".localized
+            resizingHandleView.accessibilityHint = "Accessibility.Drawer.ResizingHandle.Hint.Collapse".localized
+        } else {
+            resizingHandleView.accessibilityLabel = "Accessibility.Drawer.ResizingHandle.Label.Expand".localized
+            resizingHandleView.accessibilityHint = "Accessibility.Drawer.ResizingHandle.Hint.Expand".localized
+        }
+    }
+
+// MARK: Dimming View utilities
```
kubalani, line 347: nit: `MARK:- Dimming View utilities` instead of `MARK: Dimming View utilities`

ios/FluentUI/Tab Bar/TabBarItemView.swift
harrieshin: changing direction and remove this file change

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+        animator?.continueAnimation(withTimingParameters: nil, durationFactor: 0)
+    }
+
+    private func panningProgress(in verticalPos: CGFloat) {
+        if animator?.isRunning ?? false {
+           return
+        }
+
+        let windowHeight = view.window?.frame.height ?? 0
+        var progress = -1 * verticalPos / (windowHeight - collapsedHeight)
```
lcapkovic, line 318: This means that when the preferredExpandedHeight is lower than the full screen, the pan gesture no longer maps 1:1 to the finger moving on the screen and it feels kinda weird. Generally sheets will move with the finger 1:1, regardless of where the sheet is going to stop. Once the sheet reaches the expanded height, we could add resistance where the pan no longer moves it 1:1, and on release it bounces back to the max height. We could also reuse that for the resistance bounce in the opposite direction that rdeassis suggested.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    @objc public weak var delegate: BottomSheetViewControllerDelegate?
+    /// BottomSheetViewController's view height in collapsed mode.
+    @objc public var collapsedHeight: CGFloat = Constants.collapsedHeight {
+        didSet {
+            updateFrame()
+        }
+    }
+
+    /// BottomSheetViewController's view height in expanded mode. If the value is -1, it will grow to the its windows safe area height.
+    @objc public var preferredExpandedHeight: CGFloat = -1 {
```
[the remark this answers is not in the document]
lcapkovic: Also, low value here results in weird behavior where the sheet goes down when you pull up.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    /// BottomSheetViewController's view width in iPad when horizontal size is regular
+    @objc public var preferredContentWidthForPad: CGFloat = Constants.contentWidthForPad {
+        didSet {
+            updateFrame()
+        }
+    }
+
+    @objc public var showResizeHandle: Bool = true {
```
lcapkovic, line 48: Should this be public? Do we ever want an expandable sheet without the handle or vice versa? It could just be an internal side effect of whether the sheet is expandable or not.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    private var animator: UIViewPropertyAnimator?
+    private var contentViewController: UIViewController
+    private var currentState: BottomSheetViewState = .collapse
+    private let containerView: UIStackView = {
+        let view = UIStackView()
+        view.axis = .vertical
+        return view
+    }()
+
+    private lazy var dimmingView: DimmingView = {
```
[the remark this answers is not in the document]
lcapkovic: One bug I found - expanding to a low height (to keep the back button visible) and clicking the back button will leave this view in the hierarchy forever.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+    @objc public var showResizeHandle: Bool = true {
+        didSet {
+            if oldValue != showResizeHandle {
+                resizingHandleView.isHidden = !showResizeHandle
+            }
+        }
+    }
+
```
lcapkovic, line 55: Are there plans for an API to make the sheet static (not expandable)? Kinda relates to the showResizeHandle comment.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    private func panningStart() {
+        if animator?.isRunning ?? false {
+           return
+        }
+
+        setupAnimator()
+    }
+
+    private func panningStop(in verticalPos: CGFloat, velocity: CGFloat) {
+        panGestureRecognizer.isEnabled = false
```
lcapkovic, line 283: When I swipe ~10 times, I can get the sheet in a state where it's collapsed and the animation is over, but the panGestureRecognizer is permanently stuck in a disabled state so I can't swipe anymore. Only way to get it unstuck is to click the resize handle. Maybe the animation completion handler isn't reliable?

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+            resizingHandleView.accessibilityHint = "Accessibility.Drawer.ResizingHandle.Hint.Collapse".localized
+        } else {
+            resizingHandleView.accessibilityLabel = "Accessibility.Drawer.ResizingHandle.Label.Expand".localized
+            resizingHandleView.accessibilityHint = "Accessibility.Drawer.ResizingHandle.Hint.Expand".localized
+        }
+    }
+
+// MARK: - Dimming View utilities
+    private func setupDimmingView() {
+        if let rootView = view.superview {
```
lcapkovic, line 361: I wonder if there's a possible architecture where we move everything one level down, so self.view is transparent and the bottom sheet + the dimming view are the sibling children of self.view. 
Injecting into the superview sort of imposes the same constraints on the client (view.superview should cover the full app screen because everything should dim), but we're not as explicit about it. Would also get rid of the responsibility to clean up the dimming view.
@anandrajeswaran thoughts?

lcapkovic (changes requested):

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
harrieshin: i think i saw one possible example in Word app spec that bottom sheet is locked and not expandable.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
harrieshin: i think the problem here is that as soon as we try to make self.view transparent, we run into the same accessibility issues as before.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
harrieshin: i think that could be a possibility based on variation i have sseen before.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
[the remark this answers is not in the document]
harrieshin: based on the design i have seen so far there isn't mid step that i have seen. if we do support multiple epansion steps we need appropriate accessibility string design to communicate that to our users via voice over as well. i do think we would have to extend the state enums

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+        }
+    }
+
+    @objc public var showResizeHandle: Bool = true {
+        didSet {
+            if oldValue != showResizeHandle {
+                resizingHandleView.isHidden = !showResizeHandle
+            }
+        }
+    }
```
anandrajeswaran, line 54: Use a computed property instead of a stored property so there is a single source of truth?

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    public required init?(coder aDecoder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+// MARK: - Lifecycle methods
+
+    open override func viewDidDisappear(_ animated: Bool) {
+        removeDimmingView()
+        super.viewDidDisappear(animated)
+    }
```
anandrajeswaran, line 102: I'm in general wary of doing work on viewDidDisappear. Usually it can either move to the thing that is telling it to disappear or it can move to viewWillAppear or the like so all the code to reliably set up the view lives in the same place

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+
+    open override func viewDidLoad() {
```
anandrajeswaran, line 104: This looks a bit more like a loadView than a viewDidLoad? viewDidLoad is additional configuration after setting up the hierarchy (usually from a xib), this just looks like we're setting up the hierarchy

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+        updateResizingHandleViewAccessibility()
+    }
+
+    public override func viewWillLayoutSubviews() {
+        super.viewWillLayoutSubviews()
+        if animator?.state == .active {
+            // panning started, don't update the height
+            return
+        }
+        updateFrame()
```
anandrajeswaran, line 140: This seems a bit scary - it looks like this could call setNeedsLayout internally to updateFrame and calling that from viewWillLayoutSubviews is only needed when we require a 2-pass layout where we can carefully and predictably avoid a loop. Is that actually the case here? If so we should probably add comments to help preserve that potential source of fragility

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
```diff
@@ -0,0 +1,404 @@
...
+    }
+
+// MARK: - private helpers
+
+    private func updateFrame() {
```
anandrajeswaran, line 145: Minor: updateFrame makes it sound like it only changes the frame. This does view hierarchy manipulation and changes properties on its views so could probably use a less specific name

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
anandrajeswaran: @harrieshin is that what causes the accessibility issues? I thought it had to do with the view controllers being child view controllers as opposed to sibling view controllers, but am likely not remembering correctly

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
lcapkovic: I know, I mean it might be more natural to expose that through an "isExpandable" flag, rather than referencing the resize handle. The handle is more of a UI detail of how we represent expandability.

ios/FluentUI/BottomSheet/BottomSheetViewController.swift
anandrajeswaran: Never mind. @lcapkovic corrected me offline. The accessibility issues I remember were caused by using presentViewController, which accessibility interprets to mean at least a semi-modal presentation. Curious to hear more about what kind of issues you saw when making self.view transparent (i.e. a default view that simply contains other views)

harrieshin:
closing the PR for Lukas to take over with separate branch

PR #512, opened 2021-04-07
add support for border in macOS avatarView

harrieshin:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

- outline color doesn't seem to have any color value right now. Use the same default avatar border color as iOS. Also, rename it to "borderColor"
- rename outlineView to BorderView and directly draw the border stroke
- new contentView property that either has initialsView or imageView as subview depending on the style. The contentView directly has the circle shape mask. When border is shown, there is an inset to show a gap between border stroke and its content.
- add "hasBorder" and "borderColor" properties in AvatarView

Test app changes
- update Test Avatar to include a button to dynamically show rainbow border colors
- update Test Avatar sizes to match iOS Avatar size ramp

### Verification

Border size updates when avatarSize changes
Border color changes when client changes the property color.

| light                                       | dark                                      |
|Screen Shot 2021-04-06 at 9 01 32 PM| Screen Shot 2021-04-06 at 9 01 26 PM|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/512)


macos/FluentUITestViewControllers/TestAvatarViewController.swift
```diff
@@ -68,10 +69,11 @@ class TestAvatarViewController: NSViewController {
 
 	/// Create a single avatar view from a given size and test identity
 	private static func avatarView(size: CGFloat, identity: TestIdentity) -> AvatarView {
-		return AvatarView(avatarSize: size,
+		let avatarView = AvatarView(avatarSize: size,
```
rdeassis, line 72: nit: indentation of parameters on lines 73 to 75.

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -318,6 +346,24 @@ open class AvatarView: NSView {
...
+		initialsTextField.font = textFieldFont
+
+		// This constraint will only exist if we're using the sizing view approach to center our initials
+		initialsSizingViewHeightConstraint?.constant = ceil(textFieldFont.capHeight)
+		contentViewHeightConstraint?.constant = diameter
+		contentViewWidthConstraint?.constant = diameter
+	}
+
+	private func diameterForContentCircle() -> CGFloat {
+		return hasBorder ? avatarSize - (AvatarView.borderWidth * 2) - (AvatarView.contentInset * 2) : avatarSize
```
rdeassis, line 364: Not sure if the design spec is clear on this behavior, but for the iOS Vnext the avatar size is always the same and the ring and gap/inset add to its size instead of subtracting. (if the goal is consistency across platforms)

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -368,8 +414,13 @@ open class AvatarView: NSView {
 	/// the maximum number of initials to be displayed when we don't have an image
 	static let maximumNumberOfInitials: Int = 2
 
-	/// the color used for the outline view
-	static let outlineColor = NSColor(named: "AvatarView/outlineColor", bundle: FluentUIResources.resourceBundle)!
+	/// the color used for the border
+	static let defaultBorderColor = NSColor(named: "AvatarView/borderColor", bundle: FluentUIResources.resourceBundle)!
+
+	static let borderWidth: CGFloat = 2.0
```
rdeassis, line 420: Is the ring thickness always the same value irrespective of the Avatar size? (same for the contentInset below)

rdeassis (approved):
Comments regarding topics that would be addressed on the figma spec (which I could not find a clear definition for it).

macos/FluentUI/AvatarView/AvatarView.swift
Saadnajmi: Not sure if relevant but when I asked Kaan if the size ramp on avatar for iOS and macOS should be the same, he said yeah. So we probably want to match?

macos/FluentUI/AvatarView/AvatarView.swift
harrieshin: currently that's what i did because a) we don't have size ramp. b) i don't have exact design for the ring :(

macos/FluentUI/AvatarView/AvatarView.swift
harrieshin: that's interesting. so no longer avatar vnext will be 72 x 72 it will be 80 x 80 if there is a border ring? but again i am following existing iOS avatar calculation than vnext for sure.

harrieshin: following up with design team

harrieshin: i think the real direction should come once vnext Swiftui avatar comes to macOS this fall and we will actually have the sizeRamp for macOS avatar view which we don't right now. So, to keep avatarSize intact, i will do the subtraction rather than addition direction.

PR #514, opened 2021-04-09
Add Floating Action styles to vNext button

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added the accent floating action button and the subtle floating action buttons as styles to the vNext button.

### Verification

| FloatingActionButtoniPhoneLight | FloatingActionButtoniPhoneDark |
| FloatingActionButtonNotchediPhoneLight | FloatingActionButtonNotchediPhoneDark |
| FloatingActionButtoniPadLight | FloatingActionButtoniPadDark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/514)


ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -22,6 +22,7 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
         let state = targetButton.state
         let isDisabled = state.isDisabled
         let isPressed = configuration.isPressed
+        let isActive = isDisabled || isPressed
```
harrieshin, line 25: why is isActive = isDisabled or isPressed?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -37,9 +38,11 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
                     .font(Font(tokens.textFont))
                     .fontWeight(.medium)
                     .multilineTextAlignment(.center)
+                    .frame(minHeight: tokens.textMinimumHeight)
```
harrieshin, line 41: works well with larger text?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -52,7 +55,15 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
                                     :
                         AnyView(RoundedRectangle(cornerRadius: tokens.borderRadius)
                                     .fill(Color(isDisabled ? tokens.disabledBackgroundColor :
-                                                    (isPressed ? tokens.highlightedBackgroundColor : tokens.backgroundColor)))))
+                                                    (isPressed ? tokens.highlightedBackgroundColor : tokens.backgroundColor)))
+                                    .shadow(color: isActive ? tokens.pressedShadow1Color : tokens.restShadow1Color,
```
harrieshin, line 59: if feels little weird to me that we are setting shadowcolor (none color) for regular button as well

ios/FluentUI/Vnext/Button/Button.swift
huwilkes: The button has a smaller shadow when it's in a state other than rest. I thought "isActive" was a bit better than "isNotAtRest". But there's probably a better option.

ios/FluentUI/Vnext/Button/Button.swift
huwilkes: The text size of the button doesn't change when you change the system font size. Though, we don't support the large content viewer, either, so I should probably add that.

ios/FluentUI/Vnext/Button/Button.swift
huwilkes: I don't know if there is a way around that if we're adding the floating button as a style through design tokens. It seems like it would go against the design of design tokens to try and check what style we're making a button for.

kubalani:
We normally add vnext-ptototype tag on PR's to indicate it isn't merged to main.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -84,11 +84,11 @@ Typography:
   subheadline: "Font(subheadline)"
 
 Border:
-  radius: [ none: 0pt, small: 2pt, medium: 4pt, large: 8pt, xlarge: 12pt ]
+  radius: [ none: 0pt, small: 2pt, medium: 4pt, large: 8pt, xLarge: 12pt , circleSmall: 24pt, circleLarge: 26pt]
```
kubalani, line 87: Is `circleSmall` and `circleLarge` defined in Figma?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -10,10 +10,15 @@ class ButtonDemoController: DemoController {
 
     override func viewDidLoad() {
         super.viewDidLoad()
+        view.backgroundColor = Colors.surfaceSecondary
 
         for size in MSFButtonSize.allCases {
             addTitle(text: "\(size.description.capitalized) size")
             for style in MSFButtonStyle.allCases {
+                let floatingStyle = style == .accentFloating || style == .subtleFloating
```
kubalani, line 18: nit: If you would like to filter doing it before the loop would be easy to understand.

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: We have discussed this today and we've got a bug where large text is only updated after the app is restarted. I remember seeing the text size getting updated as we changed the large text config in the Settings app. We'll need to fix that bug.

tools/sgen/input/FluentUIStyle.yml
huwilkes: We have a definition of circle as 50% the height, but I don't think that actually translates to design tokens very well.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -246,6 +257,42 @@ MSFSecondaryButtonTokens extends MSFButtonTokens:
...
+  borderRadius: [ small: $Border.radius.circleSmall, medium: $Border.radius.circleSmall, large: $Border.radius.circleLarge ]
+  iconColor: [ rest: $Colors.Foreground.neutral3, hover: $Colors.Foreground.neutral3, pressed: $Colors.Foreground.neutral3, selected: $Colors.Foreground.brandRest, disabled: $Colors.Foreground.neutralDisabled ]
+  iconSize: [ small: $Icon.size.medium, medium: $Icon.size.medium, large: $Icon.size.medium ]
+  padding: [ small: $Spacing.small, medium: $Spacing.small, large: $Spacing.medium ]
+  textColor: [ rest: $Colors.Foreground.neutral3, hover: $Colors.Foreground.neutral3, pressed: $Colors.Foreground.neutral3, selected: $Colors.Foreground.brandRest, disabled: $Colors.Foreground.neutralDisabled ]
+  textMinimumHeight: [ small: $Icon.size.medium, medium: $Icon.size.medium, large: $Icon.size.medium ]
+  textAdditionalHorizontalPadding: [ small: $Spacing.xSmall, medium: $Spacing.xSmall, large: $Spacing.xxSmall ]
+  shadow1Color: [ rest: $Shadow.shadow8.color1, hover: $Shadow.shadow2.color1, pressed: $Shadow.shadow2.color1, selected: $Shadow.shadow2.color1, disabled: $Shadow.shadow2.color1]
```
rdeassis, line 287: Can you make MSFSubtleFloatingActionButtonTokens extend MSFAccentFloatingActionButtonTokens so you can save some of the duplicate overrides like the shadows?
I think sGen will support that with no issues.

huwilkes: I get this error when I try to build the demo app after changing the autogenerated files:
Property 'backgroundColor' with type 'FluentUIStyle.MSFSubtleFloatingActionButtonTokensAppearanceProxy.MSFSubtleFloatingActionButtonTokensbackgroundColorAppearanceProxy' cannot override a property with type 'FluentUIStyle.MSFAccentFloatingActionButtonTokensAppearanceProxy.MSFAccentFloatingActionButtonTokensbackgroundColorAppearanceProxy'

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Alternatively, we can create a category for Persona sizes as they are used in ListCell + Avatar tokens to avoid any definition confusion in the future. It will probably increase the sgen size though and I'm not sure how beneficial the tradeoff would be

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: This is good timing. I'm working on some fixes of animation related problems in the Avatar.
I'll need to add a conditional modifier (#515) for a similar scenario of applying a modifier only if needed instead of creating a neutral value when it does not apply.

@huwilkes you'll be able to use it and remove the none values of the shadow tokens (adding it only to the floating button styles).

rdeassis: You can use **.modifyIf()** do something like:

```
AnyView(RoundedRectangle(cornerRadius: tokens.borderRadius)
                                    .fill(Color(isDisabled ? tokens.disabledBackgroundColor :
                                                    (isPressed ? tokens.highlightedBackgroundColor : tokens.backgroundColor)))
                                    .modifyIf(tokens.style == .accentFloating || tokens.style == .subtleFloating, modifications: { view in
                                        view.shadow(color: isActive ? tokens.pressedShadow1Color : tokens.restShadow1Color,
                                                    radius: isActive ? tokens.pressedShadow1Blur : tokens.restShadow1Blur,
                                                    x: isActive ? tokens.pressedShadow1DepthX : tokens.restShadow1DepthX,
                                                    y: isActive ? tokens.pressedShadow1DepthY : tokens.restShadow1DepthY)
                                            .shadow(color: isActive ? tokens.pressedShadow2Color : tokens.restShadow2Color,
                                                    radius: isActive ? tokens.pressedShadow2Blur : tokens.restShadow2Blur,
                                                    x: isActive ? tokens.pressedShadow2DepthX : tokens.restShadow2DepthX,
                                                    y: isActive ? tokens.pressedShadow2DepthY : tokens.restShadow2DepthY)
                                    })
                        ))
```

ios/FluentUI/Vnext/Button/Button.swift
harrieshin: if this boolean variable is about shadow than the variable name should be more geared towards shawdow? either state enum has func like shouldShowSmallShadow()->Bool or something that will give more context how this variable is needed?

huwilkes: isDisabled is coming from the state, which isPressed is coming from the configuration, so adding the function to the state enum probably wouldn't work. But I did change the name.

tools/sgen/input/FluentUIStyle.yml
huwilkes: Switched the implementation to use .clipShape(Capsule()) instead

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -53,6 +58,17 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
                         AnyView(RoundedRectangle(cornerRadius: tokens.borderRadius)
                                     .fill(Color(isDisabled ? tokens.disabledBackgroundColor :
                                                     (isPressed ? tokens.highlightedBackgroundColor : tokens.backgroundColor)))))
+        .modifyIf(tokens.style == .accentFloating || tokens.style == .subtleFloating, { view in
```
kubalani, line 61: this condition is referenced [here](https://github.com/microsoft/fluentui-apple/blob/76e2d98aef406d73c1a09d2cce434684c123fe3e/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift#L18) too

You can add convenience method in enum `isFloatingStyle`

huwilkes: I added the var, but I didn't think making it more accessible than internal was a good idea, so I didn't change the use in the demo controller. I don't think the users will generally have a need to access an internal var that will tell them what type of style they're trying to make.

ios/FluentUI/Vnext/Button/ButtonTokens.swift
```diff
@@ -15,6 +15,8 @@ import SwiftUI
     case accentFloating
     /// For use with small and large sizes only
     case subtleFloating
+
+    var isFloatingStyle: Bool { return self == .accentFloating || self == .subtleFloating }
```
rdeassis, line 19: style nit: can you add a new line after the opening curly brace and another one before the closing curly brace?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -37,9 +38,13 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
                     .font(Font(tokens.textFont))
                     .fontWeight(.medium)
                     .multilineTextAlignment(.center)
+                    .frame(minHeight: tokens.textMinimumHeight)
                 }
         }
         .padding(tokens.padding)
+        .modifyIf(state.text?.isEmpty == false, { view in
```
rdeassis, line 45: nit: !state.text?.isEmpty instead?

ios/FluentUI/Vnext/Button/ButtonTokens.swift
```diff
@@ -29,6 +35,8 @@ class MSFButtonTokens: MSFTokensBase, ObservableObject {
     @Published public var interspace: CGFloat!
     @Published public var padding: CGFloat!
     @Published public var textFont: UIFont!
+    @Published public var textMinimumHeight: CGFloat!
```
rdeassis, line 38: Are these included in the design spec (figma) now?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -12,8 +12,8 @@ Colors:
   Background:
     - neutral1: "FluentUIColor(light: NamedColor(white), dark: NamedColor(black), darkElevated: NamedColor(grey4))"
     - neutral2: "FluentUIColor(light: NamedColor(grey98), dark: NamedColor(grey4), darkElevated: NamedColor(grey8))"
-    - neutral3: "FluentUIColor(light: NamedColor(grey96), dark: NamedColor(grey8), darkElevated: NamedColor(grey16))"
-    - neutral4: "FluentUIColor(light: NamedColor(grey94), dark: NamedColor(grey16), darkElevated: NamedColor(grey24))"
+    - neutral3: "FluentUIColor(light: NamedColor(grey96), dark: NamedColor(grey8), darkElevated: NamedColor(grey12))"
```
rdeassis, line 15: Were these darkElevated and dark variations updated in the figma design spec?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -117,12 +117,13 @@ Opacity:
   clear: 0.0f
 
 Shadow:
-    shadow2: [ blur1: 2pt, color1: $Colors.Elevation.lowElevation1, x1: 0pt, y1: 1pt, blur2: 2pt, color2: $Colors.Elevation.lowElevation2, x2: 0pt, y2: 0pt ]
-    shadow4: [ blur1: 4pt, color1: $Colors.Elevation.lowElevation1, x1: 0pt, y1: 2pt, blur2: 2pt, color2: $Colors.Elevation.lowElevation2, x2: 0pt, y2: 0pt ]
-    shadow8: [ blur1: 8pt, color1: $Colors.Elevation.lowElevation1, x1: 0pt, y1: 4pt, blur2: 8pt, color2: $Colors.Elevation.lowElevation2, x2: 0pt, y2: 0pt ]
-    shadow16: [ blur1: 16pt, color1: $Colors.Elevation.lowElevation1, x1: 0pt, y1: 8pt, blur2: 8pt, color2: $Colors.Elevation.lowElevation2, x2: 0pt, y2: 0pt ]
-    shadow28: [ blur1: 28pt, color1: $Colors.Elevation.highElevation1, x1: 0pt, y1: 14pt, blur2: 8pt, color2: $Colors.Elevation.highElevation2, x2: 0pt, y2: 0pt ]
-    shadow64: [ blur1: 64pt, color1: $Colors.Elevation.highElevation1, x1: 0pt, y1: 32pt, blur2: 8pt, color2: $Colors.Elevation.highElevation2, x2: 0pt, y2: 0pt ]
+    none: [ blur1: 0pt, color1: $Colors.Shadow.clear, x1: 0pt, y1: 0pt, blur2: 0pt, color2: $Colors.Shadow.clear, x2: 0pt, y2: 0pt ]
```
rdeassis, line 120: I believe we can now remove this "none" variation since we can conditionally apply the shadows, right?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -246,6 +256,42 @@ MSFSecondaryButtonTokens extends MSFButtonTokens:
...
+  iconSize: [ small: $Icon.size.medium, medium: $Icon.size.medium, large: $Icon.size.medium ]
+  padding: [ small: $Spacing.small, medium: $Spacing.small, large: $Spacing.medium ]
+  textFont: [ small: "Font(15px, regular)", medium: "Font(15px, regular)", large: "Font(17px, regular)" ]
+  textColor: [ rest: $Colors.Foreground.neutral3, hover: $Colors.Foreground.neutral3, pressed: $Colors.Foreground.neutral3, selected: $Colors.Foreground.brandRest, disabled: $Colors.Foreground.neutralDisabled ]
+  textMinimumHeight: [ small: $Icon.size.medium, medium: $Icon.size.medium, large: $Icon.size.medium ]
+  textAdditionalHorizontalPadding: [ small: $Spacing.xSmall, medium: $Spacing.xSmall, large: $Spacing.xxSmall ]
+  shadow1Color: [ rest: $Shadow.shadow8.color1, hover: $Shadow.shadow2.color1, pressed: $Shadow.shadow2.color1, selected: $Shadow.shadow2.color1, disabled: $Shadow.shadow2.color1]
+  shadow1Blur: [ rest: $Shadow.shadow8.blur1, hover: $Shadow.shadow2.blur1, pressed: $Shadow.shadow2.blur1, selected: $Shadow.shadow2.blur1, disabled: $Shadow.shadow2.blur1 ]
```
rdeassis, line 287: Since all the shadow definitions here are the same for both floating buttons, let's just move it to MSFButtonTokens so we generate less code.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -225,14 +225,24 @@ MSFOverflowAvatarTokens extends MSFAvatarTokens:
...
+  borderRadius: [ small: $Border.radius.medium, medium: $Border.radius.large, large: $Border.radius.xLarge ]
   borderSize: [ small: $Border.size.none, medium: $Border.size.none, large: $Border.size.none ]
   iconColor: [ rest: $Colors.Foreground.brandRest, hover: $Colors.Foreground.brandHover, pressed: $Colors.Foreground.brandPressed, selected: $Colors.Foreground.brandSelected, disabled: $Colors.Foreground.brandDisabled ]
   iconSize: [ small: $Icon.size.xSmall, medium: $Icon.size.small, large: $Icon.size.small ]
   interspace: [ small: $Spacing.xxSmall, medium: $Spacing.xSmall, large: $Spacing.xSmall ]
   padding: [ small: $Spacing.xSmall, medium: $Spacing.small, large: $Spacing.large ]
   textColor: [ rest: $Colors.Foreground.brandRest, hover: $Colors.Foreground.brandHover, pressed: $Colors.Foreground.brandPressed, selected: $Colors.Foreground.brandSelected, disabled: $Colors.Foreground.brandDisabled ]
   textFont: [ small: $Typography.footnote, medium: $Typography.subheadline, large: $Typography.subheadline ]
+  textMinimumHeight: [ small: $Spacing.none, medium: $Spacing.none, large: $Spacing.none ]
```
rdeassis, line 236: I think we can also move textMinimumHeight and textAdditionalHorizontalPadding to AP_MSFButtonTokens and apply them conditionally for the floating styles.
This will save another duplicate definition for the floating tokens.

rdeassis (approved):

PR #516, opened 2021-04-14
Date/Time Picker: Fix hour and minute accessibility values

chripear:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Bug Reference
https://dev.azure.com/microsoftdesign/fluentui-native/_workitems/edit/9568

### Description of changes
VoiceOver reads out incorrect values for the hour and minute in the date/time picker, due to two separate issues:
- Localizable.stringsdict has an error in the hour pluralization rule, where the NSStringLocalizedFormatKey is defined as `"%#@hours@"`, but the variable "hours" is undefined.  The pluralization rule is defined under the key "Variable" instead.  Because the system has no definition for the "hours" variable, it returns a localized string with that variable unsubstituted, which VO reads out directly: `"%#@hours@"`.  The fix is to rename the "Variable" key to "hours" in each language's Localizable.strings.
- The expected arguments in the localized string templates are integers (NSStringFormatValueTypeKey: d), but we supplied strings instead.  Those strings get formatted into garbage integers, which VO reads out: for example, "5032945 minutes".  The fix is to supply the expected integer arguments.

### Verification
Used test app with VoiceOver.  Tested with English and French languages.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/516)

### DIFF
--- a/ios/FluentUI/Resources/Localization/ar.lproj/Localizable.stringsdict
+++ b/ios/FluentUI/Resources/Localization/ar.lproj/Localizable.stringsdict
@@ -1,78 +1,78 @@
-﻿<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
-  <dict>
-    <key>Accessibility.DateTime.Hour.Value</key>
-    <dict>
-      <key>NSStringLocalizedFormatKey</key>
-      <string>%#@hours@</string>
-      <key>Variable</key>
-      <dict>
-        <key>NSStringFormatSpecTypeKey</key>
-        <string>NSStringPluralRuleType</string>
-        <key>NSStringFormatValueTypeKey</key>
-        <string>d</string>
-        <key>zero</key>
-        <string>الساعة %d</string>
-        <key>one</key>
-        <string>الساعة %d</string>
-        <key>two</key>
-        <string>الساعة %d</string>
-        <key>few</key>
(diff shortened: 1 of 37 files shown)


Saadnajmi:
This isn't the kinda thing where changing one sting file will prompt touchdown to change the rest of them?

ios/FluentUI/Resources/Localization/ar.lproj/Localizable.stringsdict
```diff
@@ -1,78 +1,78 @@
...
-        <string>%d من النتائج التي تم العثور عليها من الدليل</string>
-      </dict>
-    </dict>
-  </dict>
-</plist>
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+	<key>Accessibility.DateTime.Hour.Value</key>
```
harrieshin, line 5: what's the indentation changes here?

harrieshin (approved):

ios/FluentUI/Resources/Localization/ar.lproj/Localizable.stringsdict
chripear: That's all Xcode.  The changes I made were through its plist editor.

chripear:
> This isn't the kinda thing where changing one string file will prompt touchdown to change the rest of them?

Talked this over with Harrie.  I'm not really sure since it's not a typical strings file change... I'm changing some metadata in the pluralization rules.  I guess we'll see what happens after I merge it...

PR #520, opened 2021-04-15
DatePicker (macOS): Add weekday to accessibility help

chripear:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
When VoiceOver reads the label for a CalendarDay, it uses the locale-and-user-settings dependent `medium` format.  For a standard US locale, an example of that description is "Dec 4, 2019" (where VoiceOver pronounces "Dec" as "December").  This change adds the weekday to that description after a short pause, via the `accessibilityHelp` property.

Alternative approaches that were considered:
- Change the calendar day grid to a table, and add the weekdays as column headers.  This would be a pretty invasive and time-consuming change.
- Use the `full` date format, which (for United States locale with no customizations) includes the weekday: for example, "Tuesday, December 4, 2019".  In the US locale, VoiceOver speaks the weekday first, which isn't necessarily the first thing the user wants to hear when trying to navigate through the calendar.  The format is customizable by the user, and might be missing the weekday for some users, not to mention other locales.

### Verification
Tested using the Date Picker section of FluentUITestApp, in Engilsh (US) and French locales.

| Before | After |
|-|-|
| Screen Shot 2021-04-15 at 3 32 24 PMScreen Shot 2021-04-15 at 3 08 48 PM<p>(short pause)</p>Screen Shot 2021-04-15 at 3 37 37 PM | Screen Shot 2021-04-15 at 3 32 24 PMScreen Shot 2021-04-15 at 3 08 48 PM<p>(short pause)</p>Screen Shot 2021-04-15 at 3 08 52 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/520)

### DIFF
--- a/macos/FluentUI/DatePicker/CalendarView.swift
+++ b/macos/FluentUI/DatePicker/CalendarView.swift
@@ -172,7 +172,7 @@ struct CalendarDay {
 
 	/// String used for accessibility scenarios, like VoiceOver
 	let accessibilityLabel: String
-	
+
 	/// String with extra descriptors for accessibility scenarios.  VoiceOver
 	/// reads this string after a short delay if focus stays on an element.
 	let accessibilityHelp: String?
--- a/macos/FluentUI/DatePicker/CalendarView.swift
+++ b/macos/FluentUI/DatePicker/CalendarView.swift
@@ -177,6 +177,8 @@ struct CalendarDay {
 	/// reads this string after a short delay if focus stays on an element.
 	let accessibilityHelp: String?
 
-	/// Secondary String representation of the day
+	/// String representation of the day on the `secondaryCalendar`, if enabled.
+	/// For example: a Chinese lunar calendar superimposed on a Gregorian
+	/// calendar allowing cross-referencing of dates between the two.
 	let secondaryLabel: String?
 }
--- a/macos/FluentUI/DatePicker/CalendarView.swift
+++ b/macos/FluentUI/DatePicker/CalendarView.swift
@@ -172,6 +172,10 @@ struct CalendarDay {
 
 	/// String used for accessibility scenarios, like VoiceOver
 	let accessibilityLabel: String
+	
+	/// String with extra descriptors for accessibility scenarios.  VoiceOver
+	/// reads this string after a short delay if focus stays on an element.
+	let accessibilityHelp: String?
 
 	/// Secondary String representation of the day
 	let secondaryLabel: String?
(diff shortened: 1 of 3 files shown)


macos/FluentUI/DatePicker/CalendarView.swift
```diff
@@ -172,6 +172,10 @@ struct CalendarDay {
 
 	/// String used for accessibility scenarios, like VoiceOver
 	let accessibilityLabel: String
+	
+	/// String with extra descriptors for accessibility scenarios.  VoiceOver
+	/// reads this string after a short delay if focus stays on an element.
+	let accessibilityHelp: String?
 
 	/// Secondary String representation of the day
 	let secondaryLabel: String?
```
harrieshin, line 181: the secondaryLabel can't be just used for accessibilityHelp for date?

chripear: The secondaryLabel seems to have a very different purpose.  I'm not entirely clear on the use case but I figured I'd better not mess with it.  Here's what it looks like when I enable the "secondaryCalendar" checkbox in the test app:

Screen Shot 2021-04-15 at 8 47 16 PM

chripear: I suppose it must be for cross-referencing dates in different calendars.  In the test app example it's Chinese Lunar vs. Gregorian... you can tell that the numerals don't sync up with each other so it's not a simple relabeling.  Anyway `secondaryLabel` stores the actual strings that go on screen in that secondary position, so no we can't use it to store the weekdays.

harrieshin: thanks for clarifying this. do you mind updating the comment for this variable for future reference?

harrieshin (approved):

macos/FluentUI/DatePicker/CalendarView.swift
Saadnajmi: secondaryLabel was for people who wanted to put a second calendar, such as one with Chinese holidays on it (I think that was the initial ask). There might be some better documentation in the actual PR @lcapkovic had for initially checking in the DatePicker the first time around.

chripear: > thanks for clarifying this. do you mind updating the comment for this variable for future reference?

Done

PR #522, opened 2021-04-16
Avatar border colors now show color

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Avatar border colors disappeared with #434. They are now back with the updated Color Set.

### Verification: (color tests not checked in)

| Default setting | Enabling border color generation: | Color inputs from PersonaData (should override default color): |
|-----|-----|-----|
| Simulator Screen Shot - iPhone 11 Pro - 2021-04-28 at 16 40 45 | Simulator Screen Shot - iPhone 11 Pro - 2021-04-28 at 16 35 39 | Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 17 57 51 |

| Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 17 55 33 | Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 17 53 35 |
| Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 17 55 04 | Simulator Screen Shot - iPhone 11 Pro - 2021-04-16 at 17 53 47 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/522)

### DIFF
--- a/ios/FluentUI/Avatar/AvatarGroupView.swift
+++ b/ios/FluentUI/Avatar/AvatarGroupView.swift
@@ -41,7 +41,7 @@ open class AvatarGroupView: UIView {
         }
     }
 
-    /// Set to true to generate border colors from InitialView ColorSet
+    /// Set to true to generate border colors from InitialView ColorSet excluding the Overflow AvatarView
     @objc open var shouldGenerateBorderColor: Bool = false {
         didSet {
             if oldValue != shouldGenerateBorderColor {
--- a/ios/FluentUI/Avatar/AvatarGroupView.swift
+++ b/ios/FluentUI/Avatar/AvatarGroupView.swift
@@ -41,6 +41,15 @@ open class AvatarGroupView: UIView {
         }
     }
 
+    /// Set to true to generate border colors from InitialView ColorSet
+    @objc open var shouldGenerateBorderColor: Bool = false {
+        didSet {
+            if oldValue != shouldGenerateBorderColor {
+                updateAvatars()
+            }
+        }
+    }
+
     /// Maximum count of avatars that can be displayed in the avatar group view.
     /// If the avatars array contains more avatars than this limit, an overflow UI will be displayed with the overflow count.
     /// The overflow UI is an avatar view with a "+" sign and the overflow count.
@@ -119,6 +128,7 @@ open class AvatarGroupView: UIView {
 
         for avatar in avatars.prefix(Int(maxDisplayedAvatars)) {
             let avatarView = AvatarView(avatarSize: avatarSize, withBorder: showBorders, style: .circle)
+            avatarView.shouldGenerateBorderColor = shouldGenerateBorderColor
             avatarView.setup(avatar: avatar)
 
             constraints.append(contentsOf: insert(avatarView: avatarView, previousAvatarView: previousAvatarView))
--- a/ios/FluentUI/Avatar/AvatarView.swift
+++ b/ios/FluentUI/Avatar/AvatarView.swift
@@ -258,7 +258,7 @@ open class AvatarView: UIView {
         }
     }
 
-    /// When true, the border color will be calculated using InitialView's ColorSet
+    /// When true, the border color will be calculated using InitialView's ColorSet excluding the Overflow AvatarView
     @objc open var shouldGenerateBorderColor: Bool = false {
         didSet {
             if oldValue != shouldGenerateBorderColor {
--- a/ios/FluentUI/Avatar/AvatarView.swift
+++ b/ios/FluentUI/Avatar/AvatarView.swift
@@ -730,7 +730,6 @@ open class AvatarView: UIView {
             if let window = window {
                 imageView.backgroundColor = style.backgroundColor(for: window)
                 imageView.tintColor = style.imageColor(for: window)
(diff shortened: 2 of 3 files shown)


ios/FluentUI/Avatar/AvatarView.swift
```diff
@@ -724,6 +731,7 @@ open class AvatarView: UIView {
             if let window = window {
                 imageView.backgroundColor = style.backgroundColor(for: window)
                 imageView.tintColor = style.imageColor(for: window)
+                borderColor = style == AvatarFallbackImageStyle.primaryFilled ? style.imageColor(for: window) : style.backgroundColor(for: window)
             }
```
harrieshin, line 735: but if user sets a border color we should respect that.

ios/FluentUI/Avatar/AvatarView.swift
```diff
@@ -587,8 +592,10 @@ open class AvatarView: UIView {
     private func updateBorderColor() {
         if let borderColor = borderColor {
             borderView.backgroundColor = borderColor
-        } else {
+        } else if (primaryText ?? "").isEmpty && (secondaryText ?? "").isEmpty {
             borderView.backgroundColor = Colors.Avatar.border
+        } else {
+            borderView.backgroundColor = initialsView.backgroundColor
```
harrieshin, line 598: >initialsView [](http://example.com/codeflow?start=41&length=12)

is this a new design direction that even if we have an image we will try to follow the intialsView background color for border?

ios/FluentUI/Avatar/AvatarView.swift
```diff
@@ -423,10 +423,11 @@ open class AvatarView: UIView {
         self.secondaryText = secondaryText
         self.presence = presence
         self.fallbackImageStyle = nil
+        setupInitialsView(convertTextToInitials: convertTextToInitials)
```
harrieshin, line 426: >setupInitialsView [](http://example.com/codeflow?start=8&length=17)

seems weird that we setupinitialviews then redetermine that we are going to show intials view

ios/FluentUI/Avatar/AvatarView.swift
rdeassis: For context: the Vnext Avatar sets the background color as the ring color by default if the style is not "overflow" and there are strings to allow the calculation of the background color.

Each Avatar Vnext style defines its own default ring color.

To prevent breaking the current behavior here, how about creating a property like "shouldCalculateRingColor" which defaults to false keeping the default border color with the gray unless the property is set to true by the client code?

harrieshin:
can we update the PR description images for future reference please?

ios/FluentUI/Avatar/AvatarView.swift
```diff
@@ -258,6 +258,12 @@ open class AvatarView: UIView {
         }
     }
 
+    @objc open var shouldCalculateRingColor: Bool = false {
+        didSet {
+            updateBorderColor()
```
harrieshin, line 263: should we only update border when variable value has changed?

ios/FluentUI/Avatar/AvatarView.swift
```diff
@@ -258,6 +258,12 @@ open class AvatarView: UIView {
         }
     }
 
+    @objc open var shouldCalculateRingColor: Bool = false {
```
harrieshin, line 261: can you add documentation on this public variable please?

harrieshin: instead of `shouldCalculateRingColor`, should we say something like `shouldGenerateBorderColor` or something like that because ring wouldn't make sense based on other AvatarView's public properties?

ios/FluentUI/Avatar/AvatarView.swift
sophialee0416: Could you clarify? I think updateBorderColor is also called whenever the border needs to be updated

harrieshin: if oldValue != shouldCaculdateRingColor { updatebordercolor()}

sophialee0416: Oh sorry okay thank you

ios/FluentUI/Avatar/AvatarView.swift
sophialee0416: Yes makes sense

ios/FluentUI/Avatar/AvatarGroupView.swift
```diff
@@ -41,6 +41,15 @@ open class AvatarGroupView: UIView {
         }
     }
 
+    /// Set to true to generate border colors from InitialView ColorSet
```
harrieshin, line 44: except for the overflow avatarView case right?

sophialee0416: Yes.. that's intended right?

harrieshin: yeah make sure you clarify that in the documentation

harrieshin (approved):

PR #534, opened 2021-04-21
V1 of empty BottomSheetController

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds a new empty bottom sheet control. This is similar to PR #510, but after talking to Harrie offline we decided that because I'll be working on the commanding bottom sheet as well, I can take over the empty bottom sheet work.

This bottom sheet has the same appearance in all screen sizes unlike the sheet in PR #510. Because the capabilities of a bottom bar are so different from an expandable sheet, I'm purely focusing on the sheet variant here to avoid complicating the API. The higher level CommandingBottomSheet control that will come in a future PR will include that automatic sheet <-> bar switching, as having a unified sheet/bar API is more feasible for that control.

Usage is shown in the demo controller, but in short the client has to:
1. Add the bottom sheet controller as a child view controller
2. Position the root view as an overlay over whatever their main content is.
3. Give us the contentViewController which will manage the bottom sheet content.
The sheet will only expand into the space the client gave the root view. The root view doesn't size itself or manipulate the view hierarchy above it. Internally, the root view is a passthrough view and the actual bottom sheet is a subview of that.

The client provides the bottom sheet content through the contentViewController parameter. The content can be sized in two ways:
1. (Default) Sized by us to fill the available space when expanded, with an additional customization through `expandedHeightFraction`. This is good for long scrolling content that can fill any space, and the client just wants it to expand to some fraction of the screen height, regardless of the screen size.
2. Self-sized through `contentViewController.preferredContentSize` and setting `respectsPreferredContentSize` to `true`. This is good for static content with well-defined height.

When panning the sheet, the content is rigid - the height doesn't change. The sheet just pulls in / out from the off-screen area.

Other notes:
- Uses spring timing which is closer to system UI feel than ease in/out
- Dragging the sheet beyond the min/max offset has a slight rubber band effect
- All animations are interruptible
- Co-operates well with scroll views when the client provides the `hostedScrollView`.
- The passthrough view preserves accessibility
 
Video:

https://user-images.githubusercontent.com/3610850/115605355-46694e80-a297-11eb-9027-c63d5ecd349a.mov

### Verification
|iPhone|image|image|
|iPad|image|image|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/534)


lcapkovic:
Why is FluentUI Native auto-added as a reviewer? Should I remove that somehow?

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
```diff
@@ -0,0 +1,54 @@
...
+    private struct Constants {
+        static let cornerRadius: CGFloat = 14
+
+        struct Shadow {
+            static let color: CGColor = UIColor.black.cgColor
+            static let opacity: Float = 0.14
+            static let radius: CGFloat = 8
+            static let offset: CGSize = CGSize(width: 0, height: 4)
+        }
+    }
```
Saadnajmi, line 53: I think we've moved away from structs for constants and just call em private vars.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+
+        return bottomSheetView
+    }()
+
+    // MARK: - Gesture handling
+
+    @objc private func handleResizingHandleViewTap(_ sender: UITapGestureRecognizer) {
+        if currentOffsetFromBottom != collapsedOffsetFromBottom {
+            animate(to: .collapsed, velocity: 0)
+            hostedScrollView?.setContentOffset(.zero, animated: true)
```
anandrajeswaran, line 137: does this animation happen concurrently with the animate(to:, velocity:) call? Do we need to sync or stage them?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+            } else {
+                move(to: .expanded, velocity: velocity)
+            }
+        } else {
+            // Velocity high enough, animate to the offset we're swiping towards
+            if velocity > 0 {
+                move(to: .collapsed, velocity: velocity)
+            } else {
+                move(to: .expanded, velocity: velocity)
+            }
```
anandrajeswaran, line 196: this is heavily nested code so could flatten by slightly de-duping: 
```
let state = velocity > 0 ? .collapsed : .expanded
move(to: state, velocity: velocity)
```
Same with above

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+        }
+    }
+
+    private func move(to targetExpansionState: BottomSheetExpansionState, animated: Bool = true, velocity: CGFloat = 0.0) {
+        if animated {
+            animate(to: targetExpansionState, velocity: velocity)
+        } else {
+            bottomSheetOffsetConstraint.constant = -(targetExpansionState == .expanded ? expandedOffsetFromBottom : collapsedOffsetFromBottom)
+            view.setNeedsLayout()
```
anandrajeswaran, line 205: Is this really needed? if not animated just changing the constraint should automatically get a layout pass scheduled.

What happens if you remove it?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+            switch targetExpansionState {
+            case .expanded:
+                delegate?.bottomSheetControllerDidExpand?(self)
+            case .collapsed:
+                delegate?.bottomSheetControllerDidCollapse?(self)
+            }
+        }
+    }
+
+    private func animate(to targetExpansionState: BottomSheetExpansionState, velocity: CGFloat = 0.0) {
```
anandrajeswaran, line 216: This is really clear to me and kinda awesome. Looking forward to trying it out elsewhere

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+        }
+
+        if animator.isRunning {
+            animator.stopAnimation(true)
+
+            // The AutoLayout constant doesn't animate, so we need to set it to whatever it should be
+            // based on the frame calculated during the interrupted animation
+            let offsetFromBottom = view.frame.height - bottomSheetView.frame.origin.y
+            bottomSheetOffsetConstraint.constant = -offsetFromBottom
+            view.setNeedsLayout()
```
anandrajeswaran, line 258: Similar comment - what happens if we don't have it. Does adjusting the constant not already kick off a layout pass?

If we need layout to happen deterministically before something else happens then we would need layoutIfNeeded?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+    // MARK: - Height constraint utils
+
+    private func updateBottomSheetHeightConstraints() {
+        let newConstraints = generateBottomSheetHeightConstraints()
+
+        NSLayoutConstraint.deactivate(bottomSheetHeightConstraints)
+        NSLayoutConstraint.activate(newConstraints)
+
+        bottomSheetHeightConstraints = newConstraints
+        view.setNeedsLayout()
```
anandrajeswaran, line 271: Another setNeedsLayout that is hopefully not needed?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+    ///
+    /// By default the root view of `contentViewController` will be sized automatically to fill the available area,
+    /// respecting the provided `preferredExpandedHeightFraction` multiplier.
+    /// Alternatively, the content can size itself by setting `respectsPreferredContentSize` to true and providing a `preferredContentSize`.
+    @objc public init(with contentViewController: UIViewController) {
+        self.contentViewController = contentViewController
+        super.init(nibName: nil, bundle: nil)
+
+        addChild(contentViewController)
+        contentViewController.didMove(toParent: self)
```
anandrajeswaran, line 31: I may be off base with this, but having the view controller hierarchy management live separate of the view hierarchy management (done in a lazy getter indirectly in loadView) seems scary to me. Can we avoid doing any view controller hierarchy manipulation until someone actually requests our view? And then the bonus is viewController and viewHierarchy manipulation are in the same place and less likely to mysteriously go out of sync?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+    }
+
+    public func gestureRecognizerShouldBegin(_ gestureRecognizer: UIGestureRecognizer) -> Bool {
+        guard let scrollView = hostedScrollView, let panGesture = gestureRecognizer as? UIPanGestureRecognizer else {
+            return true
+        }
+
+        let scrolledToTop = scrollView.contentOffset.y <= 0
+        let panningDown = panGesture.velocity(in: view).y > 0
+        let fullyExpanded = currentOffsetFromBottom >= expandedOffsetFromBottom
```
anandrajeswaran, line 367: Should we check this first and make the checks sequential so we don't have to calculate contentOffset or velocity if we don't need to?

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
```diff
@@ -0,0 +1,54 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class BottomSheetView: UIView {
```
anandrajeswaran, line 8: Possibly worth a comment explaining why we need this to exist and have a subview? The naive view of this is why do we need a subclass at all and can't we just use contentView directly if it takes our entire and we're not using any UIView subclassing hooks.

So the 3 possible outcomes:
1) Leave the way it is
2) Convert it from a subclass to a factory function to return a view with a subview
3) Convert it from a subclass to a factory function to return just one view.

If it's not the simplest option (#3), and I'm pretty sure there is a good reason it can't be, we should probably add a comment explaining the need for the subview at least.

anandrajeswaran:
Can you explain more about "The passthrough view preserves accessibility"? Maybe I'm missing some comments somewhere in code, but what would happen if we just don't put a view there at all or we had clients give us the view they want put there (i.e. have a bottomSheetViewController (fills the bottom sheet) and a contentViewController (fills where the pass-through view currently is)?

anandrajeswaran (changes requested):

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,71 @@
...
+        setupMainPersonaListView()
+        setupBottomSheet()
+    }
+
+// MARK: Setup Methods
+
+    private func setupMainPersonaListView() {
+        view.addSubview(personaListView)
+        personaListView.frame = view.bounds
+        personaListView.autoresizingMask = [.flexibleWidth, .flexibleHeight]
```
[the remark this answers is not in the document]
lcapkovic: It does not, leftover from other demos.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+
+    /// Called after the sheet fully collapsed.
+    @objc optional func bottomSheetControllerDidCollapse(_ controller: BottomSheetController)
+}
+
+@objc(MSFBottomSheetController)
+public class BottomSheetController: UIViewController {
+
+    /// Initializes the bottom sheet controller.
+    /// - Parameter contentViewController: The view controller that manages the bottom sheet content.
```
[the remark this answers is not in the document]
lcapkovic: It's the one one we stuff in the bottom sheet. I'll try to reword.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+                    resizingHandleView.isHidden = true
+                    panGestureRecognizer.isEnabled = false
+                }
+                move(to: .collapsed, animated: false)
+            }
+        }
+    }
+
+    /// Indicates if `preferredContentSize` of `contentViewController` should be respected.
+    /// Upper limits on the content size are still enforced, depending on the area available to `BottomSheetController`.
```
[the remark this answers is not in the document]
lcapkovic: If we don't respect preferredContentSize, we will size the sheet content ourselves - attempt to expand it as far as it can go.

About the upper limits - if the child's preferred height would make the expanded sheet taller than the space available to us, we'll clamp the height.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: It works without it, I just thought I was getting lucky and I should be explicitly asking for a layout. If that's not the case I'm happy to remove that.

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
lcapkovic: I explain the need for the subview in a comment below, basically if you apply both a shadow and round corners with clipsToBounds == true on the same view, it will clip its own shadow.

Not sure I like the factory approach, because it doesn't clearly connect the two. It would have to return both the root view (which is what externally is the bottom sheet and that's what you place and size in your hierarchy) and the contentView (which is where all the content needs to go. I thought about redirecting all addSubview calls to contentView.addSubview, but that feels kinda wrong. Or is it? Is there an obvious solution here I'm missing?

lcapkovic:
> Can you explain more about "The passthrough view preserves accessibility"? Maybe I'm missing some comments somewhere in code, but what would happen if we just don't put a view there at all or we had clients give us the view they want put there (i.e. have a bottomSheetViewController (fills the bottom sheet) and a contentViewController (fills where the pass-through view currently is)?

@anandrajeswaran 
The bottom sheet controller needs to own the full area it can expand into. Both to correctly constrain the sheet content height and to possibly add a dimming view / other effects in the future. The root view (Passthrough view) is what that is. 

I flipped between these two approaches quite a bit and I'm still not sure which one is best.
1 - Container VC - what you're proposing. Be a parent to both the main VC and sheet VC.
2 - Child VC - only be a parent to the sheet VC.

1 is sort of like a UITabBarViewController, 2 is sort of like an overlay where we tell the client - place us over whatever you want the sheet to expand over.
1 seemed a bit overkill to me because we don't need to own the main view hierarchy, just the space over it. We never manipulate it, we just want to draw over it. This is different from things like UITabBarViewController or UISplitViewController that manipulate the child VCs. That's why the overlay approach of 2 seemed lighter. I'm happy to be swayed either way, I just think I need more pros and cons for either side.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,376 @@
...
+            bottomSheetView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            bottomSheetOffsetConstraint
+        ])
+    }
+
+    private lazy var resizingHandleView: ResizingHandleView = {
+        let resizingHandleView = ResizingHandleView()
+        resizingHandleView.isAccessibilityElement = true
+        resizingHandleView.accessibilityTraits = .button
+        resizingHandleView.isUserInteractionEnabled = true
```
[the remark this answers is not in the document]
lcapkovic: Yep all of these are necessary.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: It is concurrent. The scroll view just scrolls all the way up while the bottom sheet is collapsing. I don't think we need to do anything more than this.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Can you elaborate on what's scary? My thinking is that once it's our child VC, we can do as we please with the root view, which includes delaying when that view is added to any hierarchy. Is there an expectation that the view will be moved immediately after addChild is called?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,71 @@
...
+
+class BottomSheetDemoController: DemoController {
+// MARK: Lifecycle methods
+
+    override func viewDidLoad() {
+        super.viewDidLoad()
+
+        setupMainPersonaListView()
+        setupBottomSheet()
+    }
```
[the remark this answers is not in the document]
lcapkovic: Keeping this one because the superclass adds a scrollview in viewDidLoad which ends up over all of our content and breaks stuff. This demo controller will change in the future to be more comprehensive and will make proper use of the superclass container views.

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
lcapkovic: Do you know of a recent example? Everywhere I look it's this. And I kind of like this style 😄

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
anandrajeswaran: Becuase we're using property animations there's no risk of a layout pass/commit causing an in-progress animation to snap to finish when the first one finishes, right?

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
anandrajeswaran: Does the contentView actually need to be returned? Just make it something like a 
```
private func makeBottomSheetByEmbedding(view:  UIView) -> UIView
```
where it takes in the view that needs to be placed into contentView and does it for you

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
anandrajeswaran: We've run into many issues over the years with the view controller hierarchy not correctly reflecting the view hierarchy which leads to subtle bugs with view controller lifecycle events not propagating and stuff like that. Having it obvious they are paired in code has helped prevent causing more of those bugs with refactoring

anandrajeswaran (approved):
> > Can you explain more about "The passthrough view preserves accessibility"? Maybe I'm missing some comments somewhere in code, but what would happen if we just don't put a view there at all or we had clients give us the view they want put there (i.e. have a bottomSheetViewController (fills the bottom sheet) and a contentViewController (fills where the pass-through view currently is)?
> 
> @anandrajeswaran
> The bottom sheet controller needs to know about the full area it can expand into. Both to correctly constrain the sheet content height and to possibly add a dimming view / other effects in the future. The root view (Passthrough view) is what that is.
> 
> I flipped between these two approaches quite a bit and I'm still not sure which one is best.
> 1 - Container VC - what you're proposing. Be a parent to both the main VC and sheet VC.
> 2 - Child VC - only be a parent to the sheet VC.
> 
> 1 is sort of like a UITabBarViewController, 2 is sort of like an overlay where we tell the client - place us over whatever you want the sheet to expand over.
> 1 seemed a bit overkill to me because we don't need to own the main view hierarchy, just the space over it. We never manipulate it, we just want to draw over it. This is different from things like UITabBarViewController or UISplitViewController that manipulate the child VCs. That's why the overlay approach of 2 seemed lighter. I'm happy to be swayed either way, I just think I need more pros and cons for either side.

Discussed offline and I could definitely go either way. Container VC seems like it would be simpler for most use cases where the contents behind it and other types of "decorators" that could show up are not very common. For complex clients who have a bunch of different surfaces they show decoration UI on using a Child VC pattern should allow them to more easily switch and animate between decorators. Up to you to decide which to optimize for

FWIW - the ability to add a dimming effect makes me a bit nervous. Isn't the design principle that this is decidedly not a modal experience?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,80 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
```
harrieshin, line 6: nit shouldn't need this

harrieshin:
nit: please add subspec for the bottomsheetcontroller in `MicrosoftFluentUI.podspec` for modularization .

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,371 @@
...
+    /// Provide this to ensure the bottom sheet pan gesture recognizer coordinates with the scroll view to enable scrolling based on current bottom sheet position and content offset.
+    @objc open var hostedScrollView: UIScrollView?
+
+    /// Indicates if the bottom sheet is expandable.
+    @objc open var isExpandable: Bool = true {
+        didSet {
+            if isExpandable != oldValue {
+                resizingHandleView.isHidden = !isExpandable
+                panGestureRecognizer.isEnabled = isExpandable
+                move(to: .collapsed, animated: false)
```
harrieshin, line 46: we need to make sure once VO double tap on the handle and the animation is done for resizing, the VO focus on the object correctly

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
harrieshin, line 8: what about the dimmingView?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,371 @@
...
+    /// Regardless of the value, the expanded height is limited by the height of `BottomSheetController`'s root view.
+    @objc open var respectsPreferredContentSize: Bool = false {
+        didSet {
+            if respectsPreferredContentSize != oldValue {
+                updateBottomSheetHeightConstraints()
+            }
+        }
+    }
+
+    /// Fraction of the available area that the bottom sheet shold take up in the expanded position.
```
harrieshin, line 61: nit: misspelled should

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,371 @@
...
+
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(bottomSheetView)
+
+        NSLayoutConstraint.activate([
+            bottomSheetView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            bottomSheetView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            bottomSheetOffsetConstraint
+        ])
+        updateBottomSheetHeightConstraints()
```
harrieshin, line 94: currently I am seeing a bug when expanded and rotating, the height of the bottomsheet on landscape isn't calculated correctly

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,371 @@
...
+
+    /// Height of the top portion of the content view that should be visible when the bottom sheet is collapsed.
+    @objc open var collapsedContentHeight: CGFloat = Constants.defaultCollapsedContentHeight
+
+    /// The object that acts as the delegate of the bottom sheet.
+    @objc open weak var delegate: BottomSheetControllerDelegate?
+
+    // MARK: - View loading
+
+    public override func loadView() {
```
harrieshin, line 80: nit for future reference can you write high-level UIView hierarchy of passthrough view, bottomsheetview, scrollview

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,80 @@
...
+
+        NSLayoutConstraint.activate([
+            personaListView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            personaListView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            personaListView.topAnchor.constraint(equalTo: view.topAnchor),
+            personaListView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        ])
+    }
+
+    private func setupBottomSheet() {
```
harrieshin, line 32: can you add more options so we can show case expand state is full screen vs half screen(preferred content size) similar to drawercontroller demo case?

harrieshin (changes requested):

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
lcapkovic: Dimming is usually used for modal sheets and isn't in specs for the persistent sheet as far as I'm aware. With this one we want the main content to be interactable even when the sheet is expanded, because often that will just be 1/3rd, 1/2 of the screen. I'm happy to add it later if there's a need.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: I think it's likely that the CommandingBottomSheet will end up being the control used by clients, and the empty one might not even have a demo controller, or it will just be one option in the controller. That's why I didn't want to spend too much time on the demo for now, so I can get to the Commanding sheet asap. Is it okay if I leave this until later?

lcapkovic: If there's a concern with having WIP demos in the main branch, I can leave out the demo entirely until things are more polished.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Which object should it focus on and can I somehow test this in simulator?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,80 @@
...
+        setupBottomSheet()
+    }
+
+// MARK: Setup Methods
+
+    private func setupMainPersonaListView() {
+        view.addSubview(personaListView)
+
+        NSLayoutConstraint.activate([
+            personaListView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
```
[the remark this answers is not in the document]
lcapkovic: As per the other comment, this demo is WIP and most likely will be totally replaced when I add the commanding bottom sheet. Let me know if it's okay to just leave it out completely for now.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
harrieshin: I think the current DrawerController behavior is that to stay in the resize handle. no there isn't a good way to test VO focus on simulator you need to do it in device

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
harrieshin: I do see dimming view in the design spec and did get confirmation from designers before.

lcapkovic: Ah I see it now. I need to talk to them more about how it should show, because I don't think it should in non-modal scenarios. The only case where it sort of makes sense is when it expands to full screen.

lcapkovic: ah, so there was an obvious solution I was missing :D

ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
```diff
@@ -15,6 +15,7 @@
 		7D0931C124AAA3D30072458A /* SideTabBarDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D0931C024AAA3D30072458A /* SideTabBarDemoController.swift */; };
 		7D23482A24D89C1C00FBE057 /* AvatarGroupViewDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7D23482924D89C1C00FBE057 /* AvatarGroupViewDemoController.swift */; };
 		7DC2FB2B24C0F4FD00367A55 /* TableViewCellFileAccessoryViewDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 7DC2FB2A24C0F4FD00367A55 /* TableViewCellFileAccessoryViewDemoController.swift */; };
+		80B1F7012628D8BB004DFEE5 /* BottomSheetDemoController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 80B1F7002628D8BB004DFEE5 /* BottomSheetDemoController.swift */; };
```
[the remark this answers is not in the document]
lcapkovic: Added the target membership.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,371 @@
...
+    /// A scroll view in `contentViewController`'s view hierarchy.
+    /// Provide this to ensure the bottom sheet pan gesture recognizer coordinates with the scroll view to enable scrolling based on current bottom sheet position and content offset.
+    @objc open var hostedScrollView: UIScrollView?
+
+    /// Indicates if the bottom sheet is expandable.
+    @objc open var isExpandable: Bool = true {
+        didSet {
+            if isExpandable != oldValue {
+                resizingHandleView.isHidden = !isExpandable
+                panGestureRecognizer.isEnabled = isExpandable
```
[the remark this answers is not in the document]
lcapkovic: Added labels same as in the DrawerController.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Solved by doing the safe thing of collapsing the sheet on any root view size changes. That way the sheet can't find itself in an invalid offset when the height changes in unexpected ways.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Can confirm focus stays on the focus stays on the resize handle.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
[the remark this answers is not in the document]
lcapkovic: Resolving the demo controller comments as I'll completely redo it with the Commanding bottom sheet.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
harrieshin: although, isn't it weird that when you have it expanded and you rotate, it will collapse? shouldn't it stay expanded?

lcapkovic: Good point, and turns it this also collapses it when only the width changes… I’ll try to preserve the expanded/collapsed state.

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
lcapkovic: Talked to design and it looks like it’s still an open question. It’s likely it will be needed in some form, but I’ll add it in a future PR once I get confirmation if that’s okay. There’s some details that need to worked out, like does it only start showing if we’re expanding into the upper third, etc.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Made a change to move to the correct offset for the new height.

ios/FluentUI/Bottom Sheet/BottomSheetView.swift
Saadnajmi: Talked offline, but ironically I couldn't find an example in the first two files I mentioned (FluentUI Apple macOS Button and Link). I guess we're bad at following it :D 

The Constants Struct is something we inherited before we settled on the ordering in aka.ms/swift-guide , for context of anyone reading this

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -0,0 +1,449 @@
...
+    }
+
+    private func handleCompletedStateChange(to targetExpansionState: BottomSheetExpansionState) {
+        switch targetExpansionState {
+        case .expanded:
+            self.delegate?.bottomSheetControllerDidExpand?(self)
+        case .collapsed:
+            self.delegate?.bottomSheetControllerDidCollapse?(self)
+        }
+        updateResizingHandleViewAccessibility()
```
harrieshin, line 303: isn't this wayyy to late? that mean when you first set up the bottomsheet the handle will just say "button". only after user interacts with it at least once, it will have the right accessibility label

harrieshin (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Good catch, fixed.

PR #540, opened 2021-04-26
Adding Large Content Viewer support to the Button.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change adds support for the Large Content Viewer to the Button (Vnext).

Two different issues were discovered during the implementation of the floating style for the Button:
 - Fonts are not scaling for when the accessibility large text setting is changed.
 - There is no SwiftUI support for Large Content Viewer when it comes to custom controls. (Apple forums thread created: https://developer.apple.com/forums/thread/678281)

Creating SwiftUI Font types based on the UIFont types hinders the scaling capability of the fonts.
Using another initializer where the font name and size (retrieved from the token UIFont type) are passed works on iOS 14 only.

The workaround for these issues is to fall back in the UIKit layer where we use the [UILargeContentViewerInteraction](https://developer.apple.com/documentation/uikit/uilargecontentviewerinteraction) to implement Large Content Viewer support for the button on iOS 13 (all styles) as the font does not scale while enabling it for the floating style only on iOS 14.

### Verification

Tested on both iOS 13 and iOS 14 on the regular scenario and with the accessibility large text settings enabled:

**- iOS 14:**

https://user-images.githubusercontent.com/68076145/116115290-15aa6000-a66f-11eb-8f2d-6e6f48be1ee5.mov

**- iOS 13:**

https://user-images.githubusercontent.com/68076145/116115417-3bd00000-a66f-11eb-9155-60f6e406874b.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/540)

### DIFF
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -3,6 +3,7 @@
 //  Licensed under the MIT License.
 //
 
+import Combine
 import UIKit
 import SwiftUI
 
@@ -21,6 +22,7 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
         let tokens = targetButton.tokens
         let state = targetButton.state
         let isDisabled = state.isDisabled
+        let isFloatingStyle = tokens.style.isFloatingStyle
         let isPressed = configuration.isPressed
         let shouldUsePressedShadow = isDisabled || isPressed
 
@@ -35,16 +37,19 @@ struct MSFButtonViewButtonStyle: ButtonStyle {
 
             if let text = state.text {
                 Text(text)
-                    .font(Font(tokens.textFont))
+                    .font(isFloatingStyle ?
+                            Font(tokens.textFont) :
+                            Font.custom(tokens.textFont.familyName,
+                                        size: tokens.textFont.fontDescriptor.pointSize))
                     .fontWeight(.medium)
                     .multilineTextAlignment(.center)
-                    .modifyIf(tokens.style.isFloatingStyle, { view in
+                    .modifyIf(isFloatingStyle, { view in
                         view.frame(minHeight: tokens.textMinimumHeight)
                     })
                 }
         }
         .padding(tokens.padding)
-        .modifyIf(tokens.style.isFloatingStyle && !(state.text?.isEmpty ?? true), { view in
+        .modifyIf(isFloatingStyle && !(state.text?.isEmpty ?? true), { view in
             view.padding(.horizontal, tokens.textAdditionalHorizontalPadding )
         })
         .frame(maxWidth: .infinity, maxHeight: .infinity)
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -142,6 +149,22 @@ public struct MSFButtonView: View {
                    theme: theme)
     }
 
+    // MARK: - UIGestureRecognizerDelegate
+
+    public func gestureRecognizer(_ gestureRecognizer: UIGestureRecognizer,
+                                  shouldRecognizeSimultaneouslyWith otherGestureRecognizer: UIGestureRecognizer) -> Bool {
+        // Prevents the Large Content Viewer gesture recognizer from cancelling the SwiftUI button gesture recognizers.
+        return true
```
harrieshin, line 157: add apple radar bug number to remove? also, how does pointer interaction work with this now?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -158,12 +181,37 @@ public struct MSFButtonView: View {
         })))
         buttonView.tokens.windowProvider = self
         view.backgroundColor = UIColor.clear
+        setupLargeContentViewer()
```
harrieshin, line 184: may be shouldEnableLargeContentViewerForOSVersion should be called before setting up largecontentviewer?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -158,12 +181,37 @@ public struct MSFButtonView: View {
...
+        textPropertySubscriber = buttonView.state.$text.sink { buttonText in
+            self.view.largeContentTitle = buttonText
+        }
     }
 
+    private let shouldEnableLargeContentViewerForOSVersion: Bool = {
+        if #available(iOS 14.0, *) {
+            return false
+        } else {
+            return true
```
harrieshin, line 207: although this is just temporary right? can we have a button with increased text based on text setting on iOS 13?

ios/FluentUI/Vnext/Button/Button.swift
harrieshin: also it could return false if content viewer isn't needed?

harrieshin (approved):

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: The constant **shouldEnableLargeContentViewerForOSVersion** is initialized when the class is initialized, which happens before we call shouldEnableLargeContentViewerForOSVersion occurs.

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: Yes. As per our conversation today, I'll work on finding a solution to make that happen in iOS 13 as well (except for the floating style which should have a fixed non-scalable text).

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: I'll add the Apple radar bug number once we have it created. I'll follow up on the process.
As far as pointer interaction support, none of the FluentUI Buttons (UIKit or Vnext) currently support iPad pointer interaction. We can work on that as a new work item.

PR #475, opened 2021-03-12
Adding Persona View Control 

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding a Persona control to vNext. A Persona Cell can display an Avatar and its contents. The cell can exist as a standalone outside of a List.

### Verification

Passed pod lib lint, updated Demo Controller, & made sure control responded to state changes.

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 11 Pro - 2021-03-10 at 19 10 19 | Simulator Screen Shot - iPhone 11 Pro - 2021-03-10 at 20 49 17 |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/475)

PR #541, opened 2021-04-26
[List vNext] Creating SwiftUI property exposure for UIView properties

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Any UIView property in the List would not be accessible in a SwiftUI-only environment. This is a follow-up from #475 where the leadingView property was updated so that it can be set in SwiftUI as well.

### Verification

The PersonaView is using several of these properties:

- The leadingView is a SwiftUI View set in the PersonaView.
- The chevron is a UIView, which is then set as a SwiftUI View in the ListCell.
Simulator Screen Shot - iPhone 11 Pro - 2021-04-26 at 15 57 23

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/541)

### DIFF
--- a/ios/FluentUI/Vnext/List/ListCell.swift
+++ b/ios/FluentUI/Vnext/List/ListCell.swift
@@ -30,7 +30,6 @@ import SwiftUI
 @objc public class MSFListCellState: NSObject, ObservableObject, Identifiable {
     public var id = UUID()
 
-    // SwiftUI View Properties
     @Published public var leadingView: AnyView?
     @Published public var titleLeadingAccessoryView: AnyView?
     @Published public var titleTrailingAccessoryView: AnyView?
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -57,6 +53,64 @@ import SwiftUI
     @objc @Published public var layoutType: MSFListCellLayoutType = .automatic
     @objc @Published public var hasDivider: Bool = false
     @objc public var onTapAction: (() -> Void)?
+
+    // Any UIView property will set the value to a type-erased SwiftUI View
```
rdeassis, line 57: Do you need this comment as you are already outlining this in the main state class documentation?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -57,6 +53,64 @@ import SwiftUI
...
+    @objc public var leadingUIView: UIView? {
+        didSet {
+            if let view = self.leadingUIView {
```
rdeassis, line 60: If the calling code sets the property to nil with the intent of removing the view. How will it work here?
Have you tested that scenario?

rdeassis, line 60: nit: I think you don't need the "self." prefix, do you?

ios/FluentUI/Vnext/Persona/PersonaView.swift
```diff
@@ -14,8 +14,8 @@ import SwiftUI
 /// `onTapAction` provides tap gesture for PersonaView.
 ///
 @objc public protocol MSFPersonaViewState: MSFAvatarState {
-    var titleTrailingAccessoryView: UIView? { get set }
-    var subtitleTrailingAccessoryView: UIView? { get set }
+    var titleTrailingAccessoryUIView: UIView? { get set }
```
rdeassis, line 17: We need to add the AnyView properties to the protocol as well. (Considering the Swift/SwiftUI only scenario).

sophialee0416: The AnyView properties are not compatible with the objc protocol so I made another child protocol.

ios/FluentUI/Vnext/List/ListCell.swift
rdeassis: How about using guard let instead?

I think it falls into the early return scenario in our swift-guide:
https://microsoft.github.io/swift-guide/EarlyReturn.html

```
    @objc public var leadingUIView: UIView? {
        didSet {
            guard let view = leadingUIView else {
                leadingView = nil
                return
            }

            leadingView = AnyView(UIViewAdapter(view))
        }
    }
```

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -57,6 +52,79 @@ import SwiftUI
...
+    }
+    @objc public var footnoteTrailingAccessoryUIView: UIView? {
+        didSet {
+            if let view = footnoteTrailingAccessoryUIView {
+                footnoteTrailingAccessoryView = AnyView(UIViewAdapter(view))
+            } else {
+                footnoteTrailingAccessoryView = nil
+            }
+        }
+    }
```
rdeassis, line 118: nit: can you add an empty new line between the properties?

rdeassis (commented):
Just some minor issues and the merge conflict and we'll be good to go.
Thanks for working on this follow up!

ios/FluentUI/Vnext/Persona/PersonaView.swift
```diff
@@ -151,8 +156,8 @@ public struct PersonaView: View {
         state = MSFPersonaViewStateImpl(avatarState: avatar.state)
         state.leadingView = AnyView(avatar)
         state.leadingViewSize = .xlarge
-        state.titleTrailingAccessoryView = state.titleTrailingAccessoryView
-        state.subtitleTrailingAccessoryView = state.subtitleTrailingAccessoryView
+        state.titleTrailingAccessoryUIView = state.titleTrailingAccessoryUIView
+        state.subtitleTrailingAccessoryUIView = state.subtitleTrailingAccessoryUIView
```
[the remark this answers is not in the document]
rdeassis: Including the onTapAction below...

rdeassis (approved):

PR #547, opened 2021-04-27
hide initialsView or Image based on displayStyle of AvatarView

harrieshin:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

In case when client team reuses the avatarView, based on the displayStyle hide the subview behind the currentView.
When contactImage was translucent, it may show the initials view behind the current view.

Also, it doesn't make sense just because it doesn't have tooltip text the avatar view isn't accessible. Keep the accessibilityrole to .image if there is a contact image.

### Verification

not part of the change but programmatically update the avatarView public properties after it has rendered

|  Screen Shot 2021-04-27 at 5 00 08 PM|Screen Shot 2021-04-27 at 4 58 55 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/547)

### DIFF
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -338,7 +338,7 @@ open class AvatarView: NSView {
 		} else {
 			toolTip = nil
 			setAccessibilityLabel(nil)
-			if hasImage {
+			if !hasImage {
 				setAccessibilityElement(false)
 				setAccessibilityRole(.unknown)
 			}
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -108,7 +108,13 @@ open class AvatarView: NSView {
 			contactImageView.image = contactImage
 
 			// Update our display style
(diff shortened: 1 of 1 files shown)


macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -108,7 +108,13 @@ open class AvatarView: NSView {
 			contactImageView.image = contactImage
 
 			// Update our display style
-			displayStyle = contactImage == nil ? .initials : .image
+			if contactImage == nil {
+				displayStyle = .initials
```
kubalani, line 112: Does this need accessibility update same as line 342?

kubalani (approved):

macos/FluentUI/AvatarView/AvatarView.swift
harrieshin: make sense. I just ended up calling updateAvatarViewContents() instead of doing it directly here.

PR #551, opened 2021-05-04
Making window provider an Environment Value.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

An instance of an object that implements the FluentUIWindowProvider protocol is passed to SwiftUI controls so it can retrieve the correct theme to compute a control's tokens.

Previously this object was passed directly to the SwiftUI control by the UIKit wrapper when it is instantiated.

A problem arises when controls contain other controls and they either need to pass the window provider as a parameter, which is very error prone, or the theme override will not work for the controls in deeper levels of the view hirarchy.

This change leverage from Environment Values and makes the instance passed to a control an environment value, which means that all the views under that control's view hierarchy will use the same window provider to retrieve the correct theme.

Fixes to the List and PersonaView which were not getting the theme overrides were also fixed.

### Verification

Changed the tokens below to **$Colors.Brand.primary** YAML file locally and changed themes to ensure the color was correctly updated.
 - Avatar **ringGapColor**
 - List **borderColor**
 - List Cell **backgroundColor** (pressed state)
 - PersonaView **sublabelColor**
 - Header **textColor** (all styles)
 - Drawer **backgroundDimmedColor**: (@kubalani This scenario is not working and another work item needs to be created to address it.)

https://user-images.githubusercontent.com/68076145/116962302-754dd000-ac5a-11eb-87c5-8d5e6843a053.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/551)

### DIFF
--- a/ios/FluentUI/Vnext/Core/Theming.swift
+++ b/ios/FluentUI/Vnext/Core/Theming.swift
@@ -51,11 +51,17 @@ class MSFTokensBase {
     private var _theme: FluentUIStyle?
 }
 
-// MARK: Theme SwiftUI Environment Value
+// MARK: SwiftUI Environment Values
 
 struct ThemeKey: EnvironmentKey {
     static var defaultValue: FluentUIStyle {
         return FluentUIThemeManager.S
+    }
+}
+
+struct WindowProviderKey: EnvironmentKey {
+    static var defaultValue: FluentUIWindowProvider? {
+        return nil
     }
 }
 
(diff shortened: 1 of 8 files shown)


sophialee0416 (approved):

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -10,10 +10,19 @@ import SwiftUI
...
     @objc @Published public var cells: [MSFListCellState] = []
-    @objc @Published public var style: MSFHeaderFooterStyle = .headerPrimary
     @objc @Published public var title: String?
     @objc @Published public var backgroundColor: UIColor?
     @objc @Published public var hasDividers: Bool = false
+    @objc @Published public var style: MSFHeaderFooterStyle = .headerPrimary {
+        didSet {
+            if style != oldValue {
+                headerTokens.style = style
+                headerTokens.updateForCurrentTheme()
```
kubalani, line 20: Do we need this for published objects?

ios/FluentUI/Vnext/Core/Theming.swift
```diff
@@ -51,23 +51,80 @@ class MSFTokensBase {
...
+    ///  retrieve the theme associated with the window that this View belongs to.
+    ///
+    ///  3. If the windowProvider is nil the default theme property will be used.
+    ///
+    /// - Parameters:
+    ///   - tokens: An instance of a subclass of MSFTokensBase that defines the design tokens used in that particular SwiftUI View.
+    ///   - theme: theme property (environment value) of the SwiftUI View.
+    ///   - windowProvider: windowProvider property  (environment value) of the SwiftUI View that will potentially retrieve the theme to be applied.
+    /// - Returns: The resulting View either with the computed theme applied to its design tokens.
+    func designTokens(_ tokens: MSFTokensBase,
```
kubalani, line 117: should we add this in `ViewModifier` as opposed to `View`

https://developer.apple.com/documentation/swiftui/viewmodifier

ios/FluentUI/Vnext/Core/Theming.swift
```diff
@@ -51,23 +51,80 @@ class MSFTokensBase {
...
+    }
+
+    /// SwiftUI Views using this modifier need to define the following environment values:
+    ///  - @Environment(\.theme) var theme: FluentUIStyle
+    ///  - @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+    ///
+    /// The instances of these environment value properties will be passed to the SwiftUI view's subviews in its hierarchy.
+    ///
+    /// The logic behind the order or precedence to retrieve the correct theme to be used is:
+    ///  1. If the SwiftUI View gets a non-default theme instance through the environment value property,
```
kubalani, line 104: If a `SwiftUI` container has an Objective-C subview, How will the subview be notified about the theme changes? I am assuming this precedence is applied for purely native SwiftUI components.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -39,7 +39,15 @@ import SwiftUI
     @Published public var footnoteTrailingAccessoryView: AnyView?
     @Published public var trailingView: AnyView?
 
-    @objc @Published public var leadingViewSize: MSFListCellLeadingViewSize = .medium
+    @objc @Published public var leadingViewSize: MSFListCellLeadingViewSize = .medium {
+        didSet {
+            if leadingViewSize != oldValue {
+                tokens.cellLeadingViewSize = leadingViewSize
+                tokens.updateForCurrentTheme()
```
kubalani, line 46: Why update theme only for specific overrides? How would a client know when there is a need to invoke theme changes manually?

ios/FluentUI/Vnext/List/List.swift
rdeassis: This is because the Tokens class is not a subscriber of its own property. And the SwiftUI control does not use it to render its contents.

ios/FluentUI/Vnext/Core/Theming.swift
rdeassis: The SwiftUI View that is exposed by the UIView will get the theme change through the token subscription to the theme change event. And yes, the theme override for a specific View and its view hierarchy is something supported only on the SwiftUI hierarchy.

ios/FluentUI/Vnext/List/ListCell.swift
rdeassis: This is the way we can implement the change of all the tokens when the style is changed. Similar to the other comment that I responded: the SwiftUI View does not use the enum value, and the tokens class does not subscribe to the @published objects, so we need to force this refresh.

There's definitely room for improvement on this design, but it's beyond the scope of this PR.

ios/FluentUI/Vnext/Core/Theming.swift
rdeassis: To be honest I don't know the benefit of doing this as we still need the extension as below to apply the modifier, but I'm adhering to Apple's recommendation.

Thanks for the suggestion!

```
extension View {
    func borderedCaption() -> some View {
        modifier(BorderedCaption())
    }
}
```

kubalani (approved):

PR #552, opened 2021-05-04
Bottom commanding controller

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds a persistent bottom commanding surface. The controller takes in collections of CommandingItems, which internally get bound to their view representations.
The commanding surface style differs depending on the horizontal size class. .regular gets the bottom bar, and .compact + .unspecified get the sheet. 

Both representations get a <5-item hero command stack that's always visible. There's a also an additional list of commands, which are placed differently depending on the surface type.
In bottom bar the command list is shown in a popover. In bottom sheet the same list is located below the hero command stack, visible when the sheet is expanded.

The CommandingItem is the only entry point for clients to modify the state of the displayed views. All the properties are mutable and changes to them are automatically reflected in their representing views. Internally, the changes are handled by mutating the views in place if possible (all cell properties, only isEnabled and isOn for hero buttons), with a fallback to a full view refresh if an in-place edit isn't possible.

Known work for future PRs:
- Hiding
- Dimming
- Integrate table view cells with switches
- Dynamic expanded offset sizing
- Better handling of bottom safe area insets - might need to make expanded content transparent when collapsed, because specs show more padding in collapsed state than we can do without showing some of the expanded area.
- CommandingItem API iteration - clients will likely need more props

https://user-images.githubusercontent.com/3610850/117049465-0065af00-acc9-11eb-8218-cd0b96990e8a.mov

### Verification

|iPhone|Simulator Screen Shot - iPhone 12 - 2021-05-04 at 11 21 22|Simulator Screen Shot - iPhone 12 - 2021-05-04 at 11 21 12
|iPad|Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2021-05-04 at 11 57 30|Simulator Screen Shot - iPad Pro (11-inch) (2nd generation) - 2021-05-04 at 11 57 43

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/552)


ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,102 @@
...
+
+        /// Toggles `isOn` on tap before calling `action`.
+        case toggle
+    }
+
+    weak var delegate: CommandingItemDelegate?
+}
+
+protocol CommandingItemDelegate: class {
+    func commandingItem(_ item: CommandingItem, didChangeTitleFrom oldValue: String)
```
harrieshin, line 96: why do we need to tell the old value via the protocol? if it didChange we should just follow the changed value?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+///
+/// `.regular` -  the surface is displayed as a floating bottom bar.
+///
+/// In both styles, `heroItems` are always presented in a horizontal stack.
+/// Items from the `expandedListSections` are either presented in an expanded sheet or a popover, depending on the current style.
+///
+@objc(MSFBottomCommandingController)
+open class BottomCommandingController: UIViewController {
+
+    /// Items to be displayed in the hero area.
```
harrieshin, line 22: might as well add documentation it can't be more than 5?

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: For me this was more of a "why not". The internal property observers have the old value available to them in didSet, and it can be used to make things more efficient, like no-oping when the value hasn't changed. The delegate is kind of an external observer in this case, so I didn't see why not give them the old value as well. They can still get the current one directly from the item, just like in local didSets.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+
+import UIKit
+
+/// Persistent commanding surface displayed at the bottom of the available area.
+///
+/// The presentation style automatically varies depending on the current horizontal `UIUserInterfaceSizeClass`:
+///
+/// `.unspecified` and `.compact` - the surface is displayed as an expandable bottom sheet.
+///
+/// `.regular` -  the surface is displayed as a floating bottom bar.
```
harrieshin, line 14: why regular/compact/unspecify rather than .bar/sheet etc?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+        commandContainer.addArrangedSubview(heroCommandStack)
+        commandContainer.addArrangedSubview(moreButtonView)
+
+        let bottomBarView = makeBottomBarByEmbedding(contentView: commandContainer)
+        bottomBarView.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(bottomBarView)
+
+        NSLayoutConstraint.activate([
+            bottomBarView.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            bottomBarView.bottomAnchor.constraint(equalTo: view.bottomAnchor, constant: -Constants.BottomBar.bottomOffset)
```
harrieshin, line 112: safe area?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -0,0 +1,160 @@
...
+        bottomCommandingVC.heroItems = heroItems
+        bottomCommandingVC.expandedListSections = expandedListSections
+
+        addChild(bottomCommandingVC)
+        view.addSubview(bottomCommandingVC.view)
+        bottomCommandingVC.didMove(toParent: self)
+
+        bottomCommandingController = bottomCommandingVC
+
+        container.addArrangedSubview(createLabelAndSwitchRow(labelText: "Expanded list items", switchAction: #selector(toggleExpandedItems), isOn: true))
```
harrieshin, line 24: instead of this, which doesn't make the demo app a11y friendly in large text mode, can you use tableview with toggles for better larger text support similar to sidetabbardemocontroller please? we are slowly cleaning up this bad pattern in the demo app.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,102 @@
...
+//
+
+import UIKit
+
+/// An object representing a command.
+///
+/// `CommandingItem` defines the high level properties and behavior of a command. Its visual representation is determined by
+/// the `BottomCommandingController`.
+@objc(MSFCommandingItem)
+open class CommandingItem: NSObject {
```
harrieshin, line 13: should commanding item be a protocol rather than nsobject so client can easily pipe through their images and titles with their backing logic?

harrieshin: similar to popupmenucontroller architecture

harrieshin:
is it just me or do I observe that after you rotate the bottomcommandingcontroller the text(title label) is no longer center aligned compared to the icons in the hero commands

ios/FluentUI/Bottom Sheet/CommandingItem.swift
harrieshin: but anytime protocol didSomthingFoo is only called when it actually did change something. this is different from property setting don't you think?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -23,8 +23,19 @@ public class BottomSheetController: UIViewController {
...
+    @objc public init(contentViewController: UIViewController) {
         self.contentViewController = contentViewController
+        self.contentView = contentViewController.view
+        super.init(nibName: nil, bundle: nil)
+    }
+
+    /// Initializes the bottom sheet controller.
+    /// - Parameter contentView: The view that's placed inside the bottom sheet.
+    ///
+    /// TODO: Add view-specific sizing info
```
harrieshin, line 35: is this related to how resize handle animates weirdly when rotating the device while expanded?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+        }
+    }
+
+    private func createAndBindHeroCommandView(with item: CommandingItem) -> UIView {
+        let tabItem = TabBarItem(title: item.title, image: item.image, selectedImage: item.selectedImage)
+        let itemView = TabBarItemView(item: tabItem, showsTitle: true)
+        itemView.alwaysShowTitleBelowImage = true
+        itemView.numberOfTitleLines = 1
+        itemView.isSelected = item.isOn
+        itemView.accessibilityTraits.insert(.button)
```
harrieshin, line 334: also need to support larger content viewer now or later for larger text support

lcapkovic:
> is it just me or do I observe that after you rotate the bottomcommandingcontroller the text(title label) is no longer center aligned compared to the icons in the hero commands

I noticed this today as well. I'll look into it. Might be a TabBarItemView bug.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+            contentView.bottomAnchor.constraint(equalTo: bottomBarView.bottomAnchor, constant: -Constants.BottomBar.heroStackTopBottomMargin)
+        ])
+
+        return bottomBarView
+    }
+
+    private func makeBottomSheetContent(headerView: UIView, expandedContentView: UIView) -> UIView {
+        let view = UIView()
+        let separator = Separator()
+        separator.translatesAutoresizingMaskIntoConstraints = false
```
Saadnajmi, line 187: sidenote: I think separator sets it's intrinsicContentSize to something like "NO_INTRINSIC_SIZE" so I wonder if it even matters if we set this  mask :D

I would still set it for consistency, I just thought that was interesting

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+        let tapGesture = UITapGestureRecognizer(target: self, action: #selector(handleMoreButtonTap(_:)))
+        moreButtonView.addGestureRecognizer(tapGesture)
+
+        NSLayoutConstraint.activate([
+            moreButtonView.widthAnchor.constraint(equalToConstant: Constants.heroButtonWidth),
+            moreButtonView.heightAnchor.constraint(equalToConstant: Constants.heroButtonHeight)
+        ])
+
+        return moreButtonView
+    }()
```
Saadnajmi, line 234: do you actually need the parenthesis here? I would expect not

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+            bottomSheetController?.collapsedContentHeight = bottomSheetHeroStackHeight
+            bottomSheetController?.isExpandable = isExpandable
+            bottomSheetHeroStackTopConstraint?.constant = bottomSheetHeroStackTopMargin
+        } else {
+            moreButtonView.isHidden = !isExpandable
+        }
+    }
+
+    private lazy var moreButtonView: UIView = {
+        let moreButtonItem = TabBarItem(title: Constants.BottomBar.moreButtonTitle, image: Constants.BottomBar.moreButtonIcon ?? UIImage())
```
Saadnajmi, line 221: What happens if we actually fall through and set an empty `UIImage()` as the image?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+    @objc private func handleHeroCommandTap(_ sender: UITapGestureRecognizer) {
+        guard let tabBarItemView = sender.view as? TabBarItemView, let binding = viewToBindingMap[tabBarItemView] else {
+            return
+        }
+
+        switch binding.item.commandType {
+        case .toggle:
+            tabBarItemView.isSelected.toggle()
+            binding.item.isOn = tabBarItemView.isSelected
+            fallthrough
```
Saadnajmi, line 274: There's a fall through keyword? :o

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+        guard let binding = itemToBindingMap[item] else {
+            return
+        }
+        let staleView = binding.view
+
+        switch binding.location {
+        case .heroSet:
+            if let stackIndex = heroCommandStack.arrangedSubviews.firstIndex(of: staleView) {
+                removeBinding(binding)
+                let newView = createAndBindHeroCommandView(with: item)
```
Saadnajmi, line 372: In context, what is a "heroCommandView"? Is that wording very office specific / documented elsewhere in this repository? If not, I would like that documented somewhere to remove a little bit of an office-ism that is assumed knowledge =)

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+            static let backgroundColor: UIColor = Colors.NavigationBar.background
+
+            static let bottomOffset: CGFloat = 30
+            static let heroStackLeadingTrailingMargin: CGFloat = 8
+            static let heroStackTopBottomMargin: CGFloat = 16
+
+            static let moreButtonIcon: UIImage? = UIImage.staticImageNamed("more-24x24")
+
+            // TODO: Localize
+            static let moreButtonTitle: String = "More"
```
Saadnajmi, line 455: So it's not forgotten, I shall put a comment here :P

ios/FluentUI/Bottom Sheet/CommandingItem.swift
Saadnajmi, line 13: Are these commands similar to those that might be in the CCB (Contextual Command Bar)? Is there opportunity for shared code there?

ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,102 @@
...
+            if selectedImage != oldValue {
+                delegate?.commandingItem(self, didChangeSelectedImageFrom: oldValue)
+            }
+        }
+    }
+
+    /// Indicates whether the command is currently on.
+    ///
+    /// When `commandType` is set to `.toggle`, the item toggles this automatically before calling `action`.
+    @objc open var isOn: Bool {
```
Saadnajmi, line 48: isOn is kinda like isSelected? As in the command is pressed?

ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,102 @@
...
+    /// Determines the behavior of this command when its triggered.
+    @objc open var commandType: CommandType {
+        didSet {
+            if commandType != oldValue {
+                delegate?.commandingItem(self, didChangeCommandTypeFrom: oldValue)
+            }
+        }
+    }
+
+    @objc public init(title: String, image: UIImage, action: @escaping (CommandingItem) -> Void, selectedImage: UIImage? = nil, isSelected: Bool = false, isEnabled: Bool = true, commandType: CommandType = .simple) {
```
Saadnajmi, line 74: What does @escaping do?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -0,0 +1,160 @@
...
+
+        NSLayoutConstraint.activate([
+            bottomCommandingVC.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
+            bottomCommandingVC.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
+            bottomCommandingVC.view.topAnchor.constraint(equalTo: view.topAnchor),
+            bottomCommandingVC.view.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+        ])
+    }
+
+    @objc private func toggleExpandedItems() {
```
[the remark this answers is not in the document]
lcapkovic: Yup, because they're being pointed to by selectors.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: You do, otherwise swift thinks you are assigning a closure to the property.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: Maybe in the future 😄 The API is designed with this in mind.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: Yup, I chose isOn, because a command can be represented as a button or switch (and maybe more in the future), and isSelected really only makes sense in buttons. So this is meant to reflect the on/off state of the command, not the underlying view. The controller then interprets this however it wants, depending on what view this command is bound to.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: It means that the closure can be called after the function returns. So any callback like this needs to be marked escaping, and Swift won't let you not mark it that way.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+
+        bottomSheetController = sheetController
+        updateExpandability()
+    }
+
+    private func makeBottomBarByEmbedding(contentView: UIView) -> UIView {
+        let bottomBarView = UIView()
+        bottomBarView.layer.shadowColor = Constants.BottomBar.Shadow.color
+        bottomBarView.layer.shadowOpacity = Constants.BottomBar.Shadow.opacity
+        bottomBarView.layer.shadowRadius = Constants.BottomBar.Shadow.radius
```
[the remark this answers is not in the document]
lcapkovic: I don't think it exists in UIKit.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: It does have an autoresizing mask tho, so without this that gets converted to constraints and it clashes with mine.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: I like it 😄

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: This just describes how we react to the system horizontal size classes, so this is not a fluent enum. I might an API to return the current style (.bar/.sheet), but it's not necessary right now.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: Makes sense, I changed it to pass the new value. I wasn't using the old one anyways and if we need it in the future it's easy to add.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: Done, thanks!

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: Yeah it looks much better now :D

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Nope, but that is a thing I will look at as well.
This is because I added a way to just provide a content view to the bottom sheet rather than a content view controller. Views don't have the preferredContentSize prop, so I need to modify the API to reconcile that. It turns out we'll be adding multiple expansion states, so sizing will get more complicated and I'll do the API changes together with that because it's related.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: Done.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -0,0 +1,160 @@
...
+            showMessage("Expanded list command tapped")
+        }
+    }
+
+    private lazy var heroItems: [CommandingItem] = [
+        CommandingItem(title: "Item 1", image: UIImage(named: "Home_24")!, action: commandAction, selectedImage: UIImage(named: "Home_Selected_24")!),
+        CommandingItem(title: "Item 2", image: UIImage(named: "Home_24")!, action: commandAction, selectedImage: UIImage(named: "Home_Selected_24")!),
+        CommandingItem(title: "Item 3", image: UIImage(named: "Home_24")!, action: commandAction, selectedImage: UIImage(named: "Home_Selected_24")!),
+        CommandingItem(title: "Item 4", image: UIImage(named: "Home_24")!, action: commandAction, selectedImage: UIImage(named: "Home_Selected_24")!),
+        CommandingItem(title: "Item 5", image: UIImage(named: "Home_24")!, action: commandAction, selectedImage: UIImage(named: "Home_Selected_24")!)
```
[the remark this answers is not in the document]
lcapkovic: Yup to all of these, the demo controller is completely different now :D

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        // On a horizontal size class change the top level sheet / bar surfaces get recreated,
+        // but the item views, containers and bindings persist and are rearranged during the individual setup functions.
+        if previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass {
+            bottomSheetController?.willMove(toParent: nil)
+            bottomSheetController?.removeFromParent()
+            bottomSheetController?.view.removeFromSuperview()
+            bottomSheetController = nil
```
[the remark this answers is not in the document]
lcapkovic: Did I get too used to it or is it actually more readable with the optionals because of no nesting? It's like in objc when you send a bunch of messages to an object that might be nil but we don't really care.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,592 @@
...
+        }
+    }
+
+    public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        // On a horizontal size class change the top level sheet / bar surfaces get recreated,
+        // but the item views, containers and bindings persist and are rearranged during the individual setup functions.
+        if previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass {
+            bottomSheetController?.willMove(toParent: nil)
```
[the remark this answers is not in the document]
lcapkovic: yeah that's better, thanks!

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: It would just show the title, but the fluent staticImage API will crash the app even before that, which is probably a good thing.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: Future PR :D This goes in the dev branch, not main yet.

ios/FluentUI/Bottom Sheet/CommandingItem.swift
lcapkovic: I wanted to do something like it, but I'm not sure how I can apply that pattern here. We need the items to notify the controller about changes, and that can't be captured by a protocol. PopupMenuController doesn't have that problem because changes to the items don't reflect in the UI after passing them to the controller.
Clients can still subclass to pipe stuff through, while keeping our property observers intact.

lcapkovic:
> > is it just me or do I observe that after you rotate the bottomcommandingcontroller the text(title label) is no longer center aligned compared to the icons in the hero commands
> 
> I noticed this today as well. I'll look into it. Might be a TabBarItemView bug.

@harrieshin It is a TabBarItemView bug, I'll have a separate PR out for that fix.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
Saadnajmi: Ah, ok. If it was going into main, I would have requested a bug #. Cool!

ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,122 @@
...
+        self.commandType = commandType
+    }
+
+    @objc public enum CommandType: Int {
+        /// Calls `action` on tap without modifying the item.
+        case simple
+
+        /// Toggles `isOn` on tap before calling `action`.
+        case toggle
+    }
```
anandrajeswaran, line 109: probably could simplify by just representing this as a bool property instead of an enum?

ios/FluentUI/Bottom Sheet/CommandingItem.swift
```diff
@@ -0,0 +1,122 @@
...
+
+protocol CommandingItemDelegate: class {
+    func commandingItem(_ item: CommandingItem, didChangeTitleTo value: String)
+    func commandingItem(_ item: CommandingItem, didChangeImageTo value: UIImage)
+    func commandingItem(_ item: CommandingItem, didChangeLargeImageTo value: UIImage?)
+    func commandingItem(_ item: CommandingItem, didChangeSelectedImageTo value: UIImage?)
+    func commandingItem(_ item: CommandingItem, didChangeCommandTypeTo value: CommandingItem.CommandType)
+    func commandingItem(_ item: CommandingItem, didChangeOnTo value: Bool)
+    func commandingItem(_ item: CommandingItem, didChangeEnabledTo value: Bool)
+}
```
anandrajeswaran, line 122: Documentation for these?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -8,6 +8,14 @@ import UIKit
 class TabBarItemView: UIView {
     let item: TabBarItem
 
+    var isEnabled: Bool = true {
+        didSet {
+            titleLabel.isEnabled = isEnabled
+            imageView.tintAdjustmentMode = isEnabled ? .automatic : .dimmed
+            isUserInteractionEnabled = isEnabled
```
anandrajeswaran, line 15: Having to call isUserInteractionEnabled is usually a sign we're doing weird things that UIButtons and related types should be doing for free for us, but I guess that's pre-existing for TabBarItemView since it inherits from UIView?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+    /// or the main bottom bar area, depending on current horizontal UIUserInterfaceSizeClass.
+    ///
+    /// Only the first 5 items will be retained and displayed.
+    @objc open var heroItems: [CommandingItem] = [] {
+        willSet {
+            clearAllItemViews(in: .heroSet)
+        }
+        didSet {
+            if heroItems.count > 5 {
+                assertionFailure("At most 5 hero commands are supported. Only the first 5 items will be used.")
```
anandrajeswaran, line 32: Should avoid assertionFailure since it actually hardcodes the full file path of the compilation unit in the outputted binary. preconditionFailure should be used instead

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+                heroItems = Array(heroItems.prefix(5))
```
anandrajeswaran, line 33: It's a bit weird that we both call out an assertion (or precondition) failure but then gracefully handle it. Just remove the graceful handling?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+            }
+
+            if isHeroCommandStackLoaded {
+                let newViews = heroItems.map { createAndBindHeroCommandView(with: $0) }
+                newViews.forEach { heroCommandStack.addArrangedSubview($0) }
```
anandrajeswaran, line 38: minor: just do this all in one loop with heroItems.forEach?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+        }
+    }
+
+    /// Sections with items to be displayed in the list area.
+    @objc open var expandedListSections: [CommandingSection] = [] {
+        willSet {
+            clearAllItemViews(in: .list)
+        }
+        didSet {
+            expandedListSections.forEach { $0.items.forEach { $0.delegate = self }}
```
anandrajeswaran, line 49: This seems like a worthy candidate to split into multiple lines visually. Quite related: having nested $0 is all sorts of confusing and maybe at least one of the forEach loops should use "<name> in" to get a more explicit name

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+        guard previousTraitCollection?.horizontalSizeClass != traitCollection.horizontalSizeClass else {
+            return
+        }
+
+        // On a horizontal size class change the top level sheet / bar surfaces get recreated,
+        // but the item views, containers and bindings persist and are rearranged during the individual setup functions.
+        bottomSheetController?.willMove(toParent: nil)
+        bottomSheetController?.removeFromParent()
+        bottomSheetController?.view.removeFromSuperview()
+        bottomSheetController = nil
```
anandrajeswaran, line 83: minor: use if let to prevent repeated optional unwraps

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+
+        bottomSheetController = sheetController
+        updateExpandability()
+    }
+
+    private func makeBottomBarByEmbedding(contentView: UIView) -> UIView {
+        let bottomBarView = UIView()
+        bottomBarView.layer.shadowColor = Constants.BottomBar.Shadow.color
+        bottomBarView.layer.shadowOpacity = Constants.BottomBar.Shadow.opacity
+        bottomBarView.layer.shadowRadius = Constants.BottomBar.Shadow.radius
```
anandrajeswaran, line 163: Can avoid redundant method invocations under the hood and yield cleaner code overall by saving bottomBarView.layer to a local instead of fetching it for each call

ios/FluentUI/Tab Bar/TabBarItemView.swift
anandrajeswaran: could just disable the gesture recognizer? Otherwise won't something like large content viewer also not work? Actually, should that work on disabled items?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+        ])
+
+        return moreButtonView
+    }()
+
+    private lazy var heroCommandStack: UIStackView = {
+        let itemViews = heroItems.map { createAndBindHeroCommandView(with: $0) }
+        let stackView = UIStackView(arrangedSubviews: itemViews)
+        stackView.translatesAutoresizingMaskIntoConstraints = false
+        stackView.axis = .horizontal
```
anandrajeswaran, line 245: Isn't this the documented default value?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+            binding.item.isOn = tabBarItemView.isSelected
+            fallthrough
+        case .simple:
+            binding.item.action(binding.item)
+        }
+    }
+
+    @objc private func handleMoreButtonTap(_ sender: UITapGestureRecognizer) {
+        let popoverContentVC = UIViewController()
+        popoverContentVC.view.addSubview(tableView)
```
anandrajeswaran, line 288: or just popoverContentVC.view = tableView? But generally with UITableViews iOS makes things so much easier for us if we just use UITableViewController

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+    }
+
+}
+
+extension BottomCommandingController: UITableViewDataSource {
+    public func numberOfSections(in tableView: UITableView) -> Int {
+        return expandedListSections.count
+    }
+    public func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        assert(section < expandedListSections.count)
```
anandrajeswaran, line 486: from the declaration I would guess this has the same badness as assertionFailure in that it seems to by default encode the #file. Would recommend just using preconditionFailure

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+            self.item = item
+            self.view = view
+            self.location = location
+        }
+    }
+
+    private class HeroItemBinding: ItemBinding {
+        let widthConstraint: NSLayoutConstraint
+
+        init(item: CommandingItem, view: UIView, location: ItemLocation, widthConstraint: NSLayoutConstraint) {
```
anandrajeswaran, line 437: Not sure how I feel about location and widthConstraint being part of the binding. In my mind "binding" is a simple map and that sort of info lives on either the view or item itself.

Maybe it's just the name "binding" though? Because that also implies some level of functionality tying things together but that functionality seems to live in the usage rather than the implementation?

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
```diff
@@ -0,0 +1,596 @@
...
+        case let booleanCell as BooleanCell:
+            if booleanCell.isOn != value {
+                booleanCell.isOn = value
+            }
+        default:
+            break
+        }
+    }
+
+    func commandingItem(_ item: CommandingItem, didChangeCommandTypeTo value: CommandingItem.CommandType) {}
```
anandrajeswaran, line 595: Probably add a comment on why this is empty at least (if not just remove it)

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: It's more of BindingMetadata. Constraints are external to the item views, and items shouldn't really know how they're represented so that also feels wrong. So they should be somewhere in this controller IMO and this was a nice way to keep them tied to a binding.

anandrajeswaran: Maybe BindingMetadata is better? Binding makes me think of what that word means in SwiftUI and the scope of this is beyond that

ios/FluentUI/Bottom Sheet/CommandingItem.swift
anandrajeswaran: That's a good point. I guess technically we'd have to treat it as a breaking change, but in practice I'd imagine builds probably wouldn't break because clients wouldn't be exhausting all cases. I'd imagine only shared code would be doing that.

That said, I'd expect future iterations to still be buttons so maybe better for this to be a bool so future "types" can be composed together with this? Like I think it's more likely we come up with options to change the styling that would impact either CommandType than it is we come up with something that is mutually exclusive with simple and toggle

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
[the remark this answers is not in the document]
Saadnajmi: Yeah, I think the main reason is to limit the number of optionals, and thus the number of places code could fail. These all happen to be accessing the same optional, which only makes me want to wrap it inside can `if let` even more.
Definitely a very swift-ism though.

ios/FluentUI/Bottom Sheet/BottomCommandingController.swift
lcapkovic: We only need to have it in a controller when it's hosted in a popover. In the sheet it also has the hero stack and probably a grid in the future, so we don't really need a UITableViewController which would force me to implement the datasource and delegate in there, forcing me to somehow pipe through relevant item data from this main controller.

anandrajeswaran (approved):

Saadnajmi (approved):

PR #561, opened 2021-05-10
Adding MSFView in List Vnext component

kubalani:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Currently List support two variants of each datasource/delegate that interacts with UIView/AnyView. 
This view consolidates them into single wrapper view `MSFView`. 

E.g Instead of 
```swift
   @Published public var leadingView: AnyView? 
   
   @objc public var leadingUIView: UIView? {
        didSet {
            guard let view = leadingUIView else {
                leadingView = nil
                return
            }

            leadingView = AnyView(UIViewAdapter(view))
        }
    }
```

We now have 

```swift
  @objc @Published public var leadingView: MSFView? 
```

usage for SwiftUI  : `leadingView.swiftView`
Objective-C : `leadingView.uiView`

This would allow extension of views without duplicating additional methods for selection/disabled state. 

### Verification

LeftNav 
Screen Shot 2021-05-10 at 3 53 05 PM

ListDemo
Screen Shot 2021-05-10 at 3 52 56 PM

Objective-C Demo
Screen Shot 2021-05-10 at 3 52 45 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/561)

### DIFF
--- a/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
+++ b/ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
@@ -58,3 +58,23 @@ struct UIViewAdapter: UIViewRepresentable {
         stackView.addArrangedSubview(makeView())
     }
 }
+
+/// This is View wrapper that contains a UIView, View for SwiftUI/UIKit components.
+@objc public class MSFView: NSObject {
+
+    // Retrieve View for SwiftUI rendering
+    public let swiftView: AnyView
+
+    // Retrieve View for UIKit rendering
+    public let UIView: UIView
+
+    public init(_ swiftView: AnyView) {
+        self.swiftView = swiftView
+        self.UIView = UIHostingController(rootView: swiftView).view
+    }
+
+    @objc public init(_ uiView: UIView) {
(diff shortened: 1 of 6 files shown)


ios/FluentUI/Vnext/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -58,3 +58,23 @@ struct UIViewAdapter: UIViewRepresentable {
...
+    // Retrieve View for UIKit rendering
+    public let UIView: UIView
+
+    public init(_ swiftView: AnyView) {
+        self.swiftView = swiftView
+        self.UIView = UIHostingController(rootView: swiftView).view
+    }
+
+    @objc public init(_ uiView: UIView) {
+        self.UIView = uiView
```
rdeassis, line 77: I think creating a UIHostingController or a UIViewAdapter anytime we create this class won't be needed in many scenarios. Which is resources we're failing to safe.

PR #573, opened 2021-05-14
Fixing large text scenarios for TableViewCellFileAccessoryView.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The Large Text scenario is broken for the TableViewCellFileAccessoryView.
The cause of this particular issue is that the customAccessoryView is not being considered to calculate the available width for the text. This interferes in the calculation as the height returned considers the entire row width and ends up to be a smaller value.

The fix is to ensure that the width of the customAccessoryView is considered so that the height of the labels returned is accurate.

### Verification

Exercised the same repro steps demonstrated in the first screenshot and ensured those scenarios look correct as demonstrated in the second screenshot.

| largetext_before | largetext_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/573)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -318,10 +318,10 @@ open class TableViewCell: UITableViewCell {
         let textAreaLeadingOffset = self.textAreaLeadingOffset(customViewSize: customViewSize, isInSelectionMode: isInSelectionMode, paddingLeading: paddingLeading)
         let textAreaTrailingOffset = self.textAreaTrailingOffset(customAccessoryView: customAccessoryView, customAccessoryViewExtendsToEdge: customAccessoryViewExtendsToEdge, accessoryType: accessoryType, paddingTrailing: paddingTrailing)
         var textAreaWidth = containerWidth - (textAreaLeadingOffset + textAreaTrailingOffset)
-        if textAreaWidth < Constants.textAreaMinWidth, let customAccessoryView = customAccessoryView {
+        if let customAccessoryView = customAccessoryView {
             let oldAccessoryViewWidth = customAccessoryView.frame.width
             let availableWidth = oldAccessoryViewWidth - (Constants.textAreaMinWidth - textAreaWidth)
-            customAccessoryView.frame.size = customAccessoryView.systemLayoutSizeFitting(CGSize(width: availableWidth, height: .infinity))
+            customAccessoryView.frame.size = customAccessoryView.systemLayoutSizeFitting(CGSize(width: availableWidth, height: customAccessoryView.frame.size.height))
             textAreaWidth += oldAccessoryViewWidth - customAccessoryView.frame.width
         }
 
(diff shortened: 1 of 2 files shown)


harrieshin:
there is another issues where if the date time string is long, it won't grow correctly. isn't the date label always single line which might not be good for smaller screen?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -318,10 +318,10 @@ open class TableViewCell: UITableViewCell {
         let textAreaLeadingOffset = self.textAreaLeadingOffset(customViewSize: customViewSize, isInSelectionMode: isInSelectionMode, paddingLeading: paddingLeading)
         let textAreaTrailingOffset = self.textAreaTrailingOffset(customAccessoryView: customAccessoryView, customAccessoryViewExtendsToEdge: customAccessoryViewExtendsToEdge, accessoryType: accessoryType, paddingTrailing: paddingTrailing)
         var textAreaWidth = containerWidth - (textAreaLeadingOffset + textAreaTrailingOffset)
-        if textAreaWidth < Constants.textAreaMinWidth, let customAccessoryView = customAccessoryView {
+        if let customAccessoryView = customAccessoryView {
             let oldAccessoryViewWidth = customAccessoryView.frame.width
             let availableWidth = oldAccessoryViewWidth - (Constants.textAreaMinWidth - textAreaWidth)
-            customAccessoryView.frame.size = customAccessoryView.systemLayoutSizeFitting(CGSize(width: availableWidth, height: .infinity))
+            customAccessoryView.frame.size = customAccessoryView.systemLayoutSizeFitting(CGSize(width: availableWidth, height: customAccessoryView.frame.size.height))
```
harrieshin, line 324: this won't affect regular tableviewcell examples customaccessoryview cases?

harrieshin (approved):

harrieshin:
nit: can we update the demo subtitle case to show all the text when the larger text size is increase without any truncation?

ios/FluentUI/Table View/TableViewCell.swift
rdeassis: I have tested the other tableview scenarios (persona list, tableviewcell demo controller) and they render correctly as the size available for the text (textAreaWidth) is proportional to the accessoryview.

rdeassis:
> nit: can we update the demo subtitle case to show all the text when the larger text size is increase without any truncation?

Here are the screenshots of it (regular and large text respectively:

Screen Shot 2021-05-14 at 4 43 56 PM
Screen Shot 2021-05-14 at 4 43 49 PM

harrieshin:
nit: are we missing a border between the cell?

kubalani (approved):

rdeassis:
> there is another issues where if the date time string is long, it won't grow correctly. isn't the date label always single line which might not be good for smaller screen?

Yes. That's another issue that we also need to resolve. This fix does not address that issue.

harrieshin:
> > there is another issues where if the date time string is long, it won't grow correctly. isn't the date label always single line which might not be good for smaller screen?
> 
> Yes. That's another issue that we also need to resolve. This fix does not address that issue.

ah thanks for the clarification. in the demo, we should definitely add cell selection toggle button too then where things get even more tighter

rdeassis:
> nit: are we missing a border between the cell?
Fixed:
Screen Shot 2021-05-14 at 4 59 07 PM

PR #566, opened 2021-05-13
Various clean up of bottomSheet-dev before merge into main

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

While the Bottom Commanding control isn't fully featured yet, the API is in a good place to unblock app teams for integration work. This PR includes various clean up for the bottomSheet-dev branch before it can be merged into main:

- Updated Bottom Sheet demo controller to use table view. 
- Split up Bottom Sheet and Bottom Commanding into separate folders
- Localizable string for the "More" button in the commanding bottom bar
- Added xcfilelist with a reference to the "more" image asset for use in the subspec
- Added .button trait to the "More" button
- CocoaPods subspecs for BottomSheet and BottomCommanding

My plan is to merge bottomSheet-dev into main after this PR is merged. Future work will go in the main branch.
### Verification
Sanity checks in the test app.
### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [x] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/566)

PR #574, opened 2021-05-18
Merge in-progress Bottom Sheet work into Main

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR just follows from #566 to do the actual merge of bottomSheet-dev into main to unblock client integration work. Everything in this PR has already been reviewed and future feature work and bug fixes will happen as PRs into main directly.

See the list of commits and corresponding PRs to see what this merges in, but basically it's:
Bottom Sheet
Bottom Commanding
Corresponding demo controllers
Corresponding cocoapod subspecs

### Verification
I did the same squash merge locally and did sanity checks in the test app to verify there won't be any weird conflicts.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/574)


anandrajeswaran (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -0,0 +1,586 @@
...
+        NSLayoutConstraint.deactivate(heroCommandWidthConstraints)
+        heroCommandStack.distribution = .fillEqually
+
+        let commandStackContainer = UIView()
+        commandStackContainer.addSubview(heroCommandStack)
+
+        let sheetController = BottomSheetController(contentView: makeBottomSheetContent(headerView: commandStackContainer, expandedContentView: tableView))
+        sheetController.hostedScrollView = tableView
+        sheetController.collapsedContentHeight = bottomSheetHeroStackHeight
+        sheetController.expandedHeightFraction = Constants.BottomSheet.expandedFraction
```
harrieshin, line 133: is there a way for clients to put expanded height based on number of cells they want to expand? it's not always going to be clear cut of .5 or 1

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -0,0 +1,586 @@
...
+        isHeroCommandStackLoaded = true
+        return stackView
+    }()
+
+    private lazy var tableView: UITableView = {
+        let tableView = UITableView(frame: .zero, style: .grouped)
+        tableView.translatesAutoresizingMaskIntoConstraints = false
+        tableView.separatorStyle = .none
+        tableView.alwaysBounceVertical = false
+        tableView.isAccessibilityElement = true
```
harrieshin, line 257: we need this?

ios/FluentUI/Bottom Commanding/CommandingItem.swift
```diff
@@ -0,0 +1,127 @@
...
+        didSet {
+            if largeImage != oldValue {
+                delegate?.commandingItem(self, didChangeLargeImageTo: largeImage)
+            }
+        }
+    }
+
+    /// Indicates whether the command is currently on.
+    ///
+    /// When `isToggleable` is `true`, this property is toggled automatically before `action` is called.
```
harrieshin, line 58: is it be isOn?

ios/FluentUI/Bottom Commanding/CommandingItem.swift
```diff
@@ -0,0 +1,127 @@
...
+    @objc open var isEnabled: Bool {
+        didSet {
+            if isEnabled != oldValue {
+                delegate?.commandingItem(self, didChangeEnabledTo: isEnabled)
+            }
+        }
+    }
+
+    /// Indicates whether `isOn` should be toggled automatically before `action` is called.
+    @objc open var isToggleable: Bool {
```
harrieshin, line 77: this one only make sense if it not used for heroitem right and used for boolean cell? should we note that?

ios/FluentUI/Bottom Commanding/CommandingItem.swift
```diff
@@ -0,0 +1,127 @@
...
+            if largeImage != oldValue {
+                delegate?.commandingItem(self, didChangeLargeImageTo: largeImage)
+            }
+        }
+    }
+
+    /// Indicates whether the command is currently on.
+    ///
+    /// When `isToggleable` is `true`, this property is toggled automatically before `action` is called.
+    @objc open var isOn: Bool {
```
harrieshin, line 59: should it be isSelected? or is it for booleancell only?

Saadnajmi (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
harrieshin: might as well have respectsPreferredContentSize case in the demo app for demonstration and testability

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -0,0 +1,586 @@
...
+        addBinding(binding)
+
+        return itemView
+    }
+
+    private func setupTableViewCell(_ cell: TableViewCell, with item: CommandingItem) {
+        let iconView = UIImageView(image: item.image)
+        iconView.tintColor = Constants.tableViewIconTintColor
+        cell.setup(title: item.title, subtitle: "", footer: "", customView: iconView, customAccessoryView: nil, accessoryType: .none)
+        cell.bottomSeparatorType = .none
```
harrieshin, line 352: i think we should check if the cell is the last cell of the section and show full border. I thought if we had multiple sections we do have full separators

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -0,0 +1,258 @@
...
+        } else if expandedListSections.contains(where: { $0.items.contains(item) }) {
+            showMessage("Expanded list command tapped")
+        }
+    }
+
+    private func showMessage(_ message: String) {
+        let alert = UIAlertController(title: message, message: nil, preferredStyle: .alert)
+        present(alert, animated: true)
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + 1) {
```
harrieshin, line 157: do we need this?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -0,0 +1,150 @@
...
+        optionTableView.register(BooleanCell.self, forCellReuseIdentifier: BooleanCell.identifier)
+        optionTableView.register(ActionsCell.self, forCellReuseIdentifier: ActionsCell.identifier)
+        optionTableView.dataSource = self
+        optionTableView.delegate = self
+        optionTableView.separatorStyle = .none
+        view.addSubview(optionTableView)
+
+        let bottomSheetViewController = BottomSheetController(contentView: personaListView)
+        bottomSheetViewController.hostedScrollView = personaListView
+        bottomSheetViewController.expandedHeightFraction = 1.0
```
harrieshin, line 24: technically we don't need this right? because by defulat it is 1

ios/FluentUI/Bottom Commanding/CommandingItem.swift
lcapkovic: No it will work as described for both hero items and boolean cells.

ios/FluentUI/Bottom Commanding/CommandingItem.swift
lcapkovic: I felt like isOn is a better general name that covers both the boolean cell and regular button cases. isSelected really only makes sense for buttons.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: More intelligent scaling is coming in the future. This is more of a placeholder for now. In the future we'll be deciding between two expansion states and one, based on how much space the expanded content takes up.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Looks like we don't! Thanks

ios/FluentUI/Bottom Commanding/CommandingItem.swift
Saadnajmi: Thought about this more, and "selected" is a name that is also used for checkboxes, and for design tokens (button-selected vs button-rest, etc). Also I had to do a double take to think "Does isOn actually mean isEnabled?" 
I'm feeling isSelected more now..

lcapkovic: I'll point out that this was already reviewed and approved 😄 I guess I'm down to change it, but the reasoning here was that isOn is more view-agnostic. It just says the command is on, regardless of how it's represented. Switches use isOn, buttons use isSelected. isSelected doesn't make as much sense with switches. Another issue is that cell with a switch in it technically has independent isSelected and isOn states. We don't expose that through this API right now, but using isSelected would cause issues in that area in the future, for example if we wanted to support grouped selectable cells with checkboxes in them.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: Yes because this isn't subclassing DemoController anymore. The messages were the only thing we were using from it, and the way it adds a UIStackView in viewDidLoad was breaking the view hierarchy. It looks like DemoController should be updated in the future, now that we're switching to table views for the config options.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: yup

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
harrieshin: not really a fan of autodismiss. we can't just wait till user press ok or something to dismiss the alert? seems weird precedence and not very acessibiltiy friendly either.

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: I agree with that, but for this kind of thing I dislike manual dismiss even more. We just want to indicate a command was triggered, ideally without taking away focus or anything like that. Maybe a NotificationView would be best for this sort of thing?

lcapkovic: I switched to dismissible alerts for now. In the future I'd like to use a lighter toast, but there was an issue where it was overlapping the collapsed sheet so I'd need to spend more time on that.

PR #575, opened 2021-05-18
Use constraints for TabBarItemView image resizing

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### Problem
The TabBarItemView can be initialized with the `canResizeImage` enabled, which allows the item view to resize the icon image to whatever size it needs (24/28 depending on title visibility and screen orientation). 

The current implementation of this resizing logic has a few bugs that caused issues when I used this view in the BottomCommandingController:
1. The custom size is only applied inside a sizeThatFits override. The docs for this say "This method does not resize the receiver" so we shouldn't have this side effect here. It also means the resizing is never applied if sizeThatFits is not called.
2. The actual resizing is done by setting a frame with the target width and height and zero origin. This means that most of the time the resizing won't even stick, because any future layout pass will override the frame back to the original size and position. Zero origin also isn't what we always want, so when the frame change does stick (because sometimes the change doesn't trigger a relayout), the image view ends up in the wrong position.

Before orientation change:
Screen Shot 2021-05-17 at 7 35 52 PM

After orientation change:
(sizeThaFits was called which set the image frame to zero origin and didn't trigger a relayout)
Screen Shot 2021-05-17 at 7 36 54 PM

The logic of when a relayout of the parent stack is actually done is kind of weird. I found that:
- when you set a frame with the same dimensions but a different origin, this doesn't trigger a relayout and your frame change sticks (for a bit) - that's the bug above
- when you set a frame with different dimensions, this triggers a relayout of the parent stack, which immediately reverses any of those frame changes

#### Fix
Use constraints for the custom width/height so AutoLayout knows what we want and it can always generate the correct image view frames.
### Verification
Local testing using hardcoded `alwaysShowTitleBelowImage` in `TabBarView`. Also used incorrectly sized image assets to test the resizing.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/575)

### DIFF
--- a/ios/FluentUI/Tab Bar/TabBarItemView.swift
+++ b/ios/FluentUI/Tab Bar/TabBarItemView.swift
@@ -173,7 +173,6 @@ class TabBarItemView: UIView {
         let imageView = UIImageView(frame: .zero)
         imageView.contentMode = .scaleAspectFit
         imageView.tintColor = Constants.unselectedColor
-        imageView.contentMode = .scaleAspectFit
 
         if canResizeImage {
             let sizeConstraints = (
--- a/ios/FluentUI/Tab Bar/TabBarItemView.swift
+++ b/ios/FluentUI/Tab Bar/TabBarItemView.swift
@@ -176,17 +176,18 @@ class TabBarItemView: UIView {
         imageView.contentMode = .scaleAspectFit
 
         if canResizeImage {
-            imageViewWidthHeightConstraints = [
-                imageView.widthAnchor.constraint(equalToConstant: suggestImageSize),
-                imageView.heightAnchor.constraint(equalToConstant: suggestImageSize)
-            ]
-            NSLayoutConstraint.activate(imageViewWidthHeightConstraints)
+            let sizeConstraints = (
+                width: imageView.widthAnchor.constraint(equalToConstant: suggestImageSize),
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -172,14 +169,26 @@ class TabBarItemView: UIView {
         return container
     }()
 
-    private let imageView: UIImageView = {
+    private lazy var imageView: UIImageView = {
         let imageView = UIImageView(frame: .zero)
         imageView.contentMode = .scaleAspectFit
         imageView.tintColor = Constants.unselectedColor
+        imageView.contentMode = .scaleAspectFit
```
anandrajeswaran, line 176: duplicate of line 174?

anandrajeswaran (approved):

ios/FluentUI/Tab Bar/TabBarItemView.swift
lcapkovic: I am dumb.

harrieshin (approved):

PR #577, opened 2021-05-19
Add  UISwitch `isEnabled` property in BooleanCell for customization

ForeverQ:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

According to the requirement of an in-progress feature in Outlook iOS in Microsoft, we need to have the disabled style of UISwitch inside BooleanCell. So we make this change then callers can customize.

### Verification

I included the update locally and did sanity checks in the test app to verify the cell works as expected with and without the property been setup.

| `isSwitchEnabled` == true                                | `isSwitchEnabled` == false                            |
|Simulator Screen Shot - iPhone 12 Pro - 2021-05-19 at 22 01 19|Simulator Screen Shot - iPhone 12 Pro - 2021-05-19 at 22 00 35|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/577)

### DIFF
--- a/ios/FluentUI/Other Cells/BooleanCell.swift
+++ b/ios/FluentUI/Other Cells/BooleanCell.swift
@@ -29,8 +29,19 @@ open class BooleanCell: TableViewCell {
 
     /// Updates the cell accessory view's `isEnabled` property
     @objc public var isSwitchEnabled: Bool {
-      get { return `switch`.isEnabled }
-      set { `switch`.isEnabled = newValue }
+      get {
+        return `switch`.isEnabled && isEnabled
+      }
+      set {
+        `switch`.isEnabled = newValue
+        updateAccessibility()
+      }
+    }
+
+    @objc open override var isEnabled: Bool {
+        didSet {
+            updateAccessibility()
+        }
     }
 
     /// `onValueChanged` is called when the cell accessory view's value has changed
@@ -41,6 +52,26 @@ open class BooleanCell: TableViewCell {
         `switch`.addTarget(self, action: #selector(handleOnSwitchValueChanged), for: .valueChanged)
         return `switch`
     }()
+
+    @objc public override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        super.init(style: style, reuseIdentifier: reuseIdentifier)
+        updateAccessibility()
+    }
+
+    public required init?(coder aDecoder: NSCoder) {
+        super.init(coder: aDecoder)
+        updateAccessibility()
+    }
+
+    /// Sets up the content of the cell
+    ///
+    /// - Parameters:
+    ///   - title: Text that appears as a single line
+    ///   - customView: An optional custom view that appears on the leading edge, adjacent to the start of the `title` text
+    ///   - isOn: A boolean value describing the `isOn` state of the accessory view
+    @objc open func setup(title: String, customView: UIView? = nil, isOn: Bool = false) {
+        setup(title: title, customView: customView, isOn: isOn, isSwitchEnabled: true)
+    }
 
     /// Sets up the content of the cell
     ///
@@ -63,6 +94,14 @@ open class BooleanCell: TableViewCell {
         super.didMoveToWindow()
         if let window = window {
             `switch`.onTintColor = Colors.primary(for: window)
+        }
+    }
+
+    private func updateAccessibility() {
+        if isEnabled && isSwitchEnabled {
+            accessibilityTraits.remove(.notEnabled)
+        } else {
+            accessibilityTraits.insert(.notEnabled)
         }
     }
 }
--- a/ios/FluentUI/Other Cells/BooleanCell.swift
+++ b/ios/FluentUI/Other Cells/BooleanCell.swift
@@ -15,6 +15,8 @@ public typealias MSBooleanCell = BooleanCell
 
  Use the `isOn` property to update the cell accessory view's `isOn` property.
 
+ Use the `isSwitchEnabled` property to update the cell accessory view's `isEnabled` property.
+
  Use `onValueChanged` as a callback for an event that changes the value of the accessory view.
  */
 @objc(MSFBooleanCell)
@@ -23,6 +25,12 @@ open class BooleanCell: TableViewCell {
     @objc public var isOn: Bool {
         get { return `switch`.isOn }
         set { `switch`.isOn = newValue }
+    }
+
+    /// Updates the cell accessory view's `isEnabled` property
+    @objc public var isSwitchEnabled: Bool {
+      get { return `switch`.isEnabled }
+      set { `switch`.isEnabled = newValue }
     }
 
     /// `onValueChanged` is called when the cell accessory view's value has changed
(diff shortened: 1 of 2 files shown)


harrieshin:
shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?

lcapkovic:
> shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?

Disabling the entire cell including the switch seems to already work through the superclass isEnabled property:
Screen Shot 2021-05-19 at 1 43 41 PM

This PR seems to want to only disable the switch, although I'm not exactly sure when that would be a better design than just disabling the entire cell.

harrieshin:
> > shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?
> 
> Disabling the entire cell including the switch seems to already work through the superclass isEnabled property:
> Screen Shot 2021-05-19 at 1 43 41 PM
> 
> This PR seems to want to only disable the switch, although I'm not exactly sure when that would be a better design than just disabling the entire cell.

yeah i think it would be better if the booleancell is overriding the isEnable property of the superclass and align the switch accordingly and not have the additional isSwitchEnable property

lcapkovic:
> > > shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?
> > 
> > 
> > Disabling the entire cell including the switch seems to already work through the superclass isEnabled property:
> > Screen Shot 2021-05-19 at 1 43 41 PM
> > This PR seems to want to only disable the switch, although I'm not exactly sure when that would be a better design than just disabling the entire cell.
> 
> yeah i think it would be better if the booleancell is overriding the isEnable property of the superclass and align the switch accordingly and not have the additional isSwitchEnable property

The override isn't even necessary, it works without any code change.

harrieshin:
> > > > shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?
> > > 
> > > 
> > > Disabling the entire cell including the switch seems to already work through the superclass isEnabled property:
> > > Screen Shot 2021-05-19 at 1 43 41 PM
> > > This PR seems to want to only disable the switch, although I'm not exactly sure when that would be a better design than just disabling the entire cell.
> > 
> > 
> > yeah i think it would be better if the booleancell is overriding the isEnable property of the superclass and align the switch accordingly and not have the additional isSwitchEnable property
> 
> The override isn't even necessary, it works without any code change.

and the accessibility work fine (VoiceOver) even if we didn't change the switch disabled?

ForeverQ:
> > > > > shouldn't the entire cell be disabled? the text color is lighter and the switch is disabled (rather than just switch to be disabled)?
> > > > 
> > > > 
> > > > Disabling the entire cell including the switch seems to already work through the superclass isEnabled property:
> > > > Screen Shot 2021-05-19 at 1 43 41 PM
> > > > This PR seems to want to only disable the switch, although I'm not exactly sure when that would be a better design than just disabling the entire cell.
> > > 
> > > 
> > > yeah i think it would be better if the booleancell is overriding the isEnable property of the superclass and align the switch accordingly and not have the additional isSwitchEnable property
> > 
> > 
> > The override isn't even necessary, it works without any code change.
> 
> and the accessibility work fine (VoiceOver) even if we didn't change the switch disabled?

I tested locally, if the whole cell is disabled, VoiceOver anounces the state with one more word "dimmed". 

For the intention of this PR, it is indeed to only disable the switch and it is from [this feature spec](https://www.figma.com/file/vt80h6qPZOkqAGHfnUxXdb/AADC?node-id=1432%3A12859) (The spec is not essential for our discussion but feel free to take a look if you have permissions), an in-progress Outlook feature. I assume it won't easily challenge the designer unless we're unable to implement that.

In my opinion, both disabling the whole cell and only the switch make sense to design perspective in different scenarios. So at least this change provides one more alternative on customization which is meaningful. The only trick I found just now is that it seems no indication in VoiceOver when only switch is disabled. So I decide to fix this part afterwards.

What do you think about moving this PR forward other than the VoiceOver issue I just found?

harrieshin:
`no indication in VoiceOver when only switch is disabled. So I decide to fix this part afterwards.`

What does the default iOS UITableViewCell with UISwitch behaves in VoiceOver? if you can, can you address VO issue in this PR? we always want to make sure VO experience is good.

ios/FluentUI/Other Cells/BooleanCell.swift
```diff
@@ -25,6 +27,12 @@ open class BooleanCell: TableViewCell {
         set { `switch`.isOn = newValue }
     }
 
+    /// Updates the cell accessory view's `isEnabled` property
+    @objc public var isSwitchEnabled: Bool {
+      get { return `switch`.isEnabled }
```
harrieshin, line 32: should we also take into account of the super class isEnabled? so that when it is dimmed, switch also reflects that

ios/FluentUI/Other Cells/BooleanCell.swift
```diff
@@ -40,9 +48,11 @@ open class BooleanCell: TableViewCell {
     ///   - title: Text that appears as a single line
     ///   - customView: An optional custom view that appears on the leading edge, adjacent to the start of the `title` text
     ///   - isOn: A boolean value describing the `isOn` state of the accessory view
-    @objc open func setup(title: String, customView: UIView? = nil, isOn: Bool = false) {
+    ///   - isSwitchEnabled: A boolean value describing the `isEnabled` state of the accessory view
+    @objc open func setup(title: String, customView: UIView? = nil, isOn: Bool = false, isSwitchEnabled: Bool = true) {
```
harrieshin, line 52: can we create a convenience method that doesn't take in this new optional variable? the reason I ask this, is that the existing objc caller(clients) need to update the new param otherwise when they upgrade the library version it will have a compiler error.

lcapkovic:
> `no indication in VoiceOver when only switch is disabled. So I decide to fix this part afterwards.`
> 
> What does the default iOS UITableViewCell with UISwitch behaves in VoiceOver? if you can, can you address VO issue in this PR? we always want to make sure VO experience is good.

I just tested things with Accessibility Inspector and VO in the settings app. It seems like for system cells with switches, the entire cell is an accessibility element of type "switch" and the UISwitch itself doesn't get VO focus. This means that it would be hard to distinguish the "entire cell disabled" and "only switch disabled" variants because from a11y perspective, the entire cell is the switch.  That might not be a problem for this PR if the goal is just to distinguish the two variants visually. 
To achieve good a11y here I think the only change BooleanCell needs is to have type "switch" instead of "button", and for it to appear disabled to VO whether it's just the switch or the entire cell.

ios/FluentUI/Other Cells/BooleanCell.swift
ForeverQ: I think it's a good point, will update.

ForeverQ:
Sorry for the late update, I've been working on another priority of task. It seems the accessibility type of "button" was inserted in super class "TableViewCell" [here](https://github.com/microsoft/fluentui-apple/blob/bfacc5d819ac7e659d679303a5220eb72fa7bb59/ios/FluentUI/Table%20View/TableViewCell.swift#L1379). To avoid breaking any previous context, I don't plan to involve that part in this PR.

For a good a11y here, I updated the accessibilityTraits in BooleanCell to let us have the indicator when switch is disabled. I tested locally that now either disabling the "switch" inside or the entire "BooleanCell" will let VoiceOver anounce the state with "dimmed". So I think we're good on accessibility requirement.

harrieshin (approved):

ForeverQ:
Thanks for reminding, it's now updated.

ForeverQ:
@harrieshin Can wen merge this PR? It seems I don't have the access to click the merge button.

PR #579, opened 2021-05-19
Add support for boolean cells in `BottomCommandingController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds support for `BooleanCell` in `BottomCommandingController`. These are used when a `CommandingItem` has `isToggleable` set to `true`.

`CommandingItem.isToggleable` property change is internally handled in a special way, because the cell type changes so a full cell reload through the `UITableView` is necessary. 
### Verification

Updated the test app to include boolean cells. Also added a button to toggle random isOn states to make sure the binding works.

|Simulator Screen Shot - iPhone 12 - 2021-05-19 at 13 48 15|Simulator Screen Shot - iPhone 12 - 2021-05-19 at 13 47 01|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/579)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -257,6 +257,7 @@ open class BottomCommandingController: UIViewController {
         tableView.sectionFooterHeight = 0
         tableView.backgroundColor = Constants.tableViewBackgroundColor
         tableView.register(TableViewCell.self, forCellReuseIdentifier: TableViewCell.identifier)
+        tableView.register(BooleanCell.self, forCellReuseIdentifier: BooleanCell.identifier)
         tableView.register(TableViewHeaderFooterView.self, forHeaderFooterViewReuseIdentifier: TableViewHeaderFooterView.identifier)
         tableView.delegate = self
         tableView.dataSource = self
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -546,7 +568,16 @@ extension BottomCommandingController: CommandingItemDelegate {
     }
 
     func commandingItem(_ item: CommandingItem, didChangeToggleableTo value: Bool) {
-        reloadView(from: item)
+        guard let binding = itemToBindingMap[item] else {
+            return
+        }
+
+        // We need a UITableView.reloadRows call here because item.isToggleable changes the cell type
```
harrieshin, line 575: should we really support this scenario? who would do this? maybe togglable isn't something that is mutable?

harrieshin (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Keeping all the props mutable is good for OfficeSpace integration, because it sets user interface element properties one by one during the binding process. You're right, it doesn't really make much sense in regular use, but I don't see the harm in keeping it mutable.

anandrajeswaran: Well OfficeSpace doesn't have to. The general design is immutable properties get set by the factory and mutable properties are set during binding. We technically could eliminate the need for mutability in our usage if we don't think this is a realistic or intended scenario

anandrajeswaran (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Ah, I missed that. The factory would know whether it's creating an FSBooleanChoice so it can just init the item with the correct isToggleable value. Cool! I'll remove the mutability then.

PR #582, opened 2021-05-20
AvatarGroup vNext initial prototype

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Introducing AvatarGroup in vNext. This PR is showcasing the basics of Stack and Pile styles in AvatarGroup. MaxAvatars, Overflow, and borders are not included here.

### Verification

| Legacy                                       | vNext                                      |
| Simulator Screen Shot - iPhone 11 - 2021-05-20 at 03 02 12 |Simulator Screen Shot - iPhone 11 - 2021-05-20 at 03 02 27 |
| Simulator Screen Shot - iPhone 11 - 2021-05-20 at 03 03 03 | Simulator Screen Shot - iPhone 11 - 2021-05-20 at 03 04 17 |

### Pull request checklist

This PR has considered:


huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,58 @@
...
+    }
+
+    private func insertAvatarViews(style: MSFAvatarGroupStyle, showBorders: Bool) {
+        var constraints: [NSLayoutConstraint] = []
+
+        for size in MSFAvatarSize.allCases.reversed() {
+            let containerView = UIView(frame: .zero)
+
+            let avatarGroup = MSFAvatarGroup(style: style, size: size)
+            for index in 0...2 {
```
huwilkes, line 27: "for index in 0...2" seems a bit odd to me. Is there a reason we don't want to do all of the sample personas, or pass in the number of sample personas we'd like to use?

sophialee0416: I'm eventually going to add in a maxAvatarCount + overflowCount to determine how many personas to iterate over (not limited to just 3). I should have specified that, sorry!

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,58 @@
...
+import UIKit
+
+// MARK: - AvatarGroupDemoController
+
+class AvatarGroupDemoController: DemoController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        addTitle(text: "Avatar Stack")
+        insertAvatarViews(style: .stack, showBorders: false)
+        addTitle(text: "Avatar Pile")
```
harrieshin, line 16: shouldn't it also add without border to match the old version?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,96 @@
...
+                    })
+                    .padding(.trailing, tokens.interspace)
+            }
+        }
+    }
+
+    private struct AvatarCutout: Shape {
+        var xOrigin: CGFloat
+        var yOrigin: CGFloat
+        var cutoutSize: CGFloat
```
harrieshin, line 50: common on these variable? what is cutoutsize?

harrieshin:
update the description you are fixing group to square?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
sophialee0416: Yes I didn't add the border changes per our conversation offline earlier. I'll add that to the description as well!

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -230,6 +230,15 @@ MSFOverflowAvatarTokens extends MSFAvatarTokens:
...
   ringDefaultColor: $Colors.Background.neutralDisabled
 
+AP_MSFAvatarGroupTokens:
+  interspace: [ xSmall: $Spacing.xxSmall, small: $Spacing.xxSmall, medium: $Spacing.xSmall, large: $Spacing.xSmall, xlarge: $Spacing.xSmall, xxlarge: $Spacing.xSmall ]
+  ringInnerGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  ringThickness: [ xSmall: $Border.size.thin, small: $Border.size.thin, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  ringOuterGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  
+MSFAvatarGroupStackTokens extends MSFAvatarGroupTokens:
+  interspace: [ xSmall: -2pt, small: -2pt, medium: -4pt, large: -8pt, xlarge: -12pt, xxlarge: -12pt ]
```
kubalani, line 240: Use spacing variables instead constants for delta?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,96 @@
...
+}
+
+/// Properties that make up AvatarGroup content
+@objc public class MSFAvatarGroupState: NSObject, ObservableObject {
+    @objc @Published public var avatars: [MSFAvatarStateImpl] = []
+}
+
+/// View that represents the AvatarGroup
+public struct AvatarGroup: View {
+    @Environment(\.theme) var theme: FluentUIStyle
```
rdeassis, line 21: We need to have the window provider environment value here. (#551 for reference)

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,96 @@
...
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+
+/// Properties that make up AvatarGroup content
+@objc public class MSFAvatarGroupState: NSObject, ObservableObject {
+    @objc @Published public var avatars: [MSFAvatarStateImpl] = []
```
rdeassis, line 16: We want to expose just the protocol instead of the MSFAvatarStateImpl.

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -106,18 +106,18 @@ import SwiftUI
...
-    @Published var secondaryText: String?
-    @Published var ringColor: UIColor?
-    @Published var backgroundColor: UIColor?
-    @Published var foregroundColor: UIColor?
-    @Published var presence: MSFAvatarPresence = .none
-    @Published var hasPointerInteraction: Bool = false
-    @Published var isRingVisible: Bool = false
-    @Published var isTransparent: Bool = true
-    @Published var isOutOfOffice: Bool = false
+public class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
```
rdeassis, line 109: This should remain internal and we need to expose it as an MSFAvatarState.
We can create public APIs to create new avatar state internally and return the instance as a reference to the MSFAvatarState protocol.

ios/FluentUI/Vnext/Avatar/AvatarTokens.swift
```diff
@@ -10,7 +10,7 @@ import SwiftUI
 @objc public enum MSFAvatarStyle: Int, CaseIterable {
     case `default`
     case accent
-    case group
+    case square
```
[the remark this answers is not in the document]
sophialee0416: I see. I renamed it because the tokens I created would be named MSFAvatarGroupTokens which I felt were too similar to MSFGroupAvatarTokens? It's definitely not a permanent change and I'd love to discuss other names to use

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -230,6 +230,15 @@ MSFOverflowAvatarTokens extends MSFAvatarTokens:
   foregroundDefaultColor: $Colors.Foreground.neutral3
   ringDefaultColor: $Colors.Background.neutralDisabled
 
+AP_MSFAvatarGroupTokens:
+  interspace: [ xSmall: $Spacing.xxSmall, small: $Spacing.xxSmall, medium: $Spacing.xSmall, large: $Spacing.xSmall, xlarge: $Spacing.xSmall, xxlarge: $Spacing.xSmall ]
+  ringInnerGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  ringThickness: [ xSmall: $Border.size.thin, small: $Border.size.thin, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
```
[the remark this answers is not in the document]
sophialee0416: I think I only needed these 4 tokens so I did not because I think extending would take up more space? Would it be better to extend?

DrTacoMD (changes requested):

ios/FluentUI/Vnext/Avatar/AvatarTokens.swift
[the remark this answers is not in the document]
rdeassis: I think we should keep the style name as "group" since it's what matches the figma definition by the designers. In case we redefine names later, we can have a PR just for the renaming of types.

tools/sgen/input/FluentUIStyle.yml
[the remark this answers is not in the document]
rdeassis: I just don't see a reason why the Avatar will be modified when it's being displayed in the group.
The AvatarGroup should be just a collection of Avatars that will still be compliant with the original avatar design tokens. 
Is there any reason behind modifying the Avatars when they're displayed in the group?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,262 @@
...
+            addRow(text: size.description, items: [containerView], textStyle: .footnote, textWidth: 100)
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    private var vSamplePersonas: [MSFAvatarState] = {
+        var personas: [MSFAvatarState] = []
+        for index in 0...samplePersonas.count - 1 {
+            let state = MSFAvatar().state
```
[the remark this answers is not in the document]
rdeassis: I agree. The avatars should be created internally by the AvatarGroup and their state should be exposed to the calling code so they can change the state properties of individual avatars (without having access to the Avatars directly).

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,262 @@
...
+    private func insertAvatarViews(style: MSFAvatarGroupStyle, showBorders: Bool, mixed: Bool = false) {
+        var constraints: [NSLayoutConstraint] = []
+
+        for size in MSFAvatarSize.allCases.reversed() {
+            let containerView = UIView(frame: .zero)
+
+            let avatarGroup = MSFAvatarGroup(style: style, size: size)
+            var border = showBorders
+            for index in 0...avatarCount - 1 {
+                vSamplePersonas[index].isRingVisible = border
```
rdeassis, line 189: We could showcase the ability to set individual border settings to avatars in the control (as this is one of the scenarios that will be required for the AvatarGroup).

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
rdeassis: Have you added the designTokens modifier to AvatarGroup?
Without it the AvatarGroup overriden tokens won't be applied.

Can you make sure you test it by setting a property based on the window theme to ensure it changes accordingly?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    public var body: some View {
+        let size = tokens.size.size
+        let x: CGFloat = size + tokens.interspace - tokens.ringOuterGap - tokens.ringThickness
+        let avatars = state.avatars
+        let maxDisplayedAvatars = avatars.prefix(Int(state.maxDisplayedAvatars)).count
+        let overflowCount = (avatars.count > maxDisplayedAvatars ? UInt(avatars.count - maxDisplayedAvatars) : 0) + state.overflowCount
+        HStack(spacing: 0) {
+            ForEach(0 ..< maxDisplayedAvatars, id: \.self) { index in
+                let modify = tokens.style == .stack && (overflowCount > 0 || index + 1 < maxDisplayedAvatars)
```
rdeassis, line 41: Can we have a more descriptive name for this variable? 
Something like "shouldApplyCutout" or "needsCutout"?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,119 @@
...
+                AvatarView(style: .default, size: tokens.size, state: avatars[index])
+                    .modifyIf(modify, { view in
+                        view.mask(AvatarCutout(xOrigin: x,
```
rdeassis, line 44: Have you tested the cutout in the scenario where the avatars have presence set to something other than .none?
We should follow up on whether that's a valid scenario for the Stack style and at least call it out appropriately in the documentation of the control.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,119 @@
...
+    @objc public init(style: MSFAvatarGroupStyle = .stack, size: MSFAvatarSize = .large, theme: FluentUIStyle? = nil) {
+        super.init()
+
+        avatarGroupView = AvatarGroup(style: style, size: size)
+        hostingController = UIHostingController(rootView: AnyView(avatarGroupView
+                                                                    .windowProvider(self)
+                                                                    .modifyIf(theme != nil, { personaView in
+                                                                        personaView.customTheme(theme!)
+                                                                    })))
+        view.backgroundColor = UIColor.clear
```
rdeassis, line 97: It's safer to call the disableSafeAreaInsets here so we keep applying the workaround for scenarios where it's in the headers or anywhere close to the safe area limits.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,119 @@
...
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+
+/// Properties that make up AvatarGroup content
+@objc public class MSFAvatarGroupState: NSObject, ObservableObject {
```
rdeassis, line 15: Can we have the a public protocol named MSFAvatarGroupState and switch this to an internal class called MSFAvatarGroupStateImpl?
We should apply this pattern to avoid exposing our state implementation publicly.

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -168,6 +168,31 @@ public struct AvatarView: View {
         self.tokens = state.tokens
     }
 
+    public init(style: MSFAvatarStyle,
```
rdeassis, line 171: I agree with @DrTacoMD that we should not add an initializer parameter of type MSAvatarState and copy the properties.

I'm leaning towards pursuing the possibility of storing the AvatarView types in the state object and just expose their states publicly.
Then the rendering body property will just use the AvatarViews inside the state object here.

Can you explore this option?

DrTacoMD: Interesting, but I think I'd push back a little bit on this. There's a conceptual clarity that comes from our current separation, where state objects don't contain SwiftUI views directly, so all ownership is unidirectional. If we store `AvatarView` objects inside the state then we invert that relationship.

sophialee0416:
Closing PR due to design changes from feedback

PR #584, opened 2021-05-22
Add support for bottom safe area inset in BottomSheetController 

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### Problem
In our (and basically any) usage of a persistent bottom sheet, there is some notion of a header which is always visible, and main content which is only visible when you expand the sheet. The problem is that with the chunky iPhone bottom safe area inset, offsetting the collapsed sheet high enough to have the header in the safe area would mean revealing a part of the expanded content. This looks bad, so we need to hide the expanded content when the sheet is collapsed.

Example of the problematic case:
Screen Shot 2021-05-21 at 6 18 10 PM

#### Fix
To solve this, the sheet needs know about the distinct header + expanded content views. I added a new init which lets clients provide these two separate views. I also removed the old init methods (contentView and contentViewController), because they just complicate the implementation and I think most clients can get by with the split content pattern. It's the only way to always enforce good safe area handling. What's also good is that this would let us implement other cool effects in the future, like expanded content that gradually overlaps the header as the sheet expands.

Solution:

https://user-images.githubusercontent.com/3610850/119210727-a142c100-ba62-11eb-8b5f-7ab51d914370.mov

### Verification

Modified test app to use the header / expanded content split pattern. Also verified this works well in `BottomCommandingController`.

|Simulator Screen Shot - iPhone 12 - 2021-05-21 at 18 25 37Simulator Screen Shot - iPhone 12 - 2021-05-21 at 18 25 48|Simulator Screen Shot - iPhone 12 - 2021-05-21 at 18 38 59Simulator Screen Shot - iPhone 12 - 2021-05-21 at 18 26 43

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/584)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
@@ -5,7 +5,6 @@
 
 import FluentUI
 
-class BottomSheetDemoController: UIViewController {
 
     override func loadView() {
         view = UIView()
@@ -19,7 +18,7 @@ class BottomSheetDemoController: UIViewController {
         optionTableView.separatorStyle = .none
         view.addSubview(optionTableView)
 
-        let bottomSheetViewController = BottomSheetController(contentView: personaListView)
+        let bottomSheetViewController = BottomSheetController(sheetHeaderContentView: headerView, sheetExpandedContentView: personaListView)
         bottomSheetViewController.hostedScrollView = personaListView
 
         self.bottomSheetViewController = bottomSheetViewController
@@ -57,6 +56,20 @@ class BottomSheetDemoController: UIViewController {
         personaListView.personaList = samplePersonas
         personaListView.translatesAutoresizingMaskIntoConstraints = false
         return personaListView
+    }()
+
+    private let headerView: UIView = {
+        let view = UIView()
+        view.backgroundColor = .systemTeal
+        view.heightAnchor.constraint(equalToConstant: 70).isActive = true
+        return view
+    }()
+
+    private let expandedContentView: UIView = {
+        let view = UIView()
+        view.backgroundColor = .systemIndigo
+//        view.heightAnchor.constraint(equalToConstant: 400).isActive = true
+        return view
     }()
 
     private var bottomSheetViewController: BottomSheetController?
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
@@ -60,15 +60,26 @@ import FluentUI
 
     private let headerView: UIView = {
         let view = UIView()
-        view.backgroundColor = .systemTeal
-        view.heightAnchor.constraint(equalToConstant: 70).isActive = true
+        view.backgroundColor = .systemGray6
+        view.heightAnchor.constraint(equalToConstant: headerHeight).isActive = true
+
+        let label = UILabel()
+        label.text = "Sheet header view"
+        label.font = .systemFont(ofSize: 24)
+        label.translatesAutoresizingMaskIntoConstraints = false
+
+        view.addSubview(label)
+
+        NSLayoutConstraint.activate([
+            label.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            label.centerYAnchor.constraint(equalTo: view.centerYAnchor)
+        ])
         return view
     }()
 
     private let expandedContentView: UIView = {
         let view = UIView()
         view.backgroundColor = .systemIndigo
-//        view.heightAnchor.constraint(equalToConstant: 400).isActive = true
         return view
     }()
 
@@ -81,6 +92,8 @@ import FluentUI
             DemoItem(title: "Half screen expansion height", type: .action, action: #selector(halfScreenExpandedOffset))
         ]
     }()
+
+    private static let headerHeight: CGFloat = 70
 
     private enum DemoItemType {
         case action
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
@@ -68,7 +68,6 @@ class BottomSheetDemoController: UIViewController {
 
         let label = Label()
         label.text = "Header view"
-        label.font = .systemFont(ofSize: 24)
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -61,23 +55,11 @@ public class BottomSheetController: UIViewController {
         }
     }
 
-    /// Indicates if `preferredContentSize` of `contentViewController` should be respected.
-    /// Regardless of the value, the expanded height is limited by the height of `BottomSheetController`'s root view.
-    @objc open var respectsPreferredContentSize: Bool = false {
```
lcapkovic, line 66: This is going away for now because we don't support content UIViewControllers anymore. New sizing API (other than the current `expandedHeightFraction`) will be added together with multiple expansion states.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -19,8 +19,9 @@ class BottomSheetDemoController: UIViewController {
         optionTableView.separatorStyle = .none
         view.addSubview(optionTableView)
 
-        let bottomSheetViewController = BottomSheetController(contentView: personaListView)
+        let bottomSheetViewController = BottomSheetController(headerContentView: headerView, expandedContentView: personaListView)
```
harrieshin, line 22: this direction is bit concerning. I have seen variation of bottomsheet which only the gripper is shown behind the tabbarview and expand the sheet accordingly

harrieshin: seems like this kind of initialization is fine for commandingcontroller which has clear division of headercontentview and expanded content view

lcapkovic: What if I made the header optional? That should cover that usage scenario.

lcapkovic: I think forcing this type of division through the bottom sheet API is good, because it lets us enforce proper handling of the bottom safe area. Anything that shows a persistent sheet has some notion of a header (whatever shows in the collapsed state). And in the scenario you described above, there's just no header, which still fits this pattern.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -55,10 +56,30 @@ class BottomSheetDemoController: UIViewController {
...
         let personaListView = PersonaListView()
         personaListView.personaList = samplePersonas
+        personaListView.backgroundColor = Colors.NavigationBar.background
         personaListView.translatesAutoresizingMaskIntoConstraints = false
         return personaListView
     }()
 
+    private let headerView: UIView = {
+        let view = UIView()
+        view.backgroundColor = .systemGray6
```
harrieshin, line 66: please use fluent color as possible

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -55,10 +56,30 @@ class BottomSheetDemoController: UIViewController {
...
+        view.heightAnchor.constraint(equalToConstant: headerHeight).isActive = true
+
+        let label = UILabel()
```
harrieshin, line 69: please use fluent label if possible to handle larger text

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -55,10 +56,30 @@ class BottomSheetDemoController: UIViewController {
...
     }()
 
+    private let headerView: UIView = {
+        let view = UIView()
+        view.backgroundColor = Colors.gray100
+        view.heightAnchor.constraint(equalToConstant: headerHeight).isActive = true
+
+        let label = Label()
+        label.text = "Header view"
+        label.font = .systemFont(ofSize: 24)
```
harrieshin, line 71: please remove hardcoded font size because our demo app still goes through accessibility reviews. use one of the text ramp. i think label by default is .body

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -233,14 +226,30 @@ public class BottomSheetController: UIViewController {
...
+            targetAlpha = abs(currentOffset - collapsedOffsetFromBottom) / transitionLength
+        }
+        expandedContentView.alpha = targetAlpha
+    }
+
     @objc private func handlePan(_ sender: UIPanGestureRecognizer) {
         switch sender.state {
         case .began:
             stopAnimationIfNeeded()
+            hostedScrollView?.setContentOffset(.zero, animated: true)
```
harrieshin, line 247: doesn't it make more sense to set the content offest to zero at the end of the animation? or am I missing osomethign?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -44,7 +32,13 @@ public class BottomSheetController: UIViewController {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
-    /// A scroll view in `contentViewController`'s view hierarchy.
+    /// Top part of the sheet content that is visible in both collapsed and expanded state.
+    @objc public let headerContentView: UIView?
+
+    /// Sheet content below the header which is only visible when the sheet is expanded.
+    @objc public let expandedContentView: UIView
```
harrieshin, line 39: curious if the design will evolve to show little bit of preview because it seems like bottomsheet currently have discoverability issue.

harrieshin (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Yeah discoverability is something we're constantly trying to figure out with designers. It might be very case by case, maybe I could add a little "nudge" animation which reveals the content when the app opens, we'll see.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Yup that does make more sense, I changed it to only reset the scrollview after the sheet collapses.

PR #586, opened 2021-05-26
Avatar Vnext: Adding capability to set an image based custom ring color.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding the state property to enable a custom ring color that can be generated from an image.
This is the last remaining property needed to ensure the Avatar Vnext supports all the customization provided by the current AvatarView (UIKit).

Note: I'll update the demo controller (in a further PR) to use a Table View as we're getting more setting toggles which don't work well with Large Text accessibility modes.

### Verification

Added a demo app setting to control the property and exercised the scenarios to ensure the animations and visuals are correct:

| avatar_image_ring_light | avatar_image_ring_dark |

https://user-images.githubusercontent.com/68076145/119601268-aa52cb80-bd9d-11eb-9750-c00c3b4356a6.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/586)

### DIFF
--- a/ios/FluentUI/Vnext/Avatar/Avatar.swift
+++ b/ios/FluentUI/Vnext/Avatar/Avatar.swift
@@ -235,6 +235,10 @@ public struct AvatarView: View {
                 // transitions during the animation as the ImagePaint scale value is not animatable.
                 let ringMaxSize = avatarImageSize + (tokens.ringInnerGap + tokens.ringThickness) * 2
                 let scaleFactor = ringMaxSize / imageBasedRingColor.size.width
+
+                // ImagePaint is being used as creating a Color struct from a UIColor created with
+                // the patternImage initializer (https://developer.apple.com/documentation/uikit/uicolor/1621933-init)
+                // does not render any content.
                 Circle()
                     .strokeBorder(ImagePaint(image: Image(uiImage: imageBasedRingColor),
                                              scale: scaleFactor),
--- a/ios/FluentUI/Vnext/Avatar/Avatar.swift
+++ b/ios/FluentUI/Vnext/Avatar/Avatar.swift
@@ -96,6 +96,7 @@ import SwiftUI
     var primaryText: String? { get set }
     var secondaryText: String? { get set }
     var ringColor: UIColor? { get set }
+    var imageBasedRingColor: UIImage? { get set }
     var backgroundColor: UIColor? { get set }
     var foregroundColor: UIColor? { get set }
     var presence: MSFAvatarPresence { get set }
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -96,6 +96,7 @@ import SwiftUI
     var primaryText: String? { get set }
     var secondaryText: String? { get set }
     var ringColor: UIColor? { get set }
+    var imageBasedRingColor: UIImage? { get set }
```
harrieshin, line 99: do we really need the image based ring? I was wondering similar to Mac if UIColor (uiimage backed) is enough for avatar rainbow ring.

rdeassis: I tried using a UIColor creates with an image pattern (https://developer.apple.com/documentation/uikit/uicolor/1621933-init) but it does not seem to work with the SwiftUI Color. It just does not render anything.

I also found a report of the same thing in a stack overflow thread:
- https://stackoverflow.com/questions/60749545/swiftui-cant-create-color-by-uicolorpatternimage)

Alternatively we could create a gradient in SwiftUI, but that would make the rainbow ring a very specific property and potentially not match other controls that could be using an image to create the pattern.

harrieshin: oh wow. that's suprising to know. maybe we should add note why we are doing thing in comment then?

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -224,6 +226,24 @@ public struct AvatarView: View {
             }
         }
 
+        @ViewBuilder
+        var avatarRingView: some View {
```
harrieshin, line 230: i understand in avatarBody you check if the style is group but should there be a check or comment this variable is not available in group style?

harrieshin (approved):

ios/FluentUI/Vnext/Avatar/Avatar.swift
rdeassis: The job of the check is being done by the one you mentioned in the avatarbody computed variable. The avatarRingView is not computed unless it's being requested in the avatarBody.
I can definitely add a comment to make it explicit that the ring is not supported in the group style.

ios/FluentUI/Vnext/Avatar/Avatar.swift
Saadnajmi: Sounds like an Apple Bug?

rdeassis: I think it's unlikely that Apple will consider this as a bug since they provide the ImagePaint to accomplish that job.

PR #594, opened 2021-06-03
Reverting fix done in #573 due to regression.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change reverts the fix done in #573 as it caused a regression in client apps where the text labels were squeezed into a smaller width by the accessory view on larger screens (iPad).

Let's revert the regressing change for now and prepare a proper fix in a separate PR.

### Verification

Ran the demo app containing the reverted code to ensure the Table View Cell works correctly and confirming that the bug fixed by #573 reproduces again.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/594)


harrieshin:
can you write description that says which commit has you are reverting rather than the PR number, please?

harrieshin (approved):

harrieshin:
make sure when you squash, it also has the commit hash description for keeping the history.

PR #597, opened 2021-06-08
[vNext] Initial submission of PersonaButton

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Initial version of a vNext replacement to `ContactCollectionView`. To reduce confusion, we're consolidating on the term `Persona` for all cards representing a person, including these.

### Verification

This version of the control is more in-line with redlines, hence some minor discrepancies with the legacy `ContactCollectionView`.

| Legacy                                       | vNext                                      |
| Simulator Screen Shot - iPhone 12 mini - 2021-06-07 at 22 15 16 | Simulator Screen Shot - iPhone 12 mini - 2021-06-17 at 11 44 35 |
| Simulator Screen Shot - iPhone 12 mini - 2021-06-07 at 22 15 19 | Simulator Screen Shot - iPhone 12 mini - 2021-06-17 at 11 44 39 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/597)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -346,3 +346,17 @@ AP_MSFDrawerTokens:
...
+  avatarInterspace: [ small: $Spacing.xSmall, large: $Spacing.small ]
+  backgroundColor: $Colors.Background.neutral1
+  bottomPadding: $Spacing.xSmall
+  labelColor: $Colors.Foreground.neutral1
+  labelFont: [ small: $Typography.caption1, large: $Typography.subheadline ]
+  sublabelColor: $Colors.Foreground.neutral3
+  sublabelFont: $Typography.footnote
+  topPadding: $Spacing.xSmall
+
+AP_MSFPersonaGridTokens:
```
sophialee0416, line 360: Should GridItem extend from Grid?

DrTacoMD: Good question! I'm picturing it as a compositional model: the `Grid` is actually the full collection of faces, while each `GridItem` maps to an individual persona.

I'll see if I can think of any better names to make that relationship clearer, but I'm also open to suggestions!

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+
+    @objc public init(size: MSFPersonaGridSize = .large,
+                      theme: FluentUIStyle? = nil) {
+        super.init()
+
+        personaGridItem = PersonaGridItem(size: size)
+        hostingController = UIHostingController(rootView: AnyView(personaGridItem
+                                                                    .windowProvider(self)
+                                                                    .modifyIf(theme != nil, { personaGridItem in
+                                                                        personaGridItem.customTheme(theme!)
```
rdeassis, line 133: This will only work if you have the theme and windowProvider environment values that I mentioned on the other comment.

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+
+        super.init()
+    }
+}
+
+public struct PersonaGridItem: View {
+    @ObservedObject var tokens: MSFPersonaGridItemTokens
+    @ObservedObject var state: MSFPersonaGridItemViewStateImpl
+    let avatarView: AvatarView
+    let gridSize: MSFPersonaGridSize
```
rdeassis, line 80: This should be part of the state object as the client code could set a new size and have the view animating its size change.

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItemTokens.swift
```diff
@@ -0,0 +1,86 @@
...
+        switch self {
+        case .large:
+            return true
+        case .small:
+            return false
+        }
+    }
+}
+
+/// Representation of design tokens to buttons at runtime which interfaces with the Design Token System auto-generated code.
```
rdeassis, line 32: nit: "buttons" on the documentation

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItemTokens.swift
```diff
@@ -0,0 +1,86 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Pre-defined sizes of the avatar
```
rdeassis, line 8: nit: documentation mentioning "avatar"

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+        gridSize = size
+    }
+
+    public var body: some View {
+        VStack(spacing: 0) {
+            avatarView
+                .padding(.top, tokens.topPadding)
+                .padding(.bottom, tokens.avatarInterspace)
+            Text(state.primaryText ?? "")
+                .font(Font(tokens.labelFont))
```
rdeassis, line 96: Is this supposed to support Large Text accessibility settings? (I believe so)
We have to use our scalableFont modifier for that. (The Button Vnext uses it on styles other than the Floating Action Button).

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+}
+
+public struct PersonaGridItem: View {
+    @ObservedObject var tokens: MSFPersonaGridItemTokens
+    @ObservedObject var state: MSFPersonaGridItemViewStateImpl
+    let avatarView: AvatarView
+    let gridSize: MSFPersonaGridSize
+
+    public init(size: MSFPersonaGridSize) {
+        tokens = MSFPersonaGridItemTokens(size: size)
```
rdeassis, line 83: Why aren't we using the tokens instance from the state object?

ios/FluentUI/Vnext/PersonaGrid/PersonaGridTokens.swift
```diff
@@ -0,0 +1,18 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+/// Representation of design tokens to buttons at runtime which interfaces with the Design Token System auto-generated code.
```
rdeassis, line 8: nit: documentation mentions "buttons"

ios/FluentUI/Vnext/PersonaGrid/PersonaGridTokens.swift
```diff
@@ -0,0 +1,18 @@
...
+/// Updating these properties causes the SwiftUI button to update its view automatically.
+class MSFPersonaGridTokens: MSFTokensBase, ObservableObject {
+    @objc open func didChangeAppearanceProxy() {
+        updateForCurrentTheme()
+    }
+
+    override func updateForCurrentTheme() {
+        // TODO: 5171471 - Create PersonaGrid
```
rdeassis, line 16: ADO reference number. (same as the other comment)

tools/sgen/input/FluentUIStyle.yml
rdeassis: I think it makes sense to keep them as separate APs (like the AP_MSFListTokens and the AP_MSFListCellTokens).
The Grid can't exist without the GridItem, but the GridItem can be used as a standalone control.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -346,3 +346,17 @@ AP_MSFDrawerTokens:
   cornerRadius: $Border.radius.xxLarge
   verticalShadowOffset: $Spacing.xSmall
   horizontalShadowOffset: $Spacing.small
+
+AP_MSFPersonaGridItemTokens:
```
rdeassis, line 350: Talking about the naming discussion: 

1. The term "Grid" makes me think of a matrix instead of a horizontal or vertical list of items. I think I like "Collection" or "Carousel" as you suggested before.

2. I was thinking about another name, so here's the suggestion: How about "Persona Tag" or "Persona Badge"? Not sure I like it a lot, but just throwing the idea out there.

Maybe "PersonaBadge" and "PersonaBadgeCollection" for the names?
Let me know your thoughts.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -346,3 +346,17 @@ AP_MSFDrawerTokens:
...
+  avatarInterspace: [ small: $Spacing.xSmall, large: $Spacing.small ]
+  backgroundColor: $Colors.Background.neutral1
+  bottomPadding: $Spacing.xSmall
```
rdeassis, line 353: Should we have padding on the sides here?
I'm mostly thinking of the scenario where the single item is used as a standalone control.

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -903,6 +922,19 @@
 			path = "Bottom Commanding";
 			sourceTree = "<group>";
 		};
+		922AF01926615B03005DB168 /* PersonaGrid */ = {
```
rdeassis, line 925: Can you also create the subspecs on the podspec file for these 2 controls?

ios/FluentUI/Vnext/PersonaGrid/PersonaGrid.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+struct PersonaGrid: View {
+    var body: some View {
+        // TODO: 5171471 - Create PersonaGrid
```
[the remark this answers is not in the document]
mischreiber: Force of habit :)

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// `MSFPersonaGridItemState` contains PersonaGridItem properties in addition to a subset of the MSFAvatarState protocol.
+///
+/// `onTapAction` provides tap gesture for PersonaGridItem.
+///
+@objc public protocol MSFPersonaGridItemState {
```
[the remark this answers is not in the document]
rdeassis: I think that it is more value added right from the beginning if we allow the client code to control everything related to the ring and presence. I believe the only one worth not exposing is the size of the avatar as it's bound to the 2 size options of the control itself.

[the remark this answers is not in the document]
mischreiber: Seems reasonable! That said, the easiest way to get all those properties will be to do what `PersonaView` does and have `MSFPersonaGridItemState` extend from `MSFAvatarState`, so `size` will be exposed regardless.

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+        get {
+            return avatarState.secondaryText
+        }
+
+        set {
+            avatarState.secondaryText = newValue
+        }
+    }
+
+    var size: MSFAvatarSize {
```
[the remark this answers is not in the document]
mischreiber: It will trigger a re-layout, but it won't currently trigger the text appearing or disappearing. I'd really rather hide all this stuff away from the public interface, but I can only think of three options:
1. We implement the full `MSFAvatarState` protocol, per https://github.com/microsoft/fluentui-apple/pull/597#discussion_r648612510 (and thus make this property public)
2. We manually specify which properties from `Avatar` we want to make public, in which case we risk having API drift if new props are added to `Avatar`.
3. We split the `MSFAvatarState` protocol into two: one for styling and one for... sizing, I guess?

Option 3 is cleanest for `Persona`, but feels kinda arbitrary. The other two have their own drawbacks. What are your thoughts?

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+    init(size: MSFPersonaGridSize, avatarState: MSFAvatarState) {
+        self.avatarState = avatarState
+        self.tokens = MSFPersonaGridItemTokens(size: size)
+
+        super.init()
+    }
+}
+
+public struct PersonaGridItem: View {
+    @ObservedObject var tokens: MSFPersonaGridItemTokens
```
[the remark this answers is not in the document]
mischreiber: Got it! 👍

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+            if gridSize.canShowSubtitle {
+                Text(state.secondaryText ?? "")
+                    .font(Font(tokens.sublabelFont))
+                    .foregroundColor(Color(tokens.sublabelColor))
+            }
+            Spacer(minLength: tokens.bottomPadding)
+        }
+        .background(Color(tokens.appearanceProxy.backgroundColor))
+        .onTapGesture {
+            if let action = state.onTapAction {
```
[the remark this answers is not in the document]
mischreiber: Normally yeah, and I can switch it here. Just felt a little odd inside a closure in SwiftUI:
```
.onTapGesture {
    guard let action = state.onTapAction else {
        return
    }
    action()
}
```

But it works, so no argument here.

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+                .foregroundColor(Color(tokens.labelColor))
+            if gridSize.canShowSubtitle {
+                Text(state.secondaryText ?? "")
+                    .font(Font(tokens.sublabelFont))
+                    .foregroundColor(Color(tokens.sublabelColor))
+            }
+            Spacer(minLength: tokens.bottomPadding)
+        }
+        .background(Color(tokens.appearanceProxy.backgroundColor))
+        .onTapGesture {
```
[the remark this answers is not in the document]
mischreiber: Love it!

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
mischreiber: Agreed. I'd originally thought of `gridSize` being immutable at creation, but there's no reason it should have to be.

tools/sgen/input/FluentUIStyle.yml
mischreiber: Chewed on it a bit and yeah, I like those. I think Carousel is a bit too specific of a control, which this technically isn't.

tools/sgen/input/FluentUIStyle.yml
mischreiber: Looks nicer too, and will keep the tap zones from having no-tap "gutters" once they're a collection.

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
mischreiber: Because I made an oopsie 🥉

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
```diff
@@ -0,0 +1,146 @@
...
+
+public struct PersonaGridItem: View {
+    @ObservedObject var tokens: MSFPersonaGridItemTokens
+    @ObservedObject var state: MSFPersonaGridItemViewStateImpl
+    let avatarView: AvatarView
+    let gridSize: MSFPersonaGridSize
+
+    public init(size: MSFPersonaGridSize) {
+        tokens = MSFPersonaGridItemTokens(size: size)
+        avatarView = AvatarView(style: .default, size: size.avatarSize)
```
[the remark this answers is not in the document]
mischreiber: Yeah you're right that there's a collision here between the intended architecture and the real world.

If it's okay with you I'd like to move this train of thought to the next PR covering `PersonaBadgeCollection`, where it'll be easier to rationalize pros and cons of moving the controls around.

Plus it gives me a little breathing room to try and think of a best-of-both-worlds solution 😎

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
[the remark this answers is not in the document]
rdeassis: I'm leaning towards defining all the properties that we want to expose in the MSFPersonaBadgeState and forwarding what we need to forward in the implementation to the avatar state.

One of the reasons for that is when we need for example to have the "Signed Out" state of the individual use of the persona badge:
The calling code may want to display "Sign In" in the main label, and the Avatar can't display the "S" initials. It will instead need to be set to a specific avatar style and both primary and secondary text properties need to be set to an empty value. This would be more a Persona Badge implementation detail that we can do in its state object instead of just exposing all the Avatar state which would make it harder for the calling code to implement this scenario on its own.

We don't have to implement this functionality just yet, but I think the control will be ready for that enhancement if we don't just expose the AvatarState directly.

The other benefit is the one we already discussed about the size. We can just choose not to expose the Avatar size and control it internally when the PersonaBadge size is set.

What do you think?

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaBadgeCollectionDemoController.swift
```diff
@@ -0,0 +1,138 @@
...
+        case .large:
+            cell = tableView.dequeueReusableCell(withIdentifier: PersonaBadgeCollectionDemoController.largeBadgeReuseIdentifier, for: indexPath)
+        case .small:
+            cell = tableView.dequeueReusableCell(withIdentifier: PersonaBadgeCollectionDemoController.smallBadgeReuseIdentifier, for: indexPath)
+        }
+
+        // Create the PersonaBadge to be displayed
+        let personaBadge = MSFPersonaBadgeView(size: size)
+        let persona = personas[indexPath.item]
+        personaBadge.state.image = persona.avatarImage
```
rdeassis, line 80: Are you reproducing this on iOS13?

Screen Shot 2021-06-11 at 10 14 42 AM

ios/FluentUI/Vnext/PersonaBadge/PersonaBadge.swift
```diff
@@ -0,0 +1,261 @@
...
+
+public struct PersonaBadge: View {
+    @Environment(\.theme) var theme: FluentUIStyle
+    @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+    @ObservedObject var tokens: MSFPersonaBadgeTokens
+    @ObservedObject var state: MSFPersonaBadgeViewStateImpl
+    let avatarView: AvatarView
+
+    public init(size: MSFPersonaBadgeSize) {
+        self.avatarView = AvatarView(style: .default, size: size.avatarSize)
```
rdeassis, line 195: Following up on the discussion of having the Views in the State object or not (GitHub won't allow me to respond on the outdated thread :( ):

We can indeed discuss more about the pros and cons of doing that. And I think by the time the Collection is implemented, it would be ideal to have that pattern figured out as the collection will be dynamic and animations will have to follow on the state change of the collection of badges. 👍

rdeassis (commented):
This is looking great! :)
Could you also do the theming test by locally replacing the tokens like below and see if they reflect the changes accordingly?

AP_MSFPersonaBadgeTokens:
...
  **backgroundColor: $Colors.Brand.primary**

For reference you can check #382 for how to test it in the demo app.

ios/FluentUI/Vnext/PersonaBadge/PersonaBadge.swift
```diff
@@ -0,0 +1,261 @@
...
+
+    var onTapAction: (() -> Void)? { get set }
+}
+
+/// Properties that make up PersonaBadge content
+class MSFPersonaBadgeViewStateImpl: NSObject, ObservableObject, Identifiable, MSFPersonaBadgeState {
+    let tokens: MSFPersonaBadgeTokens
+
+    public var onTapAction: (() -> Void)?
+    private var avatarState: MSFAvatarState
```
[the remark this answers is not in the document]
mischreiber: I wonder if this is something that can be linted for...?

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaBadgeCollectionDemoController.swift
mischreiber: I do 😿  Very strange.

I'll start looking at it, but if this is something we've seen before I'd be happy for any advice 🙇

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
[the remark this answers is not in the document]
mischreiber: Makes perfect sense to me! That's actually what I had in my first iteration; it was just a much smaller subset of properties. I'll scrub through and see what makes sense in this new more-expanded world.

ios/FluentUI/Vnext/PersonaBadge/PersonaBadge.swift
mischreiber: Sounds good!

That said, my ideal would be something like the following:
- The `Group` State tracks a list of View States, each of which is somehow uniquely identifiable
- At the View level of the Group, we declare PersonaBadge views for each View State. SwiftUI doesn't recreate Views unnecessarily because it doesn't need to: it can tell which View maps to which View State and does the right thing

That's dependent on being able to create View States independently from Views, and I don't believe that's something we've built yet. But it feels cleanest of all options.

Let's talk more offline - I'll draft up a simple proposal 🤓

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaBadgeCollectionDemoController.swift
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    // MARK: - Helpers
+
+    private func personaBadgeSize(for section: Int) -> MSFPersonaBadgeSize {
+        return (section % 2 == 0) ? .small : .large
+    }
+
+    private func showsFullGrid(at indexPath: IndexPath) -> Bool {
+        // TODO
```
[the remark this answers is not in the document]
mischreiber: Actually @harrieshin pointed out that the Badge name is no good either, since that has a specific meaning for Android. So we'll need something else new still ☹️

ios/FluentUI/Vnext/PersonaGrid/PersonaGridItem.swift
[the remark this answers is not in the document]
mischreiber: As discussed in a different thread, we'll scope it a bit more tightly than the full `MSFAvatarState` protocol.

ios/FluentUI/Vnext/PersonaBadge/PersonaBadge.swift
```diff
@@ -0,0 +1,261 @@
...
+                    .foregroundColor(Color(tokens.labelColor))
+                if state.badgeSize.canShowSubtitle {
+                    Text(state.secondaryText ?? "")
+                        .scalableFont(font: tokens.sublabelFont)
+                        .foregroundColor(Color(tokens.sublabelColor))
+                }
+                Spacer(minLength: tokens.padding)
+            }
+        }
+        .background(Color(tokens.appearanceProxy.backgroundColor))
```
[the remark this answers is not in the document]
mischreiber: Good catch, that's a leftover from an older iteration of the control

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -346,3 +346,16 @@ AP_MSFDrawerTokens:
   cornerRadius: $Border.radius.xxLarge
   verticalShadowOffset: $Spacing.xSmall
   horizontalShadowOffset: $Spacing.small
+
+AP_MSFPersonaBadgeTokens:
```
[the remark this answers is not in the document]
mischreiber: Sure I can clean it up. There's no real alphabetical ordering in the `AP_*` classes, so I'll fix all that

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaBadgeCollectionDemoController.swift
mischreiber: Yep, it was a failure of rendering mode. The `Image` inside `AvatarView` was defaulting to the `template` rendering mode when I embedded it in a `Button`, but only on iOS 13. Adding an explicit `.renderingMode(.original)` resolves the issue.

Thanks again for the catch!

MicrosoftFluentUIVnext.podspec
```diff
@@ -88,6 +88,18 @@ xcodebuild ${XCODEBUILDPARAMS} -project ${PROJECT_FILE_PATH} -target "MicrosoftF
     personaView_ios.source_files = ["ios/FluentUI/Vnext/Persona/**/*.{swift,h}"]
   end
 
+  s.subspec 'PersonaBadge_ios' do |personaBadge_ios|
```
harrieshin, line 91: update the naming?

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
```diff
@@ -0,0 +1,138 @@
...
+        personaBadge.state.secondaryText = persona.secondaryText
+        personaBadge.state.onTapAction = { [weak self, personaBadge] in
+            let alert = UIAlertController(title: nil, message: "PersonaButton tapped: \(personaBadge.state.primaryText ?? "(none)")", preferredStyle: .alert)
+            alert.addAction(UIAlertAction(title: "OK", style: .default))
+            self?.present(alert, animated: true)
+        }
+
+        cell.contentView.addSubview(personaBadge.view)
+        cell.selectionStyle = .none
+        cell.backgroundColor = .clear
```
rdeassis, line 91: Minor detail: can we set the background color with a different color so we can see the padding and the background of the PersonaButton?

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -0,0 +1,273 @@
...
+
+    let tokens: MSFPersonaButtonTokens
+
+    var backgroundColor: UIColor? {
+        get {
+            return avatarState.backgroundColor
+        }
+        set {
+            avatarState.backgroundColor = newValue
+            objectWillChange.send()
```
rdeassis, line 44: I think you don't need to call it here. The avatarState implementation does it as a @Published property in an ObservableObject.

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
```diff
@@ -0,0 +1,138 @@
...
+            fatalError("showsFullCarousel() should always return false right now!")
+        } else {
+            cell = self.tableView(tableView, badgeCellForRowAt: indexPath)
+        }
+
+        return cell
+    }
+
+    override func tableView(_ tableView: UITableView, titleForHeaderInSection section: Int) -> String? {
+        switch personaBadgeSize(for: section) {
```
rdeassis, line 48: nit: update naming "personaBadge"

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
```diff
@@ -0,0 +1,138 @@
...
+        let size = personaBadgeSize(for: indexPath.section)
+        switch size {
+        case .large:
+            cell = tableView.dequeueReusableCell(withIdentifier: PersonaButtonCarouselDemoController.largeBadgeReuseIdentifier, for: indexPath)
+        case .small:
+            cell = tableView.dequeueReusableCell(withIdentifier: PersonaButtonCarouselDemoController.smallBadgeReuseIdentifier, for: indexPath)
+        }
+
+        // Create the PersonaButton to be displayed
+        let personaBadge = MSFPersonaButtonView(size: size)
```
rdeassis, line 78: nit: update naming "personaBadge"

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -0,0 +1,273 @@
...
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// `MSFPersonaButtonState` contains PersonaButton properties in addition to a subset of the MSFAvatarState protocol.
+///
+/// `onTapAction` provides tap gesture for PersonaButton.
+///
+@objc public protocol MSFPersonaButtonState: MSFAvatarState {
```
rdeassis, line 12: I believe the intent of implementing the properties below is to not inherit the MSFAvatarState?

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -0,0 +1,273 @@
...
+        let state = MSFPersonaButtonViewStateImpl(size: size, avatarState: avatarView.state)
+        self.state = state
+        self.tokens = state.tokens
+    }
+
+    public var body: some View {
+        let action = state.onTapAction ?? {}
+        Button(action: action) {
+            VStack(spacing: 0) {
+                avatarView
```
rdeassis, line 218: I think we don't need this anymore since you're adding the internal initializer for the Avatar.
I think we can create the avatar state inside the MSFPersonaButtonViewStateImpl, remove the avatar state param from the initializer and just pass the avatar state created here, right?

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -0,0 +1,273 @@
...
+
+    @objc open var state: MSFPersonaButtonState {
+        return self.personaBadge.state
+    }
+
+    @objc public init(size: MSFPersonaButtonSize = .large,
+                      theme: FluentUIStyle? = nil) {
+        super.init()
+
+        personaBadge = PersonaButton(size: size)
```
rdeassis, line 256: nit: naming "Badge"

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -0,0 +1,273 @@
...
+
+/// UIKit wrapper that exposes the SwiftUI PersonaButton implementation
+@objc open class MSFPersonaButtonView: NSObject, FluentUIWindowProvider {
+
+    @objc open var view: UIView {
+        return hostingController.view
+    }
+
+    @objc open var state: MSFPersonaButtonState {
+        return self.personaBadge.state
```
rdeassis, line 249: nit: naming "Badge"

ios/FluentUI/Vnext/PersonaButton/PersonaButtonTokens.swift
```diff
@@ -0,0 +1,84 @@
...
+    var avatarSize: MSFAvatarSize {
+        switch self {
+        case .large:
+            return .xxlarge
+        case .small:
+            return .xlarge
+        }
+    }
+
+    var canShowSubtitle: Bool {
```
rdeassis, line 22: nit: shouldShowSubtitle instead?

ios/FluentUI/Vnext/PersonaButton/PersonaButtonTokens.swift
```diff
@@ -0,0 +1,84 @@
...
+        switch self {
+        case .large:
+            return true
+        case .small:
+            return false
+        }
+    }
+}
+
+/// Representation of design tokens to persona badges at runtime which interfaces with the Design Token System auto-generated code.
```
rdeassis, line 32: nit: "badges" in the comment

ios/FluentUI/Vnext/PersonaButton/PersonaButtonTokens.swift
```diff
@@ -0,0 +1,84 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Pre-defined sizes of the persona badge
```
rdeassis, line 8: nit: "persona badge"

rdeassis (commented):
Two other topics that I wanted to bring up:
1. **The theme overriding testing:** it's good to have a local test overriding the background color of the PersonaButton and switching themes in the Theming Demo controller and ensure the color is being reflected correctly when the theme is changed.

2. **For the next PR of the Carousel control:** ability to test animations when properties are changed in the controls is something nice to have in the demo controller. Just like we have the toggle settings for other controls, it'd be a good way to test how the control is animating state changes. The PersonaButton itself can be validated on the size change. (maybe a settings item with the 2 sizes instead of 2 sections?)

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: Ah what a mess this branch has become. Sorry, that's correct.

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
mischreiber: It should be visible already? The clear background passes through the light gray color of the table view.

Ah unless you're mostly using dark mode, where they're both black. Okay, I'll give it a little *pop*

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: Yep! This just didn't have that PR yet, but I'll merge it over

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: That works for people observing `avatarState` directly, but this is necessary if you're expecting to get notifications against the wrapped property on `MSFPersonaButtonViewStateImpl`.

That said, some of these property names are confusing since they don't map directly to what the button does, so I'll rename them.

mischreiber:
I think I've responded to all feedback, so I'm going to merge this into the vnext branch for now, but I'll gladly respond to any additional comments as needed!

rdeassis:
> This is looking great! :)
> Could you also do the theming test by locally replacing the tokens like below and see if they reflect the changes accordingly?
> 
> AP_MSFPersonaBadgeTokens:
> ...
> **backgroundColor: $Colors.Brand.primary**
> 
> For reference you can check #382 for how to test it in the demo app.

@mischreiber were you able to validate the theme override scenario? Does it work as expected (respecting the overriden theme tokens when you switch them in the Theming demo controller)?

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
rdeassis: I think SwiftUI will track that if the avatar View changes and is contained by this control.

The useful application of the objectWillChange.send() call I've seen so far is to force SwiftUI refreshes if you either don't have a @published property or you're implementing a custom property that will be observed.

These cases are just a passthrough to the @published properties of the avatar, which is what makes the objectWitllChange call not needed.

mischreiber:
> > This is looking great! :)
> > Could you also do the theming test by locally replacing the tokens like below and see if they reflect the changes accordingly?
> > AP_MSFPersonaBadgeTokens:
> > ...
> > **backgroundColor: $Colors.Brand.primary**
> > For reference you can check #382 for how to test it in the demo app.
> 
> @mischreiber were you able to validate the theme override scenario? Does it work as expected (respecting the overriden theme tokens when you switch them in the Theming demo controller)?

Yes sorry didn't reply to the original comment, but I was able to verify that changing the theme did change tokens. 

| Regular | Changed |
|-----|-----|
| Simulator Screen Shot - iPhone 12 mini - 2021-06-21 at 11 26 10 | Simulator Screen Shot - iPhone 12 mini - 2021-06-21 at 11 26 04 |

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: Well that is just the darndest thing. I would have put serious money on those calls being necessary at some point during development, but you're right: they don't seem to do anything now. I'll get them removed as part of the next big PR for the carousel.

PR #599, opened 2021-06-09
Updating search paths for running swiftlint on Apple Silicon machines

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

By default, Xcode can't find any packages installed by `homebrew` on Apple Silicon Macs, since they're now installed to a different location that isn't on the default `$PATH` (`/opt/homebrew/bin/`, rather than `/usr/local/bin`).

Workaround: explicitly add this location to `$PATH` before attempting to run `swiftlint`.

### Verification

Verified that `swiftlint` is now found on both Apple Silicon and Intel Macs, for both iOS and macOS projects.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/599)

### DIFF
--- a/ios/FluentUI.xcodeproj/project.pbxproj
+++ b/ios/FluentUI.xcodeproj/project.pbxproj
@@ -1385,7 +1385,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n    export LINTPATH=\"${LOCROOT}/./\"\n    swiftlint --config ../.swiftlint.yml\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
+			shellScript = "# Adds support for Apple Silicon brew directory\nexport PATH=\"$PATH:/opt/homebrew/bin\"\n\nif which swiftlint >/dev/null; then\n    export LINTPATH=\"${LOCROOT}/./\"\n    swiftlint --config ../.swiftlint.yml\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
 		};
 /* End PBXShellScriptBuildPhase section */
 
--- a/macos/xcode/FluentUI.xcodeproj/project.pbxproj
+++ b/macos/xcode/FluentUI.xcodeproj/project.pbxproj
@@ -803,7 +803,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n    export LINTPATH=\"${LOCROOT}/../\"\n    swiftlint --config ../../.swiftlint.yml\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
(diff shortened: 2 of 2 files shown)


ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -1385,7 +1385,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "if which swiftlint >/dev/null; then\n    export LINTPATH=\"${LOCROOT}/./\"\n    swiftlint --config ../.swiftlint.yml\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
+			shellScript = "# Adds support for Apple Silicon brew directory\nexport PATH=\"$PATH:/opt/homebrew/bin\"\n\nif which swiftlint >/dev/null; then\n    export LINTPATH=\"${LOCROOT}/./\"\n    swiftlint --config ../.swiftlint.yml\nelse\n    echo \"warning: SwiftLint not installed, download from https://github.com/realm/SwiftLint\"\nfi\n";
```
rdeassis, line 1388: Do you know what happens if SwiftLint is installed in any way other than through HomeBrew?

Is there a way to customize the installation path of HomeBrew in M1 machines so we don't need to change the project?

kubalani (approved):

ios/FluentUI.xcodeproj/project.pbxproj
DrTacoMD: There's a lot of conversation as to why HomeBrew decided to switch to `/opt/homebrew` here: https://github.com/Homebrew/brew/issues/9177. It's entirely possible to install to a new location, but it involves pulling and building the code manually, which feels more fragile to me. Plus we'll be permanently nonstandard, which feels *much* more fragile to me.

The diff here is the currently proposed fix, though it hasn't been accepted into the docs yet: https://github.com/realm/SwiftLint/pull/3527.

This shouldn't affect any other installs of SwiftLint, as long as they're in the default `$PATH`. So I'd rather we not worry about supporting every possible dev environment, with `brew` as an exception since it's the most popular macOS package manager by far.

rdeassis: Thanks for all the pointers!
I do see that there's no harm in just adding another folder to the path.
I guess my point is that maybe we could do something else locally for the M1 machines (that could be added to the readme instructions) instead of editing the projects for a specific scenario of swiftlint being installed with HomeBrew.

markavitale (approved):

ios/FluentUI.xcodeproj/project.pbxproj
mischreiber: I'm not sure that much can be done, other than getting pretty invasive into directories we shouldn't be touching. Sadly, Xcode won't pick up `PATH` overrides in `.zshrc`, `.bash_profile`, etc.

That said, I'm going to keep an eye on both SwiftLint and HomeBrew to see if their recommendations end up changing going forward, and if so then I'll update these and/or the README to reflect the newest guidance.

rdeassis: Sounds good to me. Thanks!

PR #603, opened 2021-06-11
[vNext] AvatarGroup Control

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Introducing AvatarGroup control, following feedback from #582 . Changes include token renaming and redesigning control logic.

### Verification

| Before (Legacy)                                     | After (vNext)                                      |
| Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 14 05 | Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 11 40 |
| Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 13 14 | Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 12 00 |
| Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 13 21 | Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 12 19 |
|  | Simulator Screen Shot - iPhone 11 - 2021-07-06 at 10 12 35 |

Theming verification:

https://user-images.githubusercontent.com/22566866/125138542-ae277c80-e0c3-11eb-89b9-af3bf2999500.mov

### Pull request checklist

This PR has considered:
- [x] Podspec lint check

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/603)


ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,279 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+// MARK: - AvatarGroupDemoController
+
+class AvatarGroupDemoController: DemoController {
```
rdeassis, line 11: Do you intend to change this to a table view in this PR?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+
+@objc public protocol MSFAvatarGroupState {
+    var maxDisplayedAvatars: UInt32 { get set }
```
rdeassis, line 15: Any particular reason you decided to use UInt32 for maxDisplayedAvatars and UInt for overflowCount?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+    var overflowCount: UInt { get set }
```
rdeassis, line 16: The overflowCount should be calculated automatically based on the number of avatars that the group is holding.

<avatars_in_group> - maxDisplayedAvatars = overflow_count (if greater than zero)

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+    func deleteAvatar(atIndex: Int)
+}
+
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) {
+        avatars.append(AvatarView(style: style, size: size))
+    }
+
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState {
```
rdeassis, line 30: What if the index passed is out of the bounds of the array?
We should validate this instead of letting it crash.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+            tokens.style = newValue
+        }
+    }
+
+    var tokens: MSFAvatarGroupTokens
+
+    init(style: MSFAvatarGroupStyle,
+         size: MSFAvatarSize) {
+        self.tokens = MSFAvatarGroupTokens(style: style,
+                                      size: size)
```
rdeassis, line 65: nit: indentation

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+
+@objc public protocol MSFAvatarGroupState {
```
rdeassis, line 14: For the public protocol it'd be nice to have documentation (Xcode style with /// ).

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+
+    public var body: some View {
+        let size: CGFloat = tokens.size.size
+        let x: CGFloat = size + tokens.interspace - tokens.ringOuterGap - tokens.ringThickness
+        let avatars: [AvatarView] = state.avatars
+        let maxDisplayedAvatars: Int = avatars.prefix(Int(state.maxDisplayedAvatars)).count
+        let overflowCount: Int = Int((avatars.count > maxDisplayedAvatars ? UInt(avatars.count - maxDisplayedAvatars) : 0) + state.overflowCount)
+        HStack(spacing: 0) {
+            ForEach(0 ..< maxDisplayedAvatars, id: \.self) { index in
+                let modify = tokens.style == .stack && (overflowCount > 0 || index + 1 < maxDisplayedAvatars)
```
rdeassis, line 91: Can we change the constant name to something more meaningful than "modify"?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+    }
+
+    /// Cutout shape for the succeeding Avatar in an Avatar Group in Stack style.
+    ///
+    /// xOrigin: beginning location of cutout on the x axis
+    ///
+    /// yOrigin: beginning location of cutout on the y axis
+    ///
+    /// cutoutSize: dimensions of cutout shape of the Avatar
+    private struct AvatarCutout: Shape {
```
rdeassis, line 123: If I remember correctly, I asked you this on the previous PR:
In the scenario where the avatars have presence, does the cutout reflect that shape? (I think the same could be applicable in case the avatar is set for no transparency.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// UIKit wrapper that exposes the SwiftUI AvatarGroup implementation
+@objc open class MSFAvatarGroup: NSObject, FluentUIWindowProvider {
+    @objc public init(style: MSFAvatarGroupStyle = .stack, size: MSFAvatarSize = .large, theme: FluentUIStyle? = nil) {
+        super.init()
+
+        avatarGroupView = AvatarGroup(style: style, size: size)
+        hostingController = UIHostingController(rootView: AnyView(avatarGroupView
+                                                                    .windowProvider(self)
```
rdeassis, line 146: Have you tested the theme override per window?
Like... setting any token based on the theme override (theming demo controller) and see if it reflects on the control here

ios/FluentUI/Vnext/Avatar/AvatarGroupTokens.swift
```diff
@@ -0,0 +1,91 @@
...
+
+        switch size {
+        case .xsmall:
+            interspace = appearanceProxy.interspace.xSmall
+            ringInnerGap = appearanceProxy.ringInnerGap.xSmall
+            ringThickness = appearanceProxy.ringThickness.xSmall
+            ringOuterGap = appearanceProxy.ringOuterGap.xSmall
+        case .small:
+            interspace = appearanceProxy.interspace.small
+            ringInnerGap = appearanceProxy.ringInnerGap.xSmall
```
rdeassis, line 66: Shouldn't these be the small property instead of the xSmall?
(here and below for the other sizes)

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -230,6 +230,15 @@ MSFOverflowAvatarTokens extends MSFAvatarTokens:
...
   ringDefaultColor: $Colors.Background.neutralDisabled
 
+AP_MSFAvatarGroupTokens:
+  interspace: [ xSmall: $Spacing.xxSmall, small: $Spacing.xxSmall, medium: $Spacing.xSmall, large: $Spacing.xSmall, xlarge: $Spacing.xSmall, xxlarge: $Spacing.xSmall ]
+  ringInnerGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  ringThickness: [ xSmall: $Border.size.thin, small: $Border.size.thin, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  ringOuterGap: [ xSmall: $Border.size.thick, small: $Border.size.thick, medium: $Border.size.thick, large: $Border.size.thick, xlarge: $Border.size.thick, xxlarge: $Border.size.thicker ]
+  
+MSFAvatarGroupStackTokens extends MSFAvatarGroupTokens:
+  interspace: [ xSmall: -2pt, small: -2pt, medium: -4pt, large: -8pt, xlarge: -12pt, xxlarge: -12pt ]
```
rdeassis, line 240: Maybe a conversation with the design team:
Should we create some sort of named spacing tokens like "$Spacing.xSmallOverlap"?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
sophialee0416: Yes, I realized I did not address that so it is being fixed

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
sophialee0416: I wanted to maintain UInt for both but could not because when I iterate through the avatars using ForEach, the max Uint value goes out of range when cast to Int. Uint32 was a way to address that -- I could also make overflowCount UInt32?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
sophialee0416: According to legacy:

> Calculation for the overflow count: max(avatars.count() - maxDisplayedAvatars, 0) + overflowCount.

Should it still be calculated here?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
sophialee0416: The cutout is applied to the presence as well. Does this look ok?
image

harrieshin: doesn't make sense that avatar group has presence badge. the whole point of collaboration corner is to show the colorful ring based on activity.

rdeassis: I think that supporting presence on the Avatar Pile style (where the Avatars don't overlap) will be nice. Not necessarily for the coauthoring scenario though.

But still it's worth testing the avatars with rings, setting with and without transparency for the Avatar Stack style.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
markavitale: It's worth noting that the swift language guide recommends using Int anywhere possible, even for values known to be nonnegative

> NOTE
>
> Use UInt only when you specifically need an unsigned integer type with the same size as the platform’s native word size. If this isn’t the case, Int is preferred, even when the values to be stored are known to be nonnegative. A consistent use of Int for integer values aids code interoperability, avoids the need to convert between different number types, and matches integer type inference, as described in Type Safety and Type Inference.

https://docs.swift.org/swift-book/LanguageGuide/TheBasics.html#ID317

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
mischreiber: Just wondering, this will still crash, right?

Personally I'm cool with crashing for index out of bounds stuff like this (NSArray behaves similarly by throwing an unlikely-to-be-caught NSException). Not really sure what other behavior we would want anyway.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,190 @@
...
+    var style: MSFAvatarGroupStyle { get set }
+
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize)
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState
+    func deleteAvatar(atIndex: Int)
+}
+
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) {
```
mischreiber, line 39: Question about design for these. I would have assumed that these `State` objects were meant to be "dumb" property bags, without any logic. This isn't *much* logic, but it does break that assumption.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,190 @@
...
+            return avatars[atIndex].state
+        }
+    }
+
+    func deleteAvatar(atIndex: Int) {
+        avatars.remove(at: atIndex)
+    }
+
+    @Published var avatars: [AvatarView] = []
+    @Published var maxDisplayedAvatars: Int = Int(UInt32.max)
```
mischreiber, line 56: This line worries me. What does it mean to turn `UInt32.max` into an Int? We're lucky in that `Int` is a 64-bit type, so `UInt32.max` fits, but that's not really guaranteed or documented.

If we're using an `Int` type, then `Int.max` should probably be used here.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
sophialee0416: This new design change was made to address this [comment](https://github.com/microsoft/fluentui-apple/pull/582#discussion_r645686438). This is so that the consumer does not have the responsibility to create the avatar state as this class will handle it instead. @rdeassis

mischreiber: Makes sense. I agree with the earlier comments that clients shouldn't have to create views, but I'm not sure if this is the approach we should be taking. I'm actually running into similar questions regarding `PersonaCarousel` right now: what should the public interface be for creating these?

@rdeassis and I discussed very briefly offline yesterday, but we kinda settled on a pattern like the following:
1. Caller instantiates the collection (e.g. `AvatarGroup`)
2. Caller instantiates one or more data model structs/objects that represents the items to add (e.g. `Avatar`)
3. Caller fills out all properties on the local item data model(s)
4. Caller passes the now-fully-formed object(s) to the container to add it/them

Now the big sticking point: what is the object created in step 2? I don't believe we've fully come to an agreement on that point yet. Hopefully we all can discuss this further early next week.

mischreiber: (Note: that list of steps should also include an alternative where the avatar data is passed in at `init` time.)

Just for reference, check out how the "legacy" AvatarGroupView is created and populated: https://github.com/microsoft/fluentui-apple/blob/9abc10617ed457679e63734b7f7f3372295502a6/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupViewDemoController.swift#L182

sophialee0416: I noticed that and had thought about creating some sort of convenience method, should I do that here as well?

mischreiber: As discussed offline, let's hold off on anything too big at the moment, though I'm intrigued about this convenience method.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) {
+        avatars.append(AvatarView(style: style, size: size))
+    }
+
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState? {
+        if atIndex >= avatars.count {
```
mischreiber, line 44: Nit: might be a good place for a `guard` statement (same would be true for `deleteAvatar` below)

mischreiber (approved):

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,191 @@
...
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+//
+/// `MSFAvatarGroupState` defines properties in an AvatarGroup
+///
+/// `maxDisplayedAvatars`: Maximum number of avatars displayed
+///
+/// `overflowCount`: Number to be displayed in the overflow avatar
```
rdeassis, line 18: Based on discussions we had with other teams who were confused about the relation of the maxDisplayedAvatars and overflowCount, can you provide more details on the documentation of these 2 properties?

I think the main topics I'd want to clarify are:
 - **maxDisplayedAvatars** caps the number of displayed avatars and shows the remaining ones that were not displayed in the overflow avatars.
 - **overflowCount** adds to the overflow count in case the calling code did not provide all the avatars, but still want to convey more items than just the remainder of the avatars that could not be displayed due to the maxDisplayedAvatars property.

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,191 @@
...
+
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) {
+        avatars.append(AvatarView(style: style, size: size))
+    }
+
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState? {
+        if atIndex >= avatars.count {
+            print("Requested index is out of bounds!")
```
rdeassis, line 45: We should not use print statements to message an out of bounds calls as this won't get logged in the client apps.
I believe just returning a nil value will suffice.
Also, can you make sure there's documentation for these methods? (covering that this one returns nil if an out of bounds index is passed)

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,191 @@
...
+            return nil
+        } else {
+            return avatars[atIndex].state
+        }
+    }
+
+    func deleteAvatar(atIndex: Int) {
+        avatars.remove(at: atIndex)
```
rdeassis, line 53: The documentation of .remove(at:) states that the index needs to be valid.
https://developer.apple.com/documentation/swift/array/1641390-remove

We should validate before making the call.

Maybe creating a helper func to validate this case and call it from here and in getAvatarState is a good idea?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,191 @@
...
+        let size: CGFloat = tokens.size.size
+        let x: CGFloat = size + tokens.interspace - tokens.ringThickness
+        let avatars: [AvatarView] = state.avatars
+        let maxDisplayedAvatars: Int = avatars.prefix(state.maxDisplayedAvatars).count
+        let overflowCount: Int = (avatars.count > maxDisplayedAvatars ? avatars.count - maxDisplayedAvatars : 0) + state.overflowCount
+        HStack(spacing: 0) {
+            ForEach(0 ..< maxDisplayedAvatars, id: \.self) { index in
+                // If the avatar is part of Stack style and is not the last avatar in the sequence, create a cutout
+                let needsCutout = tokens.style == .stack && (overflowCount > 0 || index + 1 < maxDisplayedAvatars)
+                state.avatars[index]
```
rdeassis, line 111: After the merge of #608, we can now create the avatar states in **createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize)** and pass them to the avtars being created here instead of storing the Avatar Views in the state object.

rdeassis (changes requested):

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
rdeassis: Nevermind. I missed the point of overflowCount, which is to provide an addition to the avatars that were capped by the maxDisplayedAvatars.
I made another comment about the documentation of it. It's easy to get confused with this naming.
What are your thoughts on starting a discussion to rename this property to something like "additionalOverflowCount"?

mischreiber (commented):
Agree with other comments. Will re-review on update

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
rdeassis: As I pointed in another comment, the cutout is incorrect when the rings are set to some avatars.

Screen Shot 2021-06-25 at 10 05 38 AM

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -230,6 +230,15 @@ MSFOverflowAvatarTokens extends MSFAvatarTokens:
   foregroundDefaultColor: $Colors.Foreground.neutral3
   ringDefaultColor: $Colors.Background.neutralDisabled
 
+AP_MSFAvatarGroupTokens:
```
rdeassis, line 233: It seems like the only token we need is the interspace.
Why do we need ringInnerGap, ringThickness and ringOuterGap?

rdeassis (changes requested):

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
```diff
@@ -0,0 +1,195 @@
...
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState? {
+        guard atIndex < avatars.count else {
+            return nil
+        }
+        return avatars[atIndex]
+    }
+
+    func deleteAvatar(atIndex: Int) {
+        guard atIndex < avatars.count else {
+            fatalError("Index is out of bounds")
```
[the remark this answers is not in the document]
rdeassis: Update as per the comment on another PR:
https://github.com/microsoft/fluentui-apple/pull/614#discussion_r661025051

We can keep the fatalError as the standard for these operations. #mischreiber FYI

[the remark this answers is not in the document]
mischreiber: Dang, good find. We have a few `fatalError` calls in the codebase; I'll open up an ADO to get them replaced en masse.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,279 @@
...
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    private var vSamplePersonas: [MSFAvatarState] = {
+        var personas: [MSFAvatarState] = []
+        for index in 0...samplePersonas.count - 1 {
+            let state = MSFAvatar().state
+            state.image = samplePersonas[index].avatarImage
```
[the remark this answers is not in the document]
rdeassis: You can do:
let persona = semplePersonas[index]

To prevent accessing the vector every time you need one of the properties.

tools/sgen/input/FluentUIStyle.yml
rdeassis: Thoughts on this one?

ios/FluentUI/Vnext/Avatar/AvatarGroup.swift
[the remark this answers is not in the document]
rdeassis: This is implemented as a no-op instead of the preconditionFailure that we agreed on the thread. Could you update it?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) -> MSFAvatarState? {
+        avatars.append(MSFAvatarStateImpl(style: style, size: size))
+        return avatars.last
+    }
+
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState? {
+        guard atIndex < avatars.count else {
+            return nil
```
rdeassis, line 47: preconditionFailure instead of returning nil as per the thread in the link below:
https://github.com/microsoft/fluentui-apple/pull/603#discussion_r665741050

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+            return
+        }
+        avatars.remove(at: atIndex)
+    }
+
+    @Published var avatars: [MSFAvatarStateImpl] = []
+    @Published var maxDisplayedAvatars: Int = Int.max
+    @Published var overflowCount: Int = 0
+
+    var size: MSFAvatarSize {
```
rdeassis, line 63: Is this being used anywhere?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+//
+
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+//
```
rdeassis, line 13: nit: remove // and leave just the line break

tools/sgen/input/FluentUIStyle.yml
sophialee0416: It is used for the borders gap in a Stack

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
rdeassis: The return can be a non-optional now that we're implementing the preconditionFailure.

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+/// `getAvatarState`: Gets the state of an Avatar using an index value
+///
+/// `deleteAvatar`: Deletes an Avatar with a specified index value
+@objc public protocol MSFAvatarGroupState {
+    var maxDisplayedAvatars: Int { get set }
+    var overflowCount: Int { get set }
+    var style: MSFAvatarGroupStyle { get set }
+
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) -> MSFAvatarState?
+    func getAvatarState(_ atIndex: Int) -> MSFAvatarState?
```
rdeassis, line 34: Can you follow the same standard in the Swift language?
`getAvatarState(at index: Int) -> MSFAvatarState`

https://developer.apple.com/documentation/swift/array/1641390-remove

Here and for the deleteAvatar implementation too.

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+    func deleteAvatar(atIndex: Int)
```
rdeassis, line 35: nit: rename this to removeAvatar(at index: Int) to follow the naming standard?
https://developer.apple.com/documentation/swift/array/1641390-remove

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+        let interspace: CGFloat = tokens.interspace
+        let ringOuterGap: CGFloat = tokens.ringOuterGap
+        let ringOffset: CGFloat = tokens.ringThickness + tokens.ringInnerGap + ringOuterGap
+        let size: CGFloat = tokens.size.size
+        let x: CGFloat = size + tokens.interspace - tokens.ringThickness
+        HStack(spacing: 0) {
+            ForEach(0 ..< maxDisplayedAvatars, id: \.self) { index in
+                // If the avatar is part of Stack style and is not the last avatar in the sequence, create a cutout
+                let needsCutout = tokens.style == .stack && (overflowCount > 0 || index + 1 < maxDisplayedAvatars)
+                let currentRingCheck = avatars[index].isRingVisible
```
rdeassis, line 117: store avatars[index] in a constant?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,202 @@
...
+                let nextRingCheck = index + 1 < maxDisplayedAvatars ? avatars[index + 1].isRingVisible : false
+                AvatarView(avatars[index])
+                    .modifyIf(needsCutout, { view in
+                        view.mask(AvatarCutout(xOrigin: currentRingCheck ? x + ringOffset : x,
```
rdeassis, line 121: Can you store the results of these nested ternary operators into constants with descriptive names?
It will greatly improve the code readability.
The one on line 127 is very hard to read:

```
 .padding(.trailing, tokens.style == .stack ?
                                (currentRingCheck ? (nextRingCheck ? interspace - (ringOffset + ringOuterGap) : interspace - ringOffset) :
                                    (nextRingCheck ? interspace - ringOuterGap : interspace)) : interspace)
```

tools/sgen/input/FluentUIStyle.yml
rdeassis: Can you use the tokens defined for the Avatar instead of duplicating them for the AvatarGroup?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,289 @@
...
+    private func insertAvatarViews(style: MSFAvatarGroupStyle, showBorders: Bool, mixed: Bool = false) {
+        var constraints: [NSLayoutConstraint] = []
+
+        for size in MSFAvatarSize.allCases.reversed() {
+            let containerView = UIView(frame: .zero)
+
+            let avatarGroup = MSFAvatarGroup(style: style, size: size)
+            var border = showBorders
+            for index in 0...avatarCount - 1 {
+                vSamplePersonas[index].isRingVisible = border
```
rdeassis, line 197: nit: we don't usually have "v" to convey vectors in variable names. I think you can call this just samplePersonas.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,289 @@
...
+                if mixed {
+                    border = !border
+                }
+
+                let avatarState = avatarGroup.state.createAvatar(style: .default, size: size)
+                let samplePersona = vSamplePersonas[index]
+
+                avatarState?.accessibilityLabel = samplePersona.accessibilityLabel
```
rdeassis, line 205: It makes sense to enclose all this block (line 203 to 218) in a:
```
if let avatarState = avatarGroup.state.createAvatar(style: .default, size: size) {
}
```
Do you agree?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
rdeassis: How about creating constants for xOrigin, yOrigin and cutoutSize?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,206 @@
...
+    var style: MSFAvatarGroupStyle { get set }
+
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) -> MSFAvatarState?
+    func getAvatarState(at index: Int) -> MSFAvatarState
+    func deleteAvatar(at index: Int)
+}
+
+/// Properties that make up AvatarGroup content
+class MSFAvatarGroupStateImpl: NSObject, ObservableObject, MSFAvatarGroupState {
+    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) -> MSFAvatarState? {
```
rdeassis, line 40: Does this have to return an optional?
I think we can just make this a non-optional which will allow us not to use the if let that I suggested in the other comment:
https://github.com/microsoft/fluentui-apple/pull/603#discussion_r667107091

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,206 @@
...
+        avatars.append(MSFAvatarStateImpl(style: style, size: size))
```
rdeassis, line 41: It's safer and less error prone to store the return of the
`MSFAvatarStateImpl(style: style, size: size)`
in a constant, then add it to the array and return the constant you just added.

There is a potential risk for a race condition and multiple threads calling this at the same time could cause 2 different calls to return the same avatarstate instance.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
sophialee0416: There is already a samplePersonas array (line 248) that contains all the persona data. I added the v to indicate it was used for vNext

rdeassis: Oh! I didn't catch the idea that "v" indicates vNext.
How about changing the name to **samplePersonaStates** or something that is more descriptive like that?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,289 @@
...
+            addRow(text: size.description, items: [containerView], textStyle: .footnote, textWidth: 100)
+        }
+
+        NSLayoutConstraint.activate(constraints)
+    }
+
+    private var vSamplePersonas: [MSFAvatarState] = {
+        var personas: [MSFAvatarState] = []
+        for index in 0...samplePersonas.count - 1 {
+            let state = MSFAvatar().state
```
rdeassis, line 247: This is not how the state should be used or created.

The samplePersonas (which is just the data class) should be iterated and createAvatar should be called on each avatarGroup state to create the avatars. Creating avatarStates here is just duplicating the data storage for no added benefit.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,289 @@
...
+            let avatarGroup = MSFAvatarGroup(style: style, size: size)
+            var border = showBorders
+            for index in 0...avatarCount - 1 {
+                vSamplePersonas[index].isRingVisible = border
+                if mixed {
+                    border = !border
+                }
+
+                let avatarState = avatarGroup.state.createAvatar(style: .default, size: size)
+                let samplePersona = vSamplePersonas[index]
```
rdeassis, line 203: Instead of doing this, the code should get the samplePersonas data class and setting the state properties here:

```
let persona = samplePersonas[index]
state.image = persona.avatarImage
state.primaryText = persona.primaryText
state.secondaryText = persona.secondaryText
```

ios/FluentUI/Vnext/AvatarGroup/AvatarGroupTokens.swift
```diff
@@ -47,6 +47,7 @@ class MSFAvatarGroupTokens: MSFTokensBase, ObservableObject {
     override func updateForCurrentTheme() {
         let currentTheme = theme
         var appearanceProxy: AppearanceProxyType
+        let avatarTokens = MSFAvatarTokens(style: .default, size: size)
```
rdeassis, line 50: You don't need to create a new MSFAvatarTokens instance:

`avatarAppearanceProxy = currentTheme.MSFAvatarTokens`

And assign it from the different sizes below.

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
rdeassis: What's the word on this comment?

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -30,15 +30,15 @@ import SwiftUI
     var overflowCount: Int { get set }
     var style: MSFAvatarGroupStyle { get set }
 
-    func createAvatar(style: MSFAvatarStyle, size: MSFAvatarSize) -> MSFAvatarState
+    func createAvatar() -> MSFAvatarState
```
rdeassis, line 33: MSFAvatarState allows the client code to change the size and style of the avatar. We should prevent that.
Please refer to the design approach on the PersonaButtonCarousel.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    private func insertAvatarViews(style: MSFAvatarGroupStyle, showBorders: Bool, mixed: Bool = false) {
+        var constraints: [NSLayoutConstraint] = []
+
+        for size in MSFAvatarSize.allCases.reversed() {
+            let containerView = UIView(frame: .zero)
+
+            let avatarGroup = MSFAvatarGroup(style: style, size: size)
+            var border = showBorders
```
rdeassis, line 195: Are you using this variable?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -0,0 +1,265 @@
...
+            for index in 0...avatarCount - 1 {
+                if mixed {
+                    border = !border
+                }
+
+                let avatarState = avatarGroup.state.createAvatar()
+                let samplePersona = samplePersonas[index]
+
+                avatarState.image = samplePersona.avatarImage
+                avatarState.isRingVisible = mixed ? index % 2 == 0 : showBorders
```
rdeassis, line 205: This seems to be duplicating the border variable logic above...

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -0,0 +1,203 @@
...
+
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarGroupStyle: Int, CaseIterable {
+    case stack
+    case pile
+}
+
+/// `MSFAvatarGroupState` defines properties in an AvatarGroup
```
harrieshin, line 14: for my own knowledge, is this the preferred commenting style of swiftui procotols? rather than definding these line above the actual var and functions?

harrieshin (approved):

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -6,8 +6,8 @@
 import UIKit
 import SwiftUI
 
-/// Properties available to customize the state of the avatar
-@objc public protocol MSFAvatarState {
+/// Properties available to customize the state of the Avatar in an Avatar Group
+@objc public protocol MSFAvatarGroupItemState {
```
rdeassis, line 10: These protocols need to be defined within the scope of the AvatarGroup control.

Please take a look at the [MSFPersonaButtonAppearance](https://github.com/microsoft/fluentui-apple/blob/b992a3ceaca09501e34f86205512e212182a11df/ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift#L15) and [MSFPersonaButtonData](https://github.com/microsoft/fluentui-apple/blob/b992a3ceaca09501e34f86205512e212182a11df/ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift#L36) and how they're being used.

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
rdeassis: bringing this comment to your attention...

rdeassis (approved):

PR #604, opened 2021-06-11
Exclude swiftsourceinfo files from nuget

markavitale:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes
When trying to consume an Xcode 12.5 prebuilt version of Fluent UI Apple in Xcode 13 beta, you get a warning that the source info for the module might be corrupted. If you have warnings turned on as errors, this will cause a build break in the Xcode 13 beta project trying to consume Fluent UI Apple.

However, Xcode only tries to load the swift source info files if they are present. Let's just remove them from the prebuit NuGet binaries as we don't require them.

Note: This impacts Microsoft's internal consumption of Fluent UI Apple only.

### Verification
Run the `scripts/publish_nuget.sh` script before and after the change and diff the 

| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/604)

### DIFF
--- a/scripts/prepare_for_nuget_pack.sh
+++ b/scripts/prepare_for_nuget_pack.sh
@@ -18,6 +18,14 @@ function make_dir_if_necessary()
     fi
 }
 
+# A thin wrapper around rsync to exclude specific swift source info files
+#
+# \param $1+ the standard input to rsync without any extra flags provided
+function rsync_excluding_swift_source_info_files()
+{
+    rsync -a --exclude '*.swiftsourceinfo' --prune-empty-dirs "$@"
+}
+
 # Build up our output directory in Products/nuget
 PRODUCTS_DIR="DerivedData/Build/Products"
 NUGET_OUTPUT_DIR="$PRODUCTS_DIR/nuget"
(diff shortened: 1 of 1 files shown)


harrieshin:
for my own knowledge, "swift source info files" what are these for? any impact on existing flow?

scripts/prepare_for_nuget_pack.sh
```diff
@@ -44,34 +52,34 @@ cd $PRODUCTS_DIR
...
+rsync_excluding_swift_source_info_files Debug/FluentUI.framework/ nuget/Debug-macosx/FluentUI.framework/
 
 make_dir_if_necessary "nuget/Ship-macosx"
 echo "Copy Ship-macosx Framework into nuget folder"
-rsync -a Release/FluentUI.framework/ nuget/Ship-macosx/FluentUI.framework/
+rsync_excluding_swift_source_info_files Release/FluentUI.framework/ nuget/Ship-macosx/FluentUI.framework/
 
 make_dir_if_necessary "nuget/Debug-iphoneos"
 echo "Copy Debug-iphoneos build output into nuget folder"
 rsync -a Debug-iphoneos/libFluentUI.a nuget/Debug-iphoneos/
```
harrieshin, line 63: nit: since we are using these path like 3 times with appended component should we just create an short variable so we avoid any human errors like typos?

harrieshin (approved):

anandrajeswaran (approved):

scripts/prepare_for_nuget_pack.sh
markavitale: I felt the same way and was thwarted by my questionable bash skills. I'll take another stab at simplifying this.

markavitale:
> for my own knowledge, "swift source info files" what are these for? any impact on existing flow?

[swiftsourceinfo](https://forums.swift.org/t/proposal-emitting-source-information-file-during-compilation/28794) files are essentially local file path mappings for symbols. They're useful if everything is built on the same machine to help provide better diagnostics and cross-module fixits.

Since in our particular case the compilation is happening on a different machine than those consuming the built outputs, this doesn't provide us any benefit.

scripts/prepare_for_nuget_pack.sh
markavitale: For now I'm going to skip improving this to prioritize getting a build going with these changes, but I may come back here to try and clean things up.

PR #612, opened 2021-06-22
Fixing large text mode height calculation for TableViewCellFileAccessoryView.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This fixes the issue that #573 attempted to fix. Because it caused a regression in client apps, it was reverted by #594.

The root cause of this issue is the calculation of the area available for the text. The height of the labels will be defined based on the available width, the available width value was incorrect in 2 aspects:

1. It was not considering the accessory view size after it was laid out 
2. The cell paddings were not being passed to the height calculation (which assumed the default value of 16 instead of 24).

With a larger width available, the calculated height was shorter and caused the text overlap in large text mode.

### Verification

Validation in the broken client app scenario:

| large_text_tableviewcellfileaccessoryview_before | large_text_tableviewcellfileaccessoryview_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/612)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -447,6 +447,8 @@ open class TableViewCell: UITableViewCell {
     private static func textAreaTrailingOffset(customAccessoryView: UIView?, customAccessoryViewExtendsToEdge: Bool, accessoryType: TableViewCellAccessoryType, paddingTrailing: CGFloat) -> CGFloat {
         let customAccessoryViewAreaWidth: CGFloat
         if let customAccessoryView = customAccessoryView {
+            // Trigger layout so we can have the frame calculated correctly at this point in time
+            customAccessoryView.layoutIfNeeded()
             customAccessoryViewAreaWidth = customAccessoryView.frame.width + Constants.customAccessoryViewMarginLeading
         } else {
             customAccessoryViewAreaWidth = 0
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -447,6 +447,8 @@ open class TableViewCell: UITableViewCell {
     private static func textAreaTrailingOffset(customAccessoryView: UIView?, customAccessoryViewExtendsToEdge: Bool, accessoryType: TableViewCellAccessoryType, paddingTrailing: CGFloat) -> CGFloat {
         let customAccessoryViewAreaWidth: CGFloat
         if let customAccessoryView = customAccessoryView {
+            // Trigger layout so we can have the frame calculated correctly at this point in time
+            customAccessoryView.layoutIfNeeded()
```
harrieshin, line 451: should the caller call do this before calling the static function ?

harrieshin (approved):

ios/FluentUI/Table View/TableViewCell.swift
rdeassis: I thought about this, but there are multiple places where this is called.
So, I think we can avoid potential future bugs and consolidate the check in here.

PR #613, opened 2021-06-22
Fixing TableViewCellFileAccessoryView date label overlap in large text accessibility mode.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In the Large Text accessibility mode the date label text might overlap with the accessory view buttons depending on how long it is.
This change fixes this scenario by introducing a line break to that label when XXXL, XXL or XL accessibility content sizes are enabled, so that the longer text in the bigger font size can be broken into 2 lines.

### Verification

1. Reproduced the scenario on the Demo app and verified the fix on it.
2. Validated consumption in one of our client apps which also reproduced the same issue.

| dateLabel_largetext_before | dateLabel_largetext_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/613)

### DIFF
--- a/ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
+++ b/ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
@@ -207,7 +207,9 @@ open class TableViewCellFileAccessoryView: UIView {
 
         NSLayoutConstraint.activate([
             view.topAnchor.constraint(equalTo: dateLabel.topAnchor),
-            view.bottomAnchor.constraint(equalTo: dateLabel.bottomAnchor)
+            view.bottomAnchor.constraint(equalTo: dateLabel.bottomAnchor),
+            view.leadingAnchor.constraint(equalTo: dateLabel.leadingAnchor),
+            view.trailingAnchor.constraint(equalTo: dateLabel.trailingAnchor)
         ])
 
         return view
@@ -358,6 +360,7 @@ open class TableViewCellFileAccessoryView: UIView {
     private func updateDateLabel() {
         let dateString = date?.displayString(short: (dateColumnWidth.constant < Constants.fullDateMinWidth)) ?? ""
         dateLabel.text = dateString
+        dateLabel.numberOfLines = traitCollection.preferredContentSizeCategory.dateLabelNumberOfLines
     }
 
     @objc private func updateLayout() {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
```diff
@@ -548,4 +551,15 @@ extension UIContentSizeCategory {
             return true
         }
     }
+
+    var dateLabelNumberOfLines: Int {
+        switch self {
+        case .accessibilityExtraExtraExtraLarge, .accessibilityExtraExtraLarge, .accessibilityExtraLarge:
+            // Three lines for accessibility XXXL/XXL/XL content size categories (like the title label) so that
+            // the text can fit in its label's available width without overlapping with accessory buttons.
+            return 3
```
harrieshin, line 560: should we really limit to three? might be better if we cap the font size? or adjust font one width to certain point?

harrieshin:
test on various compact/split mode? cell selection mode?

harrieshin (approved):

ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
rdeassis: I think it makes sense to keep the same font size as the other labels in the text area are using.
I also thought about setting that to unlimited, but it could potentially grow bigger in different locales.
This is the reasoning behind keeping it in sync with the title label logic of 3 lines.

rdeassis:
> test on various compact/split mode? cell selection mode?

Yes, I'm doing a lot of validation in different variations, device types, iOS versions and split screen.

ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
mischreiber: On a related note, is there any need for a line limit at all? What would happen if we just always set the limit to 0?

rdeassis: If in a certain locale the string needs more than 3 lines, the height calculation might be off. Also, this is the logic for the title label, which is more important information than the date label and is still capped at 3 lines.

harrieshin: does it already support large content viewer ?

mischreiber (approved):

ios/FluentUI/Other Cells/TableViewCellFileAccessoryView.swift
rdeassis: Currently it doesn't support the large content viewer although the content increases in size along with the other labels.
We can follow up on our offline conversation about supporting it in case the text truncates and make an additional tweak if needed.

PR #614, opened 2021-06-22
[vNext] PersonaButtonCarousel

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Implemented `PersonaButtonCarousel` and its UIKit projection, `MSFPersonaButtonCarousel`, along with expanding demo controller testability versus the legacy `ContactCollectionView`.

In implementing these controls, I also began migrating some classes/extensions into separate files to aid with readability.

### Verification

Tested on:
- iOS 13 and 14
- Small, medium, and large devices
- Small, medium, and large numbers of personas
- Adding and removing personas live while the control is visible

| CCV | PBC |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/614)

### DIFF
--- a/ios/FluentUI/Vnext/Avatar/Avatar.swift
+++ b/ios/FluentUI/Vnext/Avatar/Avatar.swift
@@ -70,33 +70,6 @@ class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
     }
 }
 
-// MARK: - PersonaData MSFAvatarStateImpl extension
-
-extension MSFAvatarStateImpl {
-    convenience init(style: MSFAvatarStyle, size: MSFAvatarSize, personaData: PersonaData) {
-        self.init(style: style, size: size)
-        self.copyProperties(from: personaData)
-    }
-
-    func copyProperties(from personaData: PersonaData) {
-        if let name = personaData.composedName {
-            self.primaryText = name.0
-            self.secondaryText = name.1
-        } else {
-            let identifier = (personaData.name.count > 0) ? personaData.name : personaData.email
-            self.primaryText = identifier
-        }
-
-        self.foregroundColor = personaData.color
-        self.image = personaData.image
-        self.imageBasedRingColor = personaData.customBorderImage
-        self.isOutOfOffice = personaData.presence == .outOfOffice
-        self.hasRingInnerGap = !personaData.hideInsideGapForBorder
-        self.presence = MSFAvatarPresence(personaData.presence)
-        self.ringColor = personaData.color
-    }
-}
-
 /// View that represents the avatar
 public struct AvatarView: View {
     @Environment(\.theme) var theme: FluentUIStyle
--- a/ios/FluentUI/Vnext/Avatar/Avatar.swift
+++ b/ios/FluentUI/Vnext/Avatar/Avatar.swift
@@ -6,89 +6,6 @@
 import UIKit
 import SwiftUI
 
-@objc public enum MSFAvatarPresence: Int, CaseIterable {
-    case none
-    case available
-    case away
-    case blocked
-    case busy
-    case doNotDisturb
-    case offline
-    case unknown
-
-    func color(isOutOfOffice: Bool) -> Color {
-        var color = UIColor.clear
-
-        switch self {
-        case .none:
-            break
-        case .available:
-            color = FluentUIThemeManager.S.Colors.Presence.available
-        case .away:
-            color = isOutOfOffice ? FluentUIThemeManager.S.Colors.Presence.outOfOffice : FluentUIThemeManager.S.Colors.Presence.away
-        case .busy:
-            color = FluentUIThemeManager.S.Colors.Presence.busy
-        case .blocked:
-            color = FluentUIThemeManager.S.Colors.Presence.blocked
-        case .doNotDisturb:
-            color = FluentUIThemeManager.S.Colors.Presence.doNotDisturb
-        case .offline:
-            color = isOutOfOffice ? FluentUIThemeManager.S.Colors.Presence.outOfOffice : FluentUIThemeManager.S.Colors.Presence.offline
-        case .unknown:
-            color = FluentUIThemeManager.S.Colors.Presence.unknown
-        }
-
-        return Color(color)
-    }
-
-    func image(isOutOfOffice: Bool) -> Image {
-        var imageName = ""
-
-        switch self {
-        case .none:
-            break
-        case .available:
-            imageName = isOutOfOffice ? "ic_fluent_presence_available_16_regular" : "ic_fluent_presence_available_16_filled"
-        case .away:
-            imageName = isOutOfOffice ? "ic_fluent_presence_oof_16_regular" : "ic_fluent_presence_away_16_filled"
-        case .busy:
-            imageName = isOutOfOffice ? "ic_fluent_presence_unknown_16_regular" : "ic_fluent_presence_busy_16_filled"
-        case .blocked:
-            imageName = "ic_fluent_presence_blocked_16_regular"
-        case .doNotDisturb:
-            imageName = isOutOfOffice ? "ic_fluent_presence_dnd_16_regular" : "ic_fluent_presence_dnd_16_filled"
-        case .offline:
-            imageName = isOutOfOffice ? "ic_fluent_presence_oof_16_regular" : "ic_fluent_presence_offline_16_regular"
-        case .unknown:
-            imageName = "ic_fluent_presence_unknown_16_regular"
-        }
-
-        return Image(imageName,
-                     bundle: FluentUIFramework.resourceBundle)
-    }
-
-    public func string() -> String? {
-        switch self {
-        case .none:
-            return nil
-        case .available:
-            return "Presence.Available".localized
-        case .away:
-            return "Presence.Away".localized
-        case .busy:
-            return "Presence.Busy".localized
-        case .blocked:
-            return "Presence.Blocked".localized
-        case .doNotDisturb:
-            return "Presence.DND".localized
-        case .offline:
-            return "Presence.Offline".localized
-        case .unknown:
-            return "Presence.Unknown".localized
-        }
-    }
-}
-
 /// Properties available to customize the state of the avatar
 @objc public protocol MSFAvatarState {
     var accessibilityLabel: String? { get set }
@@ -150,6 +67,33 @@ class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
         self.tokens = MSFAvatarTokens(style: style,
                                       size: size)
         super.init()
+    }
+}
+
+// MARK: - PersonaData MSFAvatarStateImpl extension
+
+extension MSFAvatarStateImpl {
+    convenience init(style: MSFAvatarStyle, size: MSFAvatarSize, personaData: PersonaData) {
+        self.init(style: style, size: size)
+        self.copyProperties(from: personaData)
+    }
+
+    func copyProperties(from personaData: PersonaData) {
+        if let name = personaData.composedName {
+            self.primaryText = name.0
+            self.secondaryText = name.1
+        } else {
+            let identifier = (personaData.name.count > 0) ? personaData.name : personaData.email
+            self.primaryText = identifier
+        }
+
+        self.foregroundColor = personaData.color
+        self.image = personaData.image
+        self.imageBasedRingColor = personaData.customBorderImage
+        self.isOutOfOffice = personaData.presence == .outOfOffice
+        self.hasRingInnerGap = !personaData.hideInsideGapForBorder
+        self.presence = MSFAvatarPresence(personaData.presence)
+        self.ringColor = personaData.color
     }
 }
 
--- a/ios/FluentUI/Vnext/Avatar/MSFAvatarPresence.swift
+++ b/ios/FluentUI/Vnext/Avatar/MSFAvatarPresence.swift
@@ -1,0 +1,116 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+@objc public enum MSFAvatarPresence: Int, CaseIterable {
+    case none
+    case available
+    case away
+    case blocked
+    case busy
+    case doNotDisturb
+    case offline
+    case unknown
+
+    func color(isOutOfOffice: Bool) -> Color {
+        var color = UIColor.clear
+
+        switch self {
+        case .none:
+            break
+        case .available:
+            color = FluentUIThemeManager.S.Colors.Presence.available
+        case .away:
+            color = isOutOfOffice ? FluentUIThemeManager.S.Colors.Presence.outOfOffice : FluentUIThemeManager.S.Colors.Presence.away
+        case .busy:
+            color = FluentUIThemeManager.S.Colors.Presence.busy
+        case .blocked:
+            color = FluentUIThemeManager.S.Colors.Presence.blocked
+        case .doNotDisturb:
+            color = FluentUIThemeManager.S.Colors.Presence.doNotDisturb
+        case .offline:
+            color = isOutOfOffice ? FluentUIThemeManager.S.Colors.Presence.outOfOffice : FluentUIThemeManager.S.Colors.Presence.offline
+        case .unknown:
+            color = FluentUIThemeManager.S.Colors.Presence.unknown
+        }
+
+        return Color(color)
+    }
+
+    func image(isOutOfOffice: Bool) -> Image {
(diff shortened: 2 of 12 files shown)


ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,79 @@
...
+
+    @Published var buttons: [MSFPersonaButtonStateImpl] = [] {
+        didSet {
+            // When the buttons array changes, any new entries need to have their action set
+            // to pass through to the main callback for the carousel
+            let old = Set(oldValue)
+            let new = Set(buttons)
+            let newElements = new.subtracting(old)
+
+            newElements.forEach { button in
```
rdeassis, line 32: How about adding an initializer to the PersonaButton that takes an action closure and pass the PersonaCarouselState closure that will call the client code handler for the Carousel tap event?

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,79 @@
...
+        buttonSize = size
+    }
+
     var body: some View {
-        // TODO: Create PersonaButtonCarousel
-        Text("Hello, World!")
+        ScrollView(.horizontal, showsIndicators: false) {
+            HStack(spacing: 0) {
+                ForEach(state.buttons) { buttonState in
+                    PersonaButton(state: buttonState)
```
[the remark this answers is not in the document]
rdeassis: Going back to my other comment about the action closure on the PersonaButton.
We can pass the Carousel action closure here and forward the call to the client code in case the state object closure is not nil.

ios/FluentUI/Vnext/Avatar/Avatar.swift
```diff
@@ -153,6 +70,33 @@ class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
...
+        self.foregroundColor = personaData.color
+        self.image = personaData.image
+        self.imageBasedRingColor = personaData.customBorderImage
+        self.isOutOfOffice = personaData.presence == .outOfOffice
+        self.hasRingInnerGap = !personaData.hideInsideGapForBorder
+        self.presence = MSFAvatarPresence(personaData.presence)
+        self.ringColor = personaData.color
+    }
+}
+
```
rdeassis, line 99: Just noticed this: It seems like your previous change in the avatar (addition of the **.renderingMode(.original)** modifier) broke the rendering of the icon images:

Screen Shot 2021-06-23 at 10 11 51 AM

rdeassis (changes requested):

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,79 @@
...
+///
+/// - `buttonSize`: specifies whether to use small or large avatars
+/// - `onTapAction`: provides tap gesture
+@objc public protocol MSFPersonaCarouselState {
+    var buttonSize: MSFPersonaButtonSize { get }
+    var onTapAction: ((_ personaButtonState: MSFPersonaButtonState, _ index: Int) -> Void)? { get set }
+}
+
+/// Properties that make up PersonaGrid content
+class MSFPersonaCarouselStateImpl: NSObject, ObservableObject, Identifiable, MSFPersonaCarouselState {
```
rdeassis, line 18: If you implement the ID property of the Identifiable protocol assigning a GUID to it, does the animation get smoother?
I think it's worth a shot so that SwiftUI keeps track of the state objects and takes care of the transitions for us.

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
```diff
@@ -0,0 +1,82 @@
...
+        super.init()
+
+        personaButtonCarousel = PersonaButtonCarousel(size: size, personaButtons: personas.map { personaData in
+            MSFPersonaButtonStateImpl(size: size, personaData: personaData)
+        })
+
+        personaButtonCarousel.state.onTapAction = { [weak self] (personaButtonState: MSFPersonaButtonState, index: Int) in
+            // Unfortunate that we have to have this cast, but the alternative is to expose
+            // the internal `PersonaData` property on the public interface
+            guard let personaData = (personaButtonState as? MSFPersonaButtonStateImpl)?.personaData else {
```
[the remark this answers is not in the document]
mischreiber: True! That will make bigger closures easier to read, for sure.

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,79 @@
 
 import SwiftUI
 
+/// `MSFPersonaCarouselState` contains PersonaCarousel properties
+///
+/// - `buttonSize`: specifies whether to use small or large avatars
```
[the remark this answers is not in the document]
mischreiber: Agreed, this was written while it was still read/write. Can definitely be clearer.

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -181,32 +182,46 @@ class MSFPersonaButtonViewStateImpl: NSObject, ObservableObject, Identifiable, M
...
+    }
 }
 
 public struct PersonaButton: View {
     @Environment(\.theme) var theme: FluentUIStyle
     @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
     @ObservedObject var tokens: MSFPersonaButtonTokens
-    @ObservedObject var state: MSFPersonaButtonViewStateImpl
+    @ObservedObject var state: MSFPersonaButtonStateImpl
     @ObservedObject var avatarState: MSFAvatarStateImpl
```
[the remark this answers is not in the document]
mischreiber: Love it!

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
```diff
@@ -0,0 +1,82 @@
...
+                                                                    })))
+        hostingController.disableSafeAreaInsets()
+        view.backgroundColor = UIColor.clear
+    }
+
+    @objc public func count() -> Int {
+        return self.personaButtonCarousel.state.buttons.count
+    }
+
+    @objc public func append(_ personaData: PersonaData) {
```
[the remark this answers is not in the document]
mischreiber: Discussed offline, next iteration has a solution for that (we will pass the created `PersonaButtonData` object back to the caller as a `@discardableResult`)

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
mischreiber: Now we're thinking with SwiftUI! Awesome idea 👍

ios/FluentUI/Vnext/Avatar/Avatar.swift
mischreiber: I'll open a bug and fix that in a separate PR, thanks for catching

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
mischreiber: It does not :( I'm not sure why the animation is so janky right now. I'll keep poking at it and see if I can figure anything out.

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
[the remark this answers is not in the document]
mischreiber: Yep, done and done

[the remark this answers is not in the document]
mischreiber: As for the large text, I agree that this isn't ideal. I'll experiment a bit and probably fix as a separate PR, since it affects PersonaButton too.

ios/FluentUI/Vnext/Avatar/Avatar.swift
mischreiber: Created a new PR: 

**Fix `Avatar` image rendering mode** https://github.com/microsoft/fluentui-apple/pull/617

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,65 @@
 
 import SwiftUI
 
+/// `MSFPersonaCarouselState` contains PersonaCarousel properties
+///
+/// - `buttons`: the array of buttons displayed in the carousel
```
rdeassis, line 10: nit: there's no buttons property defined on the protocol.

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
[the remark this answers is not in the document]
rdeassis: Cool. Thank you!
Could you create a work item (probably an accessibility bug in ADO) tracking the follow up of this issue?

ios/FluentUI/Vnext/Avatar/Avatar.swift
rdeassis: Thank you!

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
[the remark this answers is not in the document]
mischreiber: Filed!

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
```diff
@@ -0,0 +1,110 @@
...
+    ///   - secondaryText: The secondary text to appear in the `PersonaButton`, below `primaryText`
+    ///   - image: The image to use as the persona's avatar
+    ///
+    /// - Returns: An optional reference to the added `PersonaButton`, which can be used to set additional properties or to update later
+    @discardableResult @objc public func add(primaryText: String?, secondaryText: String?, image: UIImage?) -> MSFPersonaButtonData {
+
+        let persona = MSFPersonaButtonStateImpl(size: self.state.buttonSize)
+
+        // Set passed-in properties
+        persona.primaryText = primaryText
```
rdeassis, line 50: I think this logic should be in the state object and just called from here.
The scenario I'm thinking is a Carousel contained in another parent composite control. We would need to rely on the UIKit wrapper to add/remove PersonaButtons.

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
```diff
@@ -0,0 +1,110 @@
...
+
+    /// Adds a `PersonaButton` to the carousel, and returns an optional reference for additional property setting
+    ///
+    /// - Parameters:
+    ///   - primaryText: The primary text to appear in the `PersonaButton`
+    ///   - secondaryText: The secondary text to appear in the `PersonaButton`, below `primaryText`
+    ///   - image: The image to use as the persona's avatar
+    ///
+    /// - Returns: An optional reference to the added `PersonaButton`, which can be used to set additional properties or to update later
+    @discardableResult @objc public func add(primaryText: String?, secondaryText: String?, image: UIImage?) -> MSFPersonaButtonData {
```
rdeassis, line 45: Shouldn't we also add a func to retrieve an MSFPersonaButtonData instance like we do in the AvatarGroup?

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
```diff
@@ -0,0 +1,110 @@
...
+        }
+    }
+
+    /// Removes a `PersonaButton` from the carousel at the given index.
+    ///
+    /// - Parameters:
+    ///   - index: The index at which the `PersonaButton` to be removed can be currently found.
+    @objc public func remove(at index: Int) {
+        if index >= count {
+            fatalError("Attempting to remove item outside bounds of carousel")
```
rdeassis, line 75: If I'm not mistaken, the AvatarGroup implements a no-op behavior in this case.
We should be consistent on this behavior across our collection states. 
Thoughts?

ios/FluentUI/Vnext/Avatar/MSFAvatarPresence.swift
```diff
@@ -0,0 +1,116 @@
...
+        case .available:
+            self = .available
+        case .away:
+            self = .away
+        case .busy:
+            self = .busy
+        case .doNotDisturb:
+            self = .doNotDisturb
+        case .outOfOffice:
+            // Not supported
```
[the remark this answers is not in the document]
mischreiber: Good catch, missed this one

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
```diff
@@ -5,9 +5,64 @@
...
+    }
+
     var body: some View {
-        // TODO: Create PersonaButtonCarousel
-        Text("Hello, World!")
+        ScrollView(.horizontal, showsIndicators: false) {
+            HStack(spacing: 0) {
+                ForEach(state.buttons, id: \.self) { buttonState in
+                    PersonaButton(state: buttonState) { [weak state] in
+                        guard let state = state,
```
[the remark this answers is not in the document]
mischreiber: Not sure how the compiler would still be happy: if the guard was removed then it'd change from an optional to a non-optional var, so attempting to use it explicitly would still fail. I guess if you passed it into another optional property then it'd change the behavior? But either way I can change it if that's the team standard.

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
mischreiber: I just checked and `AvatarGroup` also implements this as a `fatalError` so we're good.

mischreiber: Regardless, I think firing a `fatalError` is safer than allowing the code to silently fail. If a client is trying to remove a `Persona` from a nonexistent index then it probably indicates an error in their own business logic somewhere.

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
mischreiber: Sure I can add that.

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
mischreiber: Agreed. That said, I intentionally didn't address this until the team has had a chance to discuss our plan going forward for how to handle SwiftUI public interfaces. But for the time being, I think it makes sense to move all of these methods (add, remove, count, etc) to the `state` object and have this layer just pass through.

ios/FluentUI/Vnext/PersonaButtonCarousel/PersonaButtonCarousel.swift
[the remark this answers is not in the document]
rdeassis: You're right.. I missed the optional factor. Thanks for updating it to "strongSelf" to keep the standard!

ios/FluentUI/Vnext/PersonaButtonCarousel/MSFPersonaButtonCarousel.swift
rdeassis: Sounds good.
I left a comment on the other PR so it's not updated to a no-op:
https://github.com/microsoft/fluentui-apple/pull/603/files#r658904309

@sophialee0416 FYI

rdeassis (approved):

PR #615, opened 2021-06-25
Add simplified inits and `tag` to `CommandingItem`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Some objc clients need an init that just takes title/image/action. Separately, for integration that I'm working on I need a simple init(), because most of the properties are filled during a separate binding process.

I also added a new tag property to help application-side tracking of the items.

### Verification
Sanity checks + manually checked the bridging header to verify the objc inits are there.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/615)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -366,8 +366,9 @@ open class BottomCommandingController: UIViewController {
 
     private func createAndBindHeroCommandView(with item: CommandingItem) -> UIView {
         let itemImage = item.image ?? UIImage()
-        let tabItem = TabBarItem(title: item.title ?? "", image: itemImage, selectedImage: item.selectedImage, largeContentImage: item.largeImage)
-        let itemView = TabBarItemView(item: tabItem, showsTitle: item.title != nil)
+        let itemTitle = item.title ?? ""
+        let tabItem = TabBarItem(title: itemTitle, image: itemImage, selectedImage: item.selectedImage, largeContentImage: item.largeImage)
+        let itemView = TabBarItemView(item: tabItem, showsTitle: itemTitle != "")
         itemView.alwaysShowTitleBelowImage = true
         itemView.numberOfTitleLines = 1
         itemView.isSelected = item.isOn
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -365,8 +365,9 @@ open class BottomCommandingController: UIViewController {
     }
 
     private func createAndBindHeroCommandView(with item: CommandingItem) -> UIView {
-        let tabItem = TabBarItem(title: item.title, image: item.image, selectedImage: item.selectedImage, largeContentImage: item.largeImage)
-        let itemView = TabBarItemView(item: tabItem, showsTitle: true)
+        let itemImage = item.image ?? UIImage()
+        let tabItem = TabBarItem(title: item.title ?? "", image: itemImage, selectedImage: item.selectedImage, largeContentImage: item.largeImage)
```
rdeassis, line 369: nit: put the **item.title ?? ""** result in a constant to avoid the duplicated calls below?

rdeassis (approved):

anandrajeswaran:
Nothing about the description seems that specific to objC though, right? If those properties are set during binding are there any clients that do want to have to set all of them in the init? Should we just simplify the init for both objC and swift?

lcapkovic:
> Nothing about the description seems that specific to objC though, right? If those properties are set during binding are there any clients that do want to have to set all of them in the init? Should we just simplify the init for both objC and swift?
> 
> Basically can we just avoid the default values (which is swift only) and just remove them from the init altogether

The demo controller itself is a great example of why having a big init with default values is great for something like the CommandingItem. It lets you create the items in one go when you're making static arrays of CommandingItems etc. And then in objc where we don't get that flexibility, we need those pre-defined inits. I don't want to limit swift clients because of objc limitations.

lcapkovic:
> Nothing about the description seems that specific to objC though, right? If those properties are set during binding are there any clients that do want to have to set all of them in the init? Should we just simplify the init for both objC and swift?
> 
> Basically can we just avoid the default values (which is swift only) and just remove them from the init altogether

After talking offline, decided to go with simplified inits that work for both swift and objc.

PR #618, opened 2021-06-28
Add image support for borderless button style

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

Files are only related to ios.

#### Button.swift
Added support for having an image on a button that is styled as `.borderless`. Also updated a comment about the size supported for an image and added spacing between the image and label of a `.borderless` button.

#### ButtonDemoController.swift
Added examples of the `.borderless` style with an image.

### Verification

Manually tested the change in the demo application to show that current uses of the button (no image specified), still work correctly alongside buttons that do have an image.

| image | image|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/618)

### DIFF
--- a/ios/FluentUI/Button/Button.swift
+++ b/ios/FluentUI/Button/Button.swift
@@ -71,9 +71,9 @@ public enum ButtonStyle: Int, CaseIterable {
         switch self {
         case .primaryFilled, .primaryOutline:
             return 10
-        case .secondaryOutline:
+        case .secondaryOutline, .borderless:
             return 8
-        case .tertiaryOutline, .borderless:
+        case .tertiaryOutline:
             return 0
         }
     }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Button/Button.swift
```diff
@@ -241,7 +241,7 @@ open class Button: UIButton {
     }
 
     private func updateImage() {
-        let isDisplayingImage = (style == .primaryFilled || style == .primaryOutline || style == .secondaryOutline) && image != nil
+        let isDisplayingImage = (style == .primaryFilled || style == .primaryOutline || style == .secondaryOutline || style == .borderless) && image != nil
```
harrieshin, line 244: nit : with this long list i wonder it will be more simplified if we check NOT .tertiaray.

harrieshin (approved):

rdeassis (approved):

PR #619, opened 2021-06-29
fix VO order for segmented control in ColorDemoController

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

fix VO order for segmented control in ColorDemoController by setting a new color provider and reloading the table view when switching themes rather than popping a new view controller which makes the VO start over at the back button when it should stay on the selected control

### Verification

tested on iPad and ensured VO maintained its focus on the newly selected segmented control (mimicking native UI behavior) and announced the proper label after

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/619)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -60,13 +60,13 @@ class DemoListViewController: UITableViewController {
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
 
-        if isFirstLaunch {
+        if DemoListViewController.isFirstLaunch {
             if let lastDemoController = UserDefaults.standard.string(forKey: DemoListViewController.lastDemoControllerKey),
                 let index = demos.firstIndex(where: { $0.title == lastDemoController }) {
                 tableView(tableView, didSelectRowAt: IndexPath(row: index, section: 0))
             }
 
-            isFirstLaunch = false
+            DemoListViewController.isFirstLaunch = false
         } else {
             UserDefaults.standard.set(nil, forKey: DemoListViewController.lastDemoControllerKey)
         }
@@ -97,6 +97,6 @@ class DemoListViewController: UITableViewController {
     }
 
     let cellReuseIdentifier: String = "TableViewCell"
-    private var isFirstLaunch: Bool = true
+    private static var isFirstLaunch: Bool = true
     private static let lastDemoControllerKey: String = "LastDemoController"
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
@@ -129,7 +129,7 @@ class ColorDemoController: UIViewController {
         super.viewDidAppear(animated)
         navigationController?.navigationBar.shadowImage = UIImage()
         if let window = view.window {
-            segmentedControl.selectedSegmentIndex = colorProviderThemedWindowTypes.firstIndex(where: { return window.isKind(of: $0.windowType) }) ?? 0
+        segmentedControl.selectedSegmentIndex = colorProviderThemes.firstIndex(where: { return window.isKind(of: $0.demoColorTheme.windowType) }) ?? 0
         }
     }
 
@@ -139,14 +139,14 @@ class ColorDemoController: UIViewController {
     }
 
     private lazy var segmentedControl: SegmentedControl = {
-        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return SegmentItem(title: $0.name) }), style: .primaryPill)
+        let segmentedControl = SegmentedControl(items: colorProviderThemes.map({ return SegmentItem(title: $0.name) }), style: .primaryPill)
         segmentedControl.addTarget(self, action: #selector(segmentedControlValueChanged(sender:)), for: .valueChanged)
         return segmentedControl
     }()
 
     @objc private func segmentedControlValueChanged(sender: Any) {
         if let segmentedControl = sender as? SegmentedControl {
-            let window = colorProviderThemedWindows[segmentedControl.selectedSegmentIndex].window
+            let window = colorProviderThemes[segmentedControl.selectedSegmentIndex].demoColorTheme.window
             if let colorProvider = window as? ColorProviding {
                 Colors.setProvider(provider: colorProvider, for: window)
             }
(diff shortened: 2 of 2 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -153,6 +153,11 @@ class ColorDemoController: UIViewController {
             if let navigationController = navigationController {
                 navigationController.popViewController(animated: false)
                 colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
+                UIAccessibility.post(notification: .screenChanged, argument: self.segmentedControl.segmentView(at:segmentedControl.selectedSegmentIndex))
+                DispatchQueue.main.asyncAfter(deadline: .now() + .milliseconds(100))
```
kubalani, line 157: Why use a timer?

harrieshin:
this bug should be fixed by "re-writting" how the viewcontroller is loaded. Instead of removing the viewcontroller and adding back the viewcontroller to update the colors, just updating the section of the democontroller that shows the primarycolors would be more appropriate.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
jeaniehuynh: From [here](https://stackoverflow.com/questions/21554056/call-ios-accessibility-voice-over-programmatically-on-a-ui-element): "If you post your announcement immediately after a user action, such as a button tap, it may get swallowed by the button's automatic voice-over announcement. In this case you should post your announcement after a suitable delay (as mentioned here). If that delay is too short, your announcement might cut into the button's announcement."

We'll be taking an alternative route to fix this bug though, so the timer won't be used in future iterations of this change.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -147,19 +146,23 @@ class ColorDemoController: UIViewController {
...
+            if let colorProvider = window as? ColorProviding {
+                Colors.setProvider(provider: colorProvider, for: window)
             }
         }
+
+        tableView.reloadData()
     }
 
+    private let tableView = UITableView(frame: .zero, style: .grouped)
     private let colorProviderThemedWindowTypes: [(name: String, windowType: UIWindow.Type)] = [("Default", DemoColorThemeDefaultWindow.self),
```
harrieshin, line 159: instead of 2 different dictionary, could we make dictionary with string and a new struct, or just array of custom object to make sure these are aligned all the time?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -223,13 +227,13 @@ class DemoColorView: UIView {
...
     }
 
-    override func didMoveToWindow() {
-        super.didMoveToWindow()
-        if let colorProvider = colorProvider,
-            let window = window {
-            backgroundColor = colorProvider(window)
-        }
-    }
+	func updateColorView(window: UIWindow)
```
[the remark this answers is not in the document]
jeaniehuynh: still trying to get it to work/stop reporting thousands of issues--just resolved the ones that the validation in the checks below pointed out

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -140,26 +139,31 @@ class ColorDemoController: UIViewController {
...
 
-    private let colorProviderThemedWindowTypes: [(name: String, windowType: UIWindow.Type)] = [("Default", DemoColorThemeDefaultWindow.self),
-                                                                                               ("Green", DemoColorThemeGreenWindow.self),
-                                                                                               ("None", UIWindow.self)]
+    private let tableView = UITableView(frame: .zero, style: .grouped)
+        private let colorProviderThemes: [(name: String, demoColorTheme: DemoColorTheme)] = [("Default", DemoColorTheme.init(window: DemoColorThemeDefaultWindow(), windowType: DemoColorThemeDefaultWindow.self)),
+                                                                                             ("Green", DemoColorTheme.init(window: DemoColorThemeGreenWindow(), windowType: DemoColorThemeGreenWindow.self)),
+                                                                                             ("None", DemoColorTheme.init(window: UIWindow(), windowType: UIWindow.self))]
+
+    struct DemoColorTheme {
```
harrieshin, line 163: this new struct should be defined somewhere else outside of this {} bracket? also mark private?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -140,26 +139,31 @@ class ColorDemoController: UIViewController {
...
-        let segmentedControl = SegmentedControl(items: colorProviderThemedWindowTypes.map({ return SegmentItem(title: $0.name) }), style: .primaryPill)
+        let segmentedControl = SegmentedControl(items: colorProviderThemes.map({ return SegmentItem(title: $0.name) }), style: .primaryPill)
         segmentedControl.addTarget(self, action: #selector(segmentedControlValueChanged(sender:)), for: .valueChanged)
         return segmentedControl
     }()
 
     @objc private func segmentedControlValueChanged(sender: Any) {
         if let segmentedControl = sender as? SegmentedControl {
-            let windowType = colorProviderThemedWindowTypes[segmentedControl.selectedSegmentIndex].windowType
-            let colorThemeHost = view.window?.windowScene?.delegate as? ColorThemeHosting
```
harrieshin, line 151: this code is important. once user change the segment control color and user goes back to the main table list, all the subsequence fluent controls should adhere to the new selected color provider. for example, color > switch to green > go back to demo main list > go to button > see all the buttons in green > go back to main list > go back to color green should be still selected in segmented control.

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -97,6 +97,6 @@ class DemoListViewController: UITableViewController {
     }
 
     let cellReuseIdentifier: String = "TableViewCell"
-    private var isFirstLaunch: Bool = true
+    private static var isFirstLaunch: Bool = true
```
harrieshin, line 100: what's the reason for this change?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -130,7 +129,7 @@ class ColorDemoController: UIViewController {
         super.viewDidAppear(animated)
         navigationController?.navigationBar.shadowImage = UIImage()
         if let window = view.window {
-            segmentedControl.selectedSegmentIndex = colorProviderThemedWindowTypes.firstIndex(where: { return window.isKind(of: $0.windowType) }) ?? 0
+            segmentedControl.selectedSegmentIndex = colorProviderThemes.firstIndex(where: { return window.isKind(of: $0.demoColorTheme.windowType) }) ?? 0
```
harrieshin, line 132: I still see the segmented control jumps between default to green when loading the color demo controller. what is missing here to get the selected index correct?

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
jeaniehuynh: Without it, the isFirstLaunch bool keeps getting set back to true when going back to the Demo List View, but it should only be true on the first launch only. Changing it to static maintains the value once it gets set to false. This fixes the issue where the Color Demo View gets re-presented after tapping the "<" back button (video I sent you on Teams).

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
jeaniehuynh: Two reasons:
1. It's being set after the view already appeared (in viewDidAppear)
2. After moving it to viewWillAppear or viewDidLoad, the green segment selector stayed on Default and didn't switch over (the isKind(of:) call returns the wrong index).

To fix this, I created a static variable in ColorDemoController that is set to the window type and check for that instead in the firstIndex(where:) call.

harrieshin (approved):

PR #621, opened 2021-07-01
Adjust width of persona button from size category

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add logic to `PersonaButton` to adjust its width based on the current size category. This also resulted in some subtle changes to how the width of the button is calculated (and the deprecation of a redundant design token), which will result in safer, more predictable layout going forward.

### Verification

Verified accessibility sizes from smallest to largest. All spacing is reasonable and legible, and performance is solid.

Also verified exceptionally short and exceptionally long names at various accessibility sizes.

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/621)

### DIFF
--- a/ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
+++ b/ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
@@ -69,28 +69,6 @@ extension FluentUIStyle {
 		}
 
 
-		// MARK: - labelWidth
-		open var labelWidth: labelWidthAppearanceProxy {
-			return labelWidthAppearanceProxy(proxy: mainProxy)
-		}
-		open class labelWidthAppearanceProxy {
-			public let mainProxy: () -> FluentUIStyle
-			public init(proxy: @escaping () -> FluentUIStyle) {
-				self.mainProxy = proxy
-			}
-
-			// MARK: - large 
-			open var large: CGFloat {
-				return CGFloat(72.0)
-			}
-
-			// MARK: - small 
-			open var small: CGFloat {
-				return CGFloat(52.0)
-			}
-		}
-
-
 		// MARK: - padding 
 		open var padding: CGFloat {
 			return mainProxy().Spacing.xSmall
--- a/ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
+++ b/ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
@@ -42,6 +42,33 @@ extension FluentUIStyle {
 			return mainProxy().Colors.Background.neutral1
 		}
 
+		// MARK: - horizontalAvatarPadding
+		open var horizontalAvatarPadding: horizontalAvatarPaddingAppearanceProxy {
+			return horizontalAvatarPaddingAppearanceProxy(proxy: mainProxy)
+		}
+		open class horizontalAvatarPaddingAppearanceProxy {
+			public let mainProxy: () -> FluentUIStyle
+			public init(proxy: @escaping () -> FluentUIStyle) {
+				self.mainProxy = proxy
+			}
+
+			// MARK: - large 
+			open var large: CGFloat {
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -227,29 +228,50 @@ public struct PersonaButton: View {
...
     @ViewBuilder
     private var avatarView: some View {
         AvatarView(state.avatarState)
             .padding(.top, tokens.padding)
             .padding(.bottom, tokens.avatarInterspace)
-            .padding(.horizontal, tokens.padding)
+    }
+
+    /// Width of the button is conditional on the current size category
+    private func adjustedWidth() -> CGFloat {
```
harrieshin, line 253: fluent is usually snaps to 4 px grid. the fact these numbers are not dividable by 4 makes me little concern. also, are we giving too much spacing in certain category? or is it coming from designer recommendation?

harrieshin:
what i am also concern about this current bigger spacing is that carousel might not look horiztonally scrollable in certain size because there isn't a "cutoff" preview of the next avatar button which users might miss....

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: These numbers come from the CCV, but I don't know where _those_ come from. I'll investigate further before merging the PR.

ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
```diff
@@ -42,60 +42,60 @@ extension FluentUIStyle {
...
-				return mainProxy().Typography.subheadline
+			open var large: CGFloat {
+				return mainProxy().Spacing.xSmall
 			}
 
 			// MARK: - small 
-			open var small: UIFont {
-				return mainProxy().Typography.caption1
+			open var small: CGFloat {
+				return mainProxy().Spacing.medium
```
harrieshin, line 62: hmmm its interesting how smaller carousel has more larger spacing...

harrieshin (approved):

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
```diff
@@ -227,38 +228,59 @@ public struct PersonaButton: View {
...
-            .padding(.top, tokens.padding)
+            .padding(.top, tokens.verticalPadding)
             .padding(.bottom, tokens.avatarInterspace)
-            .padding(.horizontal, tokens.padding)
+    }
+
+    /// Width of the button is conditional on the current size category
+    private func adjustedWidth() -> CGFloat {
+        return state.avatarState.size.size + (2 * tokens.horizontalAvatarPadding) + {
+            switch UITraitCollection.current.preferredContentSizeCategory {
```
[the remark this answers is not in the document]
mischreiber: Presumably! Now I'm curious why the old CCV did it the way I, ahem, "borrowed", but either way easy fix.

ios/FluentUI/Vnext/PersonaButton/MSFPersonaButtonTokens.generated.swift
mischreiber: Very interesting, but I think it's because we still need almost the same amount of space for text. The avatar is much smaller, but the text is only a little smaller, so the gap has to get bigger to help the label stay readable.

ios/FluentUI/Vnext/PersonaButton/PersonaButton.swift
mischreiber: Yes! Because I spent too much of my career working in Objective-C, which doesn't use computed properties the same way :)

Updating now!

rdeassis (approved):

PR #622, opened 2021-07-01
Auto-size `BottomCommandingController` based on item count

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

BottomCommandingController had a temporary solution in place to always expand the sheet to a constant fraction of screen height. This doesn't work well for clients with only a few commands, where the expected behavior is for the sheet to fit the commands table view. This PR implements auto sizing to fit the command table view, both in the sheet and bottom bar (popover) presentations.

`BottomSheetController.swift`
Add an API for preferred expanded content height. This will be used to set the height of `expandedContentView`, up to an upper limit determined by screen size.

Using the provided commands, calculate the table view height and pass this to the `BottomSheetController` or the popover presentation controller so they display at the correct size.

`BottomCommandingDemoController`
Add distinct sets of low and high # of commands to test both the fitting and fullscreen sheet variants.

`BottomSheetDemoController`
Add buttons to test the new sizing API.

https://user-images.githubusercontent.com/3610850/124060898-c96afb80-d9e2-11eb-902b-fe86e051c8e6.MP4

### Verification
Testing in the test app on simulator and device, various sanity checks in portrait / landscape, changing safe area insets.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/622)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -22,7 +22,7 @@ class BottomCommandingDemoController: UIViewController {
 
         let bottomCommandingVC = BottomCommandingController()
         bottomCommandingVC.heroItems = heroItems
-        bottomCommandingVC.expandedListSections = fewExpandedListSections
+        bottomCommandingVC.expandedListSections = shortCommandSectionList
 
         addChild(bottomCommandingVC)
         view.addSubview(bottomCommandingVC.view)
@@ -54,35 +54,36 @@ class BottomCommandingDemoController: UIViewController {
         CommandingItem(title: "Boolean Item " + String($0), image: homeImage, action: commandAction, isToggleable: true)
     }
 
-    private lazy var fewExpandedListSections: [CommandingSection] = [
+    private lazy var shortCommandSectionList: [CommandingSection] = [
         CommandingSection(title: "Section 1", items:
         Array(1...2).map {
             CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
         } + booleanCommands)
     ]
 
-    private lazy var manyExpandedListSections: [CommandingSection] = fewExpandedListSections + [
-        CommandingSection(title: "Section 2", items:
-        Array(1...7).map {
-            CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
-        }),
-        CommandingSection(title: "Section 3", items:
-        Array(1...7).map {
-            CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
-        }),
-        CommandingSection(title: "Section 4", items:
-        Array(1...7).map {
-            CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
-        })
-    ]
+    private lazy var longCommandSectionList: [CommandingSection] = shortCommandSectionList
+        + [
+            CommandingSection(title: "Section 2", items:
+            Array(1...7).map {
+                CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
+            }),
+            CommandingSection(title: "Section 3", items:
+            Array(1...7).map {
+                CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
+            }),
+            CommandingSection(title: "Section 4", items:
+            Array(1...7).map {
+                CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
+            })
+        ]
 
-    private lazy var currentExpandedListSections: [CommandingSection] = fewExpandedListSections
+    private lazy var currentExpandedListSections: [CommandingSection] = shortCommandSectionList
 
     private lazy var demoOptionItems: [DemoItem] = {
         return [DemoItem(title: "Hidden", type: .boolean, action: #selector(toggleHidden), isOn: false),
                 DemoItem(title: "Sheet more button", type: .boolean, action: #selector(toggleSheetMoreButton), isOn: true),
                 DemoItem(title: "Expanded list items", type: .boolean, action: #selector(toggleExpandedItems), isOn: true),
-                DemoItem(title: "Many expanded list items", type: .boolean, action: #selector(toggleManyExpandedItems(_:)), isOn: false),
+                DemoItem(title: "Additional expanded list items", type: .boolean, action: #selector(toggleAdditionalExpandedItems(_:)), isOn: false),
                 DemoItem(title: "Hero command isOn", type: .boolean, action: #selector(toggleHeroCommandOnOff)),
                 DemoItem(title: "Hero command isEnabled", type: .boolean, action: #selector(toggleHeroCommandEnabled), isOn: true),
                 DemoItem(title: "List command isEnabled", type: .boolean, action: #selector(toggleListCommandEnabled), isOn: true),
@@ -111,11 +112,11 @@ class BottomCommandingDemoController: UIViewController {
         }
     }
 
-    @objc private func toggleManyExpandedItems(_ sender: BooleanCell) {
+    @objc private func toggleAdditionalExpandedItems(_ sender: BooleanCell) {
         if sender.isOn {
-            currentExpandedListSections = manyExpandedListSections
+            currentExpandedListSections = longCommandSectionList
         } else {
-            currentExpandedListSections = fewExpandedListSections
+            currentExpandedListSections = shortCommandSectionList
         }
         bottomCommandingController?.expandedListSections = currentExpandedListSections
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -22,7 +22,7 @@ class BottomCommandingDemoController: UIViewController {
 
         let bottomCommandingVC = BottomCommandingController()
         bottomCommandingVC.heroItems = heroItems
-        bottomCommandingVC.expandedListSections = expandedListSections
+        bottomCommandingVC.expandedListSections = fewExpandedListSections
 
         addChild(bottomCommandingVC)
         view.addSubview(bottomCommandingVC.view)
@@ -54,20 +54,35 @@ class BottomCommandingDemoController: UIViewController {
         CommandingItem(title: "Boolean Item " + String($0), image: homeImage, action: commandAction, isToggleable: true)
     }
 
-    private lazy var expandedListSections: [CommandingSection] = [
+    private lazy var fewExpandedListSections: [CommandingSection] = [
         CommandingSection(title: "Section 1", items:
         Array(1...2).map {
             CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
-        } + booleanCommands),
-        CommandingSection(title: "Section 2", items: Array(1...7).map {
+        } + booleanCommands)
+    ]
+
+    private lazy var manyExpandedListSections: [CommandingSection] = fewExpandedListSections + [
+        CommandingSection(title: "Section 2", items:
+        Array(1...7).map {
+            CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
+        }),
+        CommandingSection(title: "Section 3", items:
+        Array(1...7).map {
+            CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
+        }),
+        CommandingSection(title: "Section 4", items:
+        Array(1...7).map {
             CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
         })
     ]
+
+    private lazy var currentExpandedListSections: [CommandingSection] = fewExpandedListSections
 
     private lazy var demoOptionItems: [DemoItem] = {
         return [DemoItem(title: "Hidden", type: .boolean, action: #selector(toggleHidden), isOn: false),
                 DemoItem(title: "Sheet more button", type: .boolean, action: #selector(toggleSheetMoreButton), isOn: true),
                 DemoItem(title: "Expanded list items", type: .boolean, action: #selector(toggleExpandedItems), isOn: true),
+                DemoItem(title: "Many expanded list items", type: .boolean, action: #selector(toggleManyExpandedItems(_:)), isOn: false),
                 DemoItem(title: "Hero command isOn", type: .boolean, action: #selector(toggleHeroCommandOnOff)),
                 DemoItem(title: "Hero command isEnabled", type: .boolean, action: #selector(toggleHeroCommandEnabled), isOn: true),
                 DemoItem(title: "List command isEnabled", type: .boolean, action: #selector(toggleListCommandEnabled), isOn: true),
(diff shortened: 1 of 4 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -54,20 +54,35 @@ class BottomCommandingDemoController: UIViewController {
...
         CommandingSection(title: "Section 1", items:
         Array(1...2).map {
             CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
-        } + booleanCommands),
-        CommandingSection(title: "Section 2", items: Array(1...7).map {
+        } + booleanCommands)
+    ]
+
+    private lazy var manyExpandedListSections: [CommandingSection] = fewExpandedListSections + [
+        CommandingSection(title: "Section 2", items:
```
harrieshin, line 65: the toggle didn't seem additive.  shouldn't it start with section 1?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -54,20 +54,35 @@ class BottomCommandingDemoController: UIViewController {
         CommandingItem(title: "Boolean Item " + String($0), image: homeImage, action: commandAction, isToggleable: true)
     }
 
-    private lazy var expandedListSections: [CommandingSection] = [
+    private lazy var fewExpandedListSections: [CommandingSection] = [
```
harrieshin, line 57: not crazy about the name. smallerSet vs largerSet? shortList vs longlist?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: It reuses the first section of hero commands from the fewExpandedListSections array.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: Haha you're totally right, this is awful.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -49,12 +49,12 @@ class BottomSheetDemoController: UIViewController {
         bottomSheetViewController?.isHidden.toggle()
     }
 
-    @objc private func fullScreenExpandedOffset() {
-        bottomSheetViewController?.expandedHeightFraction = 1.0
+    @objc private func fullScreenSheetContent() {
+        bottomSheetViewController?.preferredExpandedContentHeight = 0
```
harrieshin, line 53: bit counterintuitive for me that fullscreensheet is 0 contenthieght?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -322,6 +327,7 @@ open class BottomCommandingController: UIViewController {
             popoverContentViewController.view.addSubview(tableView)
             popoverContentViewController.modalPresentationStyle = .popover
             popoverContentViewController.popoverPresentationController?.sourceView = moreButtonView
+            popoverContentViewController.preferredContentSize = CGSize(width: 0, height: fittingTableViewHeight)
```
harrieshin, line 330: why zero width?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: I updated the naming.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I'll double check if this is documented, but providing zero width results in the system setting the width automatically, just like if I didn't provide a preferredContentSize at all.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -567,6 +590,10 @@ extension BottomCommandingController: UITableViewDataSource {
 }
 
 extension BottomCommandingController: UITableViewDelegate {
+    public func tableView(_ tableView: UITableView, heightForFooterInSection section: Int) -> CGFloat {
```
harrieshin, line 593: what is this for? not clear from the PR description

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: Any very large number will also result in a full screen sheet, which makes more sense. 0 is just a convenient distinct default number and the desired usage is really that when a client wants a full screen sheet, they just don't set the preferredExpandedContentHeight prop at all. Is there a behavior you'd prefer for the 0 case?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: UITableView adds an arbitrary 20px margin at the bottom after the last section. Doing this stops it from doing that and results in a more predictable table view sizing. I'll add a comment.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -469,28 +490,40 @@ public class BottomSheetController: UIViewController {
...
 
-        bottomSheetHeightConstraint.isActive = false
-        newConstraint.isActive = true
+            // Activate constraint which ties expandedContentView height to the layout guide height
+            preferredExpandedContentHeightConstraint.isActive = true
 
-        bottomSheetHeightConstraint = newConstraint
+        } else {
+            // Tie the sheet size to maxSheetHeightLayoutGuide.heightAnchor to make it full screen
+            preferredExpandedContentHeightConstraint.isActive = false
```
harrieshin, line 505: collapsing the bottomsheet case do not need to turn off both of these constraints?

lcapkovic: No, the sheet is rigid. The height doesn't change as you collapse it.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Turns out I can directly just set the height, so I'll do that.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
harrieshin: so then shouldn't it be done for only last section?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -91,6 +91,18 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// Preferred height of `expandedContentView`.
+    ///
+    /// The default value is 0, which results in a full screen sheet expansion.
+    @objc open var preferredExpandedContentHeight: CGFloat = 0 {
+        didSet {
+            if isViewLoaded {
+                updateSheetSizingConstraints()
```
harrieshin, line 100: either here or in side the function should we check if client is giving the wrong value like bigger than the screen size?

lcapkovic: The screen height limit is enforced through `bottomSheetView.heightAnchor.constraint(lessThanOrEqualTo: maxSheetHeightLayoutGuide.heightAnchor)`, so giving us an arbitrarily high number here is totally valid.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
anandrajeswaran: This is probably at least worth a comment in code as consumers would be using this as an example for integration. But as you mention if someone wants this behavior, the example they follow should just not set the property at all? Could add a comment saying that we're avoiding setting the property since the default value gives us full screen?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -118,25 +119,27 @@ open class BottomCommandingController: UIViewController {
...
-        }
-        bottomSheetController = nil
-        bottomBarView?.removeFromSuperview()
-        bottomBarView = nil
-
-        if traitCollection.horizontalSizeClass == .regular {
-            setupBottomBarLayout()
-        } else {
-            setupBottomSheetLayout()
+        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
```
anandrajeswaran, line 141: do we also want to check if previousTraitCollection is non-nil? Or is it intended to go through this code path if it's nil (is that the initial setup scenario?)

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -283,6 +287,7 @@ open class BottomCommandingController: UIViewController {
 
     private lazy var tableView: UITableView = {
         let tableView = UITableView(frame: .zero, style: .grouped)
+        tableView.contentInsetAdjustmentBehavior = .automatic
```
anandrajeswaran, line 290: This is the default value so no need to set?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -424,6 +430,23 @@ open class BottomCommandingController: UIViewController {
         }
     }
 
+    private func updateSheetExpandedContentHeight() {
+        if let bottomSheetController = bottomSheetController {
+            bottomSheetController.preferredExpandedContentHeight = fittingTableViewHeight + Constants.BottomSheet.expandedContentTopMargin
+        }
+    }
+
+    private var fittingTableViewHeight: CGFloat {
```
anandrajeswaran, line 439: Probably worth updating the name (and adding a comment) that this is an estimated height? "fitting" in UIKit parlance implies actually measured

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -157,16 +160,34 @@ public class BottomSheetController: UIViewController {
         overflowView.backgroundColor = Colors.NavigationBar.background
         view.addSubview(overflowView)
 
+        view.addLayoutGuide(maxSheetHeightLayoutGuide)
```
anandrajeswaran, line 163: What happens if someone changes collapsedConentHeight? Is there badness to accumulating layout guides?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -469,28 +490,40 @@ public class BottomSheetController: UIViewController {
...
 
-    private func generateBottomSheetHeightConstraint() -> NSLayoutConstraint {
-        // Fill view bounds, respecting the given height fraction
-        let constraint = bottomSheetView.heightAnchor.constraint(
-            equalTo: view.heightAnchor,
-            multiplier: expandedHeightFraction,
-            constant: view.safeAreaInsets.top - Constants.minimumTopExpandedPadding)
+    private lazy var preferredExpandedContentHeightConstraint: NSLayoutConstraint = {
+        let constraint = expandedContentView.heightAnchor.constraint(equalTo: preferredExpandedContentLayoutGuide.heightAnchor)
+        constraint.priority = .defaultHigh
```
anandrajeswaran, line 512: Any time we have a non-required constraint that is implicit ambiguity. It may be the right call (I think it is in this case), but is always worth a comment

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: We don't have section footers, so what we return here does reflect reality 😄

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Yeah all the height constraints have to be lower than required so AutoLayout can automatically enforce the max height constraint.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
anandrajeswaran: Sounds like a good case for a preconditionFailure then?

lcapkovic: Can you elaborate? This is where the delegate returns footer sizes. The table view calls this even when there are no section footers.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: From my testing the nil case doesn't seem to occur. Have you seen that happen? I'm not sure if we should be ignoring a potentially real trait change, even if previousTraitCollection is nil in some weird case after loadView.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
anandrajeswaran: Ok that piece I did not know - if a table view calls this even when there are no footers that seems like an Apple bug we should file?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
anandrajeswaran: But it's a var so this code path could get hit more than once. So then is there badness to calling addLayoutGuide multiple times with maxSheetHeightLayoutGuide. Or anything else in here where there is badness to potentially it getting repeated if collapsedContentHeight changes?

lcapkovic: This is in loadView. Are we looking at the same thing?

lcapkovic: Oh I see what happened. It's bad GitHub PR UI. When you expand the file you'll see this is in loadView.

anandrajeswaran: Oops my bad. Sorry for the confusion

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: Added a comment.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
harrieshin: could you add a comment here as well?

harrieshin (approved):

PR #623, opened 2021-07-01
Adopt the container pattern in `BottomCommandingController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

After initial integration prototyping and discussions with the team, we found that a container pattern would be a more suitable choice for `BottomCommandingController` than the current sibling pattern. The main disadvantage of the current pattern in Office is that you have to keep fixing the z-order whenever other siblings are added. The code that adds the siblings shouldn't necessarily be aware that the sheet overlay even exists. Container pattern enforces this implicitly and encourages small view controllers. We also have many other overlays in Office, and all of them use the container pattern.

Other advantages of the container pattern I found (we don't make use of these now)
- ownership of the main content view means we can apply more effects, similar to the 3D movement in full screen system sheets
- we can add additional safeAreaInsets on the child VC to account for the collapsed sheet. This would be great for child scroll views

Disadvantages
- the integration can be slightly more difficult, depending on the app's VC hierarchy
- deep nesting of view controllers when there's many overlays

### Verification
Sanity checks in the test app, verifying main content is still touchable and accessible.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/623)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -20,8 +20,7 @@ class BottomCommandingDemoController: UIViewController {
         optionTableView.delegate = self
         optionTableView.separatorStyle = .none
 
-        let bottomCommandingVC = BottomCommandingController()
-        bottomCommandingVC.contentViewController = optionTableViewController
+        let bottomCommandingVC = BottomCommandingController(with: optionTableViewController)
         bottomCommandingVC.heroItems = heroItems
         bottomCommandingVC.expandedListSections = expandedListSections
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -10,7 +10,8 @@ class BottomCommandingDemoController: UIViewController {
     override func loadView() {
         view = UIView()
 
-        let optionTableView = UITableView(frame: .zero, style: .plain)
+        let optionTableViewController = UITableViewController(style: .plain)
+        let optionTableView: UITableView = optionTableViewController.tableView
         optionTableView.translatesAutoresizingMaskIntoConstraints = false
         optionTableView.register(TableViewCell.self, forCellReuseIdentifier: TableViewCell.identifier)
         optionTableView.register(BooleanCell.self, forCellReuseIdentifier: BooleanCell.identifier)
@@ -18,9 +19,9 @@ class BottomCommandingDemoController: UIViewController {
         optionTableView.dataSource = self
         optionTableView.delegate = self
         optionTableView.separatorStyle = .none
-        view.addSubview(optionTableView)
 
         let bottomCommandingVC = BottomCommandingController()
+        bottomCommandingVC.contentViewController = optionTableViewController
         bottomCommandingVC.heroItems = heroItems
         bottomCommandingVC.expandedListSections = expandedListSections
 
@@ -31,10 +32,6 @@ class BottomCommandingDemoController: UIViewController {
         bottomCommandingController = bottomCommandingVC
 
         NSLayoutConstraint.activate([
-            optionTableView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            optionTableView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            optionTableView.topAnchor.constraint(equalTo: view.topAnchor),
-            optionTableView.bottomAnchor.constraint(equalTo: view.bottomAnchor),
             bottomCommandingVC.view.leadingAnchor.constraint(equalTo: view.leadingAnchor),
             bottomCommandingVC.view.trailingAnchor.constraint(equalTo: view.trailingAnchor),
             bottomCommandingVC.view.topAnchor.constraint(equalTo: view.topAnchor),
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -22,22 +22,23 @@ open class BottomCommandingController: UIViewController {
     /// View controller that will be displayed below the bottom commanding UI.
     @objc public var contentViewController: UIViewController? {
         didSet {
-            if let oldViewController = oldValue {
-                oldViewController.willMove(toParent: nil)
-                if oldViewController.isViewLoaded {
+            if isViewLoaded,
+               oldValue != contentViewController {
+                if let oldViewController = oldValue {
+                    oldViewController.willMove(toParent: nil)
                     oldViewController.view.removeFromSuperview()
+                    oldViewController.removeFromParent()
                 }
-                oldViewController.removeFromParent()
-            }
-            if let newContentViewController = contentViewController {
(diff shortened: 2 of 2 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -19,6 +19,29 @@ import UIKit
 @objc(MSFBottomCommandingController)
 open class BottomCommandingController: UIViewController {
 
+    /// View controller that will be displayed below the bottom commanding UI.
+    @objc public var contentViewController: UIViewController? {
```
harrieshin, line 23: should it be part of the init?

lcapkovic: For Office purposes we need the settable property, but I also added the init because we should probably encourage giving us the content VC right away.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -10,17 +10,17 @@ class BottomCommandingDemoController: UIViewController {
     override func loadView() {
         view = UIView()
 
-        let optionTableView = UITableView(frame: .zero, style: .plain)
+        let optionTableViewController = UITableViewController(style: .plain)
+        let optionTableView: UITableView = optionTableViewController.tableView
```
anandrajeswaran, line 14: Can the type here be inferred?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -19,6 +19,30 @@ import UIKit
...
+                }
+
+                if let newContentViewController = contentViewController {
+                    addChild(newContentViewController)
+                    if let rootCommandingView = rootCommandingView {
+                        let newContentView: UIView = newContentViewController.view
+                        view.insertSubview(newContentView, belowSubview: rootCommandingView)
+                        activateContentViewConstraints(for: newContentView)
+                    }
+                    newContentViewController.didMove(toParent: self)
```
anandrajeswaran, line 40: can factor this out to avoid duplicating between here and loadView?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -102,17 +126,38 @@ open class BottomCommandingController: UIViewController {
...
+
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
     // MARK: - View building and layout
 
     public override func loadView() {
-        view = BottomSheetPassthroughView()
```
anandrajeswaran, line 108: And we can't get rid of BottomSheetPassthroughView because BottomSheetController is still using it?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -102,17 +126,38 @@ open class BottomCommandingController: UIViewController {
...
         if traitCollection.horizontalSizeClass == .regular {
             setupBottomBarLayout()
         } else {
             setupBottomSheetLayout()
         }
+
+        if let contentViewController = contentViewController,
+           let rootCommandingView = rootCommandingView {
+            addChild(contentViewController)
+            let newContentView: UIView = contentViewController.view
```
anandrajeswaran, line 156: Probably just let it infer the UIView type?

anandrajeswaran (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: yup

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -19,6 +19,30 @@ import UIKit
 @objc(MSFBottomCommandingController)
 open class BottomCommandingController: UIViewController {
 
+    /// View controller that will be displayed below the bottom commanding UI.
+    @objc public var contentViewController: UIViewController? {
+        didSet {
+            if isViewLoaded,
+               oldValue != contentViewController {
```
markavitale, line 26: Would this be a little more idiomatic with a guard like this?
```
guard isViewLoaded, oldValue != contentViewController else { return }
```

I'm not 100% sure I'm connecting the braces properly so I might be missing something.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -434,6 +489,14 @@ open class BottomCommandingController: UIViewController {
 
     private var bottomSheetController: BottomSheetController?
 
+    private var rootCommandingView: UIView? {
+        if isInSheetMode {
```
markavitale, line 493: Thoughts on using ternary notation here?
```
return isInSheetMode ? bottomSheetController?.view : bottomBarView
```

markavitale (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
lcapkovic: The type has to be explicit for the implicitly unwrapped optional to work (UITableViewController.tableView is of type TableView! <- including that "!").

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Like the other comment, needed because of the implicitly unwrapped optional.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Thanks for the swifty suggestions!

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I like it.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: done

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -102,17 +121,33 @@ open class BottomCommandingController: UIViewController {
...
+    @available(*, unavailable)
+    public required init?(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
     // MARK: - View building and layout
 
     public override func loadView() {
-        view = BottomSheetPassthroughView()
+        view = UIView()
```
harrieshin, line 139: why is it no longer passthrough view?

harrieshin (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: With the container pattern we don't need it anymore, because we own the main content as well.

PR #624, opened 2021-07-06
make sure pill segmentedcontrol supports RTL

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When pillsegmentcontrol is shown RTL make sure the pill buttons shown from RTL.

### Verification

|segement_before | segment_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/624)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -442,6 +442,8 @@ open class SegmentedControl: UIControl {
 
         if style == .tabs {
             bottomSeparator.frame = CGRect(x: 0, y: frame.height - bottomSeparator.frame.height, width: frame.width, height: bottomSeparator.frame.height)
+        } else {
+            pillContainerView.flipSubviewsForRTL()
         }
 
         flipSubviewsForRTL()
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -443,6 +443,7 @@ open class SegmentedControl: UIControl {
         if style == .tabs {
             bottomSeparator.frame = CGRect(x: 0, y: frame.height - bottomSeparator.frame.height, width: frame.width, height: bottomSeparator.frame.height)
         } else {
+            // flipSubviewsForRTL only works on direct children subviews
             pillContainerView.flipSubviewsForRTL()
         }
 


ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -442,6 +442,8 @@ open class SegmentedControl: UIControl {
 
         if style == .tabs {
             bottomSeparator.frame = CGRect(x: 0, y: frame.height - bottomSeparator.frame.height, width: frame.width, height: bottomSeparator.frame.height)
+        } else {
+            pillContainerView.flipSubviewsForRTL()
```
rdeassis, line 446: Interesting... I'm wondering why the call to to flipSubviewsForRTL() in line 449 didn't take care of the pillContainerView too? (as the pillContainerView is added in the initializer in line [309](https://github.com/microsoft/fluentui-apple/blob/cfedd5edfda9d0fb34a30409746e5a174fb296b7/ios/FluentUI/SegmentedControl/SegmentedControl.swift#L309))

mischreiber: From a quick glance, I'd guess it's because `flipSubviewsForRTL` goes down exactly one subview, not recursively, so calling it on `self` as in line [449] is behaviorally different from calling it on `pillContainerView` directly.

That said, might be worth throwing a comment in here as to why that's the case, cuz I agree with @rdeassis that that's not immediately apparent, and the double invocation of `flipSubviewsForRTL` looks like a bug at first glance.

harrieshin: @mischrieber is correct. flipsubviewsforrtl only works for one layer of subviews

rdeassis (approved):

ios/FluentUI/SegmentedControl/SegmentedControl.swift
rdeassis: I took some time to look at the flipSubviewsForRTL implementation... which made me think of something else (maybe for a future change after a good perf impact analysis)

Should we consider changing the implementation of flipForRTL call flipSubviewsForRTL on itself? 
That way we would get the recursive implementation behavior.

harrieshin: i don't think it should always be recursive? we should be careful which ones we are actually flipping if we are using frame calculations. currently it is other way around right flipsubviewsforrtl is calling flipforrtl. but i guess the overall usuage for flipforrtl is pretty small and for simple uiview so if we want to make that change in the furture it isn't huge amount of work.

anandrajeswaran: I would also say that if we hit more serious RTL issues we should probably spend time migrating to autolayout from manual layout instead of making non-trivial patches to the manual layout hack.

mischreiber (approved):

PR #631, opened 2021-07-14
Basic support for Swift Package Manager

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

TL;DR: These changes make it possible to consume the Fluent library using Swift Package Manager. Once fully merged, it will be possible to simply point to `https://github.com/microsoft/fluentui-apple` in Xcode or your `Package.swift` file, and immediately start linking against the library. There are no changes to the existing builds/projects/outputs.

#### Okay, but what is this, really?

Before my change, iOS had two built outputs (framework + static lib) and macOS has one (framework), defined in each platform's respective Xcode project.

After my change, all this will remain exactly the same. Same outputs, with the same names, via the same mechanisms. Bundles, libs, frameworks: all unchanged.

What this change *does* do, however, is add a new, fully independent build system for FluentUI. Instead of emitting a library or framework, it emits a series of Swift module-based object files. There's no Xcode project or workspace, and no schemes. Instead, there's only a single defining file: `Package.swift`, that lives at the root of our project.

This file defines the relationship between modules of our code. There are four total targets defined:
1) The common resources, which live under `apple/Resources`
2) Our iOS-specific code and resources, which live under `ios/FluentUI` and depend on number 1
3) Our macOS-specific code and resources, which live under `macOS/FluentUI` and depend on number 2
4) An umbrella target, which conditionally imports and re-exports either 1+2 or 1+3 depending on platform

This parallel system will only be encountered by clients that choose to import our Swift package.

Much like Carthage builds, and unlike CocoaPods, this is currently an "all or nothing" approach. Allowing for partial packages would require significantly more work to modularize the framework. Early investigations suggested that a naïve attempt would result in an unacceptable package size increase, so that work is postponed for the time being.

Most of the changes I've made involved shuffling files around to better fit with the package requirements. In particular, some strings files needed to be moved for macOS. I'll be updating our localization pipeline to ensure that these are not seen as new assets in need of localization.

Additionally, this change does NOT affect our test apps at the moment. iOS will continue to consume the generated framework, and macOS will keep consuming the static lib + bundle produced by Xcode. Conversations are ongoing as to whether these should be updated, but it's beyond the scope of this particular change.

### Verification

I created a local test app (iOS + macOS), added my private branch's URL as the package target, and verified that both code and assets were successfully pulled down and integrated into the app.

| Xcode projects with frameworks and static lib/bundle | Xcode projects with frameworks and static lib/bundle, *plus* SPM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/631)


Package.swift
```diff
@@ -0,0 +1,68 @@
...
+
+let package = Package(
+    name: "FluentUI",
+    defaultLocalization: "en",
+    platforms: [
+        .iOS(.v13),
+        .macOS(.v10_14),
+    ],
+    products: [
+        .library(
```
Saadnajmi, line 13: Does this mean the default export for macOS would be a static library too? Currently downstream we consume it via a framework packed inside the NuGet package, but I have some local code testing how it would look to use it as a static library for macOS instead.

Package.swift
```diff
@@ -0,0 +1,68 @@
...
+            name: "FluentUI",
+            type: .static,
+            targets: [
+                "FluentUI"
+            ]
+        )
+    ],
```
Saadnajmi, line 20: Does this mean the default export for macOS would be a static library too? Currently downstream we consume it via a framework packed inside the NuGet package, but I have some local code testing how it would look to use it as a static library for macOS instead.

Package.swift
```diff
@@ -0,0 +1,68 @@
...
+                "Info.plist",
+                "Navigation/Navigation.resources.xcfilelist",
+                "Notification/Notification.resources.xcfilelist",
+                "Other Cells/OtherCells.resources.xcfilelist",
+                "Resources/Localization/CultureMapping.json",
+                "Table View/TableView.resources.xcfilelist",
+                "Tooltip/Tooltip.resources.xcfilelist",
+                "Tooltip/Tooltip.resources.xcfilelist",
+                "TwoLineTitleView/TwoLineTitleView.resources.xcfilelist",
+            ]
```
Saadnajmi, line 55: Why do we have to exclude these? (For my own knowledge's sake)

Package.swift
mischreiber: Yes, but this only applies to clients that explicitly use SPM to consume the package. The existing framework won't be affected, since our official pipeline will continue to use our existing Xcode projects/schemes/etc.

Package.swift
mischreiber: Short answer: because Xcode gives a warning if you don't 🤠

Longer answer: AFAIK the Swift packager wants to understand every file that it sees in its paths, and it doesn't know what to do with these. They're not source and they're not resources, so it gives a warning saying "I don't know what this is". Putting those files in the exclusion list lets us skip over trying to parse them.

Remember: unlike an Xcode project, where files are opt-in (by manually adding them to the project), in SPM files are opt-out, and are considered part of the package by default unless you explicitly exclude them.

Saadnajmi: Thanks for the explanation!

harrieshin:
can you add explanation on the PR description why we had to change the bundle name?

public/FluentUI.swift
```diff
@@ -0,0 +1,10 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+#if os(iOS)
+@_exported import FluentUI_ios
+#elseif os(macOS)
+@_exported import FluentUI_macos
+ #endif
```
harrieshin, line 10: is there a weird indentation here?

macos/FluentUI/Core/FluentUIResources.swift
```diff
@@ -12,10 +15,23 @@ public class FluentUIResources: NSObject {
 
 	/// The resource bundle contained within the FluentUI framework
 	@objc public static let resourceBundle: Bundle = {
+		#if SWIFT_PACKAGE
+		return Bundle.module
+		#else
 		guard let url = bundle.resourceURL?.appendingPathComponent("FluentUIResources-macos.bundle", isDirectory: true), let bundle = Bundle(url: url) else {
```
harrieshin, line 21: even after the migraton, the cocoapod path can still use fluentuiresource-macos.bundle as is?

ios/FluentUI/Core/FluentUIFramework.swift
```diff
@@ -32,10 +35,21 @@ public extension Colors {
 public class FluentUIFramework: NSObject {
     @objc public static var bundle: Bundle { return Bundle(for: self) }
     @objc public static let resourceBundle: Bundle = {
+        #if SWIFT_PACKAGE
+        return Bundle.module
+        #else
         guard let url = bundle.resourceURL?.appendingPathComponent("FluentUIResources-ios.bundle", isDirectory: true), let bundle = Bundle(url: url) else {
```
harrieshin, line 41: this -ios.bundle is still ok? any update we need to do for nugetpackage copying scripts if the bundle name is different? so devmain should know?

ios/FluentUI/Core/FluentUIFramework.swift
```diff
@@ -32,10 +35,21 @@ public extension Colors {
...
             preconditionFailure("FluentUI resource bundle is not found")
         }
         return bundle
+        #endif
+    }()
+    @objc public static let colorsBundle: Bundle = {
+        #if SWIFT_PACKAGE
+        return SharedResources.colorsBundle
```
harrieshin, line 49: why do we have a separate color bundle now?

macos/FluentUI/Core/Colors.swift
```diff
@@ -149,7 +149,7 @@ public final class Colors: NSObject {
 		case presenceUnknown
 
 		public var color: NSColor {
-			if let fluentColor = NSColor(named: "FluentColors/" + self.name, bundle: FluentUIResources.resourceBundle) {
+			if let fluentColor = NSColor(named: "FluentColors/" + self.name, bundle: FluentUIResources.colorsBundle) {
```
harrieshin, line 152: if we have color bundle itself, then should the foldername "fluentColors" is even needed?

public/FluentUI.swift
mischreiber: Yes, a pesky space snuck in. Good eye! 👀

ios/FluentUI/Core/FluentUIFramework.swift
mischreiber: We need a separate color bundle in SPM because the colors live outside the path of iOS or macOS modules.

Basically, SPM cares a lot about file directories when defining modules. The roots of the iOS and macOS modules are under the `ios/` and `macos/` directories, respectively, and SPM does not allow you to know about anything that is not under the root of your directory. Since colors live in a different root so they can be shared (under `apple/`), that has to become a separate module.

macos/FluentUI/Core/Colors.swift
mischreiber: It's still necessary because the existing builds for static lib + framework haven't changed. They still combine asset catalogs like before (which is why I need so many `#if` statements to keep the behavior separate).

If we want to change those too, then we could. But my goal for this PR was to have minimum risk for existing consumers and pipelines.

ios/FluentUI/Core/FluentUIFramework.swift
mischreiber: Nothing changed for existing pipelines, so nothing should need to change. Nuget builds will not be based off SPM right now.

macos/FluentUI/Core/FluentUIResources.swift
mischreiber: Good catch, I need to update the podspec to include the new resources path.

MicrosoftFluentUI.podspec
```diff
@@ -326,7 +326,7 @@ xcodebuild ${XCODEBUILDPARAMS} -project ${PROJECT_FILE_PATH} -target "MicrosoftF
   s.subspec 'Core_mac' do |core_mac|
     core_mac.platform = :osx
     core_mac.resource_bundle = { 'FluentUIResources-macos' => ["apple/Resources/**/*.{json,xcassets}",
-                                                             "macos/FluentUIResources-macos/**/*.{storyboard,xib,xcassets,strings,stringsdict}"] }
+                                                             "macos/FluentUI/**/*.{storyboard,xib,xcassets,strings,stringsdict}"] }
```
harrieshin, line 329: i see so macoS we need to update the podspec but not for ios?

harrieshin (approved):

harrieshin:
because of your mac bundle name changes, can you also check under .ado folder the fluentui-apple-build-zip.yml update?

MicrosoftFluentUI.podspec
mischreiber: Correct! iOS was already set up the way it needed to be. Seems like someone moved it in the past?

Package.swift
```diff
@@ -0,0 +1,68 @@
...
+            path: "apple"
+        ),
+        .target(
+            name: "FluentUI_ios",
+            dependencies: [
+                .target(name: "FluentUIResources")
+            ],
+            path: "ios/FluentUI",
+            exclude: [
+                "Avatar/Avatar.resources.xcfilelist",
```
rdeassis, line 41: I really wish they would support wildcards here.

ios/FluentUI/Core/FluentUIFramework.swift
```diff
@@ -3,6 +3,9 @@
 //  Licensed under the MIT License.
 //
 
+#if SWIFT_PACKAGE
```
rdeassis, line 6: Suggestion: maybe add a comment to clarify that SWIFT_PACKAGE is defined when the code is being built by Swift Package Manager?

rdeassis (approved):

Package.swift
mischreiber: Me too 😵‍💫

PR #632, opened 2021-07-14
Add sheet sizing helpers to `BottomCommandingController` and `BottomSheetController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There are two bottom commanding use cases we need to support:
1. Attaching client-owned views to the bottom chrome.
2. Setting scroll view insets to clear the collapsed bottom chrome.

To support (1) I added a `layoutGuide` which always covers the visible portion of the commanding chrome. This allows clients to respond to dynamic sheet / bar position changes in Auto Layout.

For (2) I added `collapsedSheetHeight` and `collapsedChromeHeight` properties to `BottomSheetController` and `BottomCommandingController` which return current height of the collapsed bottom chrome inside the safe area. This allows clients to easily adjust content insets of their scroll views to clear the sheet / bar. Both of these also get delegate functions which notify about any changes.

### Verification

Updated the demo apps to use the collapsed heights to update content insets. Verified these are updated based on the delegate calls.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/632)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -305,9 +305,9 @@ extension BottomCommandingDemoController: UITableViewDataSource {
 }
 
 extension BottomCommandingDemoController: BottomCommandingControllerDelegate {
-    func bottomCommandingControllerCollapsedChromeHeightDidChange(_ bottomCommandingController: BottomCommandingController) {
+    func bottomCommandingControllerCollapsedHeightInSafeAreaDidChange(_ bottomCommandingController: BottomCommandingController) {
         if let tableView = mainTableViewController?.tableView {
-            tableView.contentInset.bottom = bottomCommandingController.collapsedChromeHeight
+            tableView.contentInset.bottom = bottomCommandingController.collapsedHeightInSafeArea
         }
     }
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -11,6 +11,8 @@ class BottomCommandingDemoController: UIViewController {
         view = UIView()
 
         let optionTableViewController = UITableViewController(style: .plain)
+        mainTableViewController = optionTableViewController
+
         let optionTableView: UITableView = optionTableViewController.tableView
         optionTableView.translatesAutoresizingMaskIntoConstraints = false
         optionTableView.register(TableViewCell.self, forCellReuseIdentifier: TableViewCell.identifier)
@@ -23,6 +25,7 @@ class BottomCommandingDemoController: UIViewController {
         let bottomCommandingVC = BottomCommandingController(with: optionTableViewController)
         bottomCommandingVC.heroItems = heroItems
         bottomCommandingVC.expandedListSections = shortCommandSectionList
+        bottomCommandingVC.delegate = self
 
         addChild(bottomCommandingVC)
         view.addSubview(bottomCommandingVC.view)
@@ -37,6 +40,8 @@ class BottomCommandingDemoController: UIViewController {
             bottomCommandingVC.view.bottomAnchor.constraint(equalTo: view.bottomAnchor)
         ])
     }
+
+    private var mainTableViewController: UITableViewController?
 
     private lazy var heroItems: [CommandingItem] = {
         return Array(1...4).map {
@@ -296,5 +301,13 @@ extension BottomCommandingDemoController: UITableViewDataSource {
         }
 
         return UITableViewCell()
+    }
+}
+
+extension BottomCommandingDemoController: BottomCommandingControllerDelegate {
+    func bottomCommandingControllerCollapsedChromeHeightDidChange(_ bottomCommandingController: BottomCommandingController) {
+        if let tableView = mainTableViewController?.tableView {
+            tableView.contentInset.bottom = bottomCommandingController.collapsedChromeHeight
+        }
     }
 }
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -133,6 +133,7 @@ open class BottomCommandingController: UIViewController {
     @objc public let commandingLayoutGuide = UILayoutGuide()
 
     /// Height of the portion of the collapsed commanding UI that's in the safe area.
+    /// When using the bottom bar style, this will include the entire height of the bottom bar.
     ///
     /// Valid after the root view is loaded.
     ///
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -5,6 +5,13 @@
 
 import UIKit
 
+@objc(MSFBottomCommandingControllerDelegate)
+public protocol BottomCommandingControllerDelegate: AnyObject {
+
+    /// Called when `collapsedChromeHeight` changes.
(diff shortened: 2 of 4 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -11,6 +11,8 @@ class BottomCommandingDemoController: UIViewController {
         view = UIView()
 
         let optionTableViewController = UITableViewController(style: .plain)
```
harrieshin, line 13: why not just use maintableviewcontroller directly?

lcapkovic: Because the property is an optional and it's a pain dealing with optionals when configuring the view controller.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -122,6 +129,27 @@ open class BottomCommandingController: UIViewController {
         }
     }
 
+    /// A layout guide that covers the on-screen portion of the current commanding view.
+    @objc public let layoutGuide = UILayoutGuide()
```
harrieshin, line 133: is this layoutguide only applicaable to bottomsheetviewcontroller?

lcapkovic: It will wrap both the bar and the sheet, depending on screen size.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -122,6 +129,27 @@ open class BottomCommandingController: UIViewController {
...
+
+    /// Height of the portion of the collapsed bottom chrome that's in the safe area.
+    ///
+    /// Valid after the root view is loaded.
+    ///
+    /// Use this to adjust `contentInsets` on your scroll views. This height won't change when the commanding UI is hidden or expanded.
+    @objc public var collapsedChromeHeight: CGFloat {
```
harrieshin, line 140: couple of things: i don't think this variable name is super clear for other people what portion of the UI this is about for both iPad and iphone cases. Also, is there a need for clients that they need to know the collapsedchromeheight even in expanded cases? it can't be the "current content height" of the control?

lcapkovic: I used the word chrome to cover both the sheet and the bar. Do you have something in mind that would be more clear? 

As for the second point, in the sheet case, I specifically want this to only give the collapsed height, because the purpose of this property is to set scroll view bottom inset and it's not desirable for that to keep changing as the sheet expands. If the clients need to dynamically track where the bottom UI is right now, that's what the layout guide is for.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
anandrajeswaran: Can we be a bit more specific in the name? commandingLayoutGuide?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
anandrajeswaran: Have we used the term "chrome" elsewhere? That doesn't feel like a very industry standard term. I like the idea of collapsedCommandingHeight

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -489,6 +537,21 @@ open class BottomCommandingController: UIViewController {
...
+    private func makeBottomBarLayoutGuideConstraints(with bottomBarView: UIView) -> [NSLayoutConstraint] {
+        let requiredConstraints = [
+            layoutGuide.leadingAnchor.constraint(equalTo: bottomBarView.leadingAnchor),
+            layoutGuide.trailingAnchor.constraint(equalTo: bottomBarView.trailingAnchor),
+            layoutGuide.bottomAnchor.constraint(equalTo: view.bottomAnchor),
+            layoutGuide.topAnchor.constraint(lessThanOrEqualTo: view.bottomAnchor)
+        ]
+
+        // bottomBarView will go off-screen when it's hidden, so this constraint is not always required.
+        let breakableConstraint = layoutGuide.topAnchor.constraint(equalTo: bottomBarView.topAnchor)
```
anandrajeswaran, line 549: can we be more specific than "breakableConstraint". Breakable says something about its priority, but doesn't say which constraint it is

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -541,6 +608,9 @@ open class BottomCommandingController: UIViewController {
 
     private var bottomBarHidingAnimator: UIViewPropertyAnimator?
 
+    // Constraints attaching self.layoutGuide to the current commanding surface (bar or a sheet)
+    private var layoutGuideConstraints: [NSLayoutConstraint] = [NSLayoutConstraint]()
```
anandrajeswaran, line 612: Using this syntax is a bit duplicate with providing the type hint. Could just be = []?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -9,7 +9,10 @@ import UIKit
 public protocol BottomSheetControllerDelegate: AnyObject {
 
     /// Called after the sheet moved to a new expansion state
-    @objc optional func bottomSheetControllerDidMove(to expansionState: BottomSheetExpansionState)
+    @objc optional func bottomSheetController(_ bottomSheetController: BottomSheetController, didMoveTo expansionState: BottomSheetExpansionState)
+
+    /// Called when `collapsedSheetHeight` changes.
```
anandrajeswaran, line 14: Probably more to add in the comments here about the subtleties of how this behaves in expanded and hidden states?

anandrajeswaran (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I think the linter will complain but I will try :D

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
harrieshin: collapseOffsetFromSafeArea? collapsedHeightFromSafeArea?

lcapkovic: Or collapsedHeightInSafeArea? This only has the part that's in the safe area.

harrieshin: yup, I think that is slightly better even tho it is bit hard to explain for iPad case.

lcapkovic: Yeah that's a tough one as there's no concept of collapsing in the bottom bar. I think it's acceptable because the control client will know about the sheet representation and how it relates to that.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Linter is smarter than I thought. Done!

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Added more documentation.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I included more detail in the documentation to explain this will return the bottom bar height as well.

harrieshin (approved):

PR #640, opened 2021-07-20
Fix segmented control VO for ColorDemoController

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

VoiceOver focus lands on the "Back" button after changing selected segment. Fix in #619 got lost in the later commits of my previous change. Refix this bug by reusing one theme window instead of creating new ones and pushing view controllers when switching segmented controls.

### Verification

Tested on iPad and ensured voiceover focus stayed on newly selected segment rather than jumping to "Back"

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/640)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -155,10 +156,12 @@ class ColorDemoController: UIViewController {
                 colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
             }
 
-            let window = colorProviderThemes[segmentedControl.selectedSegmentIndex].demoColorTheme.window
+            let window = colorProviderThemes[selectedSegmentIndex].demoColorTheme.window
             if let colorProvider = window as? ColorProviding {
                 Colors.setProvider(provider: colorProvider, for: window)
             }
+
+            UIAccessibility.post(notification: .screenChanged, argument: segmentedControl.segmentView(at: selectedSegmentIndex))
```
harrieshin, line 164: is it a screen change? or a layout chnage?

harrieshin: also add comments why we are adding this hack

harrieshin:
"Initial fix for this bug " instead of this phrase, write out your commit number.

ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemeWindows.swift
```diff
@@ -6,70 +6,74 @@
...
-
-    func primaryTint20Color(for window: UIWindow) -> UIColor? {
-        Colors.Palette.communicationBlueTint20.color
-    }
-
-    func primaryTint30Color(for window: UIWindow) -> UIColor? {
-        Colors.Palette.communicationBlueTint30.color
-    }
+class DemoColorThemeWindow: UIWindow, ColorProviding {
+    public static var theme: NSString = "Default"
```
rdeassis, line 10: We shouldn't merge all the themes into one. The fundamental concept of FluentUI is that we have a specific theme per window. The demo app serves as an example to be followed and we don't want client apps creating a theme that changes colors instead of setting one per window.

ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemeWindows.swift
```diff
@@ -6,70 +6,74 @@
...
-
-    func primaryShade30Color(for window: UIWindow) -> UIColor? {
-        Colors.Palette.communicationBlueShade30.color
-    }
-}
-
-class DemoColorThemeGreenWindow: UIWindow, ColorProviding {
     func primaryColor(for window: UIWindow) -> UIColor? {
-        return UIColor(named: "Colors/DemoPrimaryColor")
+        if DemoColorThemeWindow.theme == "Green" {
```
rdeassis, line 17: This is a typical pattern of 2 classes with different implementations of the same interface. We shouldn't break this pattern.

rdeassis (changes requested):

ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemeWindows.swift
jeaniehuynh: The cause of this bug is because of the different windows/pushing view controllers vs. just updating the table. I tried to just force accessibility to behave properly by posting accessibility notifications in my first commit of #619, but @harrieshin wanted to update the table instead of this hackier route. Harrie, thoughts here?

harrieshin: can you change the theme color of the window without pushing a new view controller on the stack?

jeaniehuynh: That's what I've done in this change. If we keep the original implementation (two different theme windows), I don't think so because changing the theme color would require an entirely new UIWindow and the view controller would have to be pushed onto it somehow.

harrieshin: there isn't a way to update the windows colors by having some sort of property so we can still do just updating the tableview based on "right" setting but still keep the infrastructure to have multiple theme per window?

jeaniehuynh: If you look at the implementation of the theme windows, the colors are returned as functions of the theme. I don't think it's possible to keep this implementation and show a different color from the window's theme. Even if it were possible, that would mean showing a green theme on a DemoColorThemeDefaultWindow or a default theme on a DemoColorThemeGreenWindow, which doesn't seem right either.

harrieshin: right so why can't theme be not static class variable and per instance of windows? and the colors that returns is still based on theme string(or enum)? which will still allow different windows have different themes in iPad.

jeaniehuynh: not sure if I understand--if we have different windows with different themes, doesn't that bring us back to having to change windows when user selects a different theme, which would reset the voiceover focus to the back button (initial bug)?

harrieshin: can we not change the windows but just refresh the view to grab the new color from the provider?

jeaniehuynh: the color provider _**is**_ the window, so if we don't change the window, we can't change the color.

harrieshin: the color provider doesn't have to be the window itself. color provider is just a protocol. any object can a color provider.

jeaniehuynh: but if we make the view controller the color provider for example, it wouldn't make sense for a green themed view controller to be presented on a DemoColorTheme**Default**Window, right?

harrieshin: I don't think the colordemoviewcontroller should be the color provider. color provider should be its own object and depending on the selection of the segment control, couldn't we just set appropriate provider of the existing window in the Colors.setProvider(provider:window:)? I still don't understand why you would set green theme provider in default case?

ios/FluentUI/Core/Colors.swift
```diff
@@ -492,7 +496,7 @@ public final class Colors: NSObject {
         return colorProvidersMap.object(forKey: window)?.primaryShade30Color(for: window) ?? FallbackThemeColor.primaryShade30
     }
 
-    private static var colorProvidersMap = NSMapTable<UIWindow, ColorProviding>(keyOptions: .weakMemory, valueOptions: .weakMemory)
+    private static var colorProvidersMap = NSMapTable<UIWindow, ColorProviding>()
```
harrieshin, line 499: no we shouldn't change the weak memory allocation here. The demo app should be responsible of holding on to provider instances.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -146,30 +142,28 @@ class ColorDemoController: UIViewController {
...
         if let segmentedControl = sender as? SegmentedControl {
-            let windowType = colorProviderThemes[segmentedControl.selectedSegmentIndex].demoColorTheme.windowType
-            let colorThemeHost = view.window?.windowScene?.delegate as? ColorThemeHosting
-            ColorDemoController.themeWindowType = windowType
-
-            if let navigationController = navigationController {
-                navigationController.popViewController(animated: false)
-                colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
+            let selectedSegmentIndex = segmentedControl.selectedSegmentIndex
+            let provider = colorProviderThemes[selectedSegmentIndex].demoColorTheme.window
```
harrieshin, line 146: why is provider still a window?

jeaniehuynh:
I renamed the DemoColorThemeWindows.swift file to DemoColorTheme.swift and the checks are failing now because it's still looking for the original file name. How can I fix it?

harrieshin:
> I renamed the DemoColorThemeWindows.swift file to DemoColorTheme.swift and the checks are failing now because it's still looking for the original file name. How can I fix it?

Your Xcode project file changes doesn't seem to be part of PR?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -146,30 +141,28 @@ class ColorDemoController: UIViewController {
...
+                // If provider doesn't conform to ColorProviding, fallback on default theme
+                DemoListViewController.updateColorProviderFor(window: window!, provider: DemoColorDefaultTheme() as ColorProviding)
             }
 
-            let window = colorProviderThemes[segmentedControl.selectedSegmentIndex].demoColorTheme.window
-            if let colorProvider = window as? ColorProviding {
-                Colors.setProvider(provider: colorProvider, for: window)
-            }
+            tableView.reloadData()
+            segmentedControl.didMoveToWindow()
```
harrieshin, line 157: why do you need didmovetowindow?

harrieshin: instead of the OS api, calling updateWindowSpecificColors would be more appropriate.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -146,30 +141,28 @@ class ColorDemoController: UIViewController {
...
-
-            if let navigationController = navigationController {
-                navigationController.popViewController(animated: false)
-                colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
+            let selectedSegmentIndex = segmentedControl.selectedSegmentIndex
+            let provider = colorProviderThemes[selectedSegmentIndex].demoColorTheme.provider
+            let window = self.view.window
+            ColorDemoController.providerType = colorProviderThemes[selectedSegmentIndex].demoColorTheme.providerType
+
+            if let provider = provider as? ColorProviding {
```
harrieshin, line 149: this can be cleaned up as let provider = provider ?? DemoColorDefaultTheme()  so we reduce the duplicated code? or lets just get rid of none case if you are going to do it this way? the intent of the "none" segment control is that the client doesn't give any provider so the colors are all fall back. in your previous iteration, you actually had a remove provider of the dictionary?

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -8,22 +8,25 @@ import UIKit
...
 class DemoListViewController: UITableViewController {
 
-    static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
-            Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
-        } else {
-            FluentUIFramework.initializeAppearance(with: Colors.primary(for: window))
-        }
+    private static var provider: ColorProviding = DemoColorDefaultTheme() as ColorProviding
```
harrieshin, line 11: no need for being static right?

jeaniehuynh: provider variable is used in addDemoListTo and updateColorProviderFor, which are static methods. You cannot access non-static stuff in a static method.

harrieshin: thanks for the explanation. now that you have change how window connects to demo controller, can we remove all these unnecessary static functions as much as possible?

ios/FluentUI.Demo/FluentUI.Demo/SceneDelegate.swift
```diff
@@ -5,18 +5,13 @@
...
-        updateToWindowWith(type: DemoColorThemeDefaultWindow.self, pushing: nil)
-    }
-
-    func updateToWindowWith(type: UIWindow.Type, pushing viewController: UIViewController?) {
-        let newWindow = type.init(windowScene: windowScene!)
-        DemoListViewController.addDemoListTo(window: newWindow, pushing: viewController)
-        window = newWindow
-
+        window = UIWindow.init(frame: UIScreen.main.bounds)
+        DemoListViewController.init().addDemoListTo(window: window!)
```
harrieshin, line 14: don't use ! (force unwrapping optionals) becausse it could crash run time.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -252,6 +245,6 @@ struct DemoColorSection {
 }
 
 private struct DemoColorTheme {
-    var window: UIWindow
-    var windowType: UIWindow.Type
+    var provider: NSObject
+    var providerType: NSObject.Type
```
mischreiber, line 249: Hmm... I know this is kind of a carry-over from the older version, but I really don't like the idea of passing an object type around for switching behavior, especially when we also have an instance of the object itself.

Here's an idea: instead of this struct, why don't we define a new enum for the different themes? I'd put it in `ColorDemoThemes.swift` and have individual cases for each of the three themes.

From there, you can also provide custom properties on the enum itself (Swift is cool like that), including a getter that would return the appropriate provider for the given value. Then you could do away with both the struct and the `colorProviderThemes` array, and instead use the fact that it's an interable enum to do indexing and lookup. Something like this:

```swift
enum DemoColorTheme: CaseIterable {
    case `default`
    case green
    case none

    var name: String {
        switch self {
        case .default:
            return "Default"
        case .green:
            return "Green"
        case .none:
            return "None"
        }
    }

    var provider: ColorProviding? {
        switch self {
        case .default:
            return DemoColorDefaultTheme()
        case .green:
            return DemoColorGreenTheme()
        case .none:
            return nil
        }
    }
}
```

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -8,22 +8,27 @@ import UIKit
...
 class DemoListViewController: UITableViewController {
 
-    static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
-            Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
-        } else {
-            FluentUIFramework.initializeAppearance(with: Colors.primary(for: window))
-        }
+    private var provider: ColorProviding = DemoColorDefaultTheme() as ColorProviding
```
mischreiber, line 11: This cast shouldn't be necessary; the assignment should automatically pick up that `DemoColorDefaultTheme()` is a `ColorProvider`.

mischreiber (changes requested):

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -8,22 +8,27 @@ import UIKit
...
 class DemoListViewController: UITableViewController {
 
-    static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
-            Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
-        } else {
-            FluentUIFramework.initializeAppearance(with: Colors.primary(for: window))
-        }
+    private var provider: ColorProviding = DemoColorDefaultTheme()
```
harrieshin, line 11: you don't need these 2 anymore. you just need the enum ?

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -8,22 +8,27 @@ import UIKit
...
 
-    static func addDemoListTo(window: UIWindow, pushing viewController: UIViewController?) {
-        if let colorProvider = window as? ColorProviding, let primaryColor = colorProvider.primaryColor(for: window) {
-            Colors.setProvider(provider: colorProvider, for: window)
-            FluentUIFramework.initializeAppearance(with: primaryColor, whenContainedInInstancesOf: [type(of: window)])
-        } else {
-            FluentUIFramework.initializeAppearance(with: Colors.primary(for: window))
-        }
+    private var provider: ColorProviding? = DemoColorTheme.default.provider
+    public var themeName: String = DemoColorTheme.default.name
```
mischreiber, line 12: Any reason this needs to be public? It'd be great if we could make this private.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -117,8 +113,8 @@ class ColorDemoController: UIViewController {
         view.addSubview(stackView)
         view.backgroundColor = Colors.NavigationBar.background
 
-        let window = ColorDemoController.themeWindowType
-        segmentedControl.selectedSegmentIndex = colorProviderThemes.firstIndex(where: { $0.demoColorTheme.windowType == window }) ?? 0
+        let demoListViewController = navigationController?.viewControllers.first as? DemoListViewController
+        segmentedControl.selectedSegmentIndex = DemoColorTheme.allCases.firstIndex(where: { $0.name == demoListViewController!.themeName }) ?? 0
```
mischreiber, line 117: Here's another chance to get rid of a force unwrap (`!`). If you use an `if let` when assigning `demoListViewController` then you'll be able to use it without unwrapping, and just handle the default case in an `else` block afterwards.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -139,36 +135,31 @@ class ColorDemoController: UIViewController {
...
-            if let navigationController = navigationController {
-                navigationController.popViewController(animated: false)
-                colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
+            let selectedSegmentIndex = segmentedControl.selectedSegmentIndex
+            let demoColorTheme = DemoColorTheme.allCases[selectedSegmentIndex]
+            let window = self.view.window
+
+            if let provider = demoColorTheme.provider {
+                let demoListViewController = navigationController?.viewControllers.first as? DemoListViewController
+                demoListViewController!.updateColorProviderFor(window: window!, provider: provider, themeName: demoColorTheme.name)
```
mischreiber, line 151: Another force unwrap, similar to the one before. Make this assignment part of the previous `if let` condition, and you should be good to go.

mischreiber: Actually I see two here. There's no guarantee that any view is currently hosted in a window, so `window!` is never safe. Since everything here depends on `window` existing, it might make sense to put all of this provider code in an `if let window = self.view.window` since none of it will work otherwise.

mischreiber (changes requested):

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
jeaniehuynh: It's public because it's accessed/set in the ColorDemoController when the theme is changed

mischreiber (approved):

harrieshin (approved):

harrieshin (approved):

ios/FluentUI/Core/Colors.swift
```diff
@@ -457,6 +457,10 @@ public final class Colors: NSObject {
         colorProvidersMap.setObject(provider, forKey: window)
     }
 
+    @objc public static func removeProvider(for window: UIWindow) {
```
rdeassis, line 460: Do we want to make this objc public? What would be the purpose for that?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -139,36 +138,26 @@ class ColorDemoController: UIViewController {
...
-            ColorDemoController.themeWindowType = windowType
-
-            if let navigationController = navigationController {
-                navigationController.popViewController(animated: false)
-                colorThemeHost?.updateToWindowWith(type: windowType, pushing: self)
+        if let segmentedControl = sender as? SegmentedControl, let window = self.view.window {
+            if let demoListViewController = navigationController?.viewControllers.first as? DemoListViewController {
+                demoListViewController.updateColorProviderFor(window: window, theme: DemoColorTheme.allCases[segmentedControl.selectedSegmentIndex])
+            } else {
+                // If provider doesn't conform to ColorProviding, remove mapping from dictionary and return fallback colors
```
rdeassis, line 151: How would the provider not conform to ColorProviding? Not sure I understand the comment/logic here?

jeaniehuynh: Ah, I forgot to update the comment, but now that I'm thinking about it, we don't need this anymore since we made the mapping have weak memory allocation again.

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -8,22 +8,30 @@ import UIKit
...
         let navigationController = UINavigationController(rootViewController: demoListViewController)
         window.rootViewController = navigationController
         window.makeKeyAndVisible()
+    }
 
-        if let viewController = viewController {
-            navigationController.pushViewController(viewController, animated: false)
+    func updateColorProviderFor(window: UIWindow, theme: DemoColorTheme) {
+        self.theme = theme
+        if let primaryColor = provider?.primaryColor(for: window), let provider = self.provider {
```
rdeassis, line 30: if you move the "let provider = self.provider" to be the first one, you won't need to use the optional "provider?" in the other statement, right?

`if let provider = self.provider, let primaryColor = provider.primaryColor(for: window)  {`

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -117,8 +113,11 @@ class ColorDemoController: UIViewController {
         view.addSubview(stackView)
         view.backgroundColor = Colors.NavigationBar.background
 
-        let window = ColorDemoController.themeWindowType
-        segmentedControl.selectedSegmentIndex = colorProviderThemes.firstIndex(where: { $0.demoColorTheme.windowType == window }) ?? 0
+        if let demoListViewController = navigationController?.viewControllers.first as? DemoListViewController {
```
rdeassis, line 116: nit: I think you can create a computed property that does this:

`demoListViewController = navigationController?.viewControllers.first as? DemoListViewController`

It's being used in more than one place and could be error prone in case this logic needs to be updated in the future.

```
private var currentDemoListViewController: DemoListViewController? {
    return navigationController?.viewControllers.first as? DemoListViewController
}
```

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -651,7 +651,7 @@ open class SegmentedControl: UIControl {
         }
     }
 
-    private func updateWindowSpecificColors() {
+    public func updateWindowSpecificColors() {
```
rdeassis, line 654: It doesn't seem like the client code would need to have access to this in any scenario other than the ColorDemoController. Is there a better way to trigger this update without making this func public?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
rdeassis: That's great! Thanks for investigating and removing it!

ios/FluentUI/SegmentedControl/SegmentedControl.swift
jeaniehuynh: Calling segmentedControl.didMoveToWindow() works too, but Harrie thought it'd be better to call updateWindowSpecificColors

rdeassis: I believe the price of making that a public API is higher. @harrieshin are you okay with going with the approach of calling didMoveToWindow()?

harrieshin: in this case, i don't think calling didMoveToWindow() is the right thing. (1) it is something that iOS usually calls the view not client code calling client code like invalidatelayout(). (2) in reality, it didn't move the window. we are relying on bilateral effect.

rdeassis: Yes. I agree that it doesn't seem like the right thing to do. When I asked this question one of the ideas I had in mind were more along the lines of setNeedsLayout.

jeaniehuynh: After looking at the code, didMoveToWindow() works because it's overridden in SegmentedControl.swift and it's calling the updateWindowSpecificColors() method.  I tried layoutSubviews, setNeedsLayout, setNeedsDisplay, and reloadInputViews before, but none of those work unfortunately.

PR #641, opened 2021-07-21
Add state tracking callbacks to `BottomCommandingController` and `BottomSheetController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Clients need to be notified about various state changes and what interactions triggered them for telemetry purposes. This PR adds delegate callbacks to notify about:
Sheet state changes
Bottom bar popover presentation / dismiss

Each of these send additional information about what user interaction triggered the event, if any, using the `BottomSheetInteraction` and `BottomCommandingInteraction` enums. 
`BottomCommandingController` also passes through the interaction enum coming from the sheet directly.

The PR also adds more powerful `isHidden` and `isExpanded` setters which allow clients to turn off animations and provide completion handlers. The old setters are still there, refactored to just call the bigger ones.
External changes to properties which support completion handlers will not result in state change delegate callbacks to avoid sending two callbacks about the same event.

### Verification

Local testing with a delegate plugged in to verify:
Sheet-specific interactions (swipe, grippie tap) are passed through.
Commanding-specific interactions come in correctly.
Bottom bar popover callbacks are called with correct user interactions enums, always after the animation completes.
Interrupted state change animations don't fire off a delegate callback.
No double callbacks when isHidden / isExpanded are changed.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/641)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -34,6 +34,7 @@ public protocol BottomCommandingControllerDelegate: AnyObject {
     @objc optional func bottomCommandingControllerDidDismissPopover(_ bottomCommandingController: BottomCommandingController, commandingInteraction: BottomCommandingInteraction)
 }
 
+/// User interactions that can trigger a state change.
 @objc public enum BottomCommandingInteraction: Int {
     case none
     case sheetInteraction
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -133,6 +133,8 @@ open class BottomCommandingController: UIViewController {
         }
     }
 
+    /// When in sheet layout, `BottomSheetController` holds it's own `isHidden` state which is the main
+    /// source of truth and the public getter will return that instead of this backing variable.
     private var _isHidden: Bool = false
 
     /// Indicates whether a more button is visible in the sheet style when `expandedListSections` is non-empty.
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -10,6 +10,34 @@ public protocol BottomCommandingControllerDelegate: AnyObject {
 
     /// Called when `collapsedHeightInSafeArea` changes.
     @objc optional func bottomCommandingControllerCollapsedHeightInSafeAreaDidChange(_ bottomCommandingController: BottomCommandingController)
+
+    /// Called after the bottom sheet expansion state changes.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - expansionState: The expansion state the sheet moved to.
+    ///   - commandingInteraction: If the state change was caused by user interaction, it will be indicated using this enum.
+    ///   - sheetInteraction: If `commandingInteraction` is `.sheetInteraction`, this enum will contain more information about what triggered the state change.
+    @objc optional func bottomCommandingController(_ bottomCommandingController: BottomCommandingController, sheetDidMoveTo expansionState: BottomSheetExpansionState, commandingInteraction: BottomCommandingInteraction, sheetInteraction: BottomSheetInteraction)
+
+    /// Called after the bottom bar popover is presented.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - commandingInteraction: The user interaction that caused the popover to show.
+    @objc optional func bottomCommandingControllerDidPresentPopover(_ bottomCommandingController: BottomCommandingController, commandingInteraction: BottomCommandingInteraction)
+
+    /// Called after the bottom bar popover is dismissed.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - commandingInteraction: The user interaction that caused the popover to dismiss.
+    @objc optional func bottomCommandingControllerDidDismissPopover(_ bottomCommandingController: BottomCommandingController, commandingInteraction: BottomCommandingInteraction)
+}
+
+@objc public enum BottomCommandingInteraction: Int {
+    case none
+    case sheetInteraction
+    case moreButtonTap
+    case commandTap
+    case otherUserAction
 }
 
 /// Persistent commanding surface displayed at the bottom of the available area.
@@ -213,7 +241,12 @@ open class BottomCommandingController: UIViewController {
     public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
         if presentedViewController != nil {
-            dismiss(animated: false)
+            dismiss(animated: false) { [weak self] in
+                guard let strongSelf = self else {
+                    return
+                }
+                strongSelf.delegate?.bottomCommandingControllerDidDismissPopover?(strongSelf, commandingInteraction: .none)
+            }
         }
     }
 
@@ -254,6 +287,7 @@ open class BottomCommandingController: UIViewController {
         let sheetController = BottomSheetController(headerContentView: commandStackContainer, expandedContentView: makeSheetExpandedContent(with: tableView))
         sheetController.hostedScrollView = tableView
         sheetController.isHidden = isHidden
+        sheetController.delegate = self
 
         addChild(sheetController)
         view.addSubview(sheetController.view)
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -10,6 +10,37 @@ public protocol BottomCommandingControllerDelegate: AnyObject {
...
+
+    /// Called after the bottom sheet expansion state changes.
+    ///
+    /// External changes to `isHidden` will not trigger this callback.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - expansionState: The expansion state the sheet moved to.
+    ///   - commandingInteraction: If the state change was caused by user interaction, it will be indicated using this enum.
+    ///   - sheetInteraction: If `commandingInteraction` is `.sheetInteraction`, this enum will contain more information about what triggered the state change.
+    @objc optional func bottomCommandingController(_ bottomCommandingController: BottomCommandingController, sheetDidMoveTo expansionState: BottomSheetExpansionState, commandingInteraction: BottomCommandingInteraction, sheetInteraction: BottomSheetInteraction)
```
anandrajeswaran, line 22: minor: Even if swift-lint doesn't force us to, I like the cleanliness of putting one param per line when you have this many params. Separately, we should make swift-lint force us to do so and/or just switch to swift-format

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -10,6 +10,37 @@ public protocol BottomCommandingControllerDelegate: AnyObject {
...
+
+    /// Called after the bottom bar popover is presented.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - commandingInteraction: The user interaction that caused the popover to show.
+    @objc optional func bottomCommandingControllerDidPresentPopover(_ bottomCommandingController: BottomCommandingController, commandingInteraction: BottomCommandingInteraction)
```
anandrajeswaran, line 28: minor: like above when there are multiple parameters I *think* convention is to make it `bottomCommandingController(bottomCommandingController: BottomCommandingController, didPresentPopoverFromInteraction:...`

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -10,6 +10,37 @@ public protocol BottomCommandingControllerDelegate: AnyObject {
...
+
+    /// Called after the bottom bar popover is dismissed.
+    /// - Parameters:
+    ///   - bottomCommandingController: The caller object.
+    ///   - commandingInteraction: The user interaction that caused the popover to dismiss.
+    @objc optional func bottomCommandingControllerDidDismissPopover(_ bottomCommandingController: BottomCommandingController, commandingInteraction: BottomCommandingInteraction)
+}
+
+/// User interactions that can trigger a state change.
+@objc public enum BottomCommandingInteraction: Int {
```
anandrajeswaran, line 38: Can we document these? like what does "none" mean? The commanding controller just got bored and left?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -151,6 +160,55 @@ open class BottomCommandingController: UIViewController {
...
 
+    /// Sets the `isHidden` property with a completion handler.
+    /// - Parameters:
+    ///   - isHidden: The new value.
+    ///   - animated: Indicates if the change should be animated. The default value is `true`.
+    ///   - completion: Closure to be called when the state change completes.
+    @objc public func setIsHidden(_ isHidden: Bool, animated: Bool = true, completion: ((_ isFinished: Bool) -> Void)? = nil) {
+        guard isViewLoaded else {
+            _isHidden = isHidden
+            return
```
anandrajeswaran, line 171: minor: this feels like an odd use of "guard" since it actually does some work. Though I'm definitely conservative when it comes to early returns

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -213,7 +271,12 @@ open class BottomCommandingController: UIViewController {
     public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
         if presentedViewController != nil {
-            dismiss(animated: false)
+            dismiss(animated: false) { [weak self] in
+                guard let strongSelf = self else {
+                    return
+                }
```
anandrajeswaran, line 277: minor: just use a ? on the next line instead of this to simplify?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -10,13 +10,24 @@ public protocol BottomSheetControllerDelegate: AnyObject {
...
+    ///   - bottomSheetController: The caller object.
+    ///   - expansionState: The expansion state that the sheet moved to.
+    ///   - interaction: The user interaction that caused the state change.
+    @objc optional func bottomSheetController(_ bottomSheetController: BottomSheetController, didMoveTo expansionState: BottomSheetExpansionState, interaction: BottomSheetInteraction)
 
     /// Called when `collapsedHeightInSafeArea` changes.
     @objc optional func bottomSheetControllerCollapsedHeightInSafeAreaDidChange(_ bottomSheetController: BottomSheetController)
 }
 
+/// User interactions that can trigger a state change.
```
anandrajeswaran, line 24: Similar here - document each case?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: We need to pass in a non-optional self to the delegate. That's why I'm getting it here.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: The reason I need .none is basically that objc doesn't allow optional enums (or Ints) and so I need a way to tell the delegate a state change wasn't triggered by the user. I'll try to add a comment that helps this make some sense 😄

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I think it should even be:
bottomCommandingController(_ bottomCommandingController: BottomCommandingController, didPresetPopoverFrom interaction: ...
It seemed pretty horrible to me, but you're right, that's the convention.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I tried to improve the naming and documentation.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Never mind, it's actually not that bad at all!

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: I got rid of the guard.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -88,39 +124,17 @@ open class BottomCommandingController: UIViewController {
...
+    @objc open var isHidden: Bool {
+        get {
+            return bottomSheetController?.isHidden ?? _isHidden
+        }
+        set {
+            setIsHidden(newValue)
         }
     }
 
+    private var _isHidden: Bool = false
```
markavitale, line 136: Nit: since this is complicated logic, this could probably use some minor documentation about how this interacts with the public var.

markavitale (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: done, hope the new comment somewhat helps explain it

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -88,39 +124,19 @@ open class BottomCommandingController: UIViewController {
...
-                    }
-                    newAnimator.addAnimations { [weak self] in
-                        self?.view.layoutIfNeeded()
-                    }
-
-                    newAnimator.startAnimation()
-                    bottomBarHidingAnimator = newAnimator
-                }
-            }
+    @objc open var isHidden: Bool {
```
harrieshin, line 127: isn't it bit confusing that there is IsHidden and then setishiddenwithanimation ?

PR #644, opened 2021-07-22
Add support for 2-line hero command titles in `BottomCommandingController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

As per the design spec, hero commands in the bottom sheet need to support 2-line labels. To follow the spec, the hero items now have to be top aligned and the spacing between the header content and expanded content needs to vary depending on whether there is at least one 2-line hero item title.

The changes in this PR per file:

`BottomCommandingDemoController.swift`
Add a toggle for long hero item titles

`TabBarItemView.swift`
Add a passthrough property for titleLabel.preferredMaxLayoutWidth. We need to set this on the label to ensure that the fitting width is restricted when calculating intrinsic content size. Without this, the height given will always correspond to a 1-line label, even if the actual view ends up taller with a 2-line label. The correct intrinsic content height is necessary to know when we have at least one tall hero command.

- change a few constraints in the hero command stack, because the commands now don't have a fixed height
- recalculate the header size given to the sheet controller whenever a relevant change happens
- make sure that for the sheet specifically, we don't directly call the delegate with CollapsedHeightInSafeAreaDidChange and we just forward the equivalent delegate calls from the sheet controller. This ensures no change will be missed by he client.
- also slightly change bottom bar constraints, because now we need a fixed bar height to allow the hero stack to grow vertically without growing the bar itself
- things were getting messy, so I factored all sheet sizing logic into `updateSheetSizingParameters`

### Verification
Various sanity checks in the demo app
- make sure sheet size changes properly as titles are updated
- bottom bar doesn't grow with long titles
- delegate is informed about the height changes to update scroll view insets

https://user-images.githubusercontent.com/3610850/126695895-f48e77d3-b615-40be-9f1b-b321d04f37fc.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/644)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -179,7 +179,6 @@ open class BottomCommandingController: UIViewController {
         if let contentViewController = contentViewController {
             addChildContentViewController(contentViewController)
         }
-        delegate?.bottomCommandingControllerCollapsedHeightInSafeAreaDidChange?(self)
     }
 
     public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
@@ -202,7 +201,6 @@ open class BottomCommandingController: UIViewController {
             } else {
                 setupBottomSheetLayout()
             }
-            delegate?.bottomCommandingControllerCollapsedHeightInSafeAreaDidChange?(self)
         }
 
         if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
@@ -240,6 +238,8 @@ open class BottomCommandingController: UIViewController {
         bottomBarViewBottomConstraint = bottomConstraint
         self.bottomBarView = bottomBarView
         reloadHeroCommandStack()
+
+        delegate?.bottomCommandingControllerCollapsedHeightInSafeAreaDidChange?(self)
     }
 
     private func setupBottomSheetLayout() {
@@ -253,6 +253,7 @@ open class BottomCommandingController: UIViewController {
         let sheetController = BottomSheetController(headerContentView: commandStackContainer, expandedContentView: makeSheetExpandedContent(with: tableView))
         sheetController.hostedScrollView = tableView
         sheetController.isHidden = isHidden
+        sheetController.delegate = self
 
         addChild(sheetController)
         view.addSubview(sheetController.view)
@@ -557,7 +558,6 @@ open class BottomCommandingController: UIViewController {
 
         if newHeaderHeight != oldHeaderHeight {
             bottomSheetController.collapsedContentHeight = newHeaderHeight
-            delegate?.bottomCommandingControllerCollapsedHeightInSafeAreaDidChange?(self)
         }
     }
 
@@ -830,5 +830,11 @@ extension BottomCommandingController: CommandingItemDelegate {
         default:
             break
         }
+    }
+}
+
+extension BottomCommandingController: BottomSheetControllerDelegate {
+    public func bottomSheetControllerCollapsedHeightInSafeAreaDidChange(_ bottomSheetController: BottomSheetController) {
+        delegate?.bottomCommandingControllerCollapsedHeightInSafeAreaDidChange?(self)
     }
 }
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -547,9 +547,9 @@ open class BottomCommandingController: UIViewController {
         let reducedBottomWhitespace = max(0, requiredBottomWhitespace - view.safeAreaInsets.bottom)
 
         // We need additional top margin to account for missing resizing handle when isExpandable is false
-        let addedHeaderTopMargin = isExpandable
-            ? Constants.BottomSheet.headerExpandableAddedTopMargin
-            : Constants.BottomSheet.headerNonExpandableAddedTopMargin
+        let addedHeaderTopMargin = !isExpandable
+            ? BottomSheetController.resizingHandleHeight
+            : 0
         bottomSheetHeroStackTopConstraint?.constant = Constants.BottomSheet.headerTopMargin + addedHeaderTopMargin
 
         let oldCollapsedContentHeight = bottomSheetController.collapsedContentHeight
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -534,9 +525,39 @@ open class BottomCommandingController: UIViewController {
         }
     }
 
-    private func updateSheetExpandedContentHeight() {
-        if let bottomSheetController = bottomSheetController {
-            bottomSheetController.preferredExpandedContentHeight = estimatedTableViewHeight + Constants.BottomSheet.expandedContentTopMargin
+    private func updateSheetSizingParameters() {
```
harrieshin, line 528: not sure this function name is clearer. can you add comments>

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -641,20 +654,23 @@ open class BottomCommandingController: UIViewController {
     }
 
     private struct Constants {
-        static let heroButtonHeight: CGFloat = 48
+        static let defaultHeroButtonHeight: CGFloat = 39
```
harrieshin, line 657: this number doesn't seem right. most fluent controls snaps to 4 its

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -534,9 +525,39 @@ open class BottomCommandingController: UIViewController {
...
+        let headerTotalTopMargin = Constants.BottomSheet.heroStackTopBottomMargin
+        + ( isExpandable
+            ? Constants.BottomSheet.headerExpandableAddedTopMargin
+            : Constants.BottomSheet.headerNonExpandableAddedTopMargin)
+        bottomSheetHeroStackTopConstraint?.constant = headerTotalTopMargin
+
+        let oldHeaderHeight = bottomSheetController.collapsedContentHeight
+        let newHeaderHeight = headerTotalTopMargin + maxHeroItemHeight + Constants.BottomSheet.heroStackTopBottomMargin
+
+        if newHeaderHeight != oldHeaderHeight {
```
harrieshin, line 559: so the hero height changes based on one line or 2 lines? its not the same and the padding is bit different?

lcapkovic: Yeah, it does change and it's what makes this part of the spec so hard to implement. Visually, with the sheet expanded the header size almost doesn't change. But when collapsed we need to still show all of the tall labels without leaking out of the safe area, so the collapsed height has to change.
I'm working on a separate PR that simplifies the sizing and I'll update this PR afterwards to hopefully simplify this as well.

lcapkovic: Nevermind, I did all of it here.

So, your suggestion was great and it allowed me to make things more straightforward.
The header view height is now always the same. The 1 vs. 2 line hero label now just decides how much more of it we need to show when collapsed to fulfill the bottom whitespace requirement. 
The new solution also accounts for difference in bottom safe area insets. So if there is enough of an inset, we subtract that from how much more whitespace we need. If there isn't enough, we add our own.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -670,12 +687,12 @@ open class BottomCommandingController: UIViewController {
...
 
         struct BottomSheet {
-            static let expandedFraction: CGFloat = 0.7 // Probably should be more customizable / based on content
-            static let heroStackExpandableTopMargin: CGFloat = 0
-            static let heroStackNonExpandableTopMargin: CGFloat = 16
-            static let heroStackLeadingTrailingMargin: CGFloat = 8
+            static let headerHeight: CGFloat = 64
 
-            static let expandedContentTopMargin: CGFloat = 16
+            static let headerTopMargin: CGFloat = 5
```
harrieshin, line 692: honestly, i don't quite grasp the 3 different top margin constants you defined her and also why headertopmargein is 5?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -534,12 +529,41 @@ open class BottomCommandingController: UIViewController {
...
+        let headerHeightWithoutBottomWhitespace = Constants.BottomSheet.headerTopMargin + maxHeroItemHeight
+
+        // How much more whitespace is required at the bottom of the sheet header
+        let requiredBottomWhitespace = max(0, Constants.BottomSheet.headerHeight - headerHeightWithoutBottomWhitespace)
+
+        // The safe area inset can fulfill some or all of our bottom whitespace requirement.
+        // This is how much more we need, taking the inset into account.
+        let reducedBottomWhitespace = max(0, requiredBottomWhitespace - view.safeAreaInsets.bottom)
+
+        // We need additional top margin to account for missing resizing handle when isExpandable is false
```
harrieshin, line 549: the calculation isn't like resize handle height (depending on it is expandable not expandable) + delta gap?

lcapkovic: hmm you're right, I should probably expose the current resizing handle height through the sheet controller rather than guessing through these constants.

lcapkovic: Changed it to what your suggestion is.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Changed 5 to 4. It was mostly eyeballed from the figma spec, but 4 does feel better.

harrieshin (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -619,12 +613,41 @@ open class BottomCommandingController: UIViewController {
...
+    private func updateSheetHeaderSizingParameters() {
+        guard let bottomSheetController = bottomSheetController else {
+            return
+        }
+        bottomSheetController.isExpandable = isExpandable
+
+        let maxHeroItemHeight = heroCommandStack.arrangedSubviews.map { $0.intrinsicContentSize.height }.max() ?? Constants.defaultHeroButtonHeight
+        let headerHeightWithoutBottomWhitespace = Constants.BottomSheet.headerTopMargin + maxHeroItemHeight
+
+        // How much more whitespace is required at the bottom of the sheet header
```
anandrajeswaran, line 626: padding might be a better word than whitespace, which is confusing already in dark mode?

PR #645, opened 2021-07-22
Ensure only iPad gets the bottom bar layout in `BottomCommandingController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There is a bug where the bottom bar shows in landscape on the largest iPhones. This is because these iPhones fall under the "regular" horizontal size class when in landscape, and that's currently the only condition we use to switch to the bottom bar layout.
To fix this I added an additional check to see if we're running on an iPad.

### Verification

|Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-22 at 14 46 28|Simulator Screen Shot - iPhone 12 Pro Max - 2021-07-22 at 14 45 47|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/645)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -170,11 +170,7 @@ open class BottomCommandingController: UIViewController {
         view.translatesAutoresizingMaskIntoConstraints = false
         view.addLayoutGuide(commandingLayoutGuide)
 
-        if traitCollection.horizontalSizeClass == .regular {
-            setupBottomBarLayout()
-        } else {
-            setupBottomSheetLayout()
-        }
+        setupCommandingLayout()
 
         if let contentViewController = contentViewController {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -217,6 +209,14 @@ open class BottomCommandingController: UIViewController {
         }
     }
 
+    private func setupCommandingLayout() {
+        if traitCollection.horizontalSizeClass == .regular && traitCollection.userInterfaceIdiom == .pad {
```
mischreiber, line 213: I was about to ask how you'd handle being in compact mode on an iPad, but this looks like it should do the trick!

mischreiber (approved):

PR #648, opened 2021-07-23
[vNext] AvatarGroup: Creating smoother animations

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Animations whenever the AvatarGroup would change would stutter. We added animatableData and encased the group in a ViewBuilder to prevent alignment bugs that cause the group to appear centered.

### Verification

| before | after |
|  | after1 |

#### Procedure:
- Add avatar
- Remove avatar
- Change max displayed avatars
- Change overflow count

Tested on iOS 13, 14, & 15,

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/648)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
@@ -360,7 +360,7 @@ class AvatarGroupDemoController: DemoTableViewController {
         }
     }
 
-    private var maxDisplayedAvatars: Int = 4 {
+    private var maxDisplayedAvatars: Int = 3 {
         didSet {
             if oldValue != maxDisplayedAvatars {
                 maxAvatarsTextField.text = "\(maxDisplayedAvatars)"
@@ -440,9 +440,9 @@ class AvatarGroupDemoController: DemoTableViewController {
         return textField
     }()
 
-    private var avatarCount: Int = 5 {
+    private var avatarCount: Int = 4 {
         didSet {
-            guard avatarCount != 0 || oldValue != 0 else {
+            guard oldValue != avatarCount && avatarCount >= 0 else {
                 return
             }
             AvatarGroupDemoSection.allCases.filter({ section in
@@ -482,6 +482,7 @@ class AvatarGroupDemoController: DemoTableViewController {
         guard avatarCount > 0 else {
             return
         }
+
         avatarCount -= 1
     }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
@@ -138,7 +138,8 @@ class AvatarGroupDemoController: DemoTableViewController {
             NSLayoutConstraint.activate([
                 cell.contentView.leadingAnchor.constraint(equalTo: avatarGroupView.leadingAnchor, constant: -20),
                 cell.contentView.topAnchor.constraint(equalTo: avatarGroupView.topAnchor, constant: -15),
-                cell.contentView.bottomAnchor.constraint(equalTo: avatarGroupView.bottomAnchor, constant: 15)
+                cell.contentView.bottomAnchor.constraint(equalTo: avatarGroupView.bottomAnchor, constant: 15),
+                cell.contentView.trailingAnchor.constraint(equalTo: avatarGroupView.trailingAnchor, constant: 20)
             ])
 
             cell.backgroundColor = self.isUsingAlternateBackgroundColor ? Colors.tableCellBackgroundSelected : Colors.tableCellBackground
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
@@ -441,31 +441,35 @@ class AvatarGroupDemoController: DemoTableViewController {
 
     private var avatarCount: Int = 5 {
         didSet {
-            AvatarGroupDemoSection.allCases.filter({ section in
-                return section.isDemoSection
-            }).forEach { section in
-                var avatarGroupsForCurrentSection: [AvatarGroupDemoRow: MSFAvatarGroup] = [:]
+            if avatarCount == 0 && oldValue == 0 {
+                return
+            } else {
+                AvatarGroupDemoSection.allCases.filter({ section in
+                    return section.isDemoSection
+                }).forEach { section in
+                    var avatarGroupsForCurrentSection: [AvatarGroupDemoRow: MSFAvatarGroup] = [:]
 
-                AvatarGroupDemoRow.allCases.filter({ row in
-                    return row.isDemoRow
-                }).forEach { row in
-                    let avatarGroup = demoAvatarGroupsBySection[section]![row]
-                    if oldValue < avatarCount {
-                        let avatarState = avatarGroup?.state.createAvatar()
-                        for index in 0...avatarCount - 1 {
-                            let samplePersona = samplePersonas[index]
-                            avatarState!.image = samplePersona.image
-                            avatarState!.isRingVisible = section.isMixedBorder ? index % 2 == 0 : section.showBorders
-                            avatarState!.primaryText = samplePersona.name
-                            avatarState!.secondaryText = samplePersona.email
+                    AvatarGroupDemoRow.allCases.filter({ row in
+                        return row.isDemoRow
+                    }).forEach { row in
+                        let avatarGroup = demoAvatarGroupsBySection[section]![row]
+                        if oldValue < avatarCount {
+                            let avatarState = avatarGroup?.state.createAvatar()
+                            for index in 0...avatarCount - 1 {
+                                let samplePersona = samplePersonas[index]
+                                avatarState!.image = samplePersona.image
+                                avatarState!.isRingVisible = section.isMixedBorder ? index % 2 == 0 : section.showBorders
+                                avatarState!.primaryText = samplePersona.name
+                                avatarState!.secondaryText = samplePersona.email
+                            }
+                        } else {
+                            avatarGroup?.state.removeAvatar(at: avatarCount)
                         }
-                    } else {
-                        avatarGroup?.state.removeAvatar(at: avatarCount)
-                    }
 
-                    avatarGroupsForCurrentSection.updateValue(avatarGroup!, forKey: row)
+                        avatarGroupsForCurrentSection.updateValue(avatarGroup!, forKey: row)
+                    }
+                    demoAvatarGroupsBySection.updateValue(avatarGroupsForCurrentSection, forKey: section)
                 }
-                demoAvatarGroupsBySection.updateValue(avatarGroupsForCurrentSection, forKey: section)
             }
         }
     }
@@ -475,7 +479,7 @@ class AvatarGroupDemoController: DemoTableViewController {
     }
 
     @objc private func subtractAvatarCount(_ cell: ActionsCell) {
-        avatarCount -= 1
+        avatarCount -= avatarCount == 0 ? 0 : 1
     }
 
     @objc private func toggleAlternateBackground(_ cell: BooleanCell) {
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
@@ -442,35 +442,34 @@ class AvatarGroupDemoController: DemoTableViewController {
 
     private var avatarCount: Int = 5 {
         didSet {
-            if avatarCount == 0 && oldValue == 0 {
+            guard avatarCount != 0 || oldValue != 0 else {
                 return
-            } else {
-                AvatarGroupDemoSection.allCases.filter({ section in
-                    return section.isDemoSection
-                }).forEach { section in
-                    var avatarGroupsForCurrentSection: [AvatarGroupDemoRow: MSFAvatarGroup] = [:]
+            }
+            AvatarGroupDemoSection.allCases.filter({ section in
+                return section.isDemoSection
+            }).forEach { section in
+                var avatarGroupsForCurrentSection: [AvatarGroupDemoRow: MSFAvatarGroup] = [:]
 
-                    AvatarGroupDemoRow.allCases.filter({ row in
-                        return row.isDemoRow
-                    }).forEach { row in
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -125,36 +125,48 @@ public struct AvatarGroup: View {
...
-                AvatarView(avatar)
-                    .modifyIf(needsCutout, { view in
-                        view.mask(AvatarCutout(xOrigin: xOrigin,
-                                               yOrigin: yOrigin,
-                                               cutoutSize: cutoutSize)
-                                    .fill(style: FillStyle(eoFill: true)))
-                    })
-                    .padding(.trailing, tokens.style == .stack ? stackPadding : interspace)
+        let avatarHeight = groupContainsRing() ? CGFloat(size + (ringOffset * 2)) : size
+        GeometryReader { _ in
```
rdeassis, line 129: This seems to be related to the bug you found about iOS 13:

Screen Shot 2021-07-23 at 10 32 13 AM

rdeassis: Isn't the GeometryReader supposed to be used inside the HStack?
Weren't you supposed to get the geometry proxy to calculate the final x and y coordinates?

```
GeometryReader { geoProxy in
...
}
```

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -125,36 +125,48 @@ public struct AvatarGroup: View {
...
+
+                    let ringPaddingInterspace = nextAvatarHasRing ? interspace - (ringOffset + ringOuterGap) : interspace - ringOffset
+                    let noRingPaddingInterspace = nextAvatarHasRing ? interspace - ringOuterGap : interspace
+                    let stackPadding = (currentAvatarHasRing ? ringPaddingInterspace : noRingPaddingInterspace)
+
+                    let xOrigin = currentAvatarHasRing ? x + ringOffset : x
+                    let yOrigin = currentAvatarHasRing ? (nextAvatarHasRing ? ringOuterGap : ringOffset) :
+                        (nextAvatarHasRing ? 0 - ringOffset + tokens.ringOuterGap : 0)
+                    let cutoutSize = nextAvatarHasRing ? size + ringOffset + ringOuterGap : size
+                    AvatarView(avatar)
```
rdeassis, line 146: Provided that the calculations above need to be adjusted, I thought of the following approach:

- not using the GeometryReader
- making the Avatar state implement Identifiable
- using the .offset modifier to position them in the HStack

I think that could improve the animation as well as SwiftUI would keep track of the views an animate them as the state changes:

```
ForEach(avatars.prefix(maxDisplayedAvatars), id: \.id) { avatar in
...
}
```

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -125,36 +125,43 @@ public struct AvatarGroup: View {
...
+                        .modifyIf(needsCutout, { view in
+                            view.mask(AvatarCutout(xOrigin: xOrigin,
+                                                   yOrigin: yOrigin,
+                                                   cutoutSize: cutoutSize)
+                                        .fill(style: FillStyle(eoFill: true)))
+                        })
+                        .padding(.trailing, tokens.style == .stack ? stackPadding : interspace)
```
rdeassis, line 154: So, the GeometryReader won't be able to replace the padding to position the Avatars in the View?
I thought that this would be replaced by a .frame(x: y:) modifier.

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
rdeassis: Any findings on implementing Identifiable in the Avatar state?
I'd be curious to see if this wouldn't make the animations smoother.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -220,6 +234,18 @@ public struct AvatarGroup: View {
         var yOrigin: CGFloat
         var cutoutSize: CGFloat
 
+        var animatableData: AnimatablePair<AnimatablePair<CGFloat, CGFloat>, CGFloat> {
+            get {
+                AnimatablePair(AnimatablePair(xOrigin, yOrigin), cutoutSize)
+            }
+
```
mischreiber, line 241: Nit: no blank line between `get` and `set`

mischreiber (approved):

ios/FluentUI/Vnext/AvatarGroup/AvatarGroup.swift
```diff
@@ -125,36 +125,43 @@ public struct AvatarGroup: View {
...
+                }
+                if overflowCount > 0 {
+                    createOverflow(count: overflowCount)
+                }
             }
+            .frame(width: geo.frame(in: .local).minX,
+                   height: geo.size.height,
+                   alignment: .leading)
         }
+        .frame(height: groupHeight)
```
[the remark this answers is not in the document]
rdeassis: Any findings on how to correctly update the Geometry Reader frame width?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -451,7 +451,9 @@ public struct Avatar: View {
 }
 
 /// Properties available to customize the state of the avatar
-class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
+class MSFAvatarStateImpl: NSObject, ObservableObject, Identifiable, MSFAvatarState {
```
rdeassis, line 454: Since this is already in main, the PR should target to merge to main instead of vnext-prototype.

rdeassis: Also, main already has the new demo controller based on the UITableView after the merge of #674.
Can you also bring those changes to your PR branch and test in the new demo controller? (updating the description screenshots/videos)

rdeassis (changes requested):

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -167,48 +168,64 @@ public struct AvatarGroup: View {
...
+                        let xPosition = currentAvatarHasRing ? x - ringOuterGap - ringOuterGap : x - ringOuterGap
+                        let xPositionRTL = currentAvatarHasRing ? rtlRingPaddingInterspace : rtlNoRingPaddingInterspace
+                        let xOrigin = Locale.current.isRightToLeftLayoutDirection() ? xPositionRTL : xPosition
+                        let yOrigin = sizeDiff / 2
+                        let cutoutSize = isLastDisplayed ? (ringOuterGap * 2) + imageSize : nextAvatarSize
+
+                        avatarView
+                            .modifyIf(needsCutout, { view in
+                                view.mask(Avatar.AvatarCutout(
```
rdeassis, line 206: You need to incorporate #703 here.
This state of the code still reproes the iOS 13.0/13.1 issue:

Screen Shot 2021-09-10 at 1 46 44 PM

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -167,48 +168,64 @@ public struct AvatarGroup: View {
...
+                    if overflowCount > 0 {
+                        createOverflow(count: overflowCount)
+                            .transition(AnyTransition.move(edge: .leading))
+                    }
+                }
+                .frame(width: geo.frame(in: .local).minX,
+                       height: geo.size.height,
+                       alignment: .leading)
             }
+            .frame(height: groupHeight)
```
rdeassis, line 225: I think I added a comment about this in the past.
This is not showing in the UI of the simulator, but the Geometry Reader frame width is not being set (screenshot below).
The correct approach in this case should be calculating the width of the entire control and set it here.

Screen Shot 2021-09-10 at 1 41 38 PM

rdeassis (commented):
Just tried that in iOS 13 and the alignment is still not fixed.
Is the fix you made integrated into this branch or is this a new issue?

Screen Shot 2021-11-09 at 11 47 40 AM

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -185,48 +186,80 @@ public struct AvatarGroup: View {
...
+                .frame(height: geo.size.height,
+                       alignment: .leading)
             }
+            .frame(width: calcwidth, height: groupHeight)
+        }
+
+        return avatarGroupContent
+    }
+
+    private var calcwidth: CGFloat {
```
rdeassis, line 253: The width that is calculated seems wider than needed.
Is that the root cause of the alignment issue?

Screen Shot 2021-11-19 at 10 41 13 AM

rdeassis (commented):
The alignment does not seem to be to the left of the original view.
Also, the app crashes when we try to decrease the last avatar in the control.

https://user-images.githubusercontent.com/68076145/142674910-2c9a43f0-79c8-4807-89a2-1ff61f92f4d7.mov

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -184,48 +185,83 @@ public struct AvatarGroup: View {
...
             }
-            if overflowCount > 0 {
-                createOverflow(count: overflowCount)
+            .frame(width: calcWidth(), height: groupHeight)
+        }
+
+        return avatarGroupContent
+    }
+
+    private func calcWidth() -> CGFloat {
```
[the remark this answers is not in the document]
sophialee0416: I changed this back to a method because I realized I needed a way to obtain the value of overflow to take that into the calculation

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -185,48 +186,86 @@ public struct AvatarGroup: View {
...
+                .frame(height: geo.size.height,
+                       alignment: .leading)
             }
+            .frame(width: calcWidth(hasOverflow: overflowCount > 0), height: groupHeight, alignment: .leading)
+        }
+
+        return avatarGroupContent
+    }
+
+    private func calcWidth(hasOverflow: Bool) -> CGFloat {
```
rdeassis, line 253: Is this considering added width of the avatars with rings?
Seems to me that the leading gap/padding is bigger when the avatar size is bigger.

Screen Shot 2021-11-24 at 1 44 28 PM

rdeassis: And the number of avatars displayed is also influencing the calculation:
Screen Shot 2021-11-24 at 1 54 13 PM

Screen Shot 2021-11-24 at 1 53 54 PM

rdeassis (commented):
The demo app is still crashing when we hit the "Decrease Avatar Count" button.
Can you please fix that by checking the avatar count before making the call to remove the avatar?

Screen Shot 2021-11-24 at 1 39 43 PM

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -441,31 +441,35 @@ class AvatarGroupDemoController: DemoTableViewController {
...
-                    let avatarGroup = demoAvatarGroupsBySection[section]![row]
-                    if oldValue < avatarCount {
-                        let avatarState = avatarGroup?.state.createAvatar()
-                        for index in 0...avatarCount - 1 {
-                            let samplePersona = samplePersonas[index]
-                            avatarState!.image = samplePersona.image
-                            avatarState!.isRingVisible = section.isMixedBorder ? index % 2 == 0 : section.showBorders
-                            avatarState!.primaryText = samplePersona.name
-                            avatarState!.secondaryText = samplePersona.email
+            if avatarCount == 0 && oldValue == 0 {
```
rdeassis, line 444: Can you use `guard let` for early returns here and in line 482?
**Reference:** 

rdeassis (commented):
Now that we bumped our minimum iOS version to 14, you can target this PR to the main_0.3 branch and then cherry-pick it to main once it's merged.

Also, because we're not merging it to vnext-prototype, can you remove the vnext-prototype tag from the PR?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -185,48 +186,92 @@ public struct AvatarGroup: View {
...
+        var width: CGFloat = 0
+        let maxShown = state.maxDisplayedAvatars < state.avatars.count ? state.maxDisplayedAvatars : state.avatars.count
+        let interspace: CGFloat = state.style == .stack ? tokens.interspace : tokens.interspace
+        let gapOffset: CGFloat = tokens.ringOuterGap * 2
+        let ringGapOffset: CGFloat = tokens.ringThickness + tokens.ringInnerGap + tokens.ringOuterGap
```
rdeassis, line 260: This offset is being used below on the calculation of the avatar size in case the ring is visible.
Shouldn't these be multiplied by 2 (both sides of the ring thickness, gap and outer gap?

Have you validated the ring scenario?

Screen Shot 2021-12-09 at 3 25 08 PM
Screen Shot 2021-12-09 at 3 25 23 PM

rdeassis: In this scenario the Avatars are getting cropped on the left side:

Screen Shot 2021-12-09 at 3 30 36 PM

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -185,48 +186,92 @@ public struct AvatarGroup: View {
...
+        let imageSize: CGFloat = tokens.size.size
+        for index in 0 ..< maxShown {
```
harrieshin, line 262: is this for loop the best and fastest way to calculate the total width of your avatarGroup? if I understand the intent of this helper function is just find the total width which doesn't necessary need the for loop because you don't need to set the individual frames of each avatar's location.  If I would have pseudo code this wouldn't be something like 

total number of avatar shown X avatar size  + ( ( count -1 /* number of inner gaps */) X gap between avatars /* it can be positive or negative depending on facepile vs face stack */ )

mischreiber: I think the problem is that different Avatars can have different sizes, depending on whether they have rings or not.

harrieshin: thanks for the explanation! something definitely should be written in comment. Although, I still think we can refactor a bit and share "true" avatar size frame logic between both this function and the other parts of geometricreader code for simplification?

rdeassis (approved):

PR #649, opened 2021-07-23
pillsegmentedcontrol mask label shouldn't be accessible

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Mask labels are rendered on top of the pill buttons to show smooth animation transition. However, it shouldn't VO accessible because user should directly interact with the pill button.

### Verification

| Screen Shot 2021-07-23 at 1 30 24 PM| Screen Shot 2021-07-23 at 1 29 49 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/649)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -554,6 +554,7 @@ open class SegmentedControl: UIControl {
         maskedLabel.text = button.currentTitle
         maskedLabel.font = button.titleLabel?.font
         maskedLabel.translatesAutoresizingMaskIntoConstraints = false
+        maskedLabel.isAccessibilityElement = false
         pillMaskedLabelsContainerView.addSubview(maskedLabel)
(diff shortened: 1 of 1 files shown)


huwilkes:
Does pillMaskedLabelsContainerView.isUserInteractionEnabled = false either not apply to children, or not disable something as an accessible element? I would guess not if we have to make this change.

huwilkes (approved):

harrieshin:
> Does pillMaskedLabelsContainerView.isUserInteractionEnabled = false either not apply to children, or not disable something as an accessible element? I would guess not if we have to make this change.
@huwilkes great equation but no. think of situation like static label or image drawn in UI. Users can't interact with it because it is just a label but VO can still readout what it is to convey the information.

mischreiber (approved):

PR #650, opened 2021-07-23
pod-publishing support for multiple branches

harrieshin:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes
to reduce the hassle of manually publishing main_0.1 for CocoaPods, update the yml file to handle both main and main_0.1 branches.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/650)

### DIFF
--- a/.github/workflows/podPublish.yml
+++ b/.github/workflows/podPublish.yml
@@ -4,9 +4,10 @@ on:
     branches:
       - main
       - main_0.1
+      - main_0.2
     tags:
-      - [0-9]+.[0-9]+.* # Run automatically on new tags matching a 0.N.M pattern
-
+      - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
+      - 0.[0-9]+.[0-9]+_main*
 jobs:
   Pod-Publish:
     runs-on: macOS-10.15
--- a/.github/workflows/podPublish.yml
+++ b/.github/workflows/podPublish.yml
@@ -1,12 +1,14 @@
 name: Pod-Publish
 on:
   push:
+    branches:
+      - main
+      - main_0.1
     tags:
       - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
 
 jobs:
   Pod-Publish:
-    if: github.event.base_ref == 'refs/heads/main'
     runs-on: macOS-10.15
     
     steps:
--- a/.github/workflows/podPublish.yml
+++ b/.github/workflows/podPublish.yml
@@ -5,7 +5,7 @@ on:
       - main
       - main_0.1
     tags:
-      - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
+      - [0-9]+.[0-9]+.* # Run automatically on new tags matching a 0.N.M pattern
 
 jobs:
   Pod-Publish:


.github/workflows/podPublish.yml
```diff
@@ -1,12 +1,14 @@
 name: Pod-Publish
 on:
   push:
+    branches:
+      - main
+      - main_0.1
     tags:
       - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
```
mischreiber, line 8: Would this match the tags we've been using for `main_0.1`? The latest tag was `0.1.34_main_0.1`. I'm not sure how strict this is regarding the start/end of the string.

If it works then I'm all for it!

mischreiber (approved):

.github/workflows/podPublish.yml
harrieshin: oh I didn't know we had a different tagging system for main_0.1 branch

mischreiber: Yeah I'm not sure why; @rdeassis might have more background of why it is a different pattern.

rdeassis (approved):

.github/workflows/podPublish.yml
rdeassis: If I remember correctly, we didn't want this to be automatic by accident, which is why we created the different tagging pattern. It'd be an extra layer of safety combined with the branches listed.

.github/workflows/podPublish.yml
```diff
@@ -1,12 +1,14 @@
 name: Pod-Publish
 on:
   push:
+    branches:
+      - main
+      - main_0.1
     tags:
-      - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
+      - [0-9]+.[0-9]+.* # Run automatically on new tags matching a 0.N.M pattern
```
rdeassis, line 8: Wouldn't this be like the below?
0.<number>.<number>[optional_suffix]

0.[0-9]+.[0-9]+*

harrieshin: + right next to * I think it is invalid regex. what I am not too confident about is that GitHub action tag filter seems to be slightly different. like I can't use \d or anything https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet. also I purpose change from 0. to [0-9]+

mischreiber: Yeah I think Rafael's _should_ work since GitHub's language is not technically regex, but I wouldn't trust it without trying it. Looks super strange, being used to regex...

harrieshin: what I am more nervous about is that I don't have a good way to test this locally. unless you all have better ideas?

rdeassis: One way I can think of in order to test this without further side effects is to create a dummy workflow that applies only to a test branch and apply the tags on that test branch.

.github/workflows/podPublish.yml
harrieshin: handled!

.github/workflows/podPublish.yml
harrieshin: verified on my test branch

.github/workflows/podPublish.yml
```diff
@@ -1,12 +1,15 @@
 name: Pod-Publish
 on:
   push:
+    branches:
+      - main
```
rdeassis, line 5: @harrieshin it seems like adding the branches here triggered a pod publish for any merge that happens into main:

https://github.com/microsoft/fluentui-apple/actions/runs/1106904940
https://github.com/microsoft/fluentui-apple/actions/runs/1117722702

Maybe we can use the release to trigger it instead?

https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#onevent_nametypes
```
on:
  release:
    # Only use the types keyword to narrow down the activity types that will trigger your workflow.
    types: [published, created, edited]
```

harrieshin: ah i think i missed the quotation to map it to refs/heads/main instead

PR #659, opened 2021-07-29
[FluentUI Vnext] Integrating Core functionality, Avatar and AvatarGroup.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This is the first integration bringing the vnext-prototype contents to main.
The components being integrated to main are:
 - Core functionality of the Design Token System.
 - Avatar (Vnext version in SwiftUI), which replaces the current AvatarView.
 - AvatarGroup (Vnext version in SwiftUI), which replaces the current AvatarGroupView.

### Verification

1. Local build and run of the Demo app.
 -- Exercising controls' functionality through their demo controllers.

2. Devmain integration local build to ensure public API changes will be accomodated.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/659)


ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -0,0 +1,646 @@
...
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        return AvatarDemoSections.allCases.count
+    }
+
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return AvatarDemoSections.allCases[section].rows.count
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let section = AvatarDemoSections.allCases[indexPath.section]
```
sophialee0416, line 36: I think organizing the demo by style then size seemed a bit cleaner, what do you think? Was it so that each table cell would have consistent height?

rdeassis: I agree it does have the benefit of uniform cell height in each section. It also shows all the different looks that the Avatar can have in a single section, which seems to be more straightforward.

mischreiber:
Noticed a weird animation on `Avatar` when adding and removing the status badge. Regression from legacy?


https://user-images.githubusercontent.com/4934719/127694702-a22dd45b-967a-4169-8308-4521eb90c8e8.mov

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -334,7 +337,11 @@ open class SideTabBar: UIView {
     }
 
     @objc private func handleAvatarViewTapped(_ recognizer: UITapGestureRecognizer) {
-        delegate?.sideTabBar?(self, didActivate: avatarView!)
+        guard let avatar = avatar else {
+            return
```
mischreiber, line 341: Did you want to resolve the other two optionals here too?

ios/FluentUI/Tab Bar/SideTabBar.swift
```diff
@@ -272,7 +275,7 @@ open class SideTabBar: UIView {
             }
 
             var previousSectionCount: Int = 0
-            if let avatar = avatarView, !avatar.isHidden {
+            if let avatar = avatar, !avatar.view.isHidden {
```
mischreiber, line 278: Shouldn't need the `!` anymore

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -0,0 +1,646 @@
...
+        let row = AvatarDemoSections.allCases[indexPath.section].rows[indexPath.row]
```
mischreiber, line 37: Could just be `let row = section.rows[indexPath.row]`

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -0,0 +1,646 @@
...
+        case .ring:
+            self.isShowingRings = isOn
+        case .ringInnerGap:
+            self.isShowingRingInnerGap = isOn
+        case .transparency:
+            self.isTransparent = isOn
+        }
+    }
+
+    private enum AvatarDemoSections: CaseIterable {
```
mischreiber, line 390: Nit: enum should be singular, so `AvatarDemoSection` and `AvatarDemoRow`

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -0,0 +1,646 @@
...
+        }
+
+        return presence!
+    }
+
+    @objc private func toggleEnablePointerInteraction(switchView: UISwitch) {
+        isPointerInteractionEnabled = switchView.isOn
+    }
+
+    @objc private func toggleShowPresence(switchView: UISwitch) {
```
mischreiber, line 294: Doesn't seem like any of these private `toggle*` functions are being used anymore

rdeassis:
> Noticed a weird animation on `Avatar` when adding and removing the status badge. Regression from legacy?
> 
>  RPReplay_Final1627667057.mov



> Noticed a weird animation on `Avatar` when adding and removing the status badge. Regression from legacy?
> 
>  RPReplay_Final1627667057.mov

As we spoke before, this is an issue caused by the change of the outer frame of the avatar which applies for any transition that causes a change on that frame. I'll issue a separate PR to fix this after the merge is done.

ios/FluentUI.Demo/FluentUI.Demo/Demos/SideTabBarDemoController.swift
```diff
@@ -227,7 +228,7 @@ extension SideTabBarDemoController: SideTabBarDelegate {
         contentViewController?.present(alert, animated: true)
     }
 
-    func sideTabBar(_ sideTabBar: SideTabBar, didActivate avatarView: AvatarView) {
+    func sideTabBar(_ sideTabBar: SideTabBar, didActivate avatarView: MSFAvatar) {
```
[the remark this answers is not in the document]
rdeassis: The naming standard in FluentUI is something that has been changed back and forth a couple of times.
The reason we decided to prefix types with MSF in Vnext was because of the Button struct in SwiftUI that clashes with the FluentUI Button.
This is the reason why the FluentUI SwiftUI button is MSFButtonView and the UIKit wrapper is MSFButton.

We can take the discussion of standards offline and update the entire code base with the decision we make on a separate work item.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -4,67 +4,492 @@
...
-        self.showsBorder = showsBorder
+        self.state = state
+        self.tokens = state.tokens
     }
 
-    @objc public convenience init(primaryText: String = "", secondaryText: String = "", image: UIImage? = nil, presence: Presence = .none, color: UIColor? = nil) {
-        self.init(primaryText: primaryText, secondaryText: secondaryText, image: image, presence: presence, color: color, showsBorder: false)
+    // This initializer should be used by internal container views. These containers should first initialize
+    // MSFAvatarStateImpl using style and size, and then use that state and this initializer in their ViewBuilder.
+    internal init(_ avatarState: MSFAvatarStateImpl) {
```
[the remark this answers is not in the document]
rdeassis: Yeah. Internal is the default access modifier here and our standard is to use the default unless we need to define it explicitly (which is not the case here, so I'll remove it).

ios/FluentUI/Tab Bar/SideTabBar.swift
rdeassis: I only want to validate the avatar because I don't want to potentially call the delegate callback with a nil instance.
Ideally we shouldn't have this optional func on the protocol and should provide a default implementation as the pattern suggests: https://docs.swift.org/swift-book/LanguageGuide/Protocols.html

We can leave the optional chaining for both the delegate instance and the optional func in the protocol here.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
rdeassis: Thanks for catching! Yes, they were all replaced by the updateSetting func.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -4,67 +4,492 @@
 //
 
 import UIKit
+import SwiftUI
 
-// MARK: Avatar
+/// UIKit wrapper that exposes the SwiftUI Avatar implementation.
+@objc open class MSFAvatar: NSObject, FluentUIWindowProvider {
```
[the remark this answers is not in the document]
rdeassis: I think we can do that as a separate PR. We can include that in the standards along with the naming convention discussed in another comment.

mischreiber (approved):

PR #660, opened 2021-07-30
[FHL] [vNext] Avatar Group SwiftUI Demo Controller

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

First pass implementation of SwiftUI demo controller for Avatar Group

### Verification

| Light | Dark |
|------|------|
| light | dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/660)


ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController_SwiftUI.swift
```diff
@@ -0,0 +1,176 @@
...
+                Text(".stack").tag(MSFAvatarGroupStyle.stack)
+            })
+            .pickerStyle(SegmentedPickerStyle())
+            .labelsHidden()
+            .frame(maxWidth: .infinity, alignment: .leading)
+
+            HStack {
+                Text("Avatar Count")
+
+                Button("+") {
```
harrieshin, line 69: can we not use text + it looks bad in larger text? either use the image or use the actual stepper? also nit pick - on left and + on right?

sophialee0416: Something that was also mentioned by @rdeassis offline was maybe using the AG protocol methods (create/remove avatars) to determine the avatar count?

mischreiber: @harrieshin agreed! I'd pulled those from the previous demo but I much prefer your suggestions.

@sophialee0416 good idea! those are not being tested right now, and this would be a great chance to do so.

mischreiber: @sophialee0416 Actually following up on your feedback: this PR removes those APIs from `AvatarGroup` directly and moves them to the `MSFAvatarGroup` wrapper, so there's no concept of calling `create`/`get`/`remove` from the SwiftUI object's state. 

Instead, users will simply update `count` to reflect the new count of avatars, and the generator will re-run to fetch the new values. 

Meanwhile, updating the properties of a single Avatar are done by simply updating the properties that were used to generate that Avatar, and SwiftUI will manage the refresh automatically. This is what we're doing with the various `avatarSettings` toggles in this file.

Let me know if you want more info and I'd be happy to sit down and go over it all in detail!

ios/FluentUI/AvatarGroup/AvatarGroupModifiers.swift
```diff
@@ -0,0 +1,19 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+public extension AvatarGroup {
```
harrieshin, line 9: why do we need this as separate extension?  it can't be part of avatar group directly?

mischreiber: No technical reason, but to me it feels nice that there's a separate place where only view modifiers are listed. @rdeassis were there any additional reasons that you separated them out for `Avatar` or was it all code organization?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -145,16 +189,26 @@ public struct AvatarGroup: View {
...
 
+    var avatars: [Avatar] {
+        return (0..<state.avatarCount).map { index in
+            self.state.avatarAtIndex(index)
+        }
+    }
+
     public var body: some View {
-        let avatars: [MSFAvatarStateImpl] = state.avatars
+        let avatars = self.avatars
```
harrieshin, line 211: we need this?

mischreiber: Yep! `self.avatars` is a calculated property, so it'll generate the list of `Avatar` views on every call. We reference this array several times in this property, so it's important to cache it locally, and I intentionally used the same variable name so we can't accidentally fall back on the generators again by looking for an array called `avatars` (that'll now default to the local one; only an explicit `self.avatars` will trigger regeneration.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController_SwiftUI.swift
```diff
@@ -0,0 +1,181 @@
...
+            .frame(maxWidth: .infinity, alignment: .leading)
+        }
+
+    }
+
+    public var body: some View {
+        VStack {
+            AvatarGroup(style: style, size: size, avatarCount: avatarCount) { index in
+                let persona = samplePersonas[index]
+                Avatar(style: .default,
```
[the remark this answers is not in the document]
rdeassis: The fact that we can use the Avatar SwiftUI View directly exposes functionality that we want to restrict through the state object.
For example, I can create avatars with presence here (even though it's not supported and the MSFAvatarGroupAvatarState does not expose presence on purpose).

For the screenshot below, I have just added a .presence(.available) modifier in line 156.

Screen Shot 2021-08-04 at 4 00 41 PM

rdeassis (changes requested):

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController_SwiftUI.swift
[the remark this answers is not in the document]
harrieshin: ^ I kind of like that you can add available like that tho in this scenario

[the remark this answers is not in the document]
rdeassis: I thought this was an unsupported scenario from previous conversations related to the cutout of the avatar that needed to include the presence outline.

But even if this was supported, there are other ways that the client code could do something we don't want:

Like changing the size or the style of individual avatars:

Screen Shot 2021-08-05 at 8 09 23 AM

Screen Shot 2021-08-05 at 8 11 08 AM

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController_SwiftUI.swift
```diff
@@ -0,0 +1,179 @@
...
+                Text(".xsmall").tag(MSFAvatarSize.xsmall)
+            }
+            .frame(maxWidth: .infinity, alignment: .leading)
+        }
+
+    }
+
+    public var body: some View {
+        VStack {
+            AvatarGroup(style: style, size: size, avatarCount: avatarCount) { avatar, index in
```
rdeassis, line 149: What I had in mind to work with the states in the creation of the AvatarGroup would be something like:

AvatarGroup(style: style,
size: size,
**avatarBuilder: (avatarGroupState: MSFAvatarGroupState) -> void**)

This way the calling code can implement the closure to iterate on any logic it wants and create the initial avatars directly from the avatarGroupState object.

mischreiber: So in this version, they'd be responsible for looping and calling `createAvatar()` on the passed-in `MSFAvatarGroupState` directly?

Interesting proposal, but I have two concerns:
1. Something feels a bit "off" about asking people to run procedural/OO method calls in the middle of a declarative statement like this. Mixing of metaphors and all that. I'd ideally prefer something that is purely declarative if at all possible.
2. I'm not sure how I feel about exposing `MSFAvatarGroupState` like this; we've been good so far about not showing our state to our SwiftUI consumers.

I toyed with having the closure yield an array of `MSFAvatarGroupAvatarState` objects, but we don't really have a clean way of yielding these without the `create` method you referenced above.

I feel like we're so close to a great paradigm...

mischreiber:
Closing for now; the code's drifted too far. I'll redo this PR at some point in the future!

PR #667, opened 2021-08-10
Add simple badge on Mac

huwilkes:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Adding a new Badge View for Mac

### Verification

The new badge displays the provided text, reads out the text as the accessibility label, and changes colors in light and dark mode.

| BadgeView_light | BadgeView_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/667)


macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+// MARK: - BadgeView
+
+@available(*, deprecated, renamed: "BadgeView")
```
harrieshin, line 10: no need

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	public required init?(coder aDecoder: NSCoder) {
+		preconditionFailure("init(coder:) has not been implemented")
+	}
+
+	/// Initializes a Fluent UI Badge View with the provided title, style, and size
+	/// - Parameters:
+	///   - title: String displayed in the badge
+	///   - style: The BadgeViewStyle, currenly only default is supported
```
harrieshin, line 29: why even have this param if default is the only thing that is supported

harrieshin: same with size

huwilkes: I thought it'd make it easier to add the other styles and sizes later if we wanted to.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+@objc(MSFBadgeViewStyle)
+public enum BadgeViewStyle: Int, CaseIterable {
+	case `default`
+}
+
+// MARK: - Colors
+
+private extension Colors {
+	struct Badge {
+		fileprivate static let defaultBackground: NSColor = Palette.communicationBlueTint40.color
```
harrieshin, line 95: well i guess for FRE cases we want it to be communication blue. but most "app" cases wouldn't be the branding color so we should directly use Primary in Colors? we could also expose background and text color so it can be customizable by clients? i think we have that for ios as well.

macos/FluentUI/Badge/BadgeView.swift
harrieshin: is that the direction that design team is going? otherwise, it is bit confusing we have that api that dosn't do anything. once we have those variation it make sense to have this convenience method added to public.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+private struct BadgeViewSizeParameters {
+	fileprivate let fontSize: CGFloat
+	fileprivate let cornerRadius: CGFloat
+	fileprivate let verticalPadding: CGFloat
+	fileprivate let horizontalPadding: CGFloat
+
+	static let small = BadgeViewSizeParameters(
+		fontSize: 11,
+		cornerRadius: 3,
+		verticalPadding: 3,
```
harrieshin, line 116: i am surprised we have paddings that aren't snapping to 4 pix grid.  is this correct?

MicrosoftFluentUI.podspec
```diff
@@ -329,6 +329,12 @@ xcodebuild ${XCODEBUILDPARAMS} -project ${PROJECT_FILE_PATH} -target "MicrosoftF
     avatarview_mac.source_files = ["macos/FluentUI/AvatarView/**/*.{swift,h}"]
   end
 
+  s.subspec 'BadgeView_mac' do |badgeview_mac|
+    badgeview_mac.platform = :osx
+    badgeview_mac.dependency 'MicrosoftFluentUI/Core_mac'
+    badgeview_mac.source_files = ["macos/FluentUI/BadgeView/**/*.{swift,h}"]
```
rdeassis, line 335: This seems to be the reason you have pod lib lint failing in the validation checks.
The pattern is looking for macos/FluentUI/**BadgeView** while you added files in the macos/FluentUI/**Badge** folder.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+			textField.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -horizontalPadding),
+			textField.topAnchor.constraint(equalTo: topAnchor, constant: verticalPadding),
+			textField.bottomAnchor.constraint(equalTo: bottomAnchor, constant: -verticalPadding)
+		])
+	}
+
+	private var textField: NSTextField
+	private var style: BadgeViewStyle
+
+	open override func updateLayer() {
```
rdeassis, line 75: nit: move this method definition up as per the file organization standard:

https://github.com/microsoft/swift-guide/blob/gh-pages/FileOrganization.md

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+			textField.textColor = Colors.Badge.defaultText
+		}
+	}
+}
+
+// MARK: - Style
+
+@objc(MSFBadgeViewStyle)
+public enum BadgeViewStyle: Int, CaseIterable {
+	case `default`
```
rdeassis, line 88: I think we can have at least "default" and "primary" styles if you end up implementing the communication blue and primary color styles as per @harrieshin's comments on color.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+
+// MARK: - Size
+
+@objc(MSFBadgeViewSize)
+public enum BadgeViewSize: Int, CaseIterable {
+	case small
+}
+
+private struct BadgeViewSizeParameters {
+	fileprivate let fontSize: CGFloat
```
rdeassis, line 108: I don't think you need fileprivate on these constants here, do you?

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+// MARK: - Style
+
+@objc(MSFBadgeViewStyle)
+public enum BadgeViewStyle: Int, CaseIterable {
+	case `default`
+}
+
+// MARK: - Colors
+
+private extension Colors {
```
rdeassis, line 93: If you omit the "private" modifier, the extension is internal to the module by default (which has been the standard we're following).
Also, I think you don't need the fileprivate on the constants....

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,126 @@
...
+		verticalPadding: 3,
+		horizontalPadding: 8
+	)
+
+	static func parameters(forSize: BadgeViewSize) -> BadgeViewSizeParameters {
+		switch forSize {
+		case .small:
+			return .small
+		}
+	}
```
Saadnajmi, line 125: I think you can implement a function on an enum, so you could move this function to inside theBadgeViewSize enum

Saadnajmi, line 125: I think you can implement a function on an enum, so you could move this function to inside theBadgeViewSize enum

macos/FluentUI/Badge/BadgeView.swift
huwilkes: That's what the design was. The total height of the control ends up being 20, so that could be why.

macos/FluentUI/Badge/BadgeView.swift
huwilkes: This is how it was set up in the Button. Though, seems like generally our styling for Mac Fluent is getting out of date, so I probably should've mostly focused on matching the iOS file styling.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,159 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+public typealias MSBadgeView = BadgeView
```
harrieshin, line 8: dont need this

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,159 @@
...
+	}
+
+	private func updateTextColors() {
+		textField.textColor = textColor
+	}
+}
+
+// MARK: - Colors
+
+extension Colors {
```
harrieshin, line 152: lets avoid making it color extension and just use private stuct?

macos/FluentUI/Badge/BadgeView.swift
huwilkes: Amy says the overall height of 20 is in fact the goal.

macos/FluentUI/Badge/BadgeView.swift
```diff
@@ -0,0 +1,154 @@
...
+			case .small:
+				return 4
+			}
+		}
+	}
+}
+
+// MARK: - Colors
+
+struct BadgeColors {
```
harrieshin, line 149: might as well explictly set it private so only this file can access.

harrieshin (approved):

harrieshin:
can you update the PR title to more descriptive text rather than your branch name please?

macos/FluentUI/Badge/BadgeView.swift
rdeassis: Or maybe even moving the struct definition to the BadgeView class and mark it private.
Maybe rename it to "Constants"?

rdeassis (approved):

PR #668, opened 2021-08-10
Fixing safe area insets when the drawer is being dragged vertically out of screen bounds.

### LINKED ISSUE #620: The safe area is not being considered while the gesture is moving the drawer
### Environment Information

- **Platform**:
  - [x] iOS
  - [ ] macOS
- **Package version(s)**: (fill this out, include which package manager you're using)
- **Xcode and OS versions**: (fill this out if relevant)

#### Actual behavior:
The safe area is not being considered while the gesture is moving the drawer

#### Expected behavior:
Consider the safe area while the gesture is moving the drawer (decreasing size)

### Priorities and help requested:

Are you willing to submit a PR to fix? Can't:(

Requested priority: Normal

Products/applications affected: Power BI

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Fixes #620.
The root cause of this issue is that the safe area insets are only provided when the view is entirely within the screen bounds.
When the user drags the drawer vertically to out of the screen, the safe area insets need to be compensated by setting the additionalSafeAreaInsets property. 

### Verification

Tested the repro scenarios and validated they were fixed and didn't cause regressions on the combination of the following matrix:

- iOS 13 and 14.
- iPhone (notch device) vs regular device
- iPad

**Before (issue reproduced):**

https://user-images.githubusercontent.com/68076145/128812729-1399a184-a0f4-46b0-b425-197a7e0fb139.mov

**After (issue fixed and validated):**

https://user-images.githubusercontent.com/68076145/128812751-a2ac98d1-993a-484d-b583-c09f5ff772ea.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/668)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -292,7 +292,23 @@ class DrawerPresentationController: UIPresentationController {
         contentView.frame = frame
 
         if let presentedView = presentedView {
-            presentedView.frame = frameForPresentedViewController(in: presentedView.superview == containerView ? contentView.frame : contentView.bounds)
+            let presentedViewFrame = frameForPresentedViewController(in: presentedView.superview == containerView ? contentView.frame : contentView.bounds)
+
+            let isVerticallyPresentedViewPartiallyOffScreen: Bool = {
+                // Calculates the origin of the presentedView frame in relation to the device screen.
+                guard let origin = presentedView.superview?.convert(presentedViewFrame.origin, to: nil) else {
+                    return false
+                }
+
+                return (presentationDirection == .down && origin.y < 0) ||
+                       (presentationDirection == .up && (origin.y + presentedViewFrame.height - UIScreen.main.bounds.height) > 0)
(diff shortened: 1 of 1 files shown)


harrieshin:
this logic is fine even if when drawercontroller has uitableviewcells like popupmenu? because cell should be edge to edge and cell's content inset should have safearea. for stackview of buttons, your logic makes sense.

rdeassis:
> this logic is fine even if when drawercontroller has uitableviewcells like popupmenu? because cell should be edge to edge and cell's content inset should have safearea. for stackview of buttons, your logic makes sense.

Yes. It just sets the insets for the entire presented controller, so it applies to a tableview if that's the content of that view controller too.

Just to double check, this is the result that you're referring to, right?

Screen Shot 2021-08-10 at 11 07 55 AM

harrieshin (approved):

PR #670, opened 2021-08-10
Update BadgeField to respect dynamic text sizing

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Removed fixed-height assumptions in `BadgeField`, allowing the UI to respond to dynamic text sizing.
- Replaced instances of `UILabel` with `FluentUI.Label`, resulting in automatic scaling.

### Verification

Verified the following:
- Small, large, and extra-extra large sizing
- UI reflowed appropriately when resized while visible
- Adding new badges via textfield input
- Dark mode

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/670)


ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -786,6 +795,11 @@ open class BadgeField: UIView {
         originalDeviceOrientation = UIDevice.current.orientation
     }
 
+    @objc private func handleContentSizeCategoryDidChange() {
+        invalidateIntrinsicContentSize()
```
harrieshin, line 799: should the order swap so you calculate the height first before triggering invalidation?

harrieshin:
while this is a fine change and we should merge it, is there a limitation of this control to actually use auto layout rather than frame calculation?

harrieshin (approved):

mischreiber:
> while this is a fine change and we should merge it, is there a limitation of this control to actually use auto layout rather than frame calculation?

I think the problem is that the badges flow from line to line, which is really hard to do with auto layout. If I were to rewrite this today, I'd explore using a `UICollectionViewFlowLayout`(though it's possible that the original author did, and found some limitation).

ios/FluentUI/Badge Field/BadgeField.swift
mischreiber: Shouldn't matter. `invalidateIntrinsicContentSize` sets a flag on the view that will require a re-layout on the next layout pass (effectively the next iteration of the run loop), but doesn't directly trigger a synchronous layout. And all this is done on the main thread, so there's no race condition to worry about.

rdeassis (approved):

rdeassis:
Shouldn't the badges also have unlimited number of lines in the scenario where the badge field has unlimited number of lines?

ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -436,18 +438,20 @@ open class BadgeField: UIView {
         }
     }
 
-    @objc open func heightThatFits(numberOfLines: Int) -> CGFloat {
+    @objc open func heightThatFits(badgeHeight: CGFloat, numberOfLines: Int) -> CGFloat {
         // Vertical margin not applicable to top line
-        let heightForAllLines = CGFloat(numberOfLines) * (Constants.badgeHeight + Constants.badgeMarginVertical)
+        let heightForAllLines = CGFloat(numberOfLines) * (badgeHeight + Constants.badgeMarginVertical)
```
rdeassis, line 443: Shouldn't the badges also have unlimited number of lines in the scenario where the badge field has unlimited number of lines?

mischreiber: Not sure. I've never seen a `BadgeView` with multiple line support, and even if they did, there's still a difference between "let `BadgeView`s flow across multiple lines" and "each `BadgeView` can use as many lines as it wants".

I'm inclined to leave the behavior as-is, but that might be an interesting follow-up for design in the future.

PR #671, opened 2021-08-11
try to see if this changes the podpublish to be always be running

harrieshin:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

follow up from https://github.com/microsoft/fluentui-apple/pull/650 to see if every merge to main branch won't trigger podpublish.

### Verification

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/671)

### DIFF
--- a/.github/workflows/podPublish.yml
+++ b/.github/workflows/podPublish.yml
@@ -2,9 +2,10 @@ name: Pod-Publish
 on:
   push:
     branches:
-      - main
-      - main_0.1
-      - main_0.2
+      # Push events to branches matching refs/heads/
+      - 'main'
+      - 'main_0.1'
+      - 'main_0.2'
     tags:
       - 0.[0-9]+.[0-9]+ # Run automatically on new tags matching a 0.N.M pattern
       - 0.[0-9]+.[0-9]+_main*


.github/workflows/podPublish.yml
```diff
@@ -2,9 +2,10 @@ name: Pod-Publish
 on:
   push:
     branches:
-      - main
-      - main_0.1
-      - main_0.2
+      # Push events to branches matching refs/heads/
+      - 'main'
```
rdeassis, line 6: Just so I understand the goal of these changes:

Is there any real difference between listing the branch names in single quotes or this is just a 'no-op' test to force a merge?

harrieshin: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-branches-and-tags 
but i think it will filter out with origin main vs actual ref/heads/main

rdeassis (approved):

PR #675, opened 2021-08-12
Button Vnext SwiftUI support, improvements.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change adds SwiftUI support to the Button Vnext:
- Creates the MSFButtonState protocol
- Renames the class MSFButtonState to MSFButtonStateImpl and makes it internal.
- Moves the MSFButtonTokens into the state object.
- Creates SwiftUI modifiers and init parameters so that the Button can be used in a SwiftUI environment.
- Updates the UIKit Demo controller to use UITableView.
- Creates the SwiftUI Demo controller.

### Verification

| Light Mode                                       | Dark mode                                      |
| Screen Shot 2021-08-11 at 7 09 57 PM | Screen Shot 2021-08-11 at 7 10 18 PM |
| Screen Shot 2021-08-11 at 7 11 58 PM | Screen Shot 2021-08-11 at 7 12 20 PM |

https://user-images.githubusercontent.com/68076145/129128180-44d10df6-bf56-4590-8a12-2581926c798e.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/675)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
@@ -1,0 +1,110 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import SwiftUI
+import UIKit
+
+class ButtonDemoControllerSwiftUI: UIHostingController<ButtonDemoView> {
+    override init?(coder aDecoder: NSCoder, rootView: ButtonDemoView) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    @objc required dynamic init?(coder aDecoder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    init() {
+        super.init(rootView: ButtonDemoView())
+        self.title = "Button Vnext (SwiftUI)"
+    }
+}
+
+struct ButtonDemoView: View {
+    @State var isDisabled: Bool = false
+    @State var text: String = "Button"
+    @State var showImage: Bool = true
+    @State var showLabel: Bool = true
+    @State var showAlert: Bool = false
+    @State var size: MSFButtonSize = .large
+    @State var style: MSFButtonStyle = .accentFloating
+
+    public var body: some View {
+        VStack {
+            MSFButtonView(action: {
+                showAlert = true
+            },
+            style: style,
+            size: size,
+            image: showImage ? UIImage(named: "Placeholder_24") : nil,
+            text: showLabel ? text : nil)
+            .isDisabled(isDisabled)
+            .fixedSize()
+            .padding()
+            .alert(isPresented: $showAlert, content: {
+                Alert(title: Text("Button tapped"))
+            })
+
+            ScrollView {
+                Group {
+                    Group {
+                        VStack(spacing: 0) {
+                            Text("Content")
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                                .font(.title)
+                            Divider()
+                        }
+
+                        TextField("Text", text: $text)
+                            .autocapitalization(.none)
+                            .disableAutocorrection(true)
+                            .textFieldStyle(RoundedBorderTextFieldStyle())
+
+                        FluentUIDemoToggle(titleKey: "Show image", isOn: $showImage)
+                        FluentUIDemoToggle(titleKey: "Show label", isOn: $showLabel)
+                        FluentUIDemoToggle(titleKey: "Disable", isOn: $isDisabled)
+                    }
+
+                    Group {
+                        VStack(spacing: 0) {
+                            Text("Style")
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                                .font(.title)
+                            Divider()
+                        }
+
+                        Picker(selection: $style, label: EmptyView()) {
+                            Text(".primary").tag(MSFButtonStyle.primary)
+                            Text(".secondary").tag(MSFButtonStyle.secondary)
+                            Text(".ghost").tag(MSFButtonStyle.ghost)
+                            Text(".accentFloating").tag(MSFButtonStyle.accentFloating)
+                            Text(".subtleFloating").tag(MSFButtonStyle.subtleFloating)
+                        }
+                        .labelsHidden()
+                        .frame(maxWidth: .infinity, alignment: .leading)
+                    }
+
+                    Group {
+                        VStack(spacing: 0) {
+                            Text("Size")
+                                .frame(maxWidth: .infinity, alignment: .leading)
+                                .font(.title)
+                            Divider()
+                        }
+
+                        Picker(selection: $size, label: EmptyView()) {
+                            Text(".large").tag(MSFButtonSize.large)
+                            Text(".medium").tag(MSFButtonSize.medium)
+                            Text(".small").tag(MSFButtonSize.small)
+                        }
+                        .labelsHidden()
+                        .frame(maxWidth: .infinity, alignment: .leading)
+                    }
+                }
+                .padding()
+            }
+        }
+    }
+}
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
@@ -33,13 +33,12 @@ struct ButtonDemoView: View {
 
     public var body: some View {
         VStack {
-            MSFButtonView(action: {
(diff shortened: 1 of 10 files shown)


ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -3,21 +3,101 @@
...
 import SwiftUI
+import UIKit
+
+/// View that represents the button
+public struct MSFButtonView: View {
+    @Environment(\.theme) var theme: FluentUIStyle
+    @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+    @ObservedObject var tokens: MSFButtonTokens
+    @ObservedObject var state: MSFButtonStateImpl
+    var action: () -> Void
```
mischreiber, line 15: Should the action be part of the state? I think that's how we've done other controls so far

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -3,21 +3,101 @@
...
+
+    public init(action: @escaping () -> Void,
```
mischreiber, line 17: Might be nice to have the action be the final parameter to the initializer so we can create the control with a trailing closure:
```swift
MSFButton(text: "Title") {
    // action stuff!
}
```

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -3,21 +3,101 @@
...
+            .disabled(state.isDisabled)
+            .frame(maxWidth: .infinity)
+            .designTokens(tokens,
+                          from: theme,
+                          with: windowProvider)
+    }
+}
+
+/// Properties that can be used to customize the appearance of the Button.
+@objc public protocol MSFButtonState {
```
mischreiber, line 43: Nit: other finalized vNext classes have put the state protocol above the view struct. Nice opportunity for alignment/readability.

ios/FluentUI/Vnext/Button/ButtonModifiers.swift
```diff
@@ -0,0 +1,26 @@
...
+    /// - Returns: The modified Button with the property set.
+    func accessibilityLabel(_ accessibilityLabel: String?) -> MSFButtonView {
+        state.accessibilityLabel = accessibilityLabel
+        return self
+    }
+
+    /// Controls whether the button is available for user interaction, renders the control accordingly.
+    /// - Parameter isDisabled: Boolean value to set the property.
+    /// - Returns: The modified Button with the property set.
+    func isDisabled(_ isDisabled: Bool) -> MSFButtonView {
```
mischreiber, line 22: Apple has a View extension called `disabled`; can we utilize/override that?

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -0,0 +1,104 @@
...
+        return hostingController.view
+    }
+
+    @objc public convenience init(style: MSFButtonStyle = .secondary,
+                                  size: MSFButtonSize = .large,
+                                  action: ((_ sender: MSFButton) -> Void)?) {
+        self.init(style: style,
+                  size: size,
+                  action: action,
+                  theme: nil)
```
mischreiber, line 31: While we're here, can we consolidate these `init` functions? The designated initializer already has a default `nil` theme, so I think this `convenience init` can just be removed.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
```diff
@@ -0,0 +1,110 @@
...
+    @State var text: String = "Button"
+    @State var showImage: Bool = true
+    @State var showLabel: Bool = true
+    @State var showAlert: Bool = false
+    @State var size: MSFButtonSize = .large
+    @State var style: MSFButtonStyle = .accentFloating
+
+    public var body: some View {
+        VStack {
+            MSFButtonView(action: {
```
mischreiber, line 36: I noticed this button isn't responding to theme color changes (it's still blue when the main vNext view is green). In fact, that seems to be a problem for all of our SwiftUI demos...

mischreiber (changes requested):

ios/FluentUI/Vnext/Button/MSFButton.swift
rdeassis: The problem with just setting the default to nil in objc exposed method signatures is that the parameter is required, which makes it more verbose to call in objc.

ios/FluentUI/Vnext/Button/ButtonModifiers.swift
rdeassis: We can't override members in extensions in Swift.
We use the disabled modifier in the SwiftUI button in the implementation:

 ```
   public var body: some View {
        Button(action: action, label: {})
            .buttonStyle(MSFButtonViewButtonStyle(tokens: tokens,
                                                  state: state))
            .disabled(state.isDisabled)
            .frame(maxWidth: .infinity)
            .designTokens(tokens,
                          from: theme,
                          with: windowProvider)
    }
```

File: ios/FluentUI/Vnext/Button/Button.swift (hunk 1 above)
Comment (rdeassis) [reply ios/FluentUI/Vnext/Button/Button.swift:15]: The challenge here is that we want to provide a sender parameter of type MSFButton to UIKit callers, but SwiftUI does not care about that and should be just an "() -> void" closure.

If we move the action to the state, we give up the control of providing the MSFButton sender param in the UIKit environment as the client code will be setting the closure directly and would need to capture the MSFButton instance on the closure event.

I think I'm okay with making the action closure simple and just have it "() -> void" everywhere, but this deserves its own separate PR.

File: ios/FluentUI/Vnext/Button/Button.swift [hunk 7]
```diff
@@ -3,21 +3,101 @@
 //  Licensed under the MIT License.
 //
 
-import Combine
-import UIKit
 import SwiftUI
+import UIKit
+
+/// View that represents the button
+public struct MSFButtonView: View {
```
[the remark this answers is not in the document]
rdeassis: Yeah. So far the Button is the only problem as far as naming conflicts we've seen.
I thought on suggesting to change the struct name to just "ButtonView" to prevent extra workarounds to fix the clash.
Do you have any suggestions?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController_SwiftUI.swift
rdeassis: Yes. This is a known issue (likely with the timing that the HostingController is being added to the window). I have created a bug in ADO to track it.

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: After we spoke offline, I've made the smaller change that does not affect the public State protocol. We'll iterate on a discussion to eventually remove the sender from the public API of the UIKit wrapper.
If that happens, we can then make the action a property of the state protocol.

ios/FluentUI/Vnext/Button/ButtonModifiers.swift
rdeassis: As we spoke offline on this one as well:
 - I'm changing the SwiftUI Button to respect the isEnabled environment value.
 - It was necessary to convert the state property to an optional so we can override it from the UIKit consumption side.
 - A non-optional wrapping property was also needed to be able to comply with the objc bridging which doesn't allow Bool? types.

ios/FluentUI/Vnext/Button/Button.swift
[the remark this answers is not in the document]
mischreiber: As discussed offline: we'll continue this conversation separately, since it's a tricky one.

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -0,0 +1,117 @@
...
+    // MARK: - Private members
+
+    private func initialize(style: MSFButtonStyle = .secondary,
+                            size: MSFButtonSize = .large,
+                            action: ((_ sender: MSFButton) -> Void)?,
+                            theme: FluentUIStyle? = nil) {
+        self.action = action
+        buttonView = MSFButtonView(style: style,
+                                   size: size) {
+            self.action?(self)
```
mischreiber, line 82: This should be a `[weak self]` capture to avoid a circular retain (glad we discussed that this morning, we missed this before!)

mischreiber (approved):

ios/FluentUI/Vnext/Button/MSFButton.swift
rdeassis: Good catch! Thanks!

PR #686, opened 2021-08-17
Add support for showing tooltip to navbar buttons

vickprat:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Expose left and right item button stack view in the navigation bar.

### Verification
The change was tested manually and also added functionality to the demo controller. Attached screenshots.

Screenshot 2021-08-18 at 1 59 48 PM 
Screenshot 2021-08-18 at 2 00 40 PM
Screenshot 2021-08-18 at 1 58 32 PM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/686)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -382,6 +382,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
             }
             cell.setup(action1Title: "Show tooltip on 3 day view button in navbar")
             cell.action1Button.addTarget(self, action: #selector(showTooltipButtonPressed), for: .touchUpInside)
+            cell.bottomSeparatorType = .full
             return cell
         }
 
@@ -458,7 +459,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
 
     @objc private func showTooltipButtonPressed() {
         let navigationBar = msfNavigationController?.msfNavigationBar
-        guard let view = navigationBar?.barButtonItemView(tag: 1) else {
+        guard let view = navigationBar?.barButtonItemView(with: 1) else {
             return
         }
         Tooltip.shared.show(with: "Tap anywhere for this tooltip to dismiss.",
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -237,6 +237,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         tableView.delegate = self
         tableView.register(TableViewCell.self, forCellReuseIdentifier: TableViewCell.identifier)
         tableView.register(BooleanCell.self, forCellReuseIdentifier: BooleanCell.identifier)
+        tableView.register(ActionsCell.self, forCellReuseIdentifier: ActionsCell.identifier)
         return tableView
     }()
 
@@ -376,14 +377,11 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
 
         if indexPath.row == 2 {
-            guard let cell = tableView.dequeueReusableCell(withIdentifier: BooleanCell.identifier, for: indexPath) as? BooleanCell else {
+            guard let cell = tableView.dequeueReusableCell(withIdentifier: ActionsCell.identifier, for: indexPath) as? ActionsCell else {
                 return UITableViewCell()
             }
-            cell.setup(title: "Show tooltip on 3 day view button in navigation bar", isOn: false)
-            cell.titleNumberOfLines = 0
-            cell.onValueChanged = { [weak self] in
-                self?.shouldShowTooltip()
-            }
+            cell.setup(action1Title: "Show tooltip on 3 day view button in navbar")
+            cell.action1Button.addTarget(self, action: #selector(showTooltipButtonPressed), for: .touchUpInside)
             return cell
         }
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -241,14 +241,17 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+    /// Making these stack views as public so as to access the arranged subviews of these stack view.
+    /// One use case is to access the arranged subview to show the tooltip.
+    @objc public let rightBarButtonItemsStackView = UIStackView()
```
harrieshin, line 246: I don't think we should directly expose like that. then there is a risk that rightbarbuttonitem and the stack view can be out of sync. shouldn't it be a helper function that returns the UIView similar to how don't expose avatar view directly but give enough functions

vickprat: My bad!! This totally makes sense. I have created a helper function to get the bar button item view with a particular tag.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -156,6 +156,19 @@ open class NavigationBar: UINavigationBar {
         return nil
     }
 
+    @objc public func barButtonItemView(tag: Int) -> UIView? {
+        let leftBarButtonItemViews = leftBarButtonItemsStackView.arrangedSubviews
+        let rightBarButtonItemViews = rightBarButtonItemsStackView.arrangedSubviews
```
harrieshin, line 161: can we just combine line 160-162 into 1 let?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -156,6 +156,19 @@ open class NavigationBar: UINavigationBar {
         return nil
     }
 
+    @objc public func barButtonItemView(tag: Int) -> UIView? {
```
harrieshin, line 159: barbuttontiemview(with tag: Int) might be more swift friendly naming

harrieshin: please add documentation

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -375,6 +376,15 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
             return cell
         }
 
+        if indexPath.row == 2 {
+            guard let cell = tableView.dequeueReusableCell(withIdentifier: ActionsCell.identifier, for: indexPath) as? ActionsCell else {
+                return UITableViewCell()
+            }
+            cell.setup(action1Title: "Show tooltip on 3 day view button in navbar")
+            cell.action1Button.addTarget(self, action: #selector(showTooltipButtonPressed), for: .touchUpInside)
```
harrieshin, line 384: nit: can you add a separator here?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -156,6 +156,19 @@ open class NavigationBar: UINavigationBar {
...
+        let leftBarButtonItemViews = leftBarButtonItemsStackView.arrangedSubviews
+        let rightBarButtonItemViews = rightBarButtonItemsStackView.arrangedSubviews
+        let totalBarButtonItemViews = leftBarButtonItemViews + rightBarButtonItemViews
```
harrieshin, line 162: in regular title mode the tooltip seems be bit off? 
Screen Shot 2021-08-18 at 2 01 37 PM

ios/FluentUI/Navigation/NavigationBar.swift
vickprat: Done

ios/FluentUI/Navigation/NavigationBar.swift
vickprat: Done

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
vickprat: Done

ios/FluentUI/Navigation/NavigationBar.swift
vickprat: Yes, this is a bit off and the reason is the navbar's height which depends on the regular or large title mode so IMO whosoever creating the tooltip should take care of this. In the demo controller, I took care of this by checking for the large title thing and then setting `y` offset as 5 for the tooltip.

There is one more thing we can do is expose an API that will tell the difference between the height of the navbar in large and regular title mode. `Constants.expandedContentHeight - Constants.normalContentHeight`

But, it's not necessary that users of our API(`barButtonItemView(with tag:)`) necessarily want offset equal to the difference between the height of the navbar in regular and large title mode like in our case the difference is `4` based on current constants value but I set offset as `5` in the demo controller in case of regular title mode because that is looking even better. (Check the below screenshots where the offset is `4` in the first screenshot and `5` in second screenshot)
Screenshot 2021-08-19 at 11 21 27 AM

harrieshin: no I don't think that analysis is correct. I think the right and left stack view have their own stack view with the custom uibutton (just like the large title case) but it isn't actually rendered on the screen when it is "system" style navigation bar. the uivew that user actually see is the real uinavigationbar baritembutton provided by iOS.

vickprat: @harrieshin I am not sure where my understanding is wrong because here's what my analysis was based on:
1. `updateViewsForLargeTitlePresentation` method in NavigationBar.swift is showing or hiding the content stack view which contains the left and right bar button items stack views.
2. In the demo controller we have the option to have the regular title for both "system" and "primary" navigation bar styles. Screenshot 2021-08-20 at 1 21 22 PM

Now, if I write code based on title condition:`let offset: CGPoint = navigationItem.usesLargeTitle ? .init(x: 0, y: 0) : .init(x: 0, y: -4)` then it works fine like shown in the screenshots.
| Screenshot 2021-08-20 at 1 25 14 PM | Screenshot 2021-08-20 at 1 25 30 PM |

harrieshin: The point I am trying to convey is that the UIView you are returning for regular title with tag="1" you are actually returning incorrect view. (in the demo app if you say view.background = .yellow for example, you will see it will be no op)the rect happens to be somewhat close so your demo code of doing -4 works for tooltip but if this public function is trying to access the actual view then they are not getting the right UIView to interact with it. so I think the logic of retrieving  the right view needs to be different for showsLargeTitle case.

vickprat: My bad!! I should have made the API to handle both the scenarios(large title or regular title). Can you please check now?

harrieshin (approved):

PR #689, opened 2021-08-19
Add an API to get the current rect of a hero item view

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A client needs to present a popover when a hero button is tapped in the bottom bar. This popover needs to attach to the hero item view, but these are internal and we don't want to expose them to clients. 
To cover this scenario, this PR adds an API to get the current rect of a given hero item view in the coordinate space of `BottomCommandingController`'s root view. This allows clients to position the popover correctly, without us exposing the actual view.

### Verification

Various sanity checks to verify the rects are correct, both in sheet and bar styles.
Added a custom popover example code to the demo controller:

Simulator Screen Shot - iPad Pro (11-inch) (3rd generation) - 2021-08-23 at 16 53 32

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/689)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -226,6 +226,21 @@ open class BottomCommandingController: UIViewController {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
+    /// Current rectangle of the view that represents the given hero item.
+    ///
+    /// - Parameter heroItem: A `CommandingItem` contained in `heroItems`.
+    /// - Returns: The current rectangle in the coordinate system of the receiver.
+    @objc public func rectFor(heroItem: CommandingItem) -> CGRect {
+        guard isViewLoaded,
+              let bindingInfo = itemToBindingMap[heroItem],
+              bindingInfo.location == .heroSet else {
+            return .null
+        }
+
+        let itemView = bindingInfo.view
+        return itemView.convert(itemView.bounds, to: view)
+    }
+
     // MARK: - View building and layout
 
     public override func loadView() {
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -226,6 +226,20 @@ open class BottomCommandingController: UIViewController {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
+    /// Current rectangle of the view that represents the given hero item.
+    ///
+    /// - Parameter heroItem: A `CommandingItem` contained in `heroItems`.
+    /// - Returns: The current rectangle in the coordinate system of the receiver.
+    @objc public func rectFor(heroItem: CommandingItem) -> CGRect {
```
rdeassis, line 233: Does it make sense to add a demo controller scenario using a tooltip to point to this rect?

lcapkovic: As far as I can see the fluent tooltip needs an anchor view and doesn't support anchor rects. I could show a simple popover instead.

rdeassis: I guess my main point is to provide a test/demo scenario for the functionality that you're adding.

lcapkovic: Added the popover example code.

rdeassis: Thank you for adding that example!

rdeassis (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -226,6 +226,20 @@ open class BottomCommandingController: UIViewController {
...
+        guard isViewLoaded,
+              let bindingInfo = itemToBindingMap[heroItem],
+              bindingInfo.location == .heroSet else {
+            return .null
```
anandrajeswaran, line 237: The conditions that return a null rect should probably be documented?

PR #690, opened 2021-08-19
Renaming MSFButtonView (SwiftUI Button View) to FluentButton.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The current SwiftUI View that represents the Fluent Vnext Button is not named in conformance to the standard in FluentUI Vnext.
The main reason for that is that SwiftUI already defines a Button struct for the SwiftUI Button View.
In order to solve the clash and still have the name closer to what the other FluentUI Vnext SwiftUI Views are named after, we're renaming it from **MSFButtonView** to just **FluentButton**.

### Verification

Smoke test on building an running the button Demo controllers.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/690)

### DIFF
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -29,13 +29,13 @@ import UIKit
 }
 
 /// View that represents the button.
-public struct MSFButtonView: View {
+public struct FluentButton: View {
     @Environment(\.theme) var theme: FluentUIStyle
     @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
     @ObservedObject var tokens: MSFButtonTokens
     @ObservedObject var state: MSFButtonStateImpl
 
-    /// Creates a MSFButtonView.
+    /// Creates a FluentButton.
     /// - Parameters:
     ///   - style: The MSFButtonStyle used by the button.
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -29,13 +29,13 @@ import UIKit
 }
 
 /// View that represents the button.
-public struct MSFButtonView: View {
+public struct FluentButton: View {
```
mischreiber, line 32: Did we want to rename this file to `FluentButton.swift` as well? I just realized we didn't discuss this earlier. (I don't have any strong opinion)

mischreiber (approved):

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: I think we can keep this file name as we're using the core "Button" name throughout the other files, including the ButtonModifiers, and the MSFButton everywhere for the objc compatible objects and YAML file too (AP_MSFButtonTokens and its sub APs).

PR #699, opened 2021-08-27
Allow dimming view tap to dismiss the bottom sheet

lukas-cap:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A single tap on the dimming view when the sheet is expanded should cause it to collapse. This PR adds a tap gesture recognizer on the dimming view to do that. I also added a corresponding `BottomSheetInteraction` enum value for correct delegate reporting of the source interaction.

Note - the gesture recognizer is implicitly only active when the dimming view is visible. It receives no touch events when the view alpha is set to 0.

### Verification

Sanity checks in the test app.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/699)


anandrajeswaran:
How is this behavior exposed to VoiceOver users?

PR #700, opened 2021-08-31
Bump Xcode and macOS versions (again)

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

We previously tried to move our minimum required Xcode to 12.5, but ran into two problems:
1. Azure Pipelines did not yet support `macos-11`, which is the minimum supported OS for Xcode 12.5.
2. One of our clients was not yet ready to move to Xcode 12.5, which would have resulted in potential build breaks due to using different versions of Swift compiler.

But now both of these blocking issues are resolved, so we're ready to re-apply @sophialee0416's change and begin using Xcode 12.5.1 as our primary build environment.

Documentation verifying that we can upgrade:

**Microsoft-hosted agents**
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

**Virtual environments:**
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode

### Verification

- Built Static Library and Framework targets
- Basic smoke test of test app
- Verified that ADO integration functions as expected

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/700)

### DIFF
--- a/.ado/templates/fluentui-apple-publish-nuget-job.yml
+++ b/.ado/templates/fluentui-apple-publish-nuget-job.yml
@@ -1,7 +1,7 @@
 jobs:
 - job: fluentui_apple_publish_nuget
   pool:
-    vmImage: 'macos-10.15'
+    vmImage: 'macos-11'
   displayName: FluentUI Apple Publish NuGet
   timeoutInMinutes: 60 # how long to run the job before automatically cancelling
   cancelTimeoutInMinutes: 5 # how much time to give 'run always even if cancelled tasks' before killing them
--- a/README.md
+++ b/README.md
@@ -16,8 +16,8 @@ Fluent UI Apple contains native UIKit and AppKit controls aligned with [Microsof
 #### Requirements
 
 - iOS 13+ or macOS 10.14+
-- Xcode 11+
-- Swift 5.0+
+- Xcode 12.5+
+- Swift 5.4+
 
 #### Using Carthage
 
--- a/scripts/xcode_select_current_version.sh
+++ b/scripts/xcode_select_current_version.sh
@@ -3,7 +3,7 @@
 if [ -n "$XCODE_PATH_OVERRIDE" ]; then # If someone calls this with the XCODE_PATH_OVERRIDE variable set to a path to a developer dir, use it instead
     XCODE_PATH="$XCODE_PATH_OVERRIDE"
 else
-    XCODE_PATH='/Applications/Xcode_12.4.app/Contents/Developer'
+    XCODE_PATH='/Applications/Xcode_12.5.1.app/Contents/Developer'
 fi
 
 echo "Running command: sudo xcode-select --switch $XCODE_PATH"
--- a/scripts/xcode_select_current_version.sh
+++ b/scripts/xcode_select_current_version.sh
@@ -3,7 +3,7 @@
 if [ -n "$XCODE_PATH_OVERRIDE" ]; then # If someone calls this with the XCODE_PATH_OVERRIDE variable set to a path to a developer dir, use it instead
     XCODE_PATH="$XCODE_PATH_OVERRIDE"
 else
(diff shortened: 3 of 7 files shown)


Saadnajmi:
If I read this right, does this mean I can bump FluentUI React Native's pipeline to also use Xcode 12.5 now that azure pipelines supports it?

scripts/xcode_select_current_version.sh
```diff
@@ -3,7 +3,7 @@
 if [ -n "$XCODE_PATH_OVERRIDE" ]; then # If someone calls this with the XCODE_PATH_OVERRIDE variable set to a path to a developer dir, use it instead
     XCODE_PATH="$XCODE_PATH_OVERRIDE"
 else
-    XCODE_PATH='/Applications/Xcode_12.4.app/Contents/Developer'
+    XCODE_PATH='/Applications/Xcode_12.5.1.app/Contents/Developer'
```
Saadnajmi, line 6: is 12.5.1 vs 12.5 important?

rdeassis: 12.5 (12E262) is the default:
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode

I think we should keep that version since it's also the blessed one being used in devmain so we keep consistency where we can.

sophialee0416: I believe it just the most up to date version that needs minimum OS 11?

README.md
```diff
@@ -16,8 +16,8 @@ Fluent UI Apple contains native UIKit and AppKit controls aligned with [Microsof
 #### Requirements
 
 - iOS 13+ or macOS 10.14+
-- Xcode 11+
-- Swift 5.0+
+- Xcode 12.5+
+- Swift 5.4+
```
rdeassis, line 20: The podspec is still on 5.0:

`  s.swift_version = "5.0"`

rdeassis: Also, are there other places in the project where 5.0 is defined where we can bump it too?

sophialee0416 (approved):

.ado/templates/fluentui-apple-publish-nuget-job.yml
```diff
@@ -1,7 +1,7 @@
 jobs:
 - job: fluentui_apple_publish_nuget
   pool:
-    vmImage: 'macos-10.15'
+    vmImage: 'macos-11'
```
rdeassis, line 4: To document the versions in this PR, can you add the links to the description?

**Microsoft-hosted agents**
https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml#software

**Virtual environments:**
https://github.com/actions/virtual-environments/blob/main/images/macos/macos-11-Readme.md#xcode

README.md
Saadnajmi: I think internally, swift version "5.x" is just read as "latest version of Swift 5" so it doesn't actually matter

scripts/xcode_select_current_version.sh
mischreiber: Yeah I was originally thinking "well might as go with the newest available" but there's definitely value to sticking with devmain, so 12.5 it is!

README.md
mischreiber: Hm well just to be safe I'll bump the podspec anyway.

There's one more place I can think of: `Package.swift` which currently names 5.3. I'll bump that to 5.4 as well.

mischreiber:
> If I read this right, does this mean I can bump FluentUI React Native's pipeline to also use Xcode 12.5 now that azure pipelines supports it?

Yes sir! Just make sure you also bump the OS version to `macos-11` at the same time.

rdeassis (approved):

PR #702, opened 2021-09-03
[List vNext] Add animation to expand/collapse trigger

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

List expand/collapse did not have an animation when triggered externally. This fix changes that so the list is able to animate expand/collapse smoothly.

### Verification

| ezgif com-video-to-gif-3 | ezgif com-video-to-gif |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/702)

### DIFF
--- a/ios/FluentUI/Vnext/List/List.swift
+++ b/ios/FluentUI/Vnext/List/List.swift
@@ -63,7 +63,7 @@ public struct MSFListView: View {
                     }
                 }
             }
-            .animation(Animation.default)
+            .animation(.default)
             .environment(\.defaultMinListRowHeight, 0)
             .frame(maxWidth: .infinity, maxHeight: .infinity)
             .designTokens(tokens,
--- a/ios/FluentUI/Vnext/List/List.swift
+++ b/ios/FluentUI/Vnext/List/List.swift
@@ -63,6 +63,7 @@ public struct MSFListView: View {
                     }
                 }
             }
+            .animation(Animation.default)
             .environment(\.defaultMinListRowHeight, 0)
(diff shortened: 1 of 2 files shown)


mischreiber (commented):
Couple thoughts from the animation:
- It looks like the list of status items are still disappearing without animation, even though the list is collapsing smoothly. It's definitely an improvement, but not quite as complete as the expand animation (where the status items fade in from 0). Anything we can do about that?
- If you watch closely when "Busy" becomes "Available", you can see that the frame for the "Available" text is animating too, giving the impression that it's writing the word character by character. Can we suppress that part of the animation?

Overall I think it's an improvement over what we currently have, but if we can make these two additional changes then it'll be flawless.

sophialee0416:
> Couple thoughts from the animation:
> 
> * It looks like the list of status items are still disappearing without animation, even though the list is collapsing smoothly. It's definitely an improvement, but not quite as complete as the expand animation (where the status items fade in from 0). Anything we can do about that?
> * If you watch closely when "Busy" becomes "Available", you can see that the frame for the "Available" text is animating too, giving the impression that it's writing the word character by character. Can we suppress that part of the animation?
> 
> Overall I think it's an improvement over what we currently have, but if we can make these two additional changes then it'll be flawless.

Good eye! Thanks for the catch. I will look into that further!

rdeassis (approved):

PR #705, opened 2021-09-09
[vNext CCB] Tokenizing Command Bar

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Replacing CCB specs with tokens.

### Verification

| Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 17 12 06 | Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 17 09 20 |
| Simulator Screen Shot - iPhone 11 Pro - 2021-09-08 at 17 11 38 | Simulator Screen Shot - iPhone 11 Pro - 2021-09-17 at 14 29 46 |

**Theming:**

https://user-images.githubusercontent.com/22566866/133855210-f53a99d0-570c-49c8-b0ae-a79241c85c61.mov

**UIPointerInteraction Testing**

https://user-images.githubusercontent.com/22566866/134067023-c1771feb-eece-4cab-9553-d15b58d81c98.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/705)


mischreiber (commented):
I tested swapping the color theme from blue to green in the test app, but the CCB did not update. Were you able to verify this scenario on your end?

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,52 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class MSFCommandBarTokens: MSFTokensBase, ObservableObject {
```
rdeassis, line 8: This class does not need to be an ObservableObject as the CCB implementation is not SwiftUI.

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,52 @@
...
+    @Published public var defaultBackgroundColor: UIColor!
```
rdeassis, line 9: Same here for the @Published properties. We don't need these if we're not using it in a SwiftUI View.

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    @objc open func didChangeAppearanceProxy() {
+        updateForCurrentTheme()
+    }
+
+    override func updateForCurrentTheme() {
+        let currentTheme = theme
+        let appearanceProxy = currentTheme.MSFCommandBarTokens
+
+        defaultBackgroundColor = appearanceProxy.backgroundColor.rest
```
rdeassis, line 41: Why are you naming these properties "default" instead of just backgroundColor and iconColor?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -11,7 +11,7 @@ import UIKit
  Set the `delegate` property to determine whether a button can be selected and deselected, and listen to selection changes.
  */
 @objc(MSFCommandBar)
-open class CommandBar: UIView {
+open class CommandBar: UIView, FluentUIWindowProvider {
```
rdeassis, line 14: You need to override [didMoveToWindow()](https://developer.apple.com/documentation/uikit/uiview/1622527-didmovetowindow) to update the tokens as they're based on the current window.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -311,6 +311,16 @@ MSFBorderedCardNudgeTokens extends MSFCardNudgeTokens:
     outlineColor: $Colors.Stroke.neutral1
     textColor: $Colors.Brand.shade20
 
+AP_MSFCommandBarTokens:
```
rdeassis, line 314: We should change the Demo Controller description to "CommandBar (Vnext)" to communicate that this control is tokenized.

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    @Published public var dismissBackgroundColor: UIColor!
+    @Published public var dismissIconColor: UIColor!
+    @Published public var groupBorderRadius: CGFloat!
+    @Published public var groupInterspace: CGFloat!
+    @Published public var iconSize: CGFloat!
+    @Published public var itemInterspace: CGFloat!
+
+    /// Notifies the drawer controller to refresh its UI to reflect its design token values
+    var themeDidUpdate: (() -> Void)?
```
rdeassis, line 23: This is not being used anywhere.

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,52 @@
...
+    @Published public var pressedIconColor: UIColor!
+
+    @Published public var dismissBackgroundColor: UIColor!
+    @Published public var dismissIconColor: UIColor!
+    @Published public var groupBorderRadius: CGFloat!
+    @Published public var groupInterspace: CGFloat!
+    @Published public var iconSize: CGFloat!
+    @Published public var itemInterspace: CGFloat!
+
+    /// Notifies the drawer controller to refresh its UI to reflect its design token values
```
rdeassis, line 22: "drawer controller"

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -28,9 +28,14 @@ open class CommandBar: UIView {
...
     @objc public init(itemGroups: [CommandBarItemGroup], leadingItem: CommandBarItem? = nil, trailingItem: CommandBarItem? = nil) {
         self.itemGroups = itemGroups
+        self.leadingItem = leadingItem
+        self.trailingItem = trailingItem
 
         super.init(frame: .zero)
+        updateView()
+    }
 
+    private func updateView() {
```
rdeassis, line 38: nit: order of members in the file (private should be down below)

https://github.com/microsoft/swift-guide/blob/gh-pages/FileOrganization.md

rdeassis (changes requested):

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -311,6 +311,16 @@ MSFBorderedCardNudgeTokens extends MSFCardNudgeTokens:
...
+  backgroundColor: [ rest: $Colors.Background.neutral4, pressed: $Colors.Background.brandRest ]
+  dismissBackgroundColor: $Colors.Background.neutral1
+  dismissIconColor: "FluentUIColor(light: $Colors.Foreground.neutral1, dark: $Colors.Foreground.neutral3)"
+  groupBorderRadius: $Border.radius.xLarge
+  groupInterspace: $Spacing.medium
+  iconColor: [ rest: $Colors.Foreground.neutral1, pressed: $Colors.Foreground.neutralInverted ]
```
rdeassis, line 320: Don't we also need tokens for the disabled state?

sophialee0416:
> I tested swapping the color theme from blue to green in the test app, but the CCB did not update. Were you able to verify this scenario on your end?

These should be fixed with the new iteration @mischreiber !

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -14,6 +14,12 @@ class CommandBarButton: UIButton {
         }
     }
 
+//    override var isFocused: Bool {
```
harrieshin, line 17: remove?

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,78 @@
...
+    public var pressedBackgroundColor: UIColor!
+    public var pressedIconColor: UIColor!
+    
+    public var selectedBackgroundColor: UIColor!
+    public var selectedIconColor: UIColor!
+    
+    public var disabledBackgroundColor: UIColor!
+    public var disabledIconColor: UIColor!
+
+    public var dismissBackgroundColor: UIColor!
```
harrieshin, line 24: what's dismiss background color for?

sophialee0416: This is used for the keyboard dismiss!

harrieshin: then there is clear mismatch in naming?

harrieshin: the demo app is using for keyboard dismiss but the real button type(style) is fixed button isn't it?

sophialee0416: I believe it means that if the button is not of persist selection type, then it will get the dismissed color properties, sorry if I am getting it confused

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,78 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+class MSFCommandBarTokens: MSFTokensBase {
+    public var backgroundColor: UIColor!
+    public var iconColor: UIColor!
+
+    public var hoverBackgroundColor: UIColor!
```
harrieshin, line 12: where is hover color being used? keyboard focus? pointer color?

sophialee0416: Updated with pointer usage

ios/FluentUI/Core/FluentUIFramework.swift
```diff
@@ -17,7 +17,7 @@ public extension Colors {
     }
 
     internal struct NavigationBar {
-        static var background = UIColor(light: surfacePrimary, dark: gray900)
+        static var background: UIColor = surfacePrimary
```
harrieshin, line 20: this has bigger consequences on all of our navigation bar colors. especially on dark mode split mode cases with dark elevated concepts. should it be part of this change? shouldn't cab has its own background color?

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -311,6 +311,15 @@ MSFBorderedCardNudgeTokens extends MSFCardNudgeTokens:
     outlineColor: $Colors.Stroke.neutral1
     textColor: $Colors.Brand.shade20
 
+AP_MSFCommandBarTokens:
+  backgroundColor: [ rest: $Colors.Background.neutral4, hover: "FluentUIColor(light: $Colors.Background.neutral5, dark: $Colors.Stroke.neutral2)", pressed: "FluentUIColor(light: $Colors.Background.neutralDisabled, dark: $Colors.Background.neutral5)", selected: $Colors.Background.brandRest, disabled: $Colors.Stroke.neutral1 ]
+  dismissIconColor: "FluentUIColor(light: $Colors.Foreground.neutral1, dark: $Colors.Foreground.neutral3)"
```
harrieshin, line 316: still reference on dismissIconColor

ios/FluentUI/Core/FluentUIFramework.swift
```diff
@@ -29,6 +29,8 @@ public extension Colors {
 
     // Objective-C support
     @objc static var navigationBarBackground: UIColor { return NavigationBar.background }
+
+    @objc static var commandBarBackground: UIColor { return surfacePrimary }
```
harrieshin, line 33: why is this here? shouldn't it be part of token pipeline?

harrieshin (approved):

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -0,0 +1,69 @@
...
+    override init() {
+        super.init()
+
+        self.themeAware = true
+
+        updateForCurrentTheme()
+    }
+
+    @objc open func didChangeAppearanceProxy() {
+        updateForCurrentTheme()
```
rdeassis, line 40: I just noticed this issue where the primary brand color is not updated in the keyboard CCB:

https://user-images.githubusercontent.com/68076145/134248265-2340c2bf-725d-46ab-ae5f-652f51c6c66b.mov

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -74,43 +82,90 @@ class CommandBarButton: UIButton {
...
 
+    private var commandBarTokens: MSFCommandBarTokens
+
     private let isPersistSelection: Bool
 
-    private var selectedTintColor: UIColor {
-        guard let window = window else {
-            return UIColor(light: Colors.communicationBlue, dark: .black)
-        }
+    private var hoverBackgroundColor: UIColor {
```
rdeassis, line 89: Do we need these computed properties? Why not just using their values directly?

ios/FluentUI/Command Bar/CommandBarTokens.swift
mischreiber: Good find! Isn't the keyboard in its own `UIWindow` instance? If the accessory bar is attached to that, then it may not have the same root window, and thus has a separate theme.

rdeassis: @mischreiber yes, that's what I suspected too. And it does repro on the current CCB (before tokenization).
@sophialee0416 if this is the case, It'd be good to set the ColorProviding theme for the keyboard window so we have a good sample code for the client apps.

sophialee0416: Resolved offline: will create a separate task for theming mismatch due to different UIWindows as a part of the theming initiative @mischreiber @rdeassis

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -311,6 +311,16 @@ MSFBorderedCardNudgeTokens extends MSFCardNudgeTokens:
     outlineColor: $Colors.Stroke.neutral1
     textColor: $Colors.Brand.shade20
 
+AP_MSFCommandBarTokens:
+  backgroundColor: [ rest: $Colors.Background.neutral4, hover: "FluentUIColor(light: $Colors.Background.neutral5, dark: $Colors.Stroke.neutral2)", pressed: "FluentUIColor(light: $Colors.Background.neutralDisabled, dark: $Colors.Background.neutral5)", selected: $Colors.Background.brandRest, disabled: $Colors.Stroke.neutral1 ]
+  barBackgroundColor: $Colors.Background.neutral1
```
rdeassis, line 316: nit on the naming:
Shouldn't we have **barBackgroundColor** as just "**backgroundColor**"?

Then the ones related to the buttons, we can go with the "item" prefix?
- itemFixedIconColor
- itemIconColor
- itemIconSize
- itemInterspace

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -103,9 +110,9 @@ open class CommandBar: UIView {
         scrollView.translatesAutoresizingMaskIntoConstraints = false
         scrollView.contentInset = UIEdgeInsets(
             top: 0,
-            left: leadingButton == nil ? CommandBar.insets.left : CommandBar.fixedButtonSpacing,
+            left: leadingButton == nil ? CommandBar.insets.left : commandBarTokens.itemInterspace,
```
rdeassis, line 113: nit: store "commandBarTokens.itemIterspace" in a constant (let) and use it in both of these lines (113 and 115).

rdeassis: Same for lines 179, 181, 193 and 195.

rdeassis (approved):

mischreiber (approved):

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -38,12 +39,18 @@ open class CommandBar: UIView {
...
+        backgroundColor = commandBarTokens.backgroundColor
         translatesAutoresizingMaskIntoConstraints = false
 
         configureHierarchy()
         updateButtonsState()
     }
 
+    open override func didMoveToWindow() {
+        super.didMoveToWindow()
+        commandBarTokens.didChangeAppearanceProxy()
```
rdeassis, line 51: Hey @sophialee0416. While reviewing the tokenization of the NotificationView earlier today I made this comment that I missed in the CommandBar review:

Could you create a quick PR to update this call to updateForCurrentTheme() instead?
Thanks!

https://github.com/microsoft/fluentui-apple/pull/715#discussion_r718039047

PR #709, opened 2021-09-16
Interactive hiding of the bottom sheet / bottom commanding

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds support for interactive changes to the `isHidden` property on `BottomSheetController` and `BottomCommandingController`. This allows clients to drive the `isHidden` state change animation, for example through a gesture recognizer or a scroll view drag. 

We don't want to expose the underlying views to let the client interactively manipulate them, because it would break internal state. We instead let them set the `isHidden` property "interactively", which gives them a `UIViewAnimating` object through which they can drive the `fractionComplete` of the hiding / unhiding animation. 
When they want to stop the interactive portion (like when the user stops scrolling), they must call `startAnimation` to let it resume and drive itself to completion. If the animation hasn't completed yet, they are free to pause it again and go back into interactive state (like if the user starts dragging again before the animation completes). They can also reverse and resume the animation to go back to the initial state (like when the user didn't scroll enough to hide the UI). This lets us fully own the animation and change it in the future without affecting the clients at all.

This should have no impact on existing clients, new APIs only.

#### Implementation
The only new API added to both controllers is:
`prepareInteractiveIsHiddenChange(_ isHidden: Bool, completion: ((_ finalPosition: UIViewAnimatingPosition) -> Void)? = nil) -> UIViewAnimating?`

`BottomSheetController`
- Mostly refactoring so there is a single entry point for all state change animations - `stateChangeAnimator(to ...`.
- Changes to the animator to make it more self-contained - it cleans up the surrounding state in its completion handler, depending on which position the animation finished in. This enables us to return it to the client and let them drive it. The only requirement on the client is to eventually finish the animation.
- We now internally also use the animator for non-animated state changes - we just create and finish it synchronously. This again ensures all the bug-prone state changing stuff happens in one place.
- Some fixes for the internal animation interrupting code where we weren't correctly completing the animator.

`BottomCommandingController`
- Analogous refactors to `BottomSheetController`, but simpler because this controller only owns the bottom bar.

`BottomSheetDemoApp`
`BottomCommandingDemoApp`
- Added an example implementation of "shy chrome" which is interactive and interruptible.
- Small changes to the table view model so the toggles always reflect the actual state.

### Verification
- Sanity checking in both demo controllers
- With slow animations enabled, testing scrolling up down, interrupting ongoing state changes
- Making sure existing APIs for `isHidden`, `isExpanded` haven't regressed.

https://user-images.githubusercontent.com/3610850/133532218-3e658e49-5f4f-42b4-a521-f2028f3ef5f0.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/709)

### DIFF
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -174,33 +174,31 @@ open class BottomCommandingController: UIViewController {
     ///   - animated: Indicates if the change should be animated. The default value is `true`.
     ///   - completion: Closure to be called when the state change completes.
     @objc public func setIsHidden(_ isHidden: Bool, animated: Bool = true, completion: ((_ isFinished: Bool) -> Void)? = nil) {
-        if isViewLoaded {
-            if isInSheetMode {
-                bottomSheetController?.setIsHidden(isHidden, animated: animated, completion: completion)
-            } else {
-                if let animator = bottomBarHidingAnimator {
-                    animator.stopAnimation(false)
-                    animator.finishAnimation(at: .current)
-                }
+        if isInSheetMode {
+            bottomSheetController?.setIsHidden(isHidden, animated: animated, completion: completion)
+        } else {
+            if isViewLoaded {
+                completeBottomBarAnimationsIfNeeded()
 
-                if animated {
-                    let animator = bottomBarHidingAnimator(to: isHidden)
-
-                    if let animator = animator {
-                        bottomBarHidingAnimator = animator
-                        animator.addCompletion { [weak self] finalPosition in
-                            self?.bottomBarHidingAnimator = nil
-                            completion?(finalPosition == .end)
-                        }
-
-                        animator.startAnimation()
-
+                if let animator = bottomBarHidingAnimator(to: isHidden) {
+                    animator.addCompletion { finalPosition in
+                        completion?(finalPosition == .end)
                     }
-                } else {
-                    bottomBarViewBottomConstraint?.constant = isHidden ? -Constants.BottomBar.hiddenBottomOffset : -Constants.BottomBar.bottomOffset
-                    bottomBarView?.isHidden = isHidden
-                    completion?(true)
+
+                    if animated {
+                        bottomBarHidingAnimator = animator
+                        animator.addCompletion { [weak self] _ in
+                            self?.bottomBarHidingAnimator = nil
+                        }
+                        animator.startAnimation()
+                    } else {
+                        animator.stopAnimation(false)
+                        animator.finishAnimation(at: .end)
+                    }
                 }
+            } else {
+                _isHidden = isHidden
+                completion?(true)
             }
         }
     }
@@ -222,10 +220,7 @@ open class BottomCommandingController: UIViewController {
         if isInSheetMode {
             animator = bottomSheetController?.setIsHiddenInteractively(isHidden, completion: completion)
         } else {
-            if let animator = bottomBarHidingAnimator {
-                animator.stopAnimation(false)
-                animator.finishAnimation(at: animator.isReversed ? .start : .end)
-            }
+            completeBottomBarAnimationsIfNeeded(skipToEnd: true)
             if isHidden != self.isHidden, let hidingAnimator = bottomBarHidingAnimator(to: isHidden) {
                 bottomBarHidingAnimator = hidingAnimator
                 hidingAnimator.addCompletion { [weak self] finalPosition in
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -167,42 +169,60 @@ public class BottomSheetController: UIViewController {
         }
     }
 
-    /// Sets the `isHidden` property with a completion handler.
+    /// Changes the `isHidden` state with a completion handler.
```
lcapkovic, line 172: This is super unnecessary but I changed the wording because "sets property" could imply the change will be reflected immediately, but in case of an animated change it will only be reflected when the animation completes...

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -677,6 +697,49 @@ open class BottomCommandingController: UIViewController {
...
+        bottomBarView.isHidden = false
+
+        animator.addAnimations { [weak self] in
+            self?.view.layoutIfNeeded()
+        }
+
+        animator.addCompletion { [weak self] finalPosition in
+            if finalPosition == .end {
+                bottomBarView.isHidden = isHidden
+                self?._isHidden = isHidden
```
rdeassis, line 723: Might be safer to capture the strong self reference here to update the property.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -167,42 +169,60 @@ public class BottomSheetController: UIViewController {
...
     }
 
+    /// Initiates an interactive `isHidden` state change driven by the returned `UIViewAnimating` object.
+    ///
+    /// The returned animator comes preloaded with all the animations required to reach the target `isHidden` state.
+    ///
+    /// You can modify the `fractionComplete` property of the animator to interactively drive the animation in the paused state.
+    /// You can also change the `isReversed` property to swap the start and target `isHidden` states.
+    ///
+    /// At the end of the interactive portion of the animation, you must call `startAnimation` on the animator to complete it non-interactively.
```
rdeassis, line 196: What does "At the end of the interactive portion of the animation" mean for the calling code? Is there a way to be more precise on what needs to be done on their end? A small code snippet if possible?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -491,57 +511,93 @@ public class BottomSheetController: UIViewController {
...
+        let originalBottomSheetHiddenState = bottomSheetView.isHidden
+        let originalBottomOffsetConstant = bottomSheetOffsetConstraint.constant
+
+        bottomSheetView.isHidden = false
+        bottomSheetOffsetConstraint.constant = -targetOffsetFromBottom
+        translationAnimator.addAnimations { [weak self] in
+            self?.view.layoutIfNeeded()
+        }
+
+        let targetExpandedContentAlpha: CGFloat = targetExpansionState == .collapsed ? 0.0 : 1.0
```
rdeassis, line 567: What about creating computed properties for targetExpandedContentAlpha and targetDimmingViewAlpha in the BottomSheetExpansionState to return these values instead?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
lcapkovic: Why is it safer? For a single access to self it just seems redundant to me. Either way, if self is still alive - change the state. If it's not - don't. The logic would be the same right?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: It means whenever they want to stop driving the animation (which they do by repeatedly setting fractionComplete whenever they want). I'll try to make it more clear.

lcapkovic: I reworded to make things more clear.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: I like that! I also changed the logic for expanded content alpha because I realized that alpha should really only be 1.0 in .expanded state.

lcapkovic:
After chat with Anand, renamed the API to `prepareInteractiveIsHiddenChange`. The previous name started with `set` which doesn't properly imply this is the beginning of an interactive process.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
rdeassis: Will there ever be a situation where they differ?
I'm asking because their implementation is the same now:

var expandedContentAlpha: CGFloat **{ self == .expanded ? 1.0 : 0.0 }**
var dimmingViewAlpha: CGFloat **{ self == .expanded ? 1.0 : 0.0 }**

Would that be a good idea to have just one property like "viewAlphaValue" and apply on both scenarios since you now changed its logic?

rdeassis (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Yeah it's kind of a happy coincidence they end up with the same alpha, but they're unrelated. If we add another expansion state (like a midpoint), they will diverge. I do agree with merging them for now tho.

PR #722, opened 2021-09-23
Clean up UIColor extension, add init from hex.

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Removing unused color extensions and adding a new one for converting from hex.

This required updating one property in `Label`, but there should be no functional difference, as the method unconditionally returned `self` anyway.

### Verification

Ensured that Label looks the same, and that everything builds as expected.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/722)

### DIFF
--- a/ios/FluentUI/Extensions/UIColor+Extensions.swift
+++ b/ios/FluentUI/Extensions/UIColor+Extensions.swift
@@ -5,35 +5,44 @@
 
 import UIKit
 
-@objc public extension UIColor {
-    private struct AssociatedKeys {
-        static var light: String = "light"
-        static var lightHighContrast: String = "lightHighContrast"
+public extension UIColor {
+
+    /// Creates a UIColor from a single 32-bit integer value.
+    ///
+    /// - Parameter hexValue: Integer value, generally represented as hexadecimal (e.g. `0x0086F0`), to use to initialize this color.
+    ///     Must be formatted as ARGB. Note: we will not utilize the alpha channel.
+    convenience init(hexValue: UInt32) {
+        self.init(
+            red: CGFloat((hexValue & 0xFF0000) >> 16) / 255.0,
+            green: CGFloat((hexValue & 0x00FF00) >> 8) / 255.0,
+            blue: CGFloat((hexValue & 0x0000FF) >> 0) / 255.0,
+            alpha: 1.0)
     }
 
-    /// Returns self on iOS 13 and later. For older iOS versions returns self for Regular Contrast mode or a specific color for Increased Contrast mode if it's defined either for this color or for one of its ancestors.
-    var current: UIColor {
-        return self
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -5,35 +5,44 @@
...
+    /// - Is the current `accessibilityContrast` `.normal` or `.high`?
+    ///
+    /// - Parameter light: The default color for a light context. Required.
+    /// - Parameter lightHighContrast: The override color for a light, high contrast context. Optional.
+    /// - Parameter lightElevated: The override color for a light, elevated context. Optional.
+    /// - Parameter lightElevatedHighContrast: The override color for a light, elevated, high contrast context. Optional.
+    /// - Parameter dark: The override color for a dark context. Optional.
+    /// - Parameter darkHighContrast: The override color for a dark, high contrast context. Optional.
+    /// - Parameter darkElevated: The override color for a dark, elevated context. Optional.
+    /// - Parameter darkElevatedHighContrast: The override color for a dark, elevated, high contrast context. Optional.
```
harrieshin, line 37: should we say increased contrast?

harrieshin (approved):

ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -5,35 +5,44 @@
 
 import UIKit
 
-@objc public extension UIColor {
-    private struct AssociatedKeys {
-        static var light: String = "light"
-        static var lightHighContrast: String = "lightHighContrast"
-    }
+public extension UIColor {
```
rdeassis, line 8: Is it possible to make the extension internal to further reduce binary size?
What would be the use case for the client apps to use/need it?

rdeassis (approved):

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: High contrast is the term that fluent is using. I'll see if we can get it adjusted, but that'll also make this a much bigger diff so I'll research now and possibly address later.

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: The main reason to make this public is that custom theme definitions may want to use it, especially the multi-color initializer. But I'll make it internal if we end up not exposing UIColor as the type for themes and just ask for a struct filled with integers instead.

PR #723, opened 2021-09-24
Fix bottom sheet bugs related to unhiding

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### Problem
While integrating the bottom sheet, I noticed a few issues where the sheet would unhide itself in edge cases where a property that changes the expansion state in its setter is changed during a hiding animation. This is basically caused by callers of `move(to:)` checking `currentExpansionState` or `isHidden`, neither of which catch the case where the sheet isn't hidden now, but it's animating to a hidden state.

#### Fix
`BottomSheetController.swift`
1. Change the source of truth for `isHidden` getter to `bottomSheetView.isHidden` - reflecting the actual state of the UI is the most simple and consistent
2. Guard all state changes through `move(to:)` in a way that stops any unhiding, unless the caller explicitly opts in by passing in `allowsUnhiding=true`. This ensures that when we add more callers of `move(to:)` in the future, it'll be impossible to accidentally unhide.

Follow the bottom sheet logic of "Only return isHidden==true when the underlying UI is actually hidden".

### Verification

Lots of sanity checking going through the bug edge cases and regular usage.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/723)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -518,7 +518,7 @@ public class BottomSheetController: UIViewController {
                       shouldNotifyDelegate: Bool = true,
                       allowUnhiding: Bool = false,
                       completion: ((UIViewAnimatingPosition) -> Void)? = nil) {
-        guard !isHiddenOrHiding || allowUnhiding else {
+        guard targetExpansionState == .hidden || !isHiddenOrHiding || allowUnhiding else {
             return
         }
 
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -77,7 +77,7 @@ public class BottomSheetController: UIViewController {
             if isExpandable != oldValue {
                 resizingHandleView.isHidden = !isExpandable
                 panGestureRecognizer.isEnabled = isExpandable
-                if isViewLoaded && !isHidden {
+                if isViewLoaded {
                     move(to: .collapsed, animated: false)
                     delegate?.bottomSheetControllerCollapsedHeightInSafeAreaDidChange?(self)
                 }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -155,20 +158,18 @@ public class BottomSheetController: UIViewController {
     ///   - animated: Indicates if the change should be animated. The default value is `true`.
     ///   - completion: Closure to be called when the state change completes.
     @objc public func setIsExpanded(_ isExpanded: Bool, animated: Bool = true, completion: ((_ isFinished: Bool) -> Void)? = nil) {
-        guard isExpanded != self.isExpanded else {
+        guard isExpanded != self.isExpanded, !isHiddenOrHiding, isExpandable else {
```
rdeassis, line 161: question out of the scope of this change:

I noticed that both setIsExpanded and setIsHidden take the same set of params:
```
setIsExpanded(_ isExpanded: Bool, animated: Bool = true, completion: ((_ isFinished: Bool) -> Void)? = nil)
setIsHidden(_ isHidden: Bool, animated: Bool = true, completion: ((_ isFinished: Bool) -> Void)? = nil)
```

Is it possible to programmatically dismiss  the bottom sheet (hide by moving it out of the screen) when it's  in the expanded state?

The suggestion that crossed my mind is to consolidated these 2 funcs (at least in the public API) in one so we can remove one public API and still provide the dismissal from expanded directly (without calling setIsExpanded(false) and then setIsHidden(true)).

Thoughts?

rdeassis (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Calling setIsHidden(true) will override isExpanded and hide the sheet regardless of its current expansion state.
The main issue is that isExpanded is not always the opposite of isHidden. When isExpanded is false, the sheet could either be visible and collapsed, or hidden.

lcapkovic: (Resolving to unblock PR, but feel free to continue discussing here.)

rdeassis: My suggestion on incorporating both APIs into one while still keeping the same concept of hidden, collapsed and extended states. 

For that the parameter would need to be the **BottomSheetExpansionState** enumeration instead of the boolean.
From a caller standpoint that could be easier to understand.

Thoughts?

PR #733, opened 2021-10-01
Fix nav bar and colors in iOS 15

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When building in Xcode 13 against iOS 15, `UINavigationBar` handles its appearance differently, and no longer defaults to a single color. It's unclear if this is a bug or an intended change, but lots of folks on the web seem to have run into the same issue.

Instead of forcing a return to the previous appearance, let's take the opportunity to switch to a modern, translucent nav bar. This gives us much nicer fade animations when scrolling and navigating through the app.

It did, however regress the layout of the segmented control in the `ColorDemoController`. So I fixed that to properly lay out against safe areas.

While I was in the color picker, I discovered a separate issue: some color swatches weren't properly updating, and were actively disappearing on color swap. This likely has to do with the fact that we reuse the same color swatch view and transplant them between iterations of the table view, which is a bit dicey and undefined in the best of situations. So instead of a full `reloadData()`, I've updated the controller to simply refresh the existing color swatches when a theme change notification is fired.

### Verification

- Verified dark and light modes.
- Checked iPads and iPhones both with and without rounded corners (to verify safe area insets).
- Tested against iOS 13, 14, and 15.

| broken | fixed |
| broken-c | fixed-c |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/733)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -21,6 +21,8 @@ class DemoListViewController: UITableViewController {
         let demoListViewController = DemoListViewController(nibName: nil, bundle: nil)
 
         let navigationController = UINavigationController(rootViewController: demoListViewController)
+        navigationController.navigationBar.isTranslucent = true
+
         window.rootViewController = navigationController
         window.makeKeyAndVisible()
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
@@ -124,13 +124,15 @@ class ColorDemoController: UIViewController {
         }
 
         NSLayoutConstraint.activate([
-            stackView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            stackView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            stackView.topAnchor.constraint(equalTo: view.topAnchor),
-            stackView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+            stackView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
+            stackView.trailingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.trailingAnchor),
+            stackView.topAnchor.constraint(equalTo: view.safeAreaLayoutGuide.topAnchor),
+            stackView.bottomAnchor.constraint(equalTo: view.safeAreaLayoutGuide.bottomAnchor)
         ])
+        
+        NotificationCenter.default.addObserver(self, selector: #selector(didChangeTheme), name: Notification.Name.didChangeTheme, object: nil)
     }
-
+    
     override func viewDidAppear(_ animated: Bool) {
         super.viewDidAppear(animated)
         navigationController?.navigationBar.shadowImage = UIImage()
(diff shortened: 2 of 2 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
```diff
@@ -124,13 +124,15 @@ class ColorDemoController: UIViewController {
         }
 
         NSLayoutConstraint.activate([
-            stackView.leadingAnchor.constraint(equalTo: view.leadingAnchor),
-            stackView.trailingAnchor.constraint(equalTo: view.trailingAnchor),
-            stackView.topAnchor.constraint(equalTo: view.topAnchor),
-            stackView.bottomAnchor.constraint(equalTo: view.bottomAnchor)
+            stackView.leadingAnchor.constraint(equalTo: view.safeAreaLayoutGuide.leadingAnchor),
```
harrieshin, line 127: do the cell look right on landscape? like selection of the cell background color and the separator goes edge to edge?

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -21,6 +21,8 @@ class DemoListViewController: UITableViewController {
         let demoListViewController = DemoListViewController(nibName: nil, bundle: nil)
 
         let navigationController = UINavigationController(rootViewController: demoListViewController)
+        navigationController.navigationBar.isTranslucent = true
```
harrieshin, line 24: any impact on ios 14 and 13?

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
mischreiber: Nope, looks the same as always on iOS 13 and 14.
Simulator Screen Shot - iPhone SE (2nd generation) - 2021-10-01 at 15 29 23

ios/FluentUI.Demo/FluentUI.Demo/Demos/ColorDemoController.swift
mischreiber: Ah interesting, no. This constrains the scroll view to the safe areas in landscape, which means the separators don't go edge to edge. (No selection in this tableview, but if so it'd have the same problem.)

Do we have a pattern for this kind of situation? Without these changes the color picker is unusable:
Simulator Screen Shot - iPhone 13 - 2021-10-01 at 15 35 59

harrieshin: we shouldn't be using stackview? but ultimately, can't we just fix our top constraint (maybe bottom constraint) to be within safearea?

mischreiber: Somehow TableHeaderFooterView doesn't have this problem, but it's also a complicated mess that I don't necessarily want to spread. I'll just do your idea for now :(

ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
```diff
@@ -21,6 +21,8 @@ class DemoListViewController: UITableViewController {
...
+
```
rdeassis, line 25: @mischreiber I just noticed that setting the isTranslucent property to true affects the edges of the SwiftUI demo controllers:

Screen Shot 2021-10-06 at 12 35 38 PM


iOS **13**
Screen Shot 2021-10-06 at 12 21 05 PM



**iOS 15**
Screen Shot 2021-10-06 at 12 20 15 PM

PR #734, opened 2021-10-01
Add FilledTemplateImageView to Fluent UI for Mac

brentpe:
### Platforms Impacted
- [ ] iOS
- [X] macOS

### Description of changes

Add a new FilledTemplateImageView to make it easy to create icon type images with different fill and border colors.

### Verification

These images were tested locally in the FluentUITestApp for Mac:
- Using different sizes
- Different border/fill colors
- Different test images.
- Setting properties to test dynamic updating.

image | image 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/734)


harrieshin (approved):

brentpe:
@araje is added to the review. #Closed

brentpe:
@nivanka is added to the review. #Closed

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,82 @@
...
+		}
+		draw(image: image!, isFillMask: false)
+	}
+
+	/// Helper to draw the opaque pixels of the image into a transparency layer
+	private func draw(image: NSImage, isFillMask: Bool) {
+		if let localContext = NSGraphicsContext.current?.cgContext {
+			localContext.beginTransparencyLayer(in: bounds, auxiliaryInfo: nil)
+			image.draw(in: bounds, from: .zero, operation: .sourceOver, fraction: 1.0, respectFlipped: true, hints: nil)
+			isFillMask ? fillColor.setFill() : contentTintColor!.setFill()
```
mischreiber, line 77: We should `guard` to acquire this color instead of using a `!`.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,82 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+/// Custom view for displaying a template style image that can be filled with a specified color. This takes in an image with a corresponding
+/// overlay mask that is used for drawing the fill. The image and its corresponding fillMask need to line up appropriately so that when
+/// the image is drawn over it's mask, its outline will perfectly overlap the edges of the mask to achieve the desired fill effect.
```
mischreiber, line 10: Nit: `its`

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,82 @@
...
+			needsDisplay = true
+		}
+	}
+
+	/// Public override used to draw the icon with the right border and fill
+	public override func draw(_ dirtyRect: NSRect) {
+
+		// Draw the image using the specified fill and border color
+		// Don't use a fillMask if it's set to clear
+		if fillColor != NSColor.clear {
```
mischreiber, line 66: Any reason we use `NSColor.clear` instead of an optional value?

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,82 @@
...
+			draw(image: fillMask, isFillMask: true)
+		}
+		draw(image: image!, isFillMask: false)
```
mischreiber, line 69: Another case where I'd prefer a guard rather than a `!` and a possible crash. There's nothing stopping a client from setting the image to `nil`, which is completely allowed in `NSImageView`.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,82 @@
...
+		}
+	}
+
+	/// Public override used to draw the icon with the right border and fill
+	public override func draw(_ dirtyRect: NSRect) {
+
+		// Draw the image using the specified fill and border color
+		// Don't use a fillMask if it's set to clear
+		if fillColor != NSColor.clear {
+			draw(image: fillMask, isFillMask: true)
```
mischreiber, line 67: The `isFillMask` only changes which color to use in the `draw(image:isFillMask:)` below; would it be cleaner just to pass in the color instead?

mischreiber (changes requested):

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Was on the fence on that one, but went with clear as I wanted to assume non null for everything.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: That was also my first inclination, but realized I'd have access to the colors anyhow. I agree it seems cleaner to pass in colors as well (might create some needed flexibility for future uses as well).

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Yeah, good point.

Saadnajmi:
A dumb question: What's the benefit of something like this over using SF Symbols? Just that we don't have support for that downstream but needed something like it?

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Also, just to clarify - since setting a nil image is valid, then I'm assuming it would be better to use if let instead (v.s. guard let).

mischreiber:
> A dumb question: What's the benefit of something like this over using SF Symbols? Just that we don't have support for that downstream but needed something like it?

Presumably this is an equivalent for Fluent icons, as opposed to Apple's symbols?

brentpe:
> A dumb question: What's the benefit of something like this over using SF Symbols? Just that we don't have support for that downstream but needed something like it?
Good question. Multicolor support in SF Symbols on Mac is only available in Monterey and above but initially the plan was to use SF Symbols and provide down rev support using this method for Big Sur and below. However, we also got feedback from design as it turns out creating custom SF Symbols is pretty expensive.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Note that we also don't want to draw with the clear color as it produces undesired effects with the transparency layer (shows up in black). So I kept that check there in addition to using an optional.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,88 @@
...
+	///   - image: the template style image or icon to be drawn.
+	///   - fillMask: the mask image used to draw the fill color.
+	///   - contentTintColor: the color to use for the main image outline.
+	///   - fillColor: the color to use for the image fill. When set to nil or clear, the fill won't be drawn.
+	@objc(initWithImage:fillMask:contentTintColor:FillColor:)
+	public init(
+		image: NSImage,
+		fillMask: NSImage,
+		contentTintColor: NSColor,
+		fillColor: NSColor
```
mischreiber, line 25: You implied in the documentation that this could be set to `nil`, but you did not declare it as optional (`NSColor?`).

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,88 @@
...
+
+		// Draw the image using the specified fill and border color
+		// Don't use a fillMask if the color is nil or clear
+		if let fillColor = fillColor {
+			if fillColor != .clear {
+				draw(image: fillMask, withColor: fillColor)
+			}
+		}
+		if let contentTintColor = contentTintColor {
+			if let image = image {
```
mischreiber, line 72: You can combine these into a single statement:
```swift
if let contentTintColor = contentTintColor,
   let image = image {
    draw(image: image, withColor: contentTintColor)
}
```

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,88 @@
...
+				draw(image: image, withColor: contentTintColor)
+			}
+		}
+	}
+
+	/// Helper to draw the opaque pixels of the image into a transparency layer.
+	private func draw(image: NSImage, withColor color: NSColor) {
```
mischreiber, line 79: Nit: `with color: NSColor` instead of `withColor color: NSColor` to reduce redundancy.

mischreiber (changes requested):

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: No, I was torn on that one actually I've grown. Seemed better to not let clients initialize with nil initially (what's the point of using this image with no fill). Then I thought about it and perhaps they want to set it to no fill initially then fill it later - which seems like a legit case.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Neat trick. Hopefully smarter minds won't be confused by its cleverness as it looks confusing to me.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Thanks. Got hung up on that one. Still getting used to Swift naming.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,87 @@
...
+@objc(MSFFilledTemplateImageView)
+open class FilledTemplateImageView: NSImageView {
+
+	/// Creates a custom template style image view with a fill.
+	/// - Parameters:
+	///   - image: the template style image or icon to be drawn.
+	///   - fillMask: the mask image used to draw the fill color.
+	///   - contentTintColor: the color to use for the main image outline.
+	///   - fillColor: the color to use for the image fill. When set to nil or clear, the fill won't be drawn.
+	@objc(initWithImage:fillMask:contentTintColor:FillColor:)
```
mischreiber, line 20: Sorry, just noticed this. Two questions:
- Why are you declaring the Objective-C method signature explicitly? Is this not what it would be generated as if you just say `@objc`? I'm not 100% myself, TBH.
- If we are sticking with this, `FillColor` should be capitalized as `fillColor`.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
mischreiber: Yeah, I hated the whole comma-separated `if` statements, and the `let x = x`, and all that. It's on the long list of "weird Swift things that are actually super common and you eventually just get used to".

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Oops, good catch on accidentally capitolizing "FillColor". Regarding the @objc(...) line, we typically leave these for better objc naming. That said, if I remove that line it actually doesn't declare any special initializer in the objc header (I figured I'd at least get some morphed Swift version of it), so definitely needed in this case.

anandrajeswaran: Did you also remove `@objc`? Should be:
```
@objc public init(with
		image: NSImage,
		fillMask: NSImage,
		contentTintColor: NSColor,
		fillColor: NSColor?
```

brentpe: But yeah, I'll add the with decorator for Swift.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -0,0 +1,87 @@
...
+		}
+		if let contentTintColor = contentTintColor,
+		   let image = image {
+			draw(image: image, with: contentTintColor)
+		}
+	}
+
+	/// Helper to draw the opaque pixels of the image into a transparency layer.
+	private func draw(image: NSImage, with color: NSColor) {
+		if let localContext = NSGraphicsContext.current?.cgContext {
```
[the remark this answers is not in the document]
brentpe: Sure, sounds good.

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
brentpe: Just saw this. Yeah, I did remove @objc! : )

macos/FluentUI/FilledTemplateImageView/FilledTemplateImageView.swift
```diff
@@ -17,8 +17,7 @@ open class FilledTemplateImageView: NSImageView {
 	///   - fillMask: the mask image used to draw the fill color.
 	///   - contentTintColor: the color to use for the main image outline.
 	///   - fillColor: the color to use for the image fill. When set to nil or clear, the fill won't be drawn.
-	@objc(initWithImage:fillMask:contentTintColor:fillColor:)
-	public init(
+	@objc public init(
 		with image: NSImage,
```
brentpe, line 21: Well, this actually makes the Objc signature like this: **initWith:fillMask:contentTintColor:fillColor**. So it skips over the "image" part. So I think I'd prefer to add back the objc decoration here.

mischreiber (approved):

PR #736, opened 2021-10-05
Update comm blue dark tint colors

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Design has provided us with new values for four out of our sixteen "communication blue" colors. Specifically, here's what changed:

|               | Light     | Dark    |
|---------|--------|--------|
| **Shade** *(foregrounds)*  | Unchanged    | Unchanged   |
| **Tint** *(backgrounds)*     | Unchanged    | **Updated** 🥳 |

Here's the mapping from old to new:

| Name | Old | New |
|-----|-----|-----|
| `tint10` | `#0078D4` | `#0074D3` |
| `tint20` | `#004C87` | `#004F90` |
| `tint30` | `#043862` | `#002848` |
| `tint40` | `#092C47` | `#001526` |

Subtle changes, but this will darken the backgrounds of some of our elements in Dark mode, increasing contrast and legibility.

### Verification

- Lots of screenshots and time in Pixelmator to verify color values. See some of the clearest examples below.
- Scrubbed the codebase to ensure that only backgrounds were affected.

| note-before | note-after |
| card-before | card-after |

### Pull request checklist

This PR has considered:

### DIFF
--- a/apple/Resources/FluentUI-apple.xcassets/FluentColors/communicationBlueTint10.colorset/Contents.json
+++ b/apple/Resources/FluentUI-apple.xcassets/FluentColors/communicationBlueTint10.colorset/Contents.json
@@ -1,23 +1,18 @@
 {
-  "info" : {
-    "version" : 1,
-    "author" : "xcode"
-  },
   "colors" : [
     {
-      "idiom" : "universal",
       "color" : {
         "color-space" : "srgb",
         "components" : {
-          "red" : "0x2B",
           "alpha" : "1.000",
           "blue" : "0xD8",
-          "green" : "0x88"
+          "green" : "0x88",
+          "red" : "0x2B"
         }
-      }
+      },
+      "idiom" : "universal"
     },
     {
-      "idiom" : "universal",
       "appearances" : [
         {
           "appearance" : "luminosity",
@@ -27,12 +22,17 @@
       "color" : {
         "color-space" : "srgb",
         "components" : {
-          "red" : "0x00",
           "alpha" : "1.000",
-          "blue" : "0xD4",
-          "green" : "0x78"
+          "blue" : "0xD3",
+          "green" : "0x74",
+          "red" : "0x00"
         }
-      }
+      },
+      "idiom" : "universal"
     }
-  ]
+  ],
+  "info" : {
+    "author" : "xcode",
+    "version" : 1
+  }
 }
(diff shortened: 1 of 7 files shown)


mischreiber:
Note: I'm currently submitting this to `vnext-prototype` with the intention of cherry-picking to `main`, but I can just as easily do it the other direction. Thoughts?

harrieshin:
with the new background color of notification toast, it doesn't have much contrast ratio to the app background color (black?). is this. correct? the app background to toast background needs to be 4.5:1 as well I thought

harrieshin:
who else uses tint10 to tint40? badge views? what else are affected?

mischreiber:
> with the new background color of notification toast, it doesn't have much contrast ratio to the app background color (black?). is this. correct? the app background to toast background needs to be 4.5:1 as well I thought

Is that the case? This matches the design in Figma, so if there's an accessibility issue then it's with the Notification view itself (which AFAIK hasn't been merged back to `main` yet, so we can address it in `vnext-prototype` first).

mischreiber:
> who else uses tint10 to tint40? badge views? what else are affected?

@harrieshin These are the only vNext controls that use these colors. However, since I'm also updating the vCurrent set here, the following controls are also affected:
- `BadgeView`
- `Button`
- `CardView`
- `PillButtonBar`

harrieshin:
> > with the new background color of notification toast, it doesn't have much contrast ratio to the app background color (black?). is this. correct? the app background to toast background needs to be 4.5:1 as well I thought
> 
> @harrieshin Is that the case? This matches the design in Figma, so if there's an accessibility issue then it's with the Notification view's design (which AFAIK hasn't been merged back to `main` yet, so we can address it in `vnext-prototype` first).

yup we should bring it up to design team.

harrieshin:
> > who else uses tint10 to tint40? badge views? what else are affected?
> 
> @harrieshin These are the only two vNext controls that use these colors. However, since I'm also updating the vCurrent set here, the following legacy controls are also affected:
> 
> * `BadgeView`
> * `Button`
> * `CardView`
> * `PillButtonBar`

and those look good for accessibility contrast ratios? I usually use https://www.tpgi.com/color-contrast-checker/

mischreiber:
Issues are lodged against design team; PM me for links.

PR #778, opened 2021-10-28
[vNext Tokens] Enable token swapping for new token system

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Now that we have discrete control tokens (#769), it's time to build out functionality for swapping them out at runtime.

#### ControlTokenFactory

The newest class here is `ControlTokenFactory`, which serves as a central token creator and cache. Alone, its only real responsibility is to cache the most recently created `ControlTokens` subclass, but it has no real knowledge of these. Instead, each component will provide an extension to the factory that can create tokens of that class. This extension will be `open`, allowing clients to subclass and provide their own custom tokens.

An example of this is now available in `ThemingDemoController`, which temporarily features a `CardNudge` in addition to its existing `Button` and `Avatar` instances.

#### ControlTokenFactoryHosting

Generally speaking, our clients will want to apply a custom set of tokens to the entire app, not just to a single instance of a control. In a pure SwiftUI app, there is a simple view modifier called `tokenFactory(_:)` that can be applied to the root of the view hierarchy to be styled.

However, most of our clients are still UIKit-based. To facilitate this, this PR adds a new protocol, `ControlTokenFactoryHosting`, which this PR also applies to UIWindow. This allows us to "hang" a custom `ControlTokenFactory` off a window. Now anyone with access to their window can also access the custom `ControlTokenFactory` to be used for styling.

#### ControlHostingContainer

There's a moderate amount of glue code needed to connect the `ControlTokenFactoryHosting`-provided `ControlTokenFactory` and the SwiftUI controls being wrapped. This is a great opportunity to pull the creeping shared logic across our various UIKit wrappers into a single base class, `ControlHostingContainer`. This class will also contain the necessary logic to observe theme change events and use them to apply styling changes to the wrapped control view.

### Verification

This video demonstrates two important features:
- Dynamic runtime theme swapping for all default-styled controls in a window.
- Specific per-control token customization which is not affected by window-level customization.

https://user-images.githubusercontent.com/4934719/139186179-c4f2944c-fb38-4c23-abff-505ac184dfd1.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/778)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemes.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemes.swift
@@ -105,19 +105,28 @@ class DemoColorGreenTheme: NSObject, ColorProviding {
 // TODO: this will be replaced with our branding color system once it exists,
 // but in the meantime it's a good demo of swapping out tokens.
 class DemoCardNudgeTokens: CardNudgeTokens {
-    let greenPrimary: ColorSet = ColorSet(light: 0x107C41, dark: 0x10893C)
+    let greenPrimary: ColorSet = ColorSet(light: Constants.greenLight,
+                                          dark: Constants.greenDark)
+
     override var accentColor: ColorSet {
         return greenPrimary
     }
+
     override var subtitleTextColor: ColorSet {
         return greenPrimary
     }
+
     override var textColor: ColorSet {
         return greenPrimary
     }
+
+    private struct Constants {
+        static let greenLight: ColorValue = 0x107C41
+        static let greenDark: ColorValue = 0x10893C
+    }
 }
 
-class DemoTokenFactory: ControlTokenFactory {
+class DemoTokenFactory: TokenFactory {
     override var cardNudgeTokens: CardNudgeTokens {
         return DemoCardNudgeTokens()
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemes.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemes.swift
@@ -101,3 +101,24 @@ class DemoColorGreenTheme: NSObject, ColorProviding {
         return UIColor(named: "Colors/DemoPrimaryShade30Color")
     }
 }
+
+// TODO: this will be replaced with our branding color system once it exists,
+// but in the meantime it's a good demo of swapping out tokens.
+class DemoCardNudgeTokens: CardNudgeTokens {
+    let greenPrimary: ColorSet = ColorSet(light: 0x107C41, dark: 0x10893C)
+    override var accentColor: ColorSet {
+        return greenPrimary
+    }
+    override var subtitleTextColor: ColorSet {
+        return greenPrimary
+    }
+    override var textColor: ColorSet {
+        return greenPrimary
+    }
+}
+
+class DemoTokenFactory: ControlTokenFactory {
+    override var cardNudgeTokens: CardNudgeTokens {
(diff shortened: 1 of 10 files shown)


ios/FluentUI.Demo/FluentUI.Demo/DemoColorThemes.swift
```diff
@@ -101,3 +101,24 @@ class DemoColorGreenTheme: NSObject, ColorProviding {
         return UIColor(named: "Colors/DemoPrimaryShade30Color")
     }
 }
+
+// TODO: this will be replaced with our branding color system once it exists,
+// but in the meantime it's a good demo of swapping out tokens.
+class DemoCardNudgeTokens: CardNudgeTokens {
+    let greenPrimary: ColorSet = ColorSet(light: 0x107C41, dark: 0x10893C)
```
sophialee0416, line 108: Should we store these color values as constants?

ios/FluentUI/Core/Theme/ControlTokenFactory.swift
```diff
@@ -0,0 +1,80 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// Creates and manages `ControlTokens` for Fluent controls.
+///
+/// This class should be overridden by clients that want to provide custom control tokens on a larger scale.
+open class ControlTokenFactory {
```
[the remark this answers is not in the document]
sophialee0416: I think the latter makes more sense if you plan on adding BrandColors! Maybe even FluentTokenFactory if we want to be specific?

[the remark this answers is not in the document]
mischreiber: I think I'll avoid using Fluent in the names of classes since we're already in the `FluentUI` package namespace, but otherwise I like the change too!

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -190,3 +216,28 @@ open class CustomStyle: FluentUIStyle {
...
+        return purplePrimary
+    }
+    override var subtitleTextColor: ColorSet {
+        return purplePrimary
+    }
+    override var textColor: ColorSet {
+        return purplePrimary
+    }
+    override var cornerRadius: CGFloat {
+        return 0.0
```
rdeassis, line 235: The Factory is currently allowing overriding only on the control token level. For the different stylesheet that would change even corner radius definitions for the entire library, the theme swapping wouldn't handle it.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -190,3 +216,28 @@ open class CustomStyle: FluentUIStyle {
...
+    }
+}
+
+class CustomTokenFactory: TokenFactory {
```
rdeassis, line 239: Have you checked requirements with the client apps (Teams/GroupMe/Skype) that need theme swapping whether the global and alias also need to reflect different values?

mischreiber:
After talking with @rdeassis offline, I'll be closing this PR and opening a new one with branding colors and without the fancy-pants token factory. Stay tuned!

PR #785, opened 2021-11-02
Fix background color and text color for a badge label in nav bar

vickprat:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
How the badge is going to look in different scenarios on icons present in the navigation bar?
**Navigation Bar with Primary Style**

Light Mode                                                
Background Color in badge: White                                               
Text Color in badge: App theme color

Dark Mode
Background Color in badge: App theme color
Text color in badge: White

**Navigation Bar with System style**

Both Light and Dark Mode
Background color in badge: Red                 
Text color in badge: White                                      

### Verification (Dark and Light Modes)
**Screenshots where the navigation bar is of primary style**
|Screenshot 2021-11-02 at 5 38 36 PM|Screenshot 2021-11-02 at 5 43 11 PM|

**Screenshots where the navigation bar is of system style**
|Screenshot 2021-11-09 at 6 25 13 PM|Screenshot 2021-11-09 at 6 24 48 PM|

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Label/BadgeLabel.swift
+++ b/ios/FluentUI/Label/BadgeLabel.swift
@@ -8,10 +8,10 @@ import UIKit
 // MARK: BadgeLabel
 
 class BadgeLabel: UILabel {
-    var shouldUseWindowColor: Bool = true
+    var shouldUseWindowColor: Bool = false
 
-    init() {
-        super.init(frame: .zero)
+    override init(frame: CGRect) {
+        super.init(frame: frame)
         initBase()
     }
 
--- a/ios/FluentUI/Label/BadgeLabel.swift
+++ b/ios/FluentUI/Label/BadgeLabel.swift
@@ -8,10 +8,9 @@ import UIKit
 // MARK: BadgeLabel
 
 class BadgeLabel: UILabel {
-    private var shouldUseWindowColor: Bool = false
+    var shouldUseWindowColors: Bool = false
 
-    init(shouldUseWindowColor: Bool = false) {
-        self.shouldUseWindowColor = shouldUseWindowColor
+    init() {
         super.init(frame: .zero)
         initBase()
     }
@@ -34,7 +33,7 @@ class BadgeLabel: UILabel {
     override func didMoveToWindow() {
         super.didMoveToWindow()
 
-        if shouldUseWindowColor {
+        if shouldUseWindowColors {
             if let window = window {
                 textColor = UIColor(light: Colors.primary(for: window), dark: .white)
                 backgroundColor = UIColor(light: .white, dark: Colors.primary(for: window))
--- a/ios/FluentUI/Label/BadgeLabel.swift
+++ b/ios/FluentUI/Label/BadgeLabel.swift
@@ -22,11 +22,21 @@ class BadgeLabel: UILabel {
     /// Base function for initialization
     private func initBase() {
         layer.masksToBounds = true
-        backgroundColor = Colors.Palette.dangerPrimary.color
-        textColor = .white
         textAlignment = .center
         font = UIFont.systemFont(ofSize: Constants.badgeFontSize, weight: .regular)
         isHidden = true
+    }
+
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        updateWindowSpecificColors()
+    }
+
+    private func updateWindowSpecificColors() {
+        if let window = window {
+            textColor = UIColor(light: Colors.primary(for: window), dark: .white)
+            backgroundColor = UIColor(light: .white, dark: Colors.primary(for: window))
+        }
     }
 
     private struct Constants {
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Label/BadgeLabel.swift
```diff
@@ -29,6 +31,17 @@ class BadgeLabel: UILabel {
         isHidden = true
     }
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        if shouldUseWindowColor {
+            if let window = window {
+                textColor = UIColor(light: Colors.primary(for: window), dark: .white)
+                backgroundColor = UIColor(light: .white, dark: Colors.primary(for: window))
```
rdeassis, line 40: Did the design team approve this specific variation?
Seems a bit odd that both the background and the badge are white.

Screen Shot 2021-11-08 at 11 50 07 AM

ios/FluentUI/Label/BadgeLabel.swift
```diff
@@ -29,6 +31,17 @@ class BadgeLabel: UILabel {
         isHidden = true
     }
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        if shouldUseWindowColor {
```
rdeassis, line 37: This could be a guard let to early return in case the property is false.

vickprat: Done!!

ios/FluentUI/Label/BadgeLabel.swift
vickprat: Fixed!! It was an issue in my code itself and so there was no need of designers discussion.

rdeassis (commented):
As we spoke offline, there's an issue in the regular title case that the toggle switch does not do anything regarding the badge on the buttons. Can you make sure this is fixed in a separate PR? (hiding or disabling the toggle):

MicrosoftTeams-image

ios/FluentUI/Label/BadgeLabel.swift
```diff
@@ -29,6 +30,16 @@ class BadgeLabel: UILabel {
         isHidden = true
     }
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        guard let window = window, shouldUseWindowColor else {
```
rdeassis, line 36: nit: can you put the shouldUseWindowColor first? then it doesn't need to access the window property if it's not needed.

harrieshin:
hey, I think the red badge should be the default, and then we should just have a way for client team to change the background color of the badge whatever they want? similar to how badge in tabbar could work.

ios/FluentUI/Label/BadgeLabel.swift
vickprat: Done!!

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -14,6 +14,12 @@ class BadgeLabelButton: UIButton {
         }
     }
 
+    var shouldUseWindowColorInBadge: Bool? {
```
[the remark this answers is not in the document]
vickprat: Now, I made it non-optional but just wanted to know that what is the problem to make this var optional?

If the intention is to make the caller forcibly set this value then should I go ahead and change the initializer itself? But again do we really need to do that?

vickprat:
> hey, I think the red badge should be the default, and then we should just have a way for client team to change the background color of the badge whatever they want? similar to how badge in tabbar could work.

Hi Harrie!! We generallly follow the window color thing as we do in case of tooltip also because there is background color both in light and dark mode and also the text color in both light and dark mode. There is an email thread started by me and suggested by Rafael on this and designers expect some other design changes as well. Let me forward that email thread to you.

Also, because this PR is a blocker for shipping the notification center feature and this week is the last week because next week will be code freeze so I am planning to improve the design as suggested in that email thread as a separate CR on FluentUI.

vickprat:
> As we spoke offline, there's an issue in the regular title case that the toggle switch does not do anything regarding the badge on the buttons. Can you make sure this is fixed in a separate PR? (hiding or disabling the toggle):
> 
> MicrosoftTeams-image

I have fixed the bug and this was happening with the rainbow on avatar thing too. This is the PR: https://github.com/microsoft/fluentui-apple/pull/793

ios/FluentUI/Navigation/BadgeLabelButton.swift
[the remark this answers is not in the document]
rdeassis: The rationale behind that is that we're already providing a default boolean value for that, so if the calling code is changing the value, it needs to be a valid boolean (instead of a possible nil value).

rdeassis (approved):
Signing off provided that @harrieshin is okay with not going with the custom color suggested for now.
@vickprat can you just confirm before we go ahead and merge this?

rdeassis:
@vickprat I'm merging this PR to unblock your work. We're expecting the follow up changes to make this conform to the new design guidelines as well as addressing @harrieshin 's [comment](https://github.com/microsoft/fluentui-apple/pull/785#issuecomment-964453453) about the custom colors (below):

`hey, I think the red badge should be the default, and then we should just have a way for client team to change the background color of the badge whatever they want? similar to how badge in tabbar could work.`

PR #790, opened 2021-11-05
Bottom sheet - Add the ability to hide/show the content when sheet is collapsed

noaMSFT:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes
For a new feature, we update the collapsed height to show the first part of the expanded content so that it's not collapsed all the way down (more than just the header is visible) and hence, we need to prevent the bottom sheet from hiding the content when in collapsed state. 

1) BottomSheetController.swift:
Added a property to control whether the content should be hidden when the sheet is in collapsed state that is true by default to match the existing behaviour. When the property is set to true, there will be no alpha value change applied to the content.

2) BottomSheetDemoController.swift:
Updated the demo controller so that there is a switch to toggle the property to hide/show the content when collapsed.

### Verification
Used the demo controller to test the following scenarios:
- not setting the property at all - the sheet behaves the same way as before, it is hidden when collapsed, and there is an alpha animation.
- setting the property to "false" shows the content when the sheet is collapsed and there is no alpha animation while the sheet being collapsed

Collapsed content hidden (toggle is off):
collapsed content hidden

Collapsed content is NOT hidden (toggle is on):
collapsed content not hidden

### Pull request checklist
none of the checks below are relevant as I didn't add controls or modified any of the controls. It's controlling the alpha changes of the content view only.

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/790)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -135,7 +135,13 @@ public class BottomSheetController: UIViewController {
     }
 
     /// Indicates if the content should be hidden when the sheet is collapsed
-	@objc open var shouldHideCollapsedContent: Bool = true
+    @objc open var shouldHideCollapsedContent: Bool = true {
+        didSet {
+            if shouldHideCollapsedContent != oldValue {
+                updateExpandedContentAlpha()
+            }
+        }
+    }
 
     /// Current height of the portion of a collapsed sheet that's in the safe area.
     @objc public var collapsedHeightInSafeArea: CGFloat {
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -134,6 +134,9 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// Indicates if the content should be hidden when the sheet is collapsed
+	@objc open var shouldHideCollapsedContent: Bool = true
+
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -134,6 +134,9 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// Indicates if the content should be hidden when the sheet is collapsed
+	@objc open var shouldHideCollapsedContent: Bool = true
```
rdeassis, line 138: nit: can you replace the tab with 4 spaces?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -134,6 +134,9 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// Indicates if the content should be hidden when the sheet is collapsed
```
rdeassis, line 137: Setting this property should probably call updateExpandedContentAlpha().

On the demo app this happens when we change the toggle switch:



https://user-images.githubusercontent.com/68076145/140813905-4cb28420-d91f-47ca-8eb6-309838ae6f6d.mov


It should update right away when the property is set.

rdeassis (approved):

lcapkovic (approved):

PR #791, opened 2021-11-06
Fix showing tooltip in demo for bottom tab bar

vickprat:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Problem: Whenever there is a title present for the tab bar item then the tooltip is unable to show for that item whereas it is successfully been shown with the item without a title.

Reason: There are two different arrays with the tab bar items, one array is being used in case of title is present and the other array is being used when the title is not present. While adding support for showing tooltip for the tab bar items, I created and stored reference of one item and then used to it show the tooltip. I used this item's reference in the array which is being used in the case when the title is not present but forgot to use it in the other array which is being used when the title is present. Instead in the other array, we are creating a new item and so tooltip is not being shown for this new item which we don't even have the reference in the method where the tooltip is being shown.

Fix: Used the item's reference in the other array too which is being used when the title is present.

### Verification
Screenshot 2021-11-06 at 11 19 06 AM

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/791)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
@@ -75,7 +75,7 @@ class TabBarViewDemoController: DemoController {
 
         if showsItemTitles {
             updatedTabBarView.items = [
-                TabBarItem(title: "Home", image: UIImage(named: "Home_24")!, selectedImage: UIImage(named: "Home_Selected_24")!),
+                homeItem,
                 TabBarItem(title: "New", image: UIImage(named: "New_24")!, selectedImage: UIImage(named: "New_Selected_24")!),
               TabBarItem(title: "Open", image: UIImage(named: "Open_24")!, selectedImage: UIImage(named: "Open_Selected_24")!)
             ]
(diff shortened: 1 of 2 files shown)


rdeassis (commented):
Regarding the description of the PR:
- We don't reference private ADO items in GitHub because it's public. Could you remove the link from it?
- Can you explain the root cause of the issue and how the changes in code fix the issue?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -75,7 +75,7 @@ class TabBarViewDemoController: DemoController {
 
         if showsItemTitles {
             updatedTabBarView.items = [
-                TabBarItem(title: "Home", image: UIImage(named: "Home_24")!, selectedImage: UIImage(named: "Home_Selected_24")!),
+                homeItem,
```
rdeassis, line 78: Can you explain the root cause of the issue and how the changes in code fix the issue? (in the description of the PR)

vickprat: I have put the description.

rdeassis (approved):

PR #797, opened 2021-11-12
let title lines of popupmenucells to grow in larger text mode

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

let popupmenucell title's line number to be increased when larger text mode is on.

### Verification

|Simulator Screen Shot - iPhone 11 Pro Max - 2021-11-12 at 12 52 18 |Simulator Screen Shot - iPhone 11 Pro Max - 2021-11-12 at 12 56 02| 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/797)

### DIFF
--- a/ios/FluentUI/Popup Menu/PopupMenuItemCell.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuItemCell.swift
@@ -106,6 +106,7 @@ class PopupMenuItemCell: TableViewCell, PopupMenuItemTemplateCell {
 
         setup(title: item.title, subtitle: item.subtitle ?? "", customView: _imageView.image != nil ? _imageView : nil, customAccessoryView: item.accessoryView)
         isEnabled = item.isEnabled
+        titleNumberOfLines = 0
 
         updateViews()
         updateAccessibilityTraits()


rdeassis (approved):

ios/FluentUI/Popup Menu/PopupMenuItemCell.swift
```diff
@@ -106,6 +106,7 @@ class PopupMenuItemCell: TableViewCell, PopupMenuItemTemplateCell {
 
         setup(title: item.title, subtitle: item.subtitle ?? "", customView: _imageView.image != nil ? _imageView : nil, customAccessoryView: item.accessoryView)
         isEnabled = item.isEnabled
+        titleNumberOfLines = 0
```
mischreiber, line 109: Just thinking out loud, do we want this unconditionally? Are there any non-a11y scenarios where we'd _want_ long strings of text to `...` rather than wrap?

mischreiber (approved):

ios/FluentUI/Popup Menu/PopupMenuItemCell.swift
harrieshin: eh....if there are reasons to be conditional, then I might add another api on the popupmenutemplatecell. however, given that most usage of popupmenu is for simple action sheet which in theory, user should know what it is about.

PR #800, opened 2021-11-17
Add accessibility label fix to badge label button

vickprat:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR is fixing the accessibility label in the badge label button. Currently, accessibility is only for the button itself but not for the badge that is coming on the button. This PR is fixing the issue by asking the clients for an accessibility label together with the badge value.

### Verification
Verified on my device. Device details: iPhone 12, iOS 15.0.2

https://user-images.githubusercontent.com/33866787/142191150-c21ddb7a-9bef-4afc-bb0c-faefdd96e57c.MP4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/800)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -472,6 +472,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         } else {
             let dismissItem = UIBarButtonItem(title: BarButtonItemTag.dismiss.title, style: .plain, target: self, action: #selector(dismissSelf))
             dismissItem.tag = BarButtonItemTag.dismiss.rawValue
+            dismissItem.accessibilityLabel = "Dismiss"
             var items = [dismissItem]
             if allowsCellSelection {
                 let selectItem = UIBarButtonItem(title: BarButtonItemTag.select.title, style: .plain, target: self, action: #selector(showSelectionMode))
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -513,9 +513,8 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
                 let badgeValue = isOn ? "12" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
             } else {
-                let badgeValue = isOn ? "12" : nil
-                let badgeAccessibilityLabel = isOn ? "new items" : nil
-                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
+                let badgeValue = isOn ? "New" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
             }
         }
         showBadgeOnBarButtonItem = isOn
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -507,14 +507,16 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
                 let badgeValue = isOn ? "12345" : nil
-                let badgeAccessibilityLabel = isOn ? "items" : nil
+                let badgeAccessibilityLabel = isOn ? "\(badgeValue ?? "Zero") items" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else if item.tag == BarButtonItemTag.threeDay.rawValue {
                 let badgeValue = isOn ? "12" : nil
-                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
+                let badgeAccessibilityLabel = isOn ? "\(badgeValue ?? "Zero") new items" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else {
                 let badgeValue = isOn ? "New" : nil
-                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
+                let badgeAccessibilityLabel = isOn ? "New feature" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             }
         }
         showBadgeOnBarButtonItem = isOn
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -507,11 +507,11 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
                 let badgeValue = isOn ? "12345" : nil
-                let badgeAccessibilityLabel = isOn ? "\(badgeValue ?? "Zero") items" : nil
+                let badgeAccessibilityLabel = isOn ? "\(badgeValue!) items" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else if item.tag == BarButtonItemTag.threeDay.rawValue {
                 let badgeValue = isOn ? "12" : nil
-                let badgeAccessibilityLabel = isOn ? "\(badgeValue ?? "Zero") new items" : nil
+                let badgeAccessibilityLabel = isOn ? "\(badgeValue!) new items" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else {
                 let badgeValue = isOn ? "New" : nil
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -506,11 +506,16 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
-                item.badgeValue = isOn ? "12345" : nil
+                let badgeValue = isOn ? "12345" : nil
+                let badgeAccessibilityLabel = isOn ? "items" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else if item.tag == BarButtonItemTag.threeDay.rawValue {
-                item.badgeValue = isOn ? "12" : nil
+                let badgeValue = isOn ? "New" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
             } else {
-                item.badgeValue = isOn ? "New" : nil
+                let badgeValue = isOn ? "12" : nil
+                let badgeAccessibilityLabel = isOn ? "new items" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             }
         }
         showBadgeOnBarButtonItem = isOn
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -505,8 +505,8 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
             return
         }
         for item in items {
-            var badgeValue: String? = nil
-            var badgeAccessibilityLabel: String? = nil
+            var badgeValue: String?
+            var badgeAccessibilityLabel: String?
             if isOn {
                 if item.tag == BarButtonItemTag.dismiss.rawValue {
                     badgeValue = "12345"
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -510,7 +510,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
                 let badgeAccessibilityLabel = isOn ? "items" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else if item.tag == BarButtonItemTag.threeDay.rawValue {
-                let badgeValue = isOn ? "New" : nil
+                let badgeValue = isOn ? "12" : nil
                 item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
             } else {
                 let badgeValue = isOn ? "12" : nil
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -506,17 +506,19 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
-                let badgeValue = isOn ? "12345" : nil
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
```diff
@@ -33,4 +34,16 @@ import UIKit
             badgeValue = nil
         }
     }
+
+    /// Set this format string in a way that takes two parameters: item's accessibility label and badge value.
+    /// Format: %@ %@ <some string>
+    /// Example: "<Activity> <5> new activities" which will be called by voice over as "Activity 5 new activities button"
+    @objc var accessibilityLabelFormatString: String? {
```
harrieshin, line 41: I don't understand this exposure. item already has accessibilitylabel that clients can override. why do we need to expose yet another label format string? we should be careful any additional public function especially in extension files

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -208,5 +208,21 @@ class BadgeLabelButton: UIButton {
 
     @objc private func badgeValueDidChange() {
         updateBadgeLabel()
+        updateAccessibilityLabel()
+    }
+
+    private func updateAccessibilityLabel() {
+        guard let item = item, let itemAccessibilityLabel = item.accessibilityLabel else {
```
harrieshin, line 215: is guard let itemaccessibilitylabel correct? sometimes do you want to set it nil.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: Actually, the client can set accessibility label as just the name of the button and that will be called when there is only a button present without badge value and if there is badge value on that button then this string can be used or the default one which I added in the Localization file.

If I will not expose this then the client can only set one string as an accessibility label string and that will be used with and without a badge. So voice over will speak -> "Dismiss 0 items button" and "Dismiss 3 items button"

ios/FluentUI/Navigation/BadgeLabelButton.swift
vickprat: I am sorry but I didn't get your comment. I did this so that we can return early whenever there is no item or if the item's accessibility label is not present.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
harrieshin: clients knows what the badge values are and they send that to fluent to update the badge. why can't they can't set it directly on the button with the badge value?

ios/FluentUI/Navigation/BadgeLabelButton.swift
harrieshin: isn't this a change in behavior. before accessibilityLabel = item.accessibilityLabel. it set it to nil if it was nil

vickprat: Done!!

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: So basically client can only set properties of UIBarButtonItem and internally we make UIButton from it and clients can set badge value on UIBarButtonItem at any point of time in the future which internally gets updated on UIButton by listening to "notification on a change in badge value"

How a client can directly inform this UIButton class about what kind of accessibility label format string should be set in two different scenarios without exposing this property.

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -208,5 +208,21 @@ class BadgeLabelButton: UIButton {
...
     @objc private func badgeValueDidChange() {
         updateBadgeLabel()
+        updateAccessibilityLabel()
+    }
+
+    private func updateAccessibilityLabel() {
+        guard let item = item else {
+            return
+        }
+        if let badgeValue = item.badgeValue, let itemAccessibilityLabel = item.accessibilityLabel {
```
harrieshin, line 218: "let itemAccessibilityLabel = item.accessibilityLabel " i think this check is still incorrect. if there isn't an explict acessibilitylabel passed by the item, we still can use the button label itself to do the string concatenation

vickprat: I didn't get what do you mean by "button label itself". The button can also be image only button.

Also, I am following your suggestion about "change in behavior" because earlier also we just used to set accessibilityLabel = item.accessibilityLabel

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: Understood!! Yes, it makes complete sense to ask the client the badge value and badge accessibility label together.

ios/FluentUI/Navigation/BadgeLabelButton.swift
vickprat: Done!!

ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -120,6 +120,9 @@
 /* Format string for tab bar item accessbility labels. Format: "<Title>, <BadgeValue> items". Example: "Home, 5 items" */
 "Accessibility.TabBarItemView.LabelFormat" = "%@, %@ items";
 
+/* Format string for tab bar item accessbility labels. Format: "<Title>, <BadgeValue> items". Example: "Dismiss, 5 items" */
```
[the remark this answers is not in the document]
vickprat: Done!!

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -208,5 +208,13 @@ class BadgeLabelButton: UIButton {
...
     @objc private func badgeValueDidChange() {
         updateBadgeLabel()
+        updateAccessibilityLabel()
+    }
+
+    private func updateAccessibilityLabel() {
+        guard let item = item else {
+            return
+        }
+        accessibilityLabel = (item.accessibilityLabel ?? "").appending(item.badgeAccessibilityLabel ?? "")
```
rdeassis, line 218: We cannot just append strings like this because of localization scenarios.
A string like "%@, %@" needs to be created so that the localization team can arrange the order and/or eventually replace the comma with any other character of the specific language.

rdeassis: Also, the localized string for concatenating both accessibility labels will be used only if the badgeAccessibilityLabel is provided.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
```diff
@@ -24,13 +25,21 @@ import UIKit
...
+        get {
+            return objc_getAssociatedObject(self, &AssociatedKeys.badgeAccessibilityLabel) as? String ?? nil
+        }
+        set {
+            objc_setAssociatedObject(self, &AssociatedKeys.badgeAccessibilityLabel, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
         }
     }
+
+    /// Use this method on bar button item's instance to set the badge value and badge accessibility label.
+    /// Set the badge value to nil to hide the red oval.
```
rdeassis, line 38: nit: using the documentation standard for each parameter might be clearer to inform about values that can be set and their consequences in the rendering of the control.

You can hit `command + option + /` to have Xcode generate the template automatically.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -505,11 +506,15 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
-                item.badgeValue = isOn ? "12345" : nil
+                let badgeValue = isOn ? "12345" : nil
+                let badgeAccessibilityLabel = isOn ? "items" : nil
```
rdeassis, line 510: Shouldn't the badgeAccessibilityLabel be set to "\(badgeValue) items" instead of just "items"?

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -505,11 +506,15 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
...
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: badgeAccessibilityLabel)
             } else if item.tag == BarButtonItemTag.threeDay.rawValue {
-                item.badgeValue = isOn ? "12" : nil
+                let badgeValue = isOn ? "12" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
             } else {
-                item.badgeValue = isOn ? "New" : nil
+                let badgeValue = isOn ? "New" : nil
+                item.setBadgeValue(badgeValue, badgeAccessibilityLabel: nil)
```
rdeassis, line 517: I think the benefit of setting the badgeAccessibilityLabel is to customize in case the value is not a number.
Would it be nice to showcase the accessibility label with a custom one here (like "New feature" or something along those lines)?

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
```diff
@@ -24,13 +25,21 @@ import UIKit
...
+    /// Complete badge accessibility label would be the concatenation of the badge value and badge accessibility label.
+    /// Complete accessibility label for the item in the navigation bar would be item's accessibility label concatenated with the complete badge accessibility label.
+    @objc func setBadgeValue(_ badgeValue: String?, badgeAccessibilityLabel: String?) {
+        self.badgeAccessibilityLabel = (badgeValue ?? "").appending(badgeAccessibilityLabel ?? "")
```
rdeassis, line 42: why is this logic duplicated here and in line 218?
It seems to me that this API should only set the properties with the parameter values.

ios/FluentUI/Navigation/BadgeLabelButton.swift
vickprat: Done!!

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: Done!!

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
vickprat: Done!!

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
vickprat: Actually, earlier my design was that we are going to ask user only the suffx string and we will put badge value on our own but now I realized that was not good instead we should be asking the complete badge accessibility label from the user. So yes, now I have fixed.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: Actually, my design was that we are going to ask the user only the suffix string and we will put badge value on our own but now I realized that was not good instead we should be asking for the complete badge accessibility label from the user. So yes, now I have fixed it.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
```diff
@@ -24,13 +25,21 @@ import UIKit
...
+        }
+        set {
+            objc_setAssociatedObject(self, &AssociatedKeys.badgeAccessibilityLabel, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
         }
     }
+
+    /// Use this method on bar button item's instance to set the badge value and badge accessibility label.
+    /// - Parameters:
+    ///   - badgeValue: Value that will be displayed in a red oval above the bar button item. Set the badgeValue to nil to hide the red oval.
+    ///   - badgeAccessibilityLabel: Accessibility label for the badge. If present, then the overall accessibility label will be item's accessibility label concatenated with the badge's accessibility label, else only the item's accessibility label.
```
rdeassis, line 40: nit: I think the documentation comment will be clearer if stated like this:

`Accessibility label for the badge. Combined with the item's accessibility label  if not nil.`

Thoughts?

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -208,5 +208,19 @@ class BadgeLabelButton: UIButton {
...
         updateBadgeLabel()
+        updateAccessibilityLabel()
+    }
+
+    private func updateAccessibilityLabel() {
+        guard let item = item else {
+            return
+        }
+        if let itemAccessibilityLabel = item.accessibilityLabel, let badgeAccessibilityLabel = item.badgeAccessibilityLabel {
+            accessibilityLabel = String(format: "Accessibility.BadgeLabelButton.LabelFormat".localized,
```
rdeassis, line 219: Can you use [localizedStringWithFormat(_:_:)](https://developer.apple.com/documentation/foundation/nsstring/1408481-localizedstringwithformat) instead of [init(format:arguments:)](https://developer.apple.com/documentation/foundation/nsstring/1407827-init)?

We made the standard to use the former because [init(format:arguments:)](https://developer.apple.com/documentation/foundation/nsstring/1407827-init) does not do any localization while formatting.

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -208,5 +208,19 @@ class BadgeLabelButton: UIButton {
...
     @objc private func badgeValueDidChange() {
         updateBadgeLabel()
+        updateAccessibilityLabel()
+    }
+
+    private func updateAccessibilityLabel() {
+        guard let item = item else {
+            return
+        }
+        if let itemAccessibilityLabel = item.accessibilityLabel, let badgeAccessibilityLabel = item.badgeAccessibilityLabel {
```
rdeassis, line 218: Is there a scenario that makes sense to have only the badgeAccessibilityLabel set?
Something like "3 new notifications, button" instead of "Notifications, 3 new items, button".

I'm questioning that because if the accessibilityLabel set to nil will always prevent the badgeAccessibilityLabel from being set.

What are your thoughts on this logic:

```
if let badgeAccessibilityLabel = item.badgeAccessibilityLabel {
  if let itemAccessibilityLabel = item.accessibilityLabel {
    String.localizedStringWithFormat("Accessibility.BadgeLabelButton.LabelFormat".localized,itemAccessibilityLabel,badgeAccessibilityLabel)
  } else {
    accessibilityLabel = badgeAccessibilityLabel
  }
} else {
    accessibilityLabel = item.accessibilityLabel
}
```

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -505,11 +506,17 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
         for item in items {
             if item.tag == BarButtonItemTag.dismiss.rawValue {
-                item.badgeValue = isOn ? "12345" : nil
+                let badgeValue = isOn ? "12345" : nil
+                let badgeAccessibilityLabel = isOn ? "\(badgeValue ?? "Zero") items" : nil
```
rdeassis, line 510: I think that setting the accessibility label with "Zero items" while the badgeValue is nil is a bit misleading from the demo perspective.
I think you can create non-optional constants here and using the ternary operator to pass them or nil (depending on the isOn argument.

ios/FluentUI/Navigation/UIBarButtonItem+BadgeValue.swift
vickprat: Yeah!! Looks more readable.

ios/FluentUI/Navigation/BadgeLabelButton.swift
vickprat: Ohh!! I didn't know about this!!

ios/FluentUI/Navigation/BadgeLabelButton.swift
vickprat: Yeah!! Makes sense.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
vickprat: But it will only call zero items if the switch is ON and still the badge value is nil and so it's not going to say zero items while actually running the demo but yes if we would do such thing that the badge value switch is on and still badge value is nil then it will call zero items. So, that's what I think but I guess your point is that we should be crashing it if that's the case or correct me If I am wrong.

rdeassis (approved):
Thanks for addressing all the comments!

PR #801, opened 2021-11-17
Adding tooltip to links in the Mac demo app.

rdeassis:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Our accessibility test team reported an issue where the Links in the demo app do not have the tooltip set.
This change makes it compliant with that requirement and add tooltip to all links in that screen.

### Verification

https://user-images.githubusercontent.com/68076145/142267639-9ee4ed59-9854-4a59-84bb-53cf6fab9a88.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/801)

### DIFF
--- a/macos/FluentUITestViewControllers/TestLinkViewController.swift
+++ b/macos/FluentUITestViewControllers/TestLinkViewController.swift
@@ -7,26 +7,41 @@ import AppKit
 import FluentUI
 
 class TestLinkViewController: NSViewController {
-	let disabledLink = Link(title: "Disabled link with hover effects")
+	let disabledLink: Link = {
+		let disabledLinkTitle = "Disabled link with hover effects"
+		let disabledLink = Link(title: disabledLinkTitle)
+		disabledLink.toolTip = disabledLinkTitle
+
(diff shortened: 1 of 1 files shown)


Saadnajmi:
The fixes we're making here convince me it was good to split out such a simple component, and make me want to replace more custom links downstream. Nice :P

Saadnajmi (approved):

Saadnajmi:
Oh wait, should we perhaps have Link set a default tooltip?

edjamesmsft (approved):

rdeassis:
> Oh wait, should we perhaps have Link set a default tooltip?

Thanks for pointing that out! 
I have considered that too, but I don't think the tooltip should be enabled by default (and that would potentially break scenarios where the Link is already used).

That was the main reason why we're treating this as a demo app accessibility bug and fixing the issue in the client app side instead of the library side.

PR #803, opened 2021-11-19
Add Fluent Divider

huwilkes:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Add a new separator that's built with SwiftUI and uses the vNext token system

| Divider_Light | Divider_Dark |
| Divider_SwiftUI_Light | Divider_SwiftUI_Dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/803)


tools/sgen/input/FluentUIStyle.yml
```diff
@@ -320,6 +320,13 @@ AP_MSFCommandBarTokens:
   itemIconColor: [ rest: $Colors.Foreground.neutral1, hover: $Colors.Foreground.neutral1, pressed: $Colors.Foreground.neutral1, selected: $Colors.Foreground.neutralInverted, disabled: $Colors.Foreground.neutralDisabled ]
   itemInterspace: $Spacing.xxxSmall
 
+AP_MSFDividerTokens:
+  color: [ rest: $Colors.Stroke.neutral1]
```
huwilkes, line 324: This was the color from the Figma spec (List -> Border), but it's even harder to see in the test app than it was in the spec. We should probably review the spec with design

ios/FluentUI/Divider/DividerTokens.swift
```diff
@@ -0,0 +1,64 @@
...
+
+import SwiftUI
+import UIKit
+
+// MARK: - DividerStyle
+
+/// Pre-defined styles of the divider
+@objc public enum MSFDividerStyle: Int {
+    case `default`
+    case shadow
```
huwilkes, line 14: The spec didn't have a shadow style, but the previous separator did, so I added one.

Saadnajmi:
Dumb question, why is it called divider instead of separator?

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -200,6 +200,9 @@
 		C708B064260A87F7007190FA /* SegmentItem.swift in Sources */ = {isa = PBXBuildFile; fileRef = C708B04B260A8696007190FA /* SegmentItem.swift */; };
 		C77A04B825F03DD1001B3EB6 /* String+Date.swift in Sources */ = {isa = PBXBuildFile; fileRef = C77A04B625F03DD1001B3EB6 /* String+Date.swift */; };
 		C77A04EE25F046EB001B3EB6 /* Date+CellFileAccessoryView.swift in Sources */ = {isa = PBXBuildFile; fileRef = C77A04EC25F046EB001B3EB6 /* Date+CellFileAccessoryView.swift */; };
+		EC02A5F127472EC300E81B3E /* MSFDividerTokens.generated.swift in Sources */ = {isa = PBXBuildFile; fileRef = EC02A5F027472EC300E81B3E /* MSFDividerTokens.generated.swift */; };
```
rdeassis, line 203: nit: can you move the Divider folder into the Vnext folder?
The context behind that is we keep all controls that are not yet in the main branch in the Vnext folder.
In the process of merging it to main, we move it to the main folder and delete the Vnext folder.

tools/sgen/input/FluentUIStyle.yml
rdeassis: That's a great point! Thanks for bringing it up.
Do you think that could bring the number of divider styles to just one? It'd be great to save on definitions of tokens for it.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -320,6 +320,13 @@ AP_MSFCommandBarTokens:
...
+  spacing: [ none: $Spacing.none, medium: $Spacing.medium]
+
+MSFShadowDividerTokens extends MSFDividerTokens:
+  color: [ rest: $Colors.Stroke.neutral2]
```
rdeassis, line 328: If there are no other states beyond "rest", we should have the color token instead of an array of colors per state.

ios/FluentUI/Divider/DividerTokens.swift
```diff
@@ -0,0 +1,64 @@
...
+    case medium
+}
+
+/// Representation of design tokens to dividers at runtime which interfaces with the Design Token System auto-generated code
+/// Updating these properties causes the SwiftUI divider to updates its view automatically
+class MSFDividerTokens: MSFTokensBase, ObservableObject {
+    @Published public var color: UIColor!
+    @Published public var padding: CGFloat!
+
+    var style: MSFDividerStyle
```
rdeassis, line 29: If we're allowing these to be changed (var instead of let), we should call updateForCurrentTheme() on the didSet closure so that the SwiftUI View will be refreshed when the properties are updated:

For example:
https://github.com/microsoft/fluentui-apple/blob/89daa063880b3622f8f0ea229940e03ea354a24d/ios/FluentUI/Avatar/AvatarTokens.swift#L68

```

    var style: MSFAvatarStyle {
        didSet {
            if oldValue != style {
                updateForCurrentTheme()
            }
        }
    }

    var size: MSFAvatarSize {
        didSet {
            if oldValue != size {
                updateForCurrentTheme()
            }
        }
    }
```

ios/FluentUI/Divider/DividerTokens.swift
rdeassis: This would be another great thing to discuss with design. It makes sense to keep the compatibility now, but if the designers want something else, this is the best time to make more significant changes.
The name of the style being "shadow" is also a bit misleading in my opinion, since it doesn't actually project any shadow.

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+            Rectangle()
+                .fill(Color(state.color ?? tokens.color))
+        }
+
+        @ViewBuilder
+        var divider: some View {
+            switch state.orientation {
+            case .horizontal:
+                dividerBackground.frame(height: thickness)
+                    .padding(EdgeInsets(top: tokens.padding, leading: 0, bottom: tokens.padding, trailing: 0))
```
rdeassis, line 94: nit: do you mind breaking the 4 parameters into multiple lines for readability? (here and line 97)

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+    @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+    @ObservedObject var state: MSFDividerStateImpl
+    @ObservedObject var tokens: MSFDividerTokens
+
+    /// Creates and initializes a SwiftUI Divider
+    /// - Parameters:
+    ///   - style: The DividerStyle value used by the Divider.
+    ///   - orientation: The DividerOrientation used by the Divider.
+    ///   - spacing: The DividerSpacing used by the Divider.
+    public init (style: MSFDividerStyle = .default, orientation: MSFDividerOrientation = .horizontal, spacing: MSFDividerSpacing = .none) {
```
rdeassis, line 75: nit: do you mind breaking the parameters into multiple lines for readability whenever we have more than 3 params? (here, line 94, line 97)

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+
+    private var hostingController: FluentUIHostingController!
+
+    private var divider: FluentDivider!
+}
+
+/// Properties that can be used to customize the appearance of the Divider.
+@objc public protocol MSFDividerState {
+    /// Sets a custom color for the Divider.
+    var color: UIColor? {get set}
```
rdeassis, line 51: nit: space between the curly braces and the characters nearby

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+        @ViewBuilder
+        var dividerBackground: some View {
+            Rectangle()
+                .fill(Color(state.color ?? tokens.color))
+        }
+
+        @ViewBuilder
+        var divider: some View {
+            switch state.orientation {
+            case .horizontal:
```
rdeassis, line 92: Given that we only have 2 orientations (vertical and horizontal), I think we could just use ternary operators on the edge insets to avoid the switch.
Another saving in binary size is that we're saving the dividerBackground property that is anotated with @ViewBuilder (since the Rectangle().fill() can be placed directly.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DividerDemoController.swift
```diff
@@ -0,0 +1,82 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class DividerDemoController: DemoController {
```
rdeassis, line 9: I think a SwiftUI demo controller could be a good showcase of the control directly in SwiftUI.

ios/FluentUI.Demo/FluentUI.Demo/Demos/DividerDemoController.swift
```diff
@@ -0,0 +1,82 @@
...
+            horizontalStack.addArrangedSubview(MSFDivider(style: style, orientation: .vertical, spacing: spacing).view)
+            let text2 = Label(style: .subhead, colorStyle: .regular)
+            text2.text = "Text 2"
+            horizontalStack.addArrangedSubview(text2)
+            container.addArrangedSubview(horizontalStack)
+        }
+    }
+
+    private enum DividerDemoCases: CaseIterable {
+        case defaultNone
```
rdeassis, line 42: Maybe add a custom color one?

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+                dividerBackground.frame(height: thickness)
+                    .padding(EdgeInsets(top: tokens.padding, leading: 0, bottom: tokens.padding, trailing: 0))
+            case .vertical:
+                dividerBackground.frame(width: thickness)
+                    .padding(EdgeInsets(top: 0, leading: tokens.padding, bottom: 0, trailing: tokens.padding))
+            }
+        }
+
+        return divider
```
rdeassis, line 101: I suspect you will need a `frame(maxWidth: .infinity)` for the horizontal scenario or `frame(maxHeight: .infinity)` for the vertical scenario when using this in a SwiftUI View directly.

huwilkes:
> 

Because the SwiftUI one is Divider, so we wanted to align names

ios/FluentUI/Divider/Divider.swift
huwilkes: I think I tried that, and it complained that infinity wasn't an allowed value, it needed to be a positive finite value. But I'll double check.

huwilkes:
> Dumb question, why is it called divider instead of separator?

Because the SiwftUI one is called Divider, and we wanted to align names.

tools/sgen/input/FluentUIStyle.yml
sophialee0416: Last time I spoke with the previous designer, they emphasized that they wanted to replicate how the UIKit separator looks. We could try doing a color comparison and seeing if they are open to darkening the Divider color so that it better matches the legacy color?

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+import UIKit
+import SwiftUI
+
+@objc public enum MSFDividerOrientation: Int {
+    case horizontal
+    case vertical
+}
+
+/// UIKit wrapper that exposes the SwiftUI Divider implementation.
+@objc open class MSFDivider: NSObject, FluentUIWindowProvider {
```
mischreiber, line 15: Would you mind moving `MSFDivider` to its own file, a la `MSFPersonaButton`?

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+    @objc open var state: MSFDividerState {
+        return divider.state
+    }
+
+    /// Creates a new MSFDivider instance.
+    ///  - Parameters:
+    ///   - style: The DividerStyle value used by the Divider.
+    ///   - orientation: The DividerOrientation used by the Divider.
+    ///   - spacing: The DividerSpacing used by the Divider.
+    @objc public init(style: MSFDividerStyle = .default, orientation: MSFDividerOrientation = .horizontal, spacing: MSFDividerSpacing = .none) {
```
mischreiber, line 32: There's no initializer variant that takes a `FluentUIStyle` for theming. Is this intentional?

ios/FluentUI/Divider/Divider.swift
mischreiber, line 101: You don't have a `.designTokens(_:from:with:)` view modifier on your `body` result, so there will be no way to apply design tokens (you'll need to address my earlier comment about missing the `FluentUIStyle`-containing `init` as well if that's something we want this control to support).

ios/FluentUI/Divider/Divider.swift
```diff
@@ -0,0 +1,137 @@
...
+            case .vertical:
+                dividerBackground.frame(width: thickness)
+                    .padding(EdgeInsets(top: 0, leading: tokens.padding, bottom: 0, trailing: tokens.padding))
+            }
+        }
+
+        return divider
+    }
+
+    private let thickness: CGFloat = UIScreen.main.devicePixel
```
mischreiber, line 104: Speculating: is the divider fainter because it's always a pixel wide, not a point wide? Or is there more going on with colors? (I openly admit I haven't looked much into it, just throwing ideas against the wall.)

huwilkes: I think that might be it. I was talking to Harrie about this earlier and she also suggested changing the simulator to pixel accurate instead of point accurate, which made the lines much clearer.

ios/FluentUI/Divider/Divider.swift
huwilkes: I meant non-negative, not positive. And I think the issue was actually that I was trying to put infinity into the width, not the max width. Seems to have no issue in maxWidth.

ios/FluentUI/Vnext/Divider/DividerTokens.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    @objc open func didChangeAppearanceProxy() {
+        updateForCurrentTheme()
+    }
+
+    override func updateForCurrentTheme() {
+        let appearanceProxy = theme.MSFDividerTokens
+
+        color = appearanceProxy.color.rest
```
huwilkes, line 37: Should just be color = appearanceProxy.color

rdeassis (commented):
I think that putting content (maybe labels) between the dividers would be good to make sure they're more visible in the demo.

Another change we can make is to provide a customized background color to help highlight the dividers.

ios/FluentUI/Vnext/Divider/MSFDivider.swift
```diff
@@ -0,0 +1,58 @@
...
+    ///   - orientation: The DividerOrientation used by the Divider.
+    ///   - spacing: The DividerSpacing used by the Divider.
+    ///   - theme: The FluentUIStyle instance representing the theme to be overriden for this Divider.
+    @objc public init(orientation: MSFDividerOrientation,
+                      spacing: MSFDividerSpacing,
+                      theme: FluentUIStyle?) {
+        super.init()
+
+        divider = FluentDivider(orientation: orientation, spacing: spacing)
+        hostingController = FluentUIHostingController(rootView: AnyView(divider
```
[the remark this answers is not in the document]
huwilkes: The overlap is from the table view, not the divider.

tools/sgen/input/FluentUIStyle.yml
```diff
@@ -320,6 +320,10 @@ AP_MSFCommandBarTokens:
   itemIconColor: [ rest: $Colors.Foreground.neutral1, hover: $Colors.Foreground.neutral1, pressed: $Colors.Foreground.neutral1, selected: $Colors.Foreground.neutralInverted, disabled: $Colors.Foreground.neutralDisabled ]
   itemInterspace: $Spacing.xxxSmall
 
+AP_MSFDividerTokens:
+  color: $Colors.Stroke.neutral1
```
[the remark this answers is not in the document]
huwilkes: That is the color in the design. Setting the simulator to pixel accurate helps to see the dividers.

ios/FluentUI/Vnext/Divider/MSFDivider.swift
```diff
@@ -0,0 +1,62 @@
...
+                                                                            .windowProvider(self)
```
rdeassis, line 45: I think we should add a call to `hostingController.disableSafeAreaInsets()` here in case the divider is used out of safe areas.

rdeassis (approved):

PR #804, opened 2021-11-19
Update bottom sheet offset when the safe area changes

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A change in safe area insets can affect the sheet height and hence the desired sheet offset, so let's make sure we update that after the next layout pass. This also has the nice side effect of ensuring the sheet has the right initial offset, because this callback is guaranteed to be called just before the view appears on screen - https://developer.apple.com/documentation/uikit/uiviewcontroller/2891116-viewsafeareainsetsdidchange

### Verification

Sanity checks in the test app with non-default initial values for isHidden and isExpanded.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/804)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -392,6 +392,7 @@ public class BottomSheetController: UIViewController {
 
     public override func viewWillTransition(to size: CGSize, with coordinator: UIViewControllerTransitionCoordinator) {
         super.viewWillTransition(to: size, with: coordinator)
+
         if size.height != view.frame.height {
             needsOffsetUpdate = true
         }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -398,6 +398,10 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    public override func viewSafeAreaInsetsDidChange() {
+        needsOffsetUpdate = true
```
anandrajeswaran, line 402: Is it worth calling super for safety? In case our inheritance changes?

anandrajeswaran (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Sure, makes sense.

rdeassis (approved):

PR #807, opened 2021-11-24
Ensure main_0.1 Demo App builds in Xcode 13 running on Apple Silicon machines.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Building the Demo App in Xcode 13 on Apple Silicon machines generates the following error:
`FluentUI.Demo.xcodeproj: error: The linked and embedded framework 'FluentUI.framework' is missing one or more architectures required by this target: x86_64. (in target 'FluentUI.Demo' from project 'FluentUI.Demo')`

This happens because the Demo App excludes the arm64 architecture in simulator builds (so it can link with the AppCenter framework successfully) and causes the app to be built in the x86_64 architecture.

The FluentUI and Pods projects currently do not exclude arm64 architecture for simulator builds, and as a result will not generate x86_64 arch builds on Apple Silicon machines because the default architecture is arm64 (ARCHS_STANDARD = arm64 armv7).

Given the context above, the fix is to set the `EXCLUDED_ARCHS[sdk=iphonesimulator*]" = arm64 ` for the FluentUI and Pods projects.

### Verification

1. Ran pod install to ensure the changes in the podfile produced the expected result in the Pods.
2. Built and ran the demo app on the simulator (iOS 15) from Xcode 13.1 in machines of both architectures:
-- Intel
-- Apple Silicon

3. Ensured that device builds and archiving was not affected by the changes as expected.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/807)


harrieshin:
hopefully it still works with older xcode as well

harrieshin (approved):

mischreiber:
> hopefully it still works with older xcode as well

I don't currently have any machines running an OS older than Big Sur, so I can't test old Xcode versions. But assuming someone is able to verify that, then this looks good!

mischreiber (approved):

rdeassis:
> hopefully it still works with older xcode as well

I shouldn't affect older versions as we're not touching any settings that were introduced in Xcode 12 or 13.
I have tested in Xcode Version 12.5 (12E262) and built with no issues.

rdeassis:
> > hopefully it still works with older xcode as well
> 
> I don't currently have any machines running an OS older than Big Sur, so I can't test old Xcode versions. But assuming someone is able to verify that, then this looks good!

Validated on Big Sur (11.3.1) and Xcode 12.5 (12E262).

PR #813, opened 2021-12-01
[vNext Tokens] New and improved `FluentTheme`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Now that we have discrete control tokens (#769), it's time to build out functionality for swapping them out at runtime.

_(Note: the eagle-eyed reviewer will recognize much of this code from #794. However, this adds full customizability for **all** tokens, including Global and Alias.)_

#### `TokenizedControl`

The newest protocol here is `TokenizedControl`, which defines how a control will expose its internal control tokens for mutation. Thus, a client can subclass a given control's token set and provide these tokens either directly on the `state` (when invoking from UIKit) or through a modifier on the `View` (in SwiftUI).

An example of this is now available in `ThemingDemoController`, which temporarily features a `CardNudge` in addition to its existing `Button` and `Avatar` instances.

#### `TokenSet`

In order to facilitate theming, `GlobalTokens` and `AliasTokens` have been refactored to allow for custom values. By default, these updated APIs will continue to return static values defined in enums (that will eventually be generated via the token pipeline). However, the new `TokenSet` base class allows for custom values to be set by a client. These overrides will be stored in a lazily-initialized dictionary, and returned if available.

#### `FluentTheme`

While some clients want to provide fully custom tokens for their controls, others simply want to apply a custom set of brand colors to the entire app, leaving other control tokens as they are. To facilitate this, a new protocol, `FluentTheme`, is available, and is piped through our tokens.

In a pure SwiftUI app, there is a simple view modifier called `fluentTheme(_:)` that can be applied to the root of the view hierarchy to be styled.

However, most of our clients are still UIKit-based. To facilitate this, this PR adds a new protocol, `FluentThemeable`, which this PR also applies to UIWindow. This allows us to "hang" a custom `FluentTheme` instantiation off a window. Now anyone with access to their window can also access the custom `FluentTheme` to be used for styling.

#### `ControlHostingContainer`

There's a moderate amount of glue code needed to connect the `Theme` and the SwiftUI controls being wrapped. This is a great opportunity to pull the creeping shared logic across our various UIKit wrappers into a single base class, `ControlHostingContainer`. This class will also contain the necessary logic to observe theme change events and use them to apply styling changes to the wrapped control view.

### Verification

This video demonstrates two important features:
- Dynamic runtime color swapping for all default-styled controls in a window.
- Specific per-control token customization which is not affected by window-level customization.

https://user-images.githubusercontent.com/4934719/139186179-c4f2944c-fb38-4c23-abff-505ac184dfd1.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/813)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
@@ -8,6 +8,10 @@ import UIKit
 
 class ThemingDemoController: DemoController {
 
+    // Store instances to keep from deallocating early
+    let cardNudge = MSFCardNudge(style: .outline, title: "Hello!")
+    let customCardNudge = MSFCardNudge(style: .outline, title: "Hello!")
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -49,6 +53,12 @@ class ThemingDemoController: DemoController {
         avatarOutlinedPrimary.state.presence = .away
 
         addRow(items: [avatarAccent.view, avatarDefault.view, avatarOutlinedPrimary.view], itemSpacing: 20)
+
+        cardNudge.state.accentText = "I'm using the token pipeline!"
+        cardNudge.state.dismissButtonAction = { _ in
+        }
+
+        addRow(items: [cardNudge.view])
 
         container.addArrangedSubview(UIView())
 
@@ -98,21 +108,26 @@ class ThemingDemoController: DemoController {
 
         addRow(items: [customThemeAvatarAccent.view, customThemeAvatarDefault.view, customThemeAvatarOutlinedPrimary.view], itemSpacing: 20)
 
+        customCardNudge.state.tokens = CustomCardNudgeTokens()
+        customCardNudge.state.accentText = "I'm using the token pipeline!"
+        customCardNudge.state.dismissButtonAction = { _ in
+        }
+
+        addRow(items: [customCardNudge.view])
+
         container.addArrangedSubview(UIView())
     }
 
     func didPressOverrideThemeButton() {
         if let window = self.view.window {
             let greenThemeColorProviding = DemoColorGreenTheme()
-            let stylesheet = ColorProvidingStyle(colorProviding: greenThemeColorProviding,
-                                                 window: window)
-            FluentUIThemeManager.setStylesheet(stylesheet: stylesheet, for: window)
+            Colors.setProvider(provider: greenThemeColorProviding, for: window)
         }
     }
 
     func didPressResetThemeButton() {
         if let window = self.view.window {
-            FluentUIThemeManager.removeStylesheet(for: window)
+            Colors.removeProvider(for: window)
         }
     }
 }
@@ -188,5 +203,24 @@ open class CustomStyle: FluentUIStyle {
         open override var xLarge: CGFloat {
             return 0.0
         }
+    }
+}
+
+class CustomCardNudgeTokens: CardNudgeTokens {
+    let purplePrimary: ColorSet = ColorSet(light: 0x6227A7)
+    override var accentColor: ColorSet {
+        return purplePrimary
+    }
+    override var outlineColor: ColorSet {
+        return purplePrimary
+    }
+    override var subtitleTextColor: ColorSet {
+        return purplePrimary
+    }
+    override var textColor: ColorSet {
+        return purplePrimary
+    }
+    override var cornerRadius: CGFloat {
+        return 0.0
     }
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
@@ -108,7 +108,7 @@ class ThemingDemoController: DemoController {
 
         addRow(items: [customThemeAvatarAccent.view, customThemeAvatarDefault.view, customThemeAvatarOutlinedPrimary.view], itemSpacing: 20)
 
-        customCardNudge.state.tokens = CustomCardNudgeTokens()
+        customCardNudge.state.overrideTokens = CustomCardNudgeTokens()
         customCardNudge.state.accentText = "I'm using the token pipeline!"
         customCardNudge.state.dismissButtonAction = { _ in
         }
--- a/ios/FluentUI/Card Nudge/CardNudge.swift
+++ b/ios/FluentUI/Card Nudge/CardNudge.swift
@@ -41,8 +41,8 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
     /// Action to be dispatched by the dismiss ("close") button on the trailing edge of the control.
     @objc var dismissButtonAction: CardNudgeButtonAction? { get set }
 
-    /// Design token set to use when drawing this control.
-    @objc var tokens: CardNudgeTokens? { get set }
+    /// Design token set for this control, to use in place of the control's default Fluent tokens.
+    @objc var overrideTokens: CardNudgeTokens? { get set }
 }
 
 /// View that represents the CardNudge.
@@ -54,7 +54,7 @@ public struct CardNudge: TokenizedControl {
     @Environment(\.fluentTheme) var fluentTheme: FluentTheme
     @ObservedObject var state: MSFCardNudgeStateImpl
     var tokens: CardNudgeTokens {
-        let tokens = state.tokens ?? fluentTheme.tokens(for: self)
+        let tokens = state.overrideTokens ?? fluentTheme.tokens(for: self)
         tokens.style = state.style
         return tokens
     }
@@ -202,8 +202,8 @@ class MSFCardNudgeStateImpl: NSObject, ObservableObject, Identifiable, MSFCardNu
     /// Action to be dispatched by the dismiss ("close") button on the trailing edge of the control.
     @Published @objc public var dismissButtonAction: CardNudgeButtonAction?
 
-    /// Design token set to use when drawing this control.
-    @Published @objc public var tokens: CardNudgeTokens?
+    /// Design token set for this control, to use in place of the control's default Fluent tokens.
+    @Published @objc public var overrideTokens: CardNudgeTokens?
 
     @objc init(style: MSFCardNudgeStyle, title: String) {
         self.style = style
--- a/ios/FluentUI/Card Nudge/CardNudge.swift
+++ b/ios/FluentUI/Card Nudge/CardNudge.swift
@@ -47,14 +47,10 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
 
 /// View that represents the CardNudge.
 public struct CardNudge: TokenizedControl {
-    public static func defaultTokens() -> CardNudgeTokens {
-        return CardNudgeTokens()
-    }
-
     @Environment(\.fluentTheme) var fluentTheme: FluentTheme
     @ObservedObject var state: MSFCardNudgeStateImpl
     var tokens: CardNudgeTokens {
-        let tokens = state.overrideTokens ?? fluentTheme.tokens(for: self)
+        let tokens = state.overrideTokens ?? fluentTheme.tokens(for: self) ?? state.defaultTokens
         tokens.style = state.style
         return tokens
     }
@@ -204,6 +200,9 @@ class MSFCardNudgeStateImpl: NSObject, ObservableObject, Identifiable, MSFCardNu
 
     /// Design token set for this control, to use in place of the control's default Fluent tokens.
     @Published @objc public var overrideTokens: CardNudgeTokens?
+
+    /// Lazily initialized default token set.
+    lazy var defaultTokens: CardNudgeTokens = .init()
 
     @objc init(style: MSFCardNudgeStyle, title: String) {
         self.style = style
--- a/ios/FluentUI/Card Nudge/CardNudge.swift
+++ b/ios/FluentUI/Card Nudge/CardNudge.swift
@@ -63,7 +63,7 @@ public struct CardNudge: TokenizedControl {
     var icon: some View {
         if let icon = state.mainIcon {
             ZStack {
-                Circle()
+                RoundedRectangle(cornerRadius: tokens.circleRadius)
                     .frame(width: tokens.circleSize, height: tokens.circleSize)
                     .dynamicForegroundColor(tokens.buttonBackgroundColor)
                 Image(uiImage: icon)
@@ -127,7 +127,7 @@ public struct CardNudge: TokenizedControl {
                 .padding(.vertical, tokens.verticalPadding)
                 .dynamicForegroundColor(tokens.accentColor)
                 .background(
-                    RoundedRectangle(cornerRadius: .infinity)
+                    RoundedRectangle(cornerRadius: tokens.circleRadius)
                         .dynamicForegroundColor(tokens.buttonBackgroundColor)
                 )
             }
--- a/ios/FluentUI/Card Nudge/CardNudge.swift
+++ b/ios/FluentUI/Card Nudge/CardNudge.swift
@@ -46,11 +46,11 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
 }
 
 /// View that represents the CardNudge.
-public struct CardNudge: TokenizedControl {
+public struct CardNudge: TokenizedControlInternal {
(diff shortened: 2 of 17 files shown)


ios/FluentUI/Card Nudge/CardNudgeModifiers.swift
```diff
@@ -52,4 +52,9 @@ public extension CardNudge {
         return self
     }
 
+    /// Design token set to use when drawing this control.
+    func customTokens(_ tokens: CardNudgeTokens) -> CardNudge {
```
huwilkes, line 56: I'm a bit confused by the comment and the actual function. To me, the comment sounds like the function should be returning the tokens? But it returns the whole CardNudge.

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -40,14 +40,24 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
...
 }
 
 /// View that represents the CardNudge.
-public struct CardNudge: View {
-    @Environment(\.theme) var theme: FluentUIStyle
-    @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+public struct CardNudge: TokenizedControl {
+    public static func defaultTokens() -> CardNudgeTokens {
+        return CardNudgeTokens()
+    }
```
huwilkes, line 52: If the client is supposed to subclass the tokens, what's the point of having this? They're supposed to have their own class, right? Why do we need to provide a way to get the default? Is the idea is they could get a default token object, modify that, then pass that back in to the class? Are we just giving them two ways to have custom tokens?

huwilkes: Also, if the token class is open, why not just have them call the init for it themselves? Seems like we could get away with just overrideTokens in TokenizedControl. As long as we have an example of customizing the default tokens by initializing them and passing them in to the control somewhere in our demo app, we could avoid adding a public static function to all of our controls.

ios/FluentUI/Card Nudge/CardNudgeModifiers.swift
mischreiber: Ah yep that can be clarified.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: Thanks for pointing this out. I was just rethinking this part of the system this morning: I think it's a little more complicated than I'd like, but it's also a complicated problem in general.

There are three places that a component (e.g. `CardNudge`) can get tokens, in priority order:
1. A custom token set from the `overrideTokens` property on its `state`.
2. A custom token set from the current `FluentTheme` object.
3. The default token set provided by Fluent.

This function is part of the mechanism for fetching and cacheing option three inside the FluentTheme - you can see the callback mechanism inside `FluentTheme.tokens(for:)` that invokes this static method if no custom tokens are availble. But I think the set of back-and-forth callbacks between `FluentTheme` and its `TokenizedControl`s is hard to follow. I'll think about simpler alternatives, but I'm also open to suggestions if anything comes to mind!

ios/FluentUI/Core/Theme/Tokens/TokenSet.swift
```diff
@@ -4,7 +4,37 @@
...
+            if valueOverrides == nil {
+                valueOverrides = [:]
+            }
+            valueOverrides?[token] = value
+        }
+    }
+
+    /// Initializes this token set with a callback to fetch its default values as needed.
+    ///
+    /// - Parameter defaultValues: The closure that can be used to lazily fetch default token values as needed.
```
rdeassis, line 33: nit: Instead of using "lazily", how about clarifying that the closure provide static default values that will be provided in case no override is provided for that specific token?

ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -83,4 +58,89 @@ public extension UIColor {
...
+
+    /// Creates a dynamic color object that returns the appropriate color value based on the current
+    /// rendering context.
+    ///
+    /// - Parameter colorSet: The set of color values that may be applied based on the current context.
+    convenience init(colorSet: ColorSet) {
+        self.init { traits -> UIColor in
+            let colorValue = colorSet.value(colorScheme: (traits.userInterfaceStyle == .dark ? .dark : .light),
+                                            contrast: traits.accessibilityContrast == .high ? .increased : .standard,
+                                            isElevated: traits.userInterfaceLevel == .elevated ? true : false)
```
rdeassis, line 82: just `traits.userInterfaceLevel == .elevated` instead of the ternary operator?

ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -83,4 +58,89 @@ public extension UIColor {
...
+    /// Creates a UIColor from a single 32-bit integer value.
+    ///
+    /// - Parameter hexValue: Integer value, generally represented as hexadecimal (e.g. `0x0086F0`), to use to initialize this color.
+    ///     Must be formatted as ARGB. Note: we will not utilize the alpha channel.
+    convenience init(colorValue: ColorValue) {
+        self.init(
+            red: CGFloat((colorValue & 0xFF0000) >> 16) / 255.0,
+            green: CGFloat((colorValue & 0x00FF00) >> 8) / 255.0,
+            blue: CGFloat((colorValue & 0x0000FF) >> 0) / 255.0,
+            alpha: 1.0)
```
rdeassis, line 71: Please remind me: Why don't we support alpha in the ColorValue?

ios/FluentUI/Card Nudge/CardNudgeModifiers.swift
rdeassis: Yeah.. Probably better to document all parameters and return type to clarify that.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -190,3 +205,22 @@ open class CustomStyle: FluentUIStyle {
         }
     }
 }
+
+class CustomCardNudgeTokens: CardNudgeTokens {
+    let purplePrimary: ColorSet = ColorSet(light: 0x6227A7)
+    override var accentColor: ColorSet {
```
rdeassis, line 211: Why we don't use the TokenSet approach for the Control tokens in the instance based customization scenario as well?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -8,6 +8,10 @@ import UIKit
 
 class ThemingDemoController: DemoController {
 
+    // Store instances to keep from deallocating early
+    let cardNudge = MSFCardNudge(style: .outline, title: "Hello!")
+    let customCardNudge = MSFCardNudge(style: .outline, title: "Hello!")
```
rdeassis, line 13: Is there a sample where you could just override a specific token that would impact all instances of the CardNudge with no need to set the tokens in the state object?

mischreiber: Not at present, no. The closest we have is the changing of some global tokens in `Colors.swift`, but nothing yet at the "all instances of this control" level. Would need some additional testing UI to host that, while this PR has so far been just testing existing functionality.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -50,6 +54,12 @@ class ThemingDemoController: DemoController {
 
         addRow(items: [avatarAccent.view, avatarDefault.view, avatarOutlinedPrimary.view], itemSpacing: 20)
 
+        cardNudge.state.accentText = "I'm using the token pipeline!"
```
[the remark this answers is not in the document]
mischreiber: Minor regression in the latest iteration. Thanks for catching!

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
mischreiber: Good question! Because a `TokenSet` mandates a consistent token type across its entire set, and by their nature Control tokens can have a variety of different token types.

As discussed offline, there is room for iteration on the Control token pattern, but that will be out of scope for this PR.

ios/FluentUI/Card Nudge/CardNudgeModifiers.swift
mischreiber: Do we want to do that for all modifiers on this (and all) controls? If so then that's gonna be a bit out of scope for this PR, but I can open a bug.

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -40,20 +40,26 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
...
+        tokens.style = state.style
+        return tokens
+    }
 
     @ViewBuilder
     var icon: some View {
         if let icon = state.mainIcon {
             ZStack {
-                Circle()
+                RoundedRectangle(cornerRadius: tokens.circleRadius)
```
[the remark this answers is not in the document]
mischreiber: Oh sure! I was testing token overrides, and I noticed that globally setting all `borderRadius` values to 0 didn't change the circle to a square as I expected. Turns out we were hardcoding the circle instead of using the existing `circleRadius` token, so I took a quick targeted fix here and on the action button.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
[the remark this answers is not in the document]
mischreiber: Okay fixed it for now. I agree that I want a new, better test controller for theming; that said, this PR has gotten large enough already, so I'd like to wait for a separate one if that's okay with you? I can revert my stuff here and stand up the new one at the same time, with far more toggles.

Alternatively, I can add "change theme" to the nav bar for existing demo controllers, so each control can test its own themeability. I kinda like that approach myself, what about you?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -190,3 +205,22 @@ open class CustomStyle: FluentUIStyle {
         }
     }
 }
+
+class CustomCardNudgeTokens: CardNudgeTokens {
```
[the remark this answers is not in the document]
mischreiber: As mentioned above, I agree that we need the scenario, but I'm personally leaning towards putting theme overrides into a global dropdown that can be accessed from each demo controller. Thoughts?

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: Limitation of the design token pipeline. Colors are pure web-style 24-bit hex, with alpha channels stored as separate tokens.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: FYI @huwilkes my response to your comment broke certain scenarios; the `FluentTheme` _needs_ to pre-configure the default tokens, to ensure they're referencing the correct `GlobalTokens` and `AliasTokens` instances.

My latest change 3b7f57a is an attempt at a compromise: there's no more public `defaultTokens()` method, but all tokens still pass through the `FluentTheme.tokens(for:)` API. Let me know if this version, along with the comments I added, make sense!

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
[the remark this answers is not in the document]
mischreiber: Screen Shot 2021-12-02 at 11 27 07 AM

ios/FluentUI/Card Nudge/CardNudge.swift
huwilkes: Ah, gotcha. That makes sense. The new implementations look good to me, but the comment about the return of `func tokens<T: TokenizedControlInternal>(for control: T) -> T.TokenType` isn't right anymore. And `/// - Parameter control: The control to fetch controls for.` should be `/// - Parameter control: The control to fetch tokens for.`. And a nit: The three different tokens do have different priorities, but I'm not sure they're really in a queue. I think just "Priority for token retrieval" should work fine, though.

huwilkes (approved):

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: Sure! I was just going for symmetry :)

ios/FluentUI/Core/Theme/Tokens/TokenSet.swift
mischreiber: Simplified this comment, and added a lengthier description about how that is expected to work.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: Great catch! Thanks, updated.

rdeassis (approved):
A consolidated note about topics that I believe to deserve a new ADO item under the feature or simply a discussion so we can reach a decision on which way to go about it:

- Demo Controller (new demos for overrides of global, alias, controls and per instance)
- Fonts in the global tokens (will be needed for tokenizing the remainder of the controls)
- Rename "state" to "configuration"
- Add alpha support to the ColorValue type

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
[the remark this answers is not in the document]
rdeassis: > 

Just out of curiosity: What caused the regression?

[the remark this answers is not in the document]
mischreiber: Sure!

And the regression was from https://github.com/microsoft/fluentui-apple/pull/813/commits/32e37318fead61b92282c2d1c3a31f6d48c89cb6 which attempted to clean up some logic around fetching control tokens from the `FluentTheme` in response to feedback. Fixed it with https://github.com/microsoft/fluentui-apple/pull/813/commits/277d678db898e547d0b1da53038fbc9dbbf369a5.

ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
```diff
@@ -0,0 +1,32 @@
...
+    ///
+    /// - Returns: A version of this control with these custom tokens applied.
+    func customTokens(_ tokens: TokenType) -> Self
+}
+
+/// Internal extension to `TokenizedControl` that adds information about `configuration`.
+protocol TokenizedControlInternal: TokenizedControl {
+    associatedtype ConfigurationType: ControlConfiguration where ConfigurationType.TokenType == TokenType
+
+    var state: ConfigurationType { get }
```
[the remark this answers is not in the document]
mischreiber: * Yep, it'll break things when this changes. I'll file a new task, but since I assume it'll come with this change I'm at least referring to the new protocols as `Configuration`. The property is still `state`.
* Yep, changing `state`=>`configuration` will be a big change, and won't be part of this PR (or indeed part of this branch at all, most likely).
* Good question. Our SwiftUI controls should be fine using Configuration, since there's no inheritance to worry about. UIKit will be trickier though, so we'll need to be careful about adding that name (though we've largely avoided that problem so far).

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: I'll have to discuss with design; right now their intent is for opacity to be stored separately from the color, which should be sufficient for our needs. Let's sync offline to discuss the pros and cons of such an approach.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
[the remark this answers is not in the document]
mischreiber: That's great, and almost exactly what I'm picturing too! I'll get a task logged for that work.

PR #821, opened 2021-12-07
Adding new items to the PR checklist.

rdeassis:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

In light of the evolution of Apple's technology and the addition of new features to the FluentUI library, some new PR validation checklist items are being suggested:

### Verification

This list is open for discussion and suggestions. Feel free to provide your thoughts on it.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/821)

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -18,7 +18,11 @@
 
 This PR has considered:
 - [ ] Light and Dark appearances
+- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
 - [ ] VoiceOver and Keyboard Accessibility
 - [ ] Internationalization and Right to Left layouts
 - [ ] Different resolutions (1x, 2x, 3x)
 - [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
+- [ ] iPad Pointer interaction
+- [ ] SwiftUI consumption (validation or new demo scenarios needed)
+- [ ] Design tokens validation (customization or override of design tokens)
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -24,5 +24,5 @@ This PR has considered:
 - [ ] Different resolutions (1x, 2x, 3x)
 - [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
 - [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
-- [ ] SwiftUI consumption (validation or new demo scenarios needed)
+- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
 - [ ] Objective-C exposure (provide it only if needed)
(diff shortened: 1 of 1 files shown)


harrieshin:
without actual public document of the design token info, I am not sure people would know what they need to check.

harrieshin:
same comment for SwiftUI validation. iPad Pointer maybe we can point to apple doc what it is...

rdeassis:
> same comment for SwiftUI validation. iPad Pointer maybe we can point to apple doc what it is...

How about adding linking this URL in the checklist item?
https://developer.apple.com/documentation/uikit/pointer_interactions

rdeassis:
> without actual public document of the design token info, I am not sure people would know what they need to check.

I see your point on not listing design tokens if there's no public documentation on our end. I can remove it.

SwiftUI seems to me as a different case as we're shipping SwiftUI controls on main. Thoughts?

harrieshin:
> > without actual public document of the design token info, I am not sure people would know what they need to check.
> 
> I see your point on not listing design tokens if there's no public documentation on our end. I can remove it.
> 
> SwiftUI seems to me as a different case as we're shipping SwiftUI controls on main. Thoughts?

I think there should some guidance here. make sure UIKit and swiftui apps can use our component. similar to objc and swift clients. but I don't know it is obvious to the community.

mischreiber:
> > > without actual public document of the design token info, I am not sure people would know what they need to check.
> > 
> > 
> > I see your point on not listing design tokens if there's no public documentation on our end. I can remove it.
> > SwiftUI seems to me as a different case as we're shipping SwiftUI controls on main. Thoughts?
> 
> I think there should some guidance here. make sure UIKit and swiftui apps can use our component. similar to objc and swift clients. but I don't know it is obvious to the community.

Interesting point: our test app generally does not exercise any of our components' Objective-C interops. It's a bit of a hole, but it would be quite a bit more work to fully test everything across all three possible consumption models:
- SwiftUI
- UIKit (Swift)
- UIKit (Objective-C)

But maybe I just talked myself into wanting that?

rdeassis:
> > > > without actual public document of the design token info, I am not sure people would know what they need to check.
> > > 
> > > 
> > > I see your point on not listing design tokens if there's no public documentation on our end. I can remove it.
> > > SwiftUI seems to me as a different case as we're shipping SwiftUI controls on main. Thoughts?
> > 
> > 
> > I think there should some guidance here. make sure UIKit and swiftui apps can use our component. similar to objc and swift clients. but I don't know it is obvious to the community.
> 
> Interesting point: our test app generally does not exercise any of our components' Objective-C interops. It's a bit of a hole, but it would be quite a bit more work to fully test everything across all three possible consumption models:
> 
> * SwiftUI
> * UIKit (Swift)
> * UIKit (Objective-C)
> 
> But maybe I just talked myself into wanting that?

I think this is a list of potential applicable tests/validations to a specific change. Not a list of mandatory test gate that all PRs need to go through.

I agree that testing all the items irrespective of the change being made is too much work with no real accomplishment.

I added the Objective-C item as the UIKit one is just the standard.

mischreiber (approved):

sophialee0416:
I know this is already approved but should we also consider pod lib lint verification in the checklist or is that an unspoken rule?

rdeassis:
> I know this is already approved but should we also consider pod lib lint verification in the checklist or is that an unspoken rule?

I think it'd be worth adding if it was not part of our PR validation checks.
Now that pod lib lint runs when we create PRs, it should cover the scenario.

PR #823, opened 2021-12-07
Implement show/hide of SwiftUI notification with UIKit

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Rewrite notification demo controller as a TableViewDemoController
Implement show/hide of SwiftUI notification with UIKit

| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-12-07 at 11 54 53 | Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-12-07 at 11 30 30 |
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-12-07 at 11 55 44|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2021-12-07 at 13 45 10|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/823)


ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
```diff
@@ -4,105 +4,467 @@
...
-                        VStack(spacing: 0) {
-                            Text("Action")
-                                .frame(maxWidth: .infinity, alignment: .leading)
-                                .font(.title)
-                            Divider()
-                        }
-                        FluentUIDemoToggle(titleKey: "Has Action Button Action", isOn: $hasActionButtonAction)
-                        FluentUIDemoToggle(titleKey: "Has Message Action", isOn: $hasMessageAction)
-                    }
+        switch row {
```
edjamesmsft, line 51: I'm not sure what the standard practice is on this team, but this switch statement is very large and nested. Would it make sense to break out the content of each of the switch statements into well-named helper methods to condense this a bit? No biggie if that's not the done thing, though :)

jeaniehuynh: I just took a look and 7 out of 9 of the demo controllers using DemoTableViewController create their rows this way, so it seems like the standard to me! Can be handled in a separate PR if needed :)

edjamesmsft: Sounds good! I think it would be beneficial for this to be condensed for easier reading, but definitely don't block on it, especially since it would be good to clean up in a number of places.

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,194 @@
...
+    // MARK: - FluentUIWindowProvider
+
+    var window: UIWindow? {
+        return self.view.window
+    }
+
+    // MARK: - Show/Hide Methods
+
+    public func showNotification(in view: UIView, completion: ((MSFNotification) -> Void)? = nil) {
+        if isShown {
```
rdeassis, line 69: Use guard for early return.

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
```diff
@@ -172,47 +168,30 @@ public struct NotificationViewSwiftUI: View {
...
-                case .actionButton:
-                    if let actionTitle = state.actionButtonTitle {
-                        if actionTitle.isEmpty {
-                            alertString = "Dismiss"
-                        } else {
-                            alertString = actionTitle
-                        }
-                    } else {
-                        alertString = "Action"
+        if let windowWidth = UIApplication.shared.keyWindow?.bounds.width {
```
rdeassis, line 171: Have you looked at the GeometryReader to retrieve the dimensions of the screen?
I think using UIKit's window is not ideal here.

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    public var style: MSFNotificationStyle {
+        get {
+            return notification.state.style
+        }
+
+        set {
+            notification.state.style = newValue
```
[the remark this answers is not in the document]
jeaniehuynh: That's what I thought, but removing this line will make the notification's colors not update. For example, this bar should be gray but it maintains the primary toast (default)'s color scheme
image

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
jeaniehuynh: As discussed offline, we will keep the UIKit's window usage here because GeometryReader relies on the bounds of the hosting controller which cannot be set accurately at this point

mischreiber: Mmmmmm I hear your point about not being able to use `GeometryReader`, but I think there's still a real scenario here for multi-window. Example: what if I have two Word windows visible, am actively typing in one (making it `keyWindow`), but the other window is trying to show a notification?

mischreiber:
Just as a heads up, I'm able to crash the SwiftUI demo with the following:
* Open the SwiftUI demo
* Show a Notification pop-up
* Go back while the demo is still visible
* Wait for the now-hidden Notification to dismiss

See video:

https://user-images.githubusercontent.com/4934719/149601649-358ee4ec-024d-445f-9cda-33dc0699e47f.mp4

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,6 +45,7 @@ public struct NotificationViewSwiftUI: View {
     @Environment(\.swiftUIInsets) private var safeAreaInsets: EdgeInsets
     @Environment(\.theme) var theme: FluentUIStyle
     @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
+    @Environment(\.viewController) var viewController: UIViewController?
```
mischreiber, line 48: I see how you're reading from this variable below, but I don't see where it's set. Does that mean the code is currently always using the default value?

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -23,3 +23,20 @@ private extension UIEdgeInsets {
...
     }
 }
+
+struct ViewControllerHolder {
+    weak var value: UIViewController?
+}
+
+struct ViewControllerKey: EnvironmentKey {
+    static var defaultValue: ViewControllerHolder {
+        return ViewControllerHolder(value: UIApplication.shared.windows.first?.rootViewController)
```
mischreiber, line 33: `-[UIApplication windows]` is deprecated; we'll want to use a different API to get a default window (maybe back to the key window?)

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
```diff
@@ -159,8 +160,8 @@ public struct NotificationViewSwiftUI: View {
 
     public var body: some View {
         // Note: we are using keyWindow's width because GeometryReader relies on the bounds of the hosting controller which cannot be set accurately at this point
```
mischreiber, line 162: Nit: update comment

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
```diff
@@ -172,54 +159,45 @@ public struct NotificationViewSwiftUI: View {
...
-                    if let actionTitle = state.actionButtonTitle {
-                        if actionTitle.isEmpty {
-                            alertString = "Dismiss"
-                        } else {
-                            alertString = actionTitle
-                        }
-                    } else {
-                        alertString = "Action"
+        // Note: we are using viewController's width because GeometryReader relies on the bounds of the hosting controller which cannot be set accurately at this point
+        if let viewWidth = viewController?.view.bounds.width {
```
mischreiber, line 163: Note that I'm still a little hesitant about having a UIKit type in here. This will make hosting this view in a "pure" SwiftUI app more difficult, as the caller may not have ready access to any UIViewControllers.

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        hostingController = FluentUIHostingController(rootView: AnyView(notification
+                                                                            .windowProvider(self)
+                                                                            .modifyIf(theme != nil, { notification in
+                                                                                notification.customTheme(theme!)
+                                                                            })))
+        hostingController.view.backgroundColor = UIColor.clear
+        hostingController.disableSafeAreaInsets()
+    }
+
+    public var state: MSFNotificationState {
```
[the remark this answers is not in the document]
jeaniehuynh: An earlier commit was giving me errors (Property cannot be marked @objc because its type cannot be represented in Objective-C), but the lines causing that error have been changed and I missed putting it back!

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
jeaniehuynh: When I create the pure SwiftUI change, I will have it use GR instead to address that issue!

Currently idea: we have `if let viewWidth = viewController?.view.bounds.width` already, if true/viewController exists, use viewController width, else,  construct view with GR.

mischreiber (approved):

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -24,17 +24,19 @@ private extension UIEdgeInsets {
...
-                $0 as? UIWindowScene
-            }
-            .flatMap {
-                $0.windows
-            }
-            .first {
-                $0.isKeyWindow
-            }
+struct ViewControllerHolder {
+    weak var value: UIViewController?
```
rdeassis, line 28: Should the SafeAreaInsets above be retrieved from the window associated with this UIViewController? (it seems to me that it should)

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -24,17 +24,19 @@ private extension UIEdgeInsets {
...
+}
+
+struct ViewControllerKey: EnvironmentKey {
+    static var defaultValue: ViewControllerHolder {
+        return ViewControllerHolder(value: UIApplication.shared.keyWindow?.rootViewController)
+    }
+}
+
+extension EnvironmentValues {
+    var viewController: UIViewController? {
```
rdeassis, line 38: nit: how about naming this as something like "ParentViewController" or "HostingViewController"?

ios/FluentUI/Notification/MSFNotificationTokens.swift
```diff
@@ -15,6 +15,25 @@ import UIKit
     case dangerToast
     case warningToast
 
+    public var styleString: String {
```
rdeassis, line 18: This should be added in an extension of the MSFNotificationStyle enum in the demo app (example in the link below):

https://github.com/microsoft/fluentui-apple/blob/7fbc76a9a60e69d84fd9026b134fc1c15725a27f/ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift#L217

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        let style = notification.tokens.style
+        let presentationOffset: CGFloat! = notification.tokens.presentationOffset
+        if style.isToast, let currentToast = MSFNotification.currentToast, currentToast.view.window != nil {
+            currentToast.hide {
+                self.showNotification(in: view, completion: completion)
+            }
+            return
+        }
+
+        hostingController = FluentUIHostingController(rootView: AnyView(notification
```
rdeassis, line 64: Why is the hosting controller being recreated here? Isn't it possible to just reassign its current root view with the environment value?

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        let completionForShow = { (_: Bool) in
+            UIAccessibility.post(notification: .layoutChanged, argument: hostingControllerView)
+            completion?(self)
+        }
+
+        if animated {
+        view.layoutIfNeeded()
+            UIView.animate(withDuration: style.animationDurationForShow, delay: 0, usingSpringWithDamping: style.animationDampingRatio, initialSpringVelocity: 0, animations: {
```
rdeassis, line 104: nit: can you break the params into multiple lines for better readability?

ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        if updatedDelay > 0 {
+            DispatchQueue.main.asyncAfter(deadline: .now() + updatedDelay) { [weak self] in
+                self?.hide(completion: completion)
+            }
+            return
+        }
+
+        let hostingControllerView = self.hostingController.view
+        if let completion = completion {
+            completionsForHide.append(completion)
```
rdeassis, line 138: Why is completionsForHide an array? Is it calling completions for other toasts that were previously requested to be hidden?

ios/FluentUI/Notification/NotificationView_SwiftUI.swift
rdeassis, line 163: One idea that came to my mind here. What if you use `windowProvider?.window?.bounds.width` here instead of passing the environment value just for that view controller?
Seems to me you'd be able to save a lot of code in the show method that presents it.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
```diff
@@ -4,105 +4,528 @@
 //
 
 import FluentUI
-import SwiftUI
+import UIKit
 
-class NotificationViewDemoControllerSwiftUI: UIHostingController<NotificationDemoView> {
-    override init?(coder aDecoder: NSCoder, rootView: NotificationDemoView) {
-        preconditionFailure("init(coder:) has not been implemented")
+class NotificationViewDemoControllerSwiftUI: DemoTableViewController, UIPickerViewDataSource, UIPickerViewDelegate {
```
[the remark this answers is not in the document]
jeaniehuynh: Correct!

ios/FluentUI/Notification/MSFNotificationView.swift
jeaniehuynh: I'm not sure, it was how whoever created the original UIKit version of Notification had implemented it, so I left it.

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -551,6 +555,15 @@
 			path = "Pill Button Bar";
 			sourceTree = "<group>";
 		};
+		4BA9B8B5279F2940007536F5 /* Notification */ = {
+			isa = PBXGroup;
+			children = (
+				43488C44270FAD0200124C71 /* NotificationView_SwiftUI.swift */,
```
rdeassis, line 561: super nit: can you arrange the files in alphabetical order in both folders?

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -7,7 +7,8 @@ import SwiftUI
 
 struct SafeAreaInsetsKey: EnvironmentKey {
     static var defaultValue: EdgeInsets {
-        UIApplication.shared.keyWindow?.safeAreaInsets.swiftUIInsets ?? EdgeInsets()
+        @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
```
rdeassis, line 10: Wouldn't be simpler to just add a computed property directly in the NotificationViewSwiftUI View since it's got the windowProvider environment value?

ios/FluentUI/Extensions/UIApplication+Extensions.swift
```diff
@@ -0,0 +1,21 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+extension UIApplication {
```
rdeassis, line 8: It seems like this extension is not being used anymore. Can you delete this file?

ios/FluentUI/Vnext/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        let style = notification.tokens.style
+        let presentationOffset: CGFloat! = notification.tokens.presentationOffset
+        if style.isToast, let currentToast = MSFNotification.currentToast, currentToast.view.window != nil {
+            currentToast.hide {
+                self.showNotification(in: view, completion: completion)
+            }
+            return
+        }
+
+        hostingController.disableSafeAreaInsets()
```
rdeassis, line 64: This is already being called in line 35. Shouldn't be needed here.

ios/FluentUI/Vnext/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        var hostingControllerView: UIView! = hostingController.view
```
rdeassis, line 65: duplicated code (lines 65 and 66)

ios/FluentUI/Vnext/Notification/MSFNotificationView.swift
```diff
@@ -0,0 +1,182 @@
...
+        hostingControllerView = hostingController.view
+        hostingControllerView.translatesAutoresizingMaskIntoConstraints = false
```
rdeassis, line 67: this will crash if hostingControllerView is nil.

you can validate it in the beginning of the func...  something like this:

```
guard let hostingControllerView = hostingController.view else {
    return
}
```

rdeassis (approved):

PR #826, opened 2021-12-08
[vNext Tokens] Update `ControlTokens` to be statically initialized, rather than subclassed

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change brings a new approach to `ControlTokens`. Before, clients were expected to subclass the token set for a desired control and provide overrides for any properties. Now, there's a simple `init()` function that takes all properties as arguments and returns a fully-formed instance of their desired `ControlTokens` class.

This has a few benefits:
- More binary size improvement! With this, there's no need to expose individual properties on tokens to Fluent consumers. This'll save us a lot, especially on controls with tons of properties (I'm looking at you, `Button`).
- We only re-render our controls if they are given a fresh new instance of `ControlTokens`, either directly or through a FluentTheme. Allowing for overrides would let clients change the value under us, which may seem like a bug to them.

### Verification

Verified all known scenarios for `CardNudge` theming (colors, light+dark, outline vs standard).

| Metric     | Before               | After                   |
|------------|----------------------|-------------------------|
| Number of public properties on `CardNudgeTokens` | 19 | 1 |
| Release device build of FluentUI.Demo | 31,868,050 bytes | 31,769,202 bytes |
| Able to modify properties behind our backs without creating a new instance of `CardNudgeTokens` | yes 💔 | no 🥳 |

 🎉🎉🎉 _**Delta: 98,848 bytes, or ~97 kB.**_ 🎉🎉🎉

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/826)


ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -46,7 +46,7 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
 }
 
 /// View that represents the CardNudge.
-public struct CardNudge: TokenizedControlInternal {
+public struct CardNudge: View, TokenizedControlInternal {
```
harrieshin, line 49: for my own knowledge, why is the View needed?

mischreiber: Because I removed it from `TokenizedControlInternal`, so I could apply that protocol to `UIView` as well.

mischreiber:
Converting to draft because `OverridableToken` is adding too much binary size. This is gonna be a tricky one to solve...

ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -6,19 +6,32 @@
...
+        // the caller does not give us a custom set themselves.
         if let aliasTokens = aliasTokens {
             self.aliasTokens = aliasTokens
+        } else if hasCustomGlobalTokens {
+            self.aliasTokens = .init()
+        }
+
+        // If we have custom global tokens, our "custom" alias tokens from above must track them.
+        if hasCustomGlobalTokens {
+            self.aliasTokens.globalTokens = self.globalTokens
```
huwilkes, line 34: Seems like we should be able to combine the two if hasCustomGlobalTokens. Or maybe move the alias tokens above the global tokens so we can set the aliasTokens.globalTokens in the  global tokens section?

mischreiber: Yeah I agree, let me see how I can clean this up a bit.

harrieshin (approved):

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
```diff
@@ -15,97 +15,122 @@ import SwiftUI
...
-    open var buttonInnerPaddingHorizontal: CGFloat {
-        return globalTokens.spacing[.small]
+        return tokens
     }
 
-    open var circleRadius: CGFloat {
-        return globalTokens.borderRadius[.circle]
-    }
+    fileprivate var _accentColorOverride: ColorSet?
+    var accentColor: ColorSet { _accentColorOverride ?? globalTokens.brandColors[.shade20] }
```
mischreiber, line 74: I'm not a big fan of having separate override storage from the actual token getter, but I'm at a loss as to how to support all three of the following features:
- Lazy default values that reference the superclass's `globalTokens` and `aliasTokens` properties.
- Internal overrides for alternate "styles" (e.g. `.outline` vs `.standard` here), which _also_ lazily reference `[global|alias]Tokens`.
- Optional storage of custom properties, of a variety of different types for different tokens (which makes flat array/dictionary storage hard).

If anyone has any ideas, I'm eager to hear them!

huwilkes: I'm still working to understand the changes, but would it be possible to just treat our different styles the same way we treat external custom tokens? Do we already do that? Seems like different styles and custom tokens should be fairly similar. Or do the custom tokens also make those three features hard to satisfy?

mischreiber: Great question! That's basically what last week's version (ea1a2bcecdaebb0d9a854ecfcec9d7bf8fea0cd9) tried, but it ran into a problem with custom global tokens. 

See, the tricky part is that custom tokens are constant values, while our styles can be dynamically pulled from the global or alias token sets, e.g. for brand colors. It's easy to override a property with `Color.red`, but I haven't been able to think of a way to override it with something like `self.globalTokens.brandColors[.primary]`, particularly since `globalTokens` won't yet be set at `init()`.

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
```diff
@@ -16,121 +16,151 @@ import SwiftUI
...
-    fileprivate var _accentColorOverride: ColorSet?
-    var accentColor: ColorSet { _accentColorOverride ?? globalTokens.brandColors[.shade20] }
+    // Required state value
+    let style: MSFCardNudgeStyle
+
+    // MARK: - Design Tokens
 
-    fileprivate var _accentIconSizeOverride: CGFloat?
-    var accentIconSize: CGFloat { _accentIconSizeOverride ?? globalTokens.iconSize[.xxSmall] }
+    lazy var accentColor: ColorSet = globalTokens.brandColors[.shade20]
```
rdeassis, line 108: These tokens won't be visible publicly for client apps to query them as read only?

The computed properties approach saves all those if lets in the initializer.
Was it changed because of any size impact?

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
```diff
@@ -16,121 +16,151 @@ import SwiftUI
...
-                              verticalPadding: CGFloat? = nil) -> CardNudgeTokens {
-
-        let tokens: CardNudgeTokens
-        switch style {
-        case .standard:
-            tokens = CardNudgeTokens()
-        case .outline:
-            tokens = OutlineCardNudgeTokens()
-        }
+    public init(style: MSFCardNudgeStyle,
```
rdeassis, line 19: I know we're not planning to expose this to Objc. But does it make sense to have the designated initializer internal and publish only convenience initializers?

ios/FluentUI/Core/Theme/Tokens/ControlTokens.swift
```diff
@@ -6,7 +6,7 @@
 import Foundation
 
 /// Base class for all Fluent control tokenization.
-open class ControlTokens: NSObject {
-    var globalTokens: GlobalTokens = GlobalTokens.shared
-    var aliasTokens: AliasTokens = AliasTokens.shared
+public class ControlTokens: NSObject {
+    lazy var globalTokens: GlobalTokens = FluentTheme.shared.globalTokens
```
rdeassis, line 10: What are pros/cons on the approaches for these references?
1. lazy var?
2. computed property (var)?
3. let (constant)?

rdeassis (approved):

ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
```diff
@@ -34,6 +34,10 @@ protocol TokenizedControlInternal: TokenizedControl {
 /// Internal protocol for all controls' internal `configuration` objects.
 protocol ControlConfiguration: NSObject, ObservableObject, Identifiable {
     associatedtype TokenType: ControlTokens
+
+    /// Custom design token set for this control, to use in place of the control's default Fluent tokens.
     var overrideTokens: TokenType? { get set }
-    var defaultTokens: TokenType { get }
+
+    /// On-demand default token set.
+    func createDefaultTokens() -> TokenType
```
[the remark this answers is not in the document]
mischreiber: I did have it as a property until the most recent few commits, but there was a brief time where it _needed_ to be generated/computed, so making it a function was a nice way to enforce that. Now that it doesn't matter, I'll switch it back to a `var`.

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
mischreiber: 1. Correct. I don't know of any scenario where anyone would want these to be readable. We can make them public if a specific customer need arises, though at great impact to our binary size.
2. Exactly! This conversion saved us ~40kB from the previous iteration. Personally I'm happy to trade slightly messier code for such a massive improvement (think about how much that'll save us when we convert `Button`!)

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
mischreiber: I did think about that, but I'm not sure what convenience initializers would look like. Ultimately there would need to be a public initializer with this same signature, at which point other convenience initializers are nice-to-have but not critical. Still, I'm open to adding a public `init(style:)` convenience initializer in addition to this one for convenience's sake (and it should be pretty tiny, binary-size-wise).

ios/FluentUI/Core/Theme/Tokens/ControlTokens.swift
mischreiber: Great question!
### lazy var

**Pros:**
- Single source of truth for the variable
- Can be directly set in our `init`
- Is able to fetch values from `globalTokens` and `aliasTokens`, which are not yet available at `init` time
- Does not need to be overridden, and thus does not need to be `public` outside Fluent library

**Cons:**
- Cannot be overridden by a subclass (e.g. `OutlineCardNudgeTokens`)
- Cannot be set at `init` time by a subclass since they need to be lazy

### computed property (var)

**Pros:**
- Is able to fetch values from `globalTokens` and `aliasTokens`, which are not yet available at `init` time
- Can be overridden by a subclass (e.g. `OutlineCardNudgeTokens`)

**Cons:**
- Cannot store override values from clients without separate storage properties (greatly increasing binary size)

### let (constant)

**Pros:**
- Easy to read

**Cons:**
- Cannot be changed by client in `init`
- Cannot be overridden by a subclass (e.g. `OutlineCardNudgeTokens`)
- Cannot be lazy and thus cannot reference `globalTokens` or `aliasTokens`, which are not available at `init`

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -203,8 +203,10 @@ class MSFCardNudgeStateImpl: NSObject, ControlConfiguration, MSFCardNudgeState {
     /// Style to draw the control.
     let style: MSFCardNudgeStyle
 
-    /// Lazily initialized default token set.
-    var defaultTokens: CardNudgeTokens { .create(style: self.style) }
+    /// On-demand default token set.
+    func createDefaultTokens() -> CardNudgeTokens {
```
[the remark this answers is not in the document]
mischreiber: Done!

PR #827, opened 2021-12-09
add `maxFontSize` for badgeview

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Add `maxFontSize` for clients who need to control the text size of the badgeview label.
Update the logic how the frame is calculated for badgeview so it will dynamically adjust based on label preferred size.

### Verification

|Simulator Screen Shot - iPhone 11 Pro Max - 2021-12-09 at 13 14 08  | Simulator Screen Shot - iPhone 11 Pro Max - 2021-12-09 at 12 00 49 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/827)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -271,6 +271,18 @@ open class BadgeView: UIView {
         }
     }
 
+    /**
+     The maximum allowed size point for the label's font. This property can be used
+     to restrict the largest size of the label when scaling due to Dynamic Type. The
+     default value is 0, indicating there is no maximum size.
+     */
+    open var maxFontSize: CGFloat = 0 {
+        didSet {
+            label.maxFontSize = maxFontSize
+            invalidateIntrinsicContentSize()
+        }
+    }
+
     open override var intrinsicContentSize: CGSize {
         return sizeThatFits(CGSize(width: CGFloat.infinity, height: CGFloat.infinity))
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -271,6 +271,18 @@ open class BadgeView: UIView {
         }
     }
 
+    /**
+     The maximum allowed size point for the label's font. This property can be used
+     to restrict the largest size of the label when scaling due to Dynamic Type. The
+     default value is 0, indicating there is no maximum size.
+     */
+    open var maxFontSize: CGFloat = 0 {
```
rdeassis, line 279: Is this going to be needed in ObjC?

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -327,15 +339,7 @@ open class BadgeView: UIView {
...
         backgroundView.frame = bounds
-
-        let labelHeight = label.font.deviceLineHeight
-        let labelSize = label.sizeThatFits(CGSize(width: CGFloat.infinity, height: CGFloat.infinity))
-        let fittingLabelWidth = UIScreen.main.roundToDevicePixels(labelSize.width)
-
-        let minLabelWidth = minWidth - size.horizontalPadding * 2
-        let maxLabelWidth = frame.width - size.horizontalPadding * 2
-        let labelWidth = max(minLabelWidth, min(maxLabelWidth, fittingLabelWidth))
-        label.frame = CGRect(x: size.horizontalPadding, y: size.verticalPadding, width: labelWidth, height: labelHeight)
```
mischreiber, line 338: Honestly I can barely make heads or tails of this--I can see _what_ it's doing, but I don't understand _why_. Is there any risk to replacing this logic for, say, extremely long or extremely short labels?

ios/FluentUI/Badge Field/BadgeView.swift
harrieshin: based on my understanding on this client usage, it seems like they are using swift. so i purposely didn't expose objc until i have to.

ios/FluentUI/Badge Field/BadgeView.swift
harrieshin: can you elaborate more for me? at least there is min width for the label so the short one it is taken care of ... but is the new logic really different from what it was doing before?

ios/FluentUI/Badge Field/BadgeView.swift
rdeassis: Thanks for confirming!

rdeassis (approved):

ios/FluentUI/Badge Field/BadgeView.swift
mischreiber: Nope you're right, my bad. Great optimization!

mischreiber (approved):

PR #830, opened 2021-12-13
BottomSheet: Adding custom accessibility labels for the handle.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

One of our client apps requested the ability to set a custom accessibility label for the Bottom Sheet handle.
This change adds 2 new properties to set the custom accessibility label of the Bottom Sheet handle for each of its visible state: expanded and collapsed.

### Verification

Ran changes in a physical device in the Voice Over scenario:

https://user-images.githubusercontent.com/68076145/145899362-ecb4f316-0550-481b-8714-66a9591df6f9.MP4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/830)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -110,6 +110,22 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// A string to optionally customize the accessibility label of the bottom sheet handle.
+    /// The message should convey the "Expand" action and will be used when the bottom sheet is collapsed.
+    @objc public var handleExpandCustomAccessibilityLabel: String? {
+        didSet {
+            updateResizingHandleViewAccessibility()
+        }
+    }
+
+    /// A string to optionally customize the accessibility label of the bottom sheet handle.
+    /// The message should convey the "Collapse" action and will be used when the bottom sheet is expanded.
+    @objc public var handleCollapseCustomAccessibilityLabel: String? {
+        didSet {
+            updateResizingHandleViewAccessibility()
+        }
+    }
+
     /// Indicates if the bottom sheet is expanded.
     ///
     /// Changes to this property are animated. A new value is reflected in the getter only after the animation completes.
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -110,6 +110,22 @@ public class BottomSheetController: UIViewController {
         }
     }
 
+    /// A string to optionally customize the accessibility label of the bottom sheet handle.
+    /// The message should convey the "Expand" action and will be used when the bottom sheet is collapsed.
+    @objc public var handleExpandCustomAccessibilityLabel: String? {
```
harrieshin, line 115: should we really separate expand and collapse label expose separately? or should we just have 1 label that clients can do whatever in what ever situation. technically the client knows the expand or collapse or 1/3 collapse etc state. they might want to convey more?

rdeassis: This would be possible if we updated the delegate to be triggered from all different ways (user gesture or different API calls) the bottom sheet can be changed:

```
/// Called after a transition to a new expansion state completes.
    ///
    /// External changes to`isExpanded` or `isHidden` will not trigger this callback.
    /// - Parameters:
    ///   - bottomSheetController: The caller object.
    ///   - expansionState: The expansion state that the sheet moved to.
    ///   - interaction: The user interaction that caused the state change.
    @objc optional func bottomSheetController(_ bottomSheetController: BottomSheetController,
                                              didMoveTo expansionState: BottomSheetExpansionState,
                                              interaction: BottomSheetInteraction)
```

rdeassis: I think I was misled by the API documentation remark.
`External changes to isExpanded or isHidden will not trigger this callback.`

I looked a bit further and I couldn't find a scenario that this delegate will not be called.

But still, I think it'd still be more error prone (bug prone) to rely on the caller to use the delegate to do it as I expect they could just set the property when creating the control and not updated it from the callback.

harrieshin: the size concern was the main driving factor my comment and BottomSheetExpansionState is not boolean so didn't know how it could really be evolved in the future.

harrieshin (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
rdeassis: I see your point... 
The bright side of it is that if not other state is created, the other state is "hidden" where the handle is not visible or accessible in the UI.

PR #832, opened 2021-12-15
[vNext Tokens] Update `CommandBarTokens`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Controls will start coming fast after this, but I wanted to call out `CommandBar` independently, as it's our first converted UIKit control.

Not much exciting happening here, honestly. The biggest item of note is the addition of an observer for the `didChangeTheme` notification, which allows us to change our theme while visible. This is an enhancement above the existing vNext control, which features no such ability.

It's also worth noting that most of the token protocols assume SwiftUI controls, and thus also expect the structure of our SwiftUI controls. In other words, they want the control to have a `state` object. There's no such object on our UIKit controls, so the control itself will conform to the `ControlConfiguration` protocol.

### Verification

Size improved from 31,768,161 -> 31,735,921, which means an improvement of **~31 kB**. Not bad!

| Default                                      | Custom colors                              |
| blue_light | green_light |
| blue_dark | green_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/832)


sophialee0416:
We had a conversation a while back about how the keyboard does not change with the theme because the notification for theme change is sent per window. Will this be changed in the new token system?

mischreiber:
> We had a conversation a while back about how the keyboard does not change with the theme because the notification for theme change is sent per window. Will this be changed in the new token system?

...yes? Let's say yes. Also please file a bug against me when you get a chance so I don't forget to fix that 😬

harrieshin:
how much did you save here? :)

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -39,16 +37,30 @@ open class CommandBar: UIView, FluentUIWindowProvider {
...
 
-        backgroundColor = commandBarTokens.backgroundColor
+        backgroundColor = UIColor(colorSet: tokens.backgroundColor)
         translatesAutoresizingMaskIntoConstraints = false
 
+        self.themeObserver = NotificationCenter.default.addObserver(forName: Notification.Name.didChangeTheme,
+                                                               object: nil,
+                                                               queue: .main) { [weak self] _ in
+            if let strongSelf = self {
+                strongSelf.updateButtonTokens()
```
harrieshin, line 47: so notifcation can be posted to all windows, but updatebuttontokens will figure/filter out if the theme changes for that window is applicable for this instance of the class?

ios/FluentUI/Core/ControlHostingContainer.swift
```diff
@@ -41,6 +41,12 @@ public class ControlHostingContainer: NSObject {
         self.updateRootView()
     }
 
+    deinit {
+        if let themeObserver = self.themeObserver {
+            NotificationCenter.default.removeObserver(themeObserver)
```
harrieshin, line 46: i thought we don't need to do any removeobserver these days. do we? https://developer.apple.com/documentation/foundation/nsnotificationcenter/1413994-removeobserver

harrieshin (approved):

ios/FluentUI/Core/ControlHostingContainer.swift
mischreiber: Unfortunately that's only the case if you use `addObserver(_:selector:name:object:)`. If you observe via `addObserver(forName:object:queue:using:)` then you do still need to deregister once you're done observing.

ios/FluentUI/Command Bar/CommandBar.swift
mischreiber: Good point. At the moment no, this is pretty brute-force logic, and will apply to the entire app. But that's a good optimization: the theme posting could/should include the window, and then we can have logic that decides whether it applies to us or not.

ios/FluentUI/Core/ControlHostingContainer.swift
mischreiber: That said, I'm adding a little more logic in the notification handler, so I'm switching to `addObserver(_:selector:name:object:)` in both these cases.

mischreiber:
> how much did you save here? :)

31,768,161 -> 31,735,921, so ~31 kB. Not bad!

PR #834, opened 2021-12-16
Convert `ColorValue` from a pure `UInt32` to a wrapper struct

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

At the start of the `vnext-tokens` effort, we decided to keep our internal `ColorValue` as a pure `UInt32`. This was fine when we were only storing 24 bits of information (8 bits each of r, g, and b), but then alpha came in and made everything trickier. ARGB? RGBA? Fractional definitions?

Well, it was nice while it lasted.

Now `ColorValue` is a simple wrapper around a `UInt32`, with two easy initializers: one that takes an old 24-bit hex value (e.g. 0x0078D4 for primary communication blue), and one that takes basic floating-point r, g, b, and a values a la `UIColor`.

### Verification

Ran through the New Token System™ controls to check for regressions. Verified light and dark mode, default and override themes. All is well!

Binary impact is minor: roughly 5 kB total. Not ideal, but necessary to ensure API clarity going forward.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/834)

### DIFF
--- a/ios/FluentUI/Core/Theme/Tokens/ColorSet.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/ColorSet.swift
@@ -7,25 +7,45 @@ import SwiftUI
 
 /// A platform-agnostic representation of a 32-bit RGBA color value.
 public struct ColorValue {
-    var r: UInt8 { UInt8(hexValue & 0xFF000000) }
-    var g: UInt8 { UInt8(hexValue & 0x00FF0000) }
-    var b: UInt8 { UInt8(hexValue & 0x0000FF00) }
-    var a: UInt8 { UInt8(hexValue & 0x000000FF) }
+    var r: CGFloat { CGFloat(hexValue & 0xFF000000) / 255.0 }
+    var g: CGFloat { CGFloat(hexValue & 0x00FF0000) / 255.0 }
+    var b: CGFloat { CGFloat(hexValue & 0x0000FF00) / 255.0 }
+    var a: CGFloat { CGFloat(hexValue & 0x000000FF) / 255.0 }
 
-    /// Creates a color value instance
-    public init(r: UInt8,
-                g: UInt8,
-                b: UInt8,
-                a: UInt8) {
-        hexValue = UInt32((r << 24) | (g << 16) | (b << 8) | a)
-    }
-
-    /// C
+    /// Creates a color value instance with the specified three-channel, 8-bit-per-channel color value, usually in hex.
+    ///
+    /// For example: `0xFF0000` represents red, `0x00FF00` green, and `0x0000FF` blue.
+    /// There is no way to specify an alpha channel via this initializer. For that, use `init(r:g:b:a)` instead.
+    ///
+    /// - Parameter hexValue: The color value to store, in 24-bit (three-channel, 8-bit) RGB.
+    ///
+    /// - Returns: A color object that stores the provided color information.
     public init(_ hexValue: UInt32) {
         self.hexValue = hexValue << 8 | 0xFF
     }
 
-    // Value is stored as RGBA.
+    /// Creates a color value instance with the specified channel values.
+    ///
+    /// Parameters work just like `UIColor`, `NSColor`, or `SwiftUI.Color`, and should all be in the range of `0.0 ≤ value ≤ 1.0`.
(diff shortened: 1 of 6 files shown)


harrieshin:
thank you for making this change. food for thought: can we unit test this?

ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -114,12 +113,9 @@ extension UIColor {
         var redValue: CGFloat = 0.0
         var greenValue: CGFloat = 0.0
         var blueValue: CGFloat = 0.0
-        if self.getRed(&redValue, green: &greenValue, blue: &blueValue, alpha: nil) {
-            // Ensure that all channels fall within 0x0 and 0xFF
-            let red = min(UInt32(redValue * 255.0), 0xFF)
-            let green = min(UInt32(greenValue * 255.0), 0xFF)
-            let blue = min(UInt32(blueValue * 255.0), 0xFF)
-            let colorValue: ColorValue = (red << 16) + (green << 8) + blue
+        var alphaValue: CGFloat = 0.0
```
harrieshin, line 116: default ain't1 ?

harrieshin (approved):

mischreiber:
> thank you for making this change. food for thought: can we unit test this?

yesssss I was just thinking about how this would be perfect for UTs (honestly most of my token work would be, but this in particular). Incoming!

ios/FluentUI/Extensions/UIColor+Extensions.swift
mischreiber: Shouldn't matter. I only need to initialize them to _something_ so I can pass the reference into `UIColor.getRed(_:green:blue:alpha:)`, and if that method fails then I don't use the values at all.

Maybe it's my C days showing, but 0 just seems more natural to initialize with than 1 🤔

harrieshin: i don't know. i feel like yeah initilizing by false is natural in c so 0. but then., if alpha is 0 then... nothing is showing in the ui world... soo....

mischreiber: Ehh true. Fiine I fix it.

ios/FluentUI/Core/Theme/Tokens/ColorSet.swift
```diff
@@ -5,10 +5,49 @@
...
+public struct ColorValue {
+    var r: CGFloat { CGFloat((hexValue & 0xFF000000) >> 24) / 255.0 }
+    var g: CGFloat { CGFloat((hexValue & 0x00FF0000) >> 16) / 255.0 }
+    var b: CGFloat { CGFloat((hexValue & 0x0000FF00) >> 8) / 255.0 }
+    var a: CGFloat { CGFloat(hexValue & 0x000000FF) / 255.0 }
+
+    /// Creates a color value instance with the specified three-channel, 8-bit-per-channel color value, usually in hex.
+    ///
+    /// For example: `0xFF0000` represents red, `0x00FF00` green, and `0x0000FF` blue.
+    /// There is no way to specify an alpha channel via this initializer. For that, use `init(r:g:b:a)` instead.
```
rdeassis, line 18: Does this silently accepts an unintended different color if the alpha is specified?
Is there a way we can add a validation that this fails if the value added specifies an alpha value?

rdeassis (approved):

ios/FluentUI/Core/Theme/Tokens/ColorSet.swift
mischreiber: Hmm, not sure what you mean. Can you provide an example of a value you'd be concerned of?

My thought is that if you specify as ARGB, the A gets bit-shifted off the end and ignored. And if you specify as RGBA, well then yeah it'll break, but that can't really be helped.

rdeassis: My comment was based on the assumption of the RGBA (probably with the order of the parameters in the back of my mind).
I agree with your thought below:
`My thought is that if you specify as ARGB, the A gets bit-shifted off the end and ignored.`

PR #835, opened 2021-12-16
Fixing Toggle issue in PillButtonBar Demo

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding an NSLayoutConstraint to each toggle button to prevent them from being pushed offscreen or overlapping when next to long blocks of text (especially when font size is increased on a small screen).

### Verification

Manually tested using the demo app in various font sizes, RTL/LTR languages, different size screens, etc.

| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/835)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PillButtonBarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PillButtonBarDemoController.swift
@@ -22,9 +22,10 @@ class PillButtonBarDemoController: DemoController {
         let disableOnBrandSwitchView = UISwitch()
         disableOnBrandSwitchView.isOn = true
         disableOnBrandSwitchView.addTarget(self, action: #selector(toggleOnBrandPills(switchView:)), for: .valueChanged)
+        NSLayoutConstraint.activate([disableOnBrandSwitchView.widthAnchor.constraint(equalToConstant: disableOnBrandSwitchView.frame.width)])
 
         container.addArrangedSubview(createLabelWithText("onBrand"))
-        addRow(items: [createLabelWithText("Enable/Disable pills in onBrand Pill Bar"), disableOnBrandSwitchView], itemSpacing: 20, centerItems: true)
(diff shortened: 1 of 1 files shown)


harrieshin:
my gut feeling says this might not be the right fix. UISwitch is one of those thigns that has very clear intrinsiccontentsize. if your width anchor fixed this issue, i feel like the problem is in the demo code's uistackview setting.

ios/FluentUI.Demo/FluentUI.Demo/Demos/PillButtonBarDemoController.swift
```diff
@@ -44,6 +46,7 @@ class PillButtonBarDemoController: DemoController {
         let disablePrimarySwitchView = UISwitch()
         disablePrimarySwitchView.isOn = true
         disablePrimarySwitchView.addTarget(self, action: #selector(togglePrimaryPills(switchView:)), for: .valueChanged)
+        NSLayoutConstraint.activate([disablePrimarySwitchView.widthAnchor.constraint(equalToConstant: disablePrimarySwitchView.frame.width)])
 
         container.addArrangedSubview(createLabelWithText("Primary"))
         addRow(items: [createLabelWithText("Enable/Disable pills in Primary Pill bar"), disablePrimarySwitchView], itemSpacing: 20, centerItems: true)
```
rdeassis, line 52: I agree with @harrieshin's comment that disabling the UISwitch constraints might not be the right fix here.
The container stack view is expanding to the point that it's not respecting the correct margins: 

Screen Shot 2022-01-04 at 9 04 24 AM

I think that the stackview should ensure that it stays within the right margins and the label view should be flexible in its width while truncating the text if needed.

rdeassis:
Closing this PR as the issue has been fixed in commit c14fcca9d4dec2c48a7f5d3277dee9c231c90137 (#880).
@huwilkes @edjamesmsft @harrieshin FYI.

PR #838, opened 2021-12-18
[vNext Tokens] Delete `DynamicColor`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

I originally created `DynamicColor` as part of #724 to provide a wrapper around `Color` that could respond to the dynamic values in a `ColorSet`. Unfortunately, it added more complexity than I really want, and I came up with a better idea anyway.

The problem is that `Color` has no public API for accepting a set of different color values for e.g. light vs dark mode or HC, and an extension can't monitor the necessary environment variables (since it can't have additional storage). Turns out the simplest solution is to just pass through `UIColor`! Apple's got some secret sauce behind the scenes to make sure a dynamic `UIColor` is always properly represented, so let's just leverage their code.

Oh, and did I mention that this change saves us almost **70 kB**?? 🤯

### Verification

Checked `CardNudge` before and after in light, dark, and "green" modes to make sure there are no regressions.

| 31,735,921 bytes | 31,666,929 bytes |

Final improvement: 68,992 bytes, or **67 kB**

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/838)


ios/FluentUI/Core/Theme/Tokens/ColorSet.swift
```diff
@@ -162,5 +162,23 @@ public struct ColorSet {
...
+    /// - Parameter colorSet: The set of color values that may be applied based on the current context.
+    init(colorSet: ColorSet) {
+        self.init(UIColor(colorSet: colorSet))
+    }
+
+    /// Creates a Color from a `ColorValue` instance.
+    ///
+    /// - Parameter colorValue: Color value to use to initialize this color.
+    init(colorValue: ColorValue) {
+        self.init(UIColor(colorValue: colorValue))
```
mischreiber, line 182: The two UIColor extensions, `init(colorSet:)` and `init(colorValue:)` are under heavy use elsewhere in the code, and are very well tested.

harrieshin:
what about "isElevated" stuff?

harrieshin:
> what about "isElevated" stuff?

I was being dumb. it was already there

harrieshin:
does color extension belong in colorset.swift?

harrieshin (approved):

Saadnajmi:
I will look at this on Monday, but I wonder if this is something that can be reused for DynamicColor in react native, which does a similar thing...

mischreiber:
> does color extension belong in colorset.swift?

Good question. I was leaning towards keeping it here to avoid having a giant catch-all `Color+Whatever.swift` extension bucket. Would you rather
1. A single `Color+Extension.swift` for all Fluent extensions
2. A specialized `Color+ColorSet.swift` just for these two initializers
3. Leaving it as-is

I'm not picky 😬

mischreiber:
> I will look at this on Monday, but I wonder if this is something that can be reused for DynamicColor in react native, which does a similar thing...

Great question! I'll take a peek there and see if there's any obvious opportunity for simplification.

I'm also down with renaming `ColorSet` to `DynamicColor` if that aligns with FURN's precedent?

Saadnajmi:
> > I will look at this on Monday, but I wonder if this is something that can be reused for DynamicColor in react native, which does a similar thing...
> 
> Great question! I'll take a peek there and see if there's any obvious opportunity for simplification.
> 
> I'm also down with renaming `ColorSet` to `DynamicColor` if that aligns with FURN's precedent?

I'm not sure if there's a FURN precedent in this case. FURN's precedent is to have separate themes for light and dark.

The React Native class I was thinking of is called RCTDynamicColor(.h/.m), and is both in our fork and Meta's repo, if you're curious.

PR #840, opened 2021-12-21
[fix] Shy Header is not shown correctly

AnYuan:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

When users scroll very fast, then stop the scroll view. velocity.y may be still greater than 0 which will cause the shy header show/hidden unexpectedly. in my understanding, we should depend on currentProgress to tell if we should expand or not.

before:
https://user-images.githubusercontent.com/1324616/146864893-546a2ca5-113b-4e18-ba0a-5c5fb0db8368.MP4

### Verification

Tested in Demo app

### Pull request checklist

This PR has considered:
- [X] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/840)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -442,16 +442,8 @@ class ShyHeaderController: UIViewController {
         let momentum: CGFloat = abs(velocity.y)
         let springVelocity = min(momentum, 3.0) // 3.0 arbitrarily decided through manual testing
 
-        /// if the user's finger is not moving, we expand/collapse based on current progress
-        /// otherwise, we use velocity to determine which direction to move the header
-        let expanding: Bool
-        if velocity.y == 0.0 {
-            let currentProgress = shyHeaderView.exposure.progress
-            expanding = currentProgress >= 0.5
-        } else {
-            expanding = velocity.y >= 0.0
-        }
-
+        /// we expand/collapse based on current progress
+        let expanding = shyHeaderView.exposure.progress >= 0.5
         let progress: CGFloat = expanding ? 1.0 : 0.0
 
         UIView.animate(withDuration: ShyHeaderController.Constants.headerShowHideAnimationDuration,


rdeassis (commented):
@AnYuan thanks for contributing to FluentUI!
The video you linked looks like the correct behavior to me.
It looks like the search bar is collapsing correctly as you scroll up and coming back normally as you scroll down.

Can you provide more details about the issue you're seeing?

Thank you!

AnYuan:
Hi @rdeassis! 

Thanks for your reply!

In the video, I scroll up the table view quickly, then stop the table view. in that case, shy header may be shown.

When I tried to debug this one, I found out that ```velocity.y``` is still greater than 0 which makes shy header show up. It is not a `always` bug.

AnYuan:
https://user-images.githubusercontent.com/1324616/148314236-d766a4b5-b231-47b8-87c3-013420c5f715.mov

in this video, when I scroll up, the shy header is still shown

sophialee0416:
> RPReplay_Final1641432865.mov 
> 
> in this video, when I scroll up, the shy header is still shown

Could you provide a few more details about how you are reproing this so we can try replicating it? (ex: device model, iOS version, specific demo controller, etc). Thanks!

rdeassis:
> @sophialee0416 sure.
> 
> iOS version: 15.2 device model: iPhone 12 demo controller: NavigationController -> Show "system" with collapsible search bar in
> 
> Thanks

Hi @AnYuan,

We're having a hard time reproducing this issue.
Are you able to consistently reproduce it?
If so, do you mind posting a video of the search bar "jump issue" repeatedly happening.

Also, are you using the latest version of the demo app? (**1.4.6** which uses the **FluentUI 0.4.6**)

I appreciate your patience here. We want to make sure this change is really necessary and not likely to potentially cause a regression in apps that already rely on it.

Thank you!

rdeassis:
I'm going to close this PR as it doesn't seem actionable right now.
@AnYuan feel free to reach out and reopen the PR if we get to a point that we're all confident to move forward with the fix.

Thanks!

PR #846, opened 2022-01-02
[vNext Tokens] Use new token system for `FluentButton`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This converts `FluentButton` to use the new token system. It's one of the most complex controls in the library (competitive with `Avatar`), which means updating it requires the most churn, but also has the largest binary size improvement (spoiler: we're saving well over a half megabyte!).

Quick rundown of notable changes:
* Deleted generated `MSFButton` tokens.
* Revamped `ButtonDemoController` to properly store the `MSFButton` objects being presented.

### Verification

In addition to the considerations below, the following were tested:
* Pixel-perfect alignment, with the exception of known bug fix for inter-element padding.
* Color swapping (blue to green, light to dark).
* Proper large content viewer for floating buttons.

| 31,767,518 bytes (30.3 MB) | 31,143,070 bytes (29.7 MB) |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/846)


ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -46,25 +47,28 @@ class ButtonDemoController: UITableViewController {
         case .textAndIcon,
              .textOnly,
              .iconOnly:
-            let cell = UITableViewCell()
+            let cell = tableView.dequeueReusableCell(withIdentifier: ButtonDemoController.cellReuseIdentifier, for: indexPath)
```
huwilkes, line 50: Should the name be more specific than just "cell", since you're dequeueing with a specific identifier?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -46,25 +47,28 @@ class ButtonDemoController: UITableViewController {
...
+            let subviews = cell.contentView.subviews
+            subviews.forEach { subview in
+                subview.removeFromSuperview()
```
huwilkes, line 53: Do we do this in the other demo controllers? If no, should? What's the point of removing all the demo button cell's subviews?

ios/FluentUI/Vnext/Button/ButtonTokens.swift
```diff
@@ -30,145 +30,291 @@ import UIKit
...
 
-    override func updateForCurrentTheme() {
-        let currentTheme = theme
-        var appearanceProxy: AppearanceProxyType
+    lazy var borderRadius: CGFloat = {
+        switch size {
+        case .small:
+            return globalTokens.borderRadius[.large]
+        case .medium:
+            return globalTokens.borderRadius[.large]
```
huwilkes, line 50: If they're the same, we should be able do something like
`case .small,
     .medium:`

ios/FluentUI/Vnext/Button/ButtonTokens.swift
```diff
@@ -30,145 +30,291 @@ import UIKit
...
+    lazy var borderSize: CGFloat = {
         switch style {
-        case .primary:
-            appearanceProxy = currentTheme.MSFPrimaryButtonTokens
         case .secondary:
-            appearanceProxy = currentTheme.MSFSecondaryButtonTokens
-        case .ghost:
-            appearanceProxy = currentTheme.MSFGhostButtonTokens
+            return globalTokens.borderSize[.thin]
+        default:
```
huwilkes, line 60: I know in devmain (or at least the FRE code) we don't like defaults, but I'm not sure how we feel about them here. Probably similarly, so it might be better to just list out all the options?

ios/FluentUI/Vnext/Button/ButtonTokens.swift
```diff
@@ -30,145 +30,291 @@ import UIKit
...
+                selected: aliasTokens.backgroundColors[.neutral1],
+                disabled: aliasTokens.backgroundColors[.neutral1]
+            )
+        default:
+            return .init(
+                rest: DynamicColor(light: ColorValue(r: 0, g: 0, b: 0, a: 0)),
+                hover: DynamicColor(light: ColorValue(r: 0, g: 0, b: 0, a: 0)),
+                pressed: DynamicColor(light: ColorValue(r: 0, g: 0, b: 0, a: 0)),
+                selected: DynamicColor(light: ColorValue(r: 0, g: 0, b: 0, a: 0)),
+                disabled: DynamicColor(light: ColorValue(r: 0, g: 0, b: 0, a: 0))
```
huwilkes, line 275: Do we not have a clear option?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -118,33 +103,33 @@ class MSFButtonStateImpl: NSObject, ObservableObject, MSFButtonState {
...
+        let titleColor: DynamicColor
+        let borderColor: DynamicColor
+        let backgroundColor: DynamicColor
         if isDisabled {
-            iconColor = tokens.disabledIconColor
-            titleColor = tokens.disabledTitleColor
-            borderColor = tokens.disabledBorderColor
-            backgroundColor = tokens.disabledBackgroundColor
+            iconColor = tokens.iconColor.disabled
+            titleColor = tokens.textColor.disabled
```
huwilkes, line 120: Should we change titleColor to textColor? It's the only one that doesn't match its token.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
mischreiber: The difference is that before this change, we were never cacheing cells for reuse, which is a pretty inefficient way to use `UITableView`. Now that I switched it to a reuse mechanism, I'm following Apple's recommendation:

> The table view's delegate in tableView(_:cellForRowAt:) should _always_ reset all content when reusing a cell.

This is not a problem for `[MSF]TableViewCell`, which manages its own subviews during its `setup(title:subtitle:footer:customView:customAccessoryView:accessoryType:)` method.

ios/FluentUI/Vnext/Button/ButtonTokens.swift
mischreiber: Not yet, but you're right that this is repeated too many times. I'll add one.

ios/FluentUI/Vnext/Button/ButtonTokens.swift
mischreiber: Interesting question. In this case, I was looking at the model of the old `MSFButtonTokens` YAML file, where the basic structure is

```base tokens
```    |
```    |----> override 1
```    |----> override 2
```    |----> override 3

So there really is a "default" set that is automatically inherited by all control types, unless explicitly overridden, and using the `default` case helps to explicitly distinguish that.

That said, `Button` is unique among Fluent controls, as far as I can tell, in that no control style actually uses the "base" values as-is, which makes parsing out what value to use where a bit challenging. Perhaps that's also something worth investigating at some point?

File: ios/FluentUI/Vnext/Button/ButtonTokens.swift (hunk 3 above)
Comment (mischreiber) [reply ios/FluentUI/Vnext/Button/ButtonTokens.swift:50]: I went back and forth on this one, but I think I'd like to keep them separate still, for a couple of reasons:

* It's easier to skim a large file like this and see, for example, all `.medium` sized values.
* The values aren't intrinsically linked; it's often just coincidence that they happen to match. As such, keeping them as separate instances seems nicer.
* There's a long-term north star goal to have this code generated, and keeping each entry separate will mean simpler generator code.

But that's just an opinion! I'm happy to reconsider if there are cases against separation that I haven't considered.

File: ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift (hunk 1 above)
Comment (mischreiber) [reply ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift:50]: Ehhhhhh I think it's okay, because (A) it's still just a `UITableViewCell` being dequeued, and (B) there are no other cells in this context, so there isn't really any room for ambiguity.

File: ios/FluentUI/Vnext/Button/ButtonTokens.swift (hunk 4 above)
Comment (huwilkes) [reply ios/FluentUI/Vnext/Button/ButtonTokens.swift:60]: Seems like it might be worth revisiting some other time to separate out the defaults, for many of the same reasons you gave in the comment about combining the .small and .medium above.

**Review by huwilkes — APPROVED** [review]

**Review by harrieshin — APPROVED** [review]

File: ios/FluentUI/Vnext/Button/Button.swift [hunk 7]
```diff
@@ -53,28 +54,27 @@ public struct FluentButton: View {
         state.text = text
         state.image = image
         self.state = state
-        self.tokens = state.tokens
+
+        // We want separate lookup keys for each permutation of `style` and `size`.
+        self.tokenKey = "\(type(of: self))_\(style.rawValue)_\(size.rawValue)"
```
rdeassis, line 59: Is there going to be a pattern for the controls going forward to identify the token keys?
I feel like we should have a standard for these.

ios/FluentUI/Vnext/Button/ButtonTokens.swift
rdeassis: My personal opinion is that having the combined cases makes the code more readable.
I do agree that it creates complexity in the code generator in the future. And I believe we can reconsider if that extra complexity is very costly in the future.
I suspect that it might be a tiny bit less code unless the compiler optimizes it.

Thoughts?

ios/FluentUI/Vnext/Button/ButtonTokens.swift
rdeassis: I believe Swiftlint has a check on switch statements that require them to be exaustive.
It's definitely a good practice in devmain code that no default is used to avoid silent bugs when new items in the enumeration are introduced.
I think we should not introduce defaults in the code and work on removing the remaining ones throughout.
Quick note: there are some enumerations named "default", which in this case is fine although a bit confusing:

`case .default:`

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -8,91 +8,38 @@ import SwiftUI
...
-        buttonView = FluentButton(style: style,
-                                  size: size) { [weak self] in
+        let buttonView = FluentButton(style: style,
+                                      size: size,
+                                      action: {})
+        state = buttonView.state
+        super.init(AnyView(buttonView))
+
+        // After initialization, set the new action to refer to our own.
+        buttonView.state.action = { [weak self] in
```
rdeassis, line 36: Not sure I understand why we're setting the action in the state later instead of in line 31.

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -8,91 +8,38 @@ import SwiftUI
...
-    ///   - action: Closure that handles the button tap event.
-    @objc public convenience init(style: MSFButtonStyle = .secondary,
-                                  size: MSFButtonSize = .large,
-                                  action: ((_ sender: MSFButton) -> Void)?) {
-        self.init(style: style,
-                  size: size,
-                  theme: nil,
-                  action: action)
-    }
+    @objc public let state: MSFButtonState
```
rdeassis, line 18: What's the reason on not making this a computed property?
Although the state object shouldn't be changed through the lifecycle of the control, it might be error prone to rely on updating this property if it is recreated.

ios/FluentUI/Core/Theme/Tokens/DynamicColor.swift
```diff
@@ -7,10 +7,7 @@ import SwiftUI
 
 /// A platform-agnostic representation of a 32-bit RGBA color value.
 public struct ColorValue {
-    var r: CGFloat { CGFloat((hexValue & 0xFF000000) >> 24) / 255.0 }
-    var g: CGFloat { CGFloat((hexValue & 0x00FF0000) >> 16) / 255.0 }
-    var b: CGFloat { CGFloat((hexValue & 0x0000FF00) >> 8) / 255.0 }
-    var a: CGFloat { CGFloat(hexValue & 0x000000FF) / 255.0 }
+    public static let clear: ColorValue = .init(r: 0.0, g: 0.0, b: 0.0, a: 0.0)
```
rdeassis, line 10: Are these changes part of the Button changes? Seems like they belong in a separate PR?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -68,21 +68,18 @@ class ThemingDemoController: DemoController {
         let overridingTheme = CustomStyle()
         let customThemeButtonPrimary = MSFButton(style: .primary,
                                                       size: .medium,
-                                                      theme: overridingTheme,
```
rdeassis, line 71: Can you remind me how you're overriding the tokens directly in the control level?
Can you also provide videos or screenshots in the validation so we know the visual results of those changes?

ios/FluentUI/Vnext/Button/ButtonTokens.swift
mischreiber: Fair enough! I don't feel especially strongly, so I'm happy to fold it together.

huwilkes: If/when we try to generate the tokens, I think we could probably group the cases together in some sort of dictionary that holds a list of the cases as the value and the return value as the key? That doesn't seem super complicated, and should let us generate the tokens grouped like this.

ios/FluentUI/Vnext/Button/Button.swift
mischreiber: Agreed, this is getting a little out of hand. I'm thinking instead of returning a string I'll have this API return an array of `Any` values, and then concat them together in the caller.

ios/FluentUI/Vnext/Button/MSFButton.swift
mischreiber: Ah yeah, sadly it's a compiler thing. We create `buttonView` before calling `super.init(_:)`, since that call requires a fully-formed `View`. But the action requires a reference to `self`, which we're not allowed to make until after calling `super.init(_:)`.

ios/FluentUI/Core/Theme/Tokens/DynamicColor.swift
mischreiber: It's just code sanitization (moving `public` above `internal`), so I figured I could roll it into this PR. Would you prefer that I exclude this change and spin up a separate PR?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
mischreiber: Control-level overrides are done by setting the `overrideTokens` property directly on the control in question. See here for an example in `ThemingDemoController`:

https://github.com/microsoft/fluentui-apple/blob/b872a8025116ac22f80d4a5dc77fb79603752714/ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift#L108

ios/FluentUI/Vnext/Button/MSFButton.swift
mischreiber: Great question! This is part of a separation in the relationship between `state` and `tokens`, since the latter (`tokens`) is no longer stored, but generated itself. It's not really clear how `state` could maintain a reference to that generation logic, so now it simply has no concept of design tokens.

Additionally, those values have now become read-only after initialization, taking a page from UIKit controls such as `UIButton`, which are initialized with an immutable style. This won't affect our SwiftUI control, which of course recreates itself in full whenever its `style` or `size` are detected to have changed.

mischreiber:
(Apologies if you saw an unrelated change merge into this branch; it's been removed!)

ios/FluentUI/Vnext/Button/MSFButton.swift
rdeassis: As we've been discussing offline, this seems problematic because the tokens need to vary if "style" and "size" enum properties on the state object change.
Looking forward for the upcoming PR to address that.

mischreiber: PR is out now! 

- https://github.com/microsoft/fluentui-apple/pull/852

PR #850, opened 2022-01-05
Updating toolBar translucency in UINavigationController

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

One of the updates brought in iOS 15 included an updated UINavigationController, which causes all navigation bars to become opaque and adopt the standard appearance instead. Setting the translucency property to true fixes this. The visual change is only noticeable in light mode.

### Verification

| Simulator Screen Shot - iPhone 11 Pro - 2022-01-05 at 12 28 24 | Simulator Screen Shot - iPhone 11 Pro - 2022-01-05 at 12 20 44 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/850)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/DemoListViewController.swift
@@ -23,6 +23,7 @@ class DemoListViewController: DemoTableViewController {
         let navigationController = UINavigationController(rootViewController: demoListViewController)
         navigationController.navigationBar.isTranslucent = true
         navigationController.navigationBar.prefersLargeTitles = true
+        navigationController.toolbar.isTranslucent = true
 
         window.rootViewController = navigationController
         window.makeKeyAndVisible()


huwilkes (approved):
Nit: If setting .isTranslucent to true fixes the problem, did the change in iOS 15 make the navigation bars opaque, rather than transparent?

sophialee0416:
> Nit: If setting .isTranslucent to true fixes the problem, did the change in iOS 15 make the navigation bars opaque, rather than transparent?

Oh true. Let me fix that wording, thank you for the catch :P

mischreiber:
@sophialee0416 does this also fix the issue where going back to the demo list after showing a toolbar results in a big blank spot at the bottom of the screen?

sophialee0416:
> @sophialee0416 does this also fix the issue where going back to the demo list after showing a toolbar results in a big blank spot at the bottom of the screen?

It does not, just the color... I'll spin up another PR for that one too.

PR #851, opened 2022-01-05
[vnext-tokens] Converting Activity Indicator to use new tokens.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Converting the Activity Indicator to use the new design tokens structure.
The Alias Tokens needed to have the addition of the BackgroundColor neutral4 variation.

### Verification

- Built and ran the demo app to validate basic use scenarios.
- Pod lib lint on MicorosoftFluentUIVnext.podspec.

https://user-images.githubusercontent.com/68076145/148306390-ec8ebea3-66e2-4936-a934-f8b09e3efff5.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/851)


ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
```diff
@@ -173,25 +132,24 @@ public struct ActivityIndicator: View {
...
-    var size: MSFActivityIndicatorSize {
-        get {
-            return tokens.size
-        }
-
-        set {
-            tokens.size = newValue
-        }
-    }
+    // TODO: Setting this property needs to cause the tokens to update
```
rdeassis, line 146: We need to track the issue of updating the design tokens when the size property changes in the state object. The same applies for other controls that use either "style" or "size" enumerations that affect which tokens are used in the control.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
```diff
@@ -13,74 +13,67 @@ import UIKit
...
 
-        defaultColor = appearanceProxy.defaultColor
+    lazy var defaultColor: DynamicColor = aliasTokens.foregroundColors[.neutral4]
 
+    lazy var side: CGFloat = {
         switch size {
         case .xSmall:
-            activityIndicatorSize = appearanceProxy.size.xSmall
-            thickness = appearanceProxy.thickness.xSmall
+            return 12
```
huwilkes, line 55: Seems odd that none of the floats reference global tokens?

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
mischreiber: +1

Got a local branch with this working; hoping to have a PR out soon.

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
```diff
@@ -72,25 +25,29 @@ import UIKit
...
-        self.tokens = state.tokens
+
+        // Creates separate lookup keys for all different sizes.
+        self.tokenKey = "\(type(of: self))_\(size.rawValue)"
     }
 
     public var body: some View {
-        let size = tokens.activityIndicatorSize
-        let color = Color(state.color ?? tokens.defaultColor)
+        let side = tokens.side
```
huwilkes, line 43: nit: side seems like an odd name to me. I get that we're trying to not have too many things named size, but I think I'd prefer length or something.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
mischreiber: Agreed. Most of these values are available in GlobalTokens using `borderSize` for thickness and `iconSize` for this one). That said, this control hasn't been formally specced in the vNext system yet, so there's not a perfect 1-1 mapping.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
```diff
@@ -13,74 +13,67 @@ import UIKit
...
+public class ActivityIndicatorTokens: ControlTokens {
+    /// Creates an instance of `ActivityIndicatorTokens` with optional token value overrides.
+    /// - Parameters:
+    ///   - size: MSFActivityIndicatorSize enumeration value that will define pre-defined values for side and thickness.
+    ///   - defaultColor: The default color of the Activity Indicator.
+    ///   - side: The value for the side of the square frame of an Activity Indicator.
+    ///   - thickness: The value for the thickness of the ActivityIndicator ring.
+    public init(size: MSFActivityIndicatorSize,
+                defaultColor: DynamicColor? = nil,
+                side: CGFloat? = nil,
```
mischreiber, line 28: Took me a second to determine what `side` meant. I know we can't reuse `size` so maybe something related to `dimension` would be more descriptive?

Saadnajmi: diameter?

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
rdeassis: The context behind this is that there are no tokens defining frame size or a standard created for control sizes so far. I'm not sure if the design team will standardize in that level of detail as this is not an icon per se.
This is not different from the current definition on the old YAML file as of now, so seems out of the scope of this change.

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
rdeassis: The reasoning behind the name is that for rectangular frames we have different values for height and width. The term in geometry for the height and width of a square is "side" which is why I chose to use it. It implies that the shape of the frame is a square of equal sides.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
rdeassis: Just copying the same response I posted in the other comment:

`The reasoning behind the name is that for rectangular frames we have different values for height and width. The term in geometry for the height and width of a square is "side" which is why I chose to use it. It implies that the shape of the frame is a square of equal sides.`

mischreiber (approved):

PR #852, opened 2022-01-06
[vNext Tokens] Update logic for token fetching to allow for style/size changes

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This introduces one big change, and several small tweaks, to the Fluent 2.0 token system.

First, the big change: `FluentButton` and `CardNudge` will once again allow you to modify their `style` (and for `FluentButton`, `size`) properties via our UIKit wrappers at runtime. This is made possible by removing the `tokenKeyComponents` property and replacing it with a push-based system to cache tokens on a control. 

This cache will be reset using a pair of new modifiers: `resolveTokens(_:)` and `resolveTokenModifier(_:value:)`. 

The former will watch for `FluentTheme` and `overrideTokens` changes, while the latter can optionally monitor _any_ variable. For `Button`, this means watching the `style` and `size` properties on the control's `state`. In either case, a new set of tokens will be generated by the new `resolvedTokens(for:fluentTheme:)` function, triggering a SwiftUI re-layout.

Additionally, some smaller changes I rolled in:
* Renamed the `customTokens(_:)` view modifier to `overrideTokens(_:)`, to better align with new nomenclature.
* Reversed the order of initializers for `CardNudgeTokens` and `CommandBarTokens` to match `ButtonTokens`, resulting in less copy and paste of property names.
* Removal of some unnecessary `@objc public` notations on properties of `MSFCardNudgeStateImpl`, left over from a much older experiment for that control.

### Verification

Here's a fun video showing the `CardNudge` controls toggle between `standard` and `outline` while visible, simply by changing the `state.style` property:

https://user-images.githubusercontent.com/4934719/148453077-f7a72188-b831-4df4-bcfa-09cdc7cff9d7.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/852)


ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -47,7 +47,7 @@ import SwiftUI
             return tokens
         } else {
             // Register the default tokens to make future lookups more efficient.
-            let tokens = control.state.defaultTokens
+            let tokens = control.defaultTokens
```
rdeassis, line 50: Have you tested the scenario where the tokens are overridden for a specific control type (line 47 above) and then had the style/size changed?

Also related to that: why are we registering a new token instance for every combination of style and size?
It seems like the overriding instance should take care of that.

mischreiber: Excellent catch. With this, control-type overrides would only apply for a given set of `tokenKeyComponents`.

A solution will tie into your second point: there's no reason that we can't just have a single registered token set per control, and inject whatever overrides are necessary to derive the right token values. In fact this code used to work like that once, but it's been a while since we drifted this (arguably more complex) direction. Let me see if I can steer us back.

rdeassis: I believe that if we have the tokens immutable, it probably does not make sense to have a property store their style/size enumerations since can be used only when initializing the properties. Right now they seem to be used only to generate the tokenKeyComponents.
If we have a single overriding type, we can just create it on demand and initialize the tokens based on the style/size params.

mischreiber: To be fair, that is effectively what we are doing already. The various token properties are initialized once, the first time they are accessed, as opposed to being generated. The current structure of the token structs is meant for readability: each property's initialization is listed alongside its declaration, rather than all existing as part of a monolithic `initializeVariables()` function. But you're right that such a method would allow for us to remove the style/size storage. Do you have a personal preference?

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
```diff
@@ -0,0 +1,58 @@
...
+import SwiftUI
+
+/// Determines which tokens to use, and ensures they refer to the correct `FluentTheme` instance.
+func resolvedTokens<T: TokenizedControlInternal>(for control: T, fluentTheme: FluentTheme) -> T.TokenType {
+    let tokens = control.state.overrideTokens ?? fluentTheme.tokens(for: control) ?? control.state.defaultTokens
+    tokens.fluentTheme = fluentTheme
+    return tokens
+}
+
+struct TokenResolver<T: TokenizedControlInternal>: ViewModifier {
```
rdeassis, line 15: nit: documentation comments (here and in the types below)

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
```diff
@@ -0,0 +1,58 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// Determines which tokens to use, and ensures they refer to the correct `FluentTheme` instance.
+func resolvedTokens<T: TokenizedControlInternal>(for control: T, fluentTheme: FluentTheme) -> T.TokenType {
+    let tokens = control.state.overrideTokens ?? fluentTheme.tokens(for: control) ?? control.state.defaultTokens
```
rdeassis, line 10: Just for confirmation: The sample of how to set a custom ControlToken for a specific theme (`fluentTheme.tokens(for: control)`) will come on a future PR?

ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -24,10 +24,8 @@ import SwiftUI
     /// - Parameters:
     ///   - tokens: A custom set of tokens.
     ///   - control: The control to use custom tokens on.
-    public func register<T: TokenizedControl>(tokens: T.TokenType, for control: T) {
-        tokens.globalTokens = globalTokens
-        tokens.aliasTokens = aliasTokens
-        controlTokens[tokenKey(control.tokenKeyComponents)] = tokens
+    public func register<T: TokenizedControl>(controlType: T.Type, tokens: @escaping (T) -> T.TokenType) {
```
rdeassis, line 27: This doesn't seem to be used anymore (probably related to the comment below on overriding control tokens per theme in a future PR).

I think it might be worth removing it and reintroducing it in the next PR when you get the control tokens customization per theme figured out.

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
```diff
@@ -17,54 +17,34 @@ import SwiftUI
...
-                minimumHeight: CGFloat? = nil,
-                outlineColor: DynamicColor? = nil,
-                outlineWidth: CGFloat? = nil,
-                subtitleTextColor: DynamicColor? = nil,
-                textColor: DynamicColor? = nil,
-                verticalPadding: CGFloat? = nil) {
-
-        self.style = style
-        super.init()
+    convenience public init(style: MSFCardNudgeStyle,
```
rdeassis, line 26: We still don't have a solution for overriding "style/size aware" values in this scenario, right?
Are you going to explore the closure approach we discussed?

rdeassis (approved):

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
mischreiber: Yep! That will be part of addressing the testability task.

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
mischreiber: I explored it a bit and while it was functional, it was also a bit unwieldy. I'll continue to explore, along with drawing up a binary size comparison for e.g. `Button`.

ios/FluentUI/Core/Theme/FluentTheme.swift
mischreiber: Interesting point! So this function was originally designed to be used exclusively by clients; I repurposed it at one point to cover internal scenarios as well, but that was never its intended goal.

That said, this API is hard to verify without the testing work I've got planned for this week or next. I've done light local verification, so I'm confident that it works, but yeah it's unfortunate that there's no proper tests yet.

Would you be comfortable with me leaving it for the moment, in order to keep my future testability PR cleaner?

ios/FluentUI/Card Nudge/CardNudgeTokens.swift
rdeassis: Sounds good! Can you create an ADO item to track this effort? I think it deserves its own ADO task as it may cause more changes throughout.

mischreiber: Done! ADO:5727871

rdeassis: Thanks! (Just removed the ADO number because of our policy of not publishing internal info in the public github.)

PR #855, opened 2022-01-07
Adopt frame based layout in BottomSheetController

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### Problem
**TL;DR: Sizing and positioning the sheet using Auto Layout was causing more problems than it was solving. Directly calculating the frame simplifies the code and sets us up for future planned additions to the bottom sheet.**

The bottom sheet will soon be used in a new experience in our apps which requires a non-rigid mode i.e. the sheet height will change as the user expands the sheet. In preparation for this change, we need more control over how the sheet is sized.

The sheet is currently sized and positioned using constraints, however it turns out this has several problems:
- the height is determined using a complicated set of UILayoutGuides, this is hard to read by new contributors and doesn't extend well to a flexible height mode
- the width sometimes has to fill parent, sometimes not, which adds more constraints
- vertical position is set using a constant,  so we don't benefit from any Auto Layout magic here
- vertical position also depends on the height, so we need to do a dance of waiting for the layout pass to complete and only updating the offset constraint after that, which can trigger another layout pass
- all animations are frame-based (this is a UIKit limitation), so when an animation is interrupted, we need to convert between constraint offsets and the intermediate frame origin
- the sheet sometimes behaves strangely while rotating the screen, it doesn't stay attached to the bottom edge when the animation is in progress
- many more bits that I can't think of ....

So we already do a lot of calculation and barely benefit from Auto Layout's declarative magic.
Manually calculating the frame for any given state is actually fairly simple and we already do most of it anyways. It has many benefits, like a single source of truth for panning / animating / static positioning, simpler code, and easy implementation of the future flexible mode.

#### Solution
- Add a single source of truth for sheet size and position at any given offset - `func sheetFrame(offset:)`.
- Add `headerContentHeight` property which was the last piece missing to let us directly calculate sheet height.
- Add an `.intermediate` sheet expansion state so we are explicit about being in between states.
- In our viewDidLayoutSubviews override, recalculate sheet size and position if we're not in `.intermediate` state. This is a very cheap constant time calculation so we can afford to do it on every layout pass.
- Remove tons of constraints and layout guides that were used for Auto Layout.
- Sheet offset is now just `bottomSheetView.frame.minY` which is an offset from top, so we need to flip a bunch of conditions from the previous constraint-based offset.
- Simplify multiple property change handlers to just call view.setNeedsLayout().

### Verification

- Sanity checks of all sheet features both in `BotomSheetController` and `BottomCommandingController`.
- Tests of animation interruptions, rotating while animating, panning while animating. 
- Interactive sheet hiding.

No visual or functional change, other than fixing a glitch on screen rotation.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/855)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -36,7 +36,7 @@ public protocol BottomSheetControllerDelegate: AnyObject {
     case expanded // Sheet is fully expanded
     case collapsed // Sheet is collapsed
     case hidden // Sheet is hidden (fully off-screen)
-    case intermediate // Sheet is in between states, only used during user interaction / animation
+    case intermediate // Sheet is between states, only used during user interaction / animation
 
     // Target alpha of related views like the sheet content or dimming view.
     var relatedViewAlpha: CGFloat { self == .expanded ? 1.0 : 0.0 }
@@ -291,7 +291,11 @@ public class BottomSheetController: UIViewController {
 
         view.addSubview(bottomSheetView)
         bottomSheetView.isHidden = isHidden
-        bottomSheetView.frame = sheetFrame(offset: offset(for: currentExpansionState))
+
+        // The non-zero frame here ensures the layout engine won't complain if it tries to calculate
+        // sheet content layout before view.bounds is set to a non-zero rect.
+        // We will set the sheet frame to a more meaningful rect after the first layout pass.
+        bottomSheetView.frame = CGRect(x: 0, y: 0, width: Constants.minSheetWidth, height: expandedSheetHeight)
 
         let overflowView = UIView()
         overflowView.translatesAutoresizingMaskIntoConstraints = false
@@ -393,7 +397,9 @@ public class BottomSheetController: UIViewController {
         return bottomSheetView
     }
 
-    public override func viewWillLayoutSubviews() {
+    public override func viewDidLayoutSubviews() {
+        // In the intermediate state a pan gesture or an animator temporarily owns the sheet frame updates,
+        // so to avoid interfering we won't update the frame here.
         if currentExpansionState != .intermediate {
             bottomSheetView.frame = sheetFrame(offset: offset(for: currentExpansionState))
             updateExpandedContentAlpha()
@@ -502,11 +508,13 @@ public class BottomSheetController: UIViewController {
         updateDimmingViewAlpha()
     }
 
+    // Source of truth for the sheet frame at a given offset from the top of the root view bounds.
+    // The output is only meaningful once view.bounds is non-zero i.e. a layout pass has occured.
     private func sheetFrame(offset: CGFloat) -> CGRect {
-        let availableWidth: CGFloat = view.frame.width
-        let sheetWidth = max(shouldAlwaysFillWidth ? availableWidth : min(Constants.maxSheetWidth, availableWidth), Constants.minSheetWidth)
+        let availableWidth: CGFloat = view.bounds.width
+        let sheetWidth = shouldAlwaysFillWidth ? availableWidth : min(Constants.maxSheetWidth, availableWidth)
 
-        return CGRect(origin: CGPoint(x: (view.frame.width - sheetWidth) / 2, y: offset),
+        return CGRect(origin: CGPoint(x: (view.bounds.width - sheetWidth) / 2, y: offset),
                       size: CGSize(width: sheetWidth, height: expandedSheetHeight))
     }
 
@@ -641,6 +649,9 @@ public class BottomSheetController: UIViewController {
     }
 
     // Vertical offset of bottomSheetView.origin for the given expansion state
+    //
+    // Note: Since .intermediate state doesn't have a well defined offset, this function will
+    // only return the current sheet offset in that case.
     private func offset(for expansionState: BottomSheetExpansionState) -> CGFloat {
         var offset: CGFloat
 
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -722,45 +715,42 @@ public class BottomSheetController: UIViewController {
...
-        constraint.priority = .defaultHigh // Lower than required so Auto Layout can enforce max sheet height
-        return constraint
-    }()
-
-    private lazy var sheetWidthConstraint: NSLayoutConstraint = bottomSheetView.widthAnchor.constraint(equalTo: view.widthAnchor)
-
-    private lazy var sheetMaxWidthConstraint: NSLayoutConstraint = bottomSheetView.widthAnchor.constraint(lessThanOrEqualToConstant: Constants.maxSheetWidth)
-
-    private lazy var maxSheetHeightLayoutGuide: UILayoutGuide = UILayoutGuide()
+    // Height of the sheet in its most expanded state
```
noaMSFT, line 718: Suggestion: 'fully expanded state' instead of 'most expanded state'

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -776,13 +766,13 @@ public class BottomSheetController: UIViewController {
 
         // Minimum padding from top when the sheet is fully expanded
         static let minimumTopExpandedPadding: CGFloat = 25.0
```
noaMSFT, line 768: Nit: for the sake of consistency - all CGFloat's should have .0 at the end like this one (25.0) or remove the .0 to be consistent with the rest of the file.

noaMSFT (approved):
Looks great! Thanks Lukas 👍

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -722,45 +715,42 @@ public class BottomSheetController: UIViewController {
...
-    }()
-
-    private lazy var sheetWidthConstraint: NSLayoutConstraint = bottomSheetView.widthAnchor.constraint(equalTo: view.widthAnchor)
-
-    private lazy var sheetMaxWidthConstraint: NSLayoutConstraint = bottomSheetView.widthAnchor.constraint(lessThanOrEqualToConstant: Constants.maxSheetWidth)
-
-    private lazy var maxSheetHeightLayoutGuide: UILayoutGuide = UILayoutGuide()
+    // Height of the sheet in its most expanded state
+    private var expandedSheetHeight: CGFloat {
+        var height: CGFloat = 0
```
rdeassis, line 720: I think this can also be a constant, right? (as it's only initialized in the if below and not changed after that)

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -626,28 +643,32 @@ public class BottomSheetController: UIViewController {
...
         translationAnimator.pauseAnimation()
         return translationAnimator
     }
 
+    // Vertical offset of bottomSheetView.origin for the given expansion state
+    //
+    // Note: Since .intermediate state doesn't have a well defined offset, this function will
+    // only return the current sheet offset in that case.
     private func offset(for expansionState: BottomSheetExpansionState) -> CGFloat {
         var offset: CGFloat
```
rdeassis, line 660: Change offset to let?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -495,25 +497,40 @@ public class BottomSheetController: UIViewController {
...
+    private func sheetFrame(offset: CGFloat) -> CGRect {
+        let availableWidth: CGFloat = view.bounds.width
+        let sheetWidth = shouldAlwaysFillWidth ? availableWidth : min(Constants.maxSheetWidth, availableWidth)
+
+        return CGRect(origin: CGPoint(x: (view.bounds.width - sheetWidth) / 2, y: offset),
+                      size: CGSize(width: sheetWidth, height: expandedSheetHeight))
     }
 
     private func translationRubberBandFactor(for currentOffset: CGFloat) -> CGFloat {
         var offLimitsOffset: CGFloat = 0.0
```
rdeassis, line 526: I think we can turn the offLimitsOffset into a constant too:

```
    private func translationRubberBandFactor(for currentOffset: CGFloat) -> CGFloat {
        let offLimitsOffset: CGFloat
        let expandedOffset = offset(for: .expanded)
        let collapsedOffset = offset(for: .collapsed)

        if currentOffset < expandedOffset {
            offLimitsOffset = min(expandedOffset - currentOffset, Constants.maxRubberBandOffset)
        } else if currentOffset > collapsedOffset {
            offLimitsOffset = min(currentOffset - collapsedOffset, Constants.maxRubberBandOffset)
        } else {
            offLimitsOffset = 0
        }

        return max(1.0 - offLimitsOffset / Constants.maxRubberBandOffset, Constants.minRubberBandScaleFactor)
    }
```

Does the change make sense?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -123,38 +134,42 @@ public class BottomSheetController: UIViewController {
...
             }
         }
     }
 
     /// Indicates if the content should be hidden when the sheet is collapsed
     @objc open var shouldHideCollapsedContent: Bool = true {
         didSet {
-            if shouldHideCollapsedContent != oldValue {
-                updateExpandedContentAlpha()
+            if shouldHideCollapsedContent != oldValue && isViewLoaded {
```
rdeassis, line 151: nit: use guard/return for these properties? (headerContentHeight, preferredExpandedContentHeight, collapsedContentHeight, shouldHideCollapsedContent, shouldAlwaysFillWidth,collapsedHeightInSafeArea)

rdeassis (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -36,6 +36,7 @@ public protocol BottomSheetControllerDelegate: AnyObject {
     case expanded // Sheet is fully expanded
     case collapsed // Sheet is collapsed
     case hidden // Sheet is hidden (fully off-screen)
+    case intermediate // Sheet is between states, only used during user interaction / animation
```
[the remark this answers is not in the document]
lcapkovic: I like it

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: I can't believe I didn't know (or forgot) that local let variables don't need to be initialized in line :D

PR #442, opened 2021-02-18
Fixing updating UIView scenario of a UIViewAdapter.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change fixes 3 different issues:
1. FluentUI project warning where the DrawerController.swift file was duplicated in build targets.
2. Size of icons on the presence submenu (main icons are now 24x24(medium) while the presence subitems are 16x16 (small))
3. Bug where updating the leadingView property of a cell state didn't update the UI properly

### Verification

The videos of before and after show the fix of items 2 and 3 described above:

https://user-images.githubusercontent.com/68076145/108406972-f8bd6080-71d7-11eb-9727-16add3fee566.mov 

https://user-images.githubusercontent.com/68076145/108407009-04108c00-71d8-11eb-8ae0-a12e524ee6d6.mov

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/442)

PR #856, opened 2022-01-07
Fixing regression introduced in UIViewAdapter.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Commit aa90caa3224f2b5ea0450c9b46c3216576393ca9 (#815) partially removed the UIStackView workaround from the UIViewAdapter class which was needed to prevent a hang on iOS 13.

Another workaround was introduced as part of #442 to address the status cell in the LeftNav which was not updating its leading view when set by the status selection below it.

After experimentations with the UIViewRepresentable in SwiftUI, I noticed that the the view returned from **func makeUIView** is never replaced or updated even the swiftUI body property is refreshed. The **func updateUIView** is called instead.

For that reason, we need a container UIView that will take care of updating its contents when **func updateUIView** is called.

Using a UIStackView as the container seems the most efficient way of doing that since it already handles all the constraints, compression resistance, scale and resizing issues.

### Verification

1. Tested the change in the vnext-prototype code to ensure the regression does not reproduce any longer.

| repro_iOS15  | fix_ios15 |
| repro_ios14 | fix_ios14 |

2. Ensured the LeftNav scenario of switching status updates the leading view correctly:

https://user-images.githubusercontent.com/68076145/148812264-d61d32fa-cdd1-4c51-94ba-e88d5dd32727.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/856)

### DIFF
--- a/ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
+++ b/ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
@@ -15,15 +15,19 @@ struct UIViewAdapter: UIViewRepresentable {
         self.makeView = makeView
     }
 
-    func makeUIView(context: Context) -> UIView {
-        let view = makeView()
-
-        // Ensures that the UIView respects the frame defined in the SwiftUI layer.
-        view.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
-        view.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
-
-        return view
+    func makeUIView(context: Context) -> UIStackView {
+        // Using a UIStackView as the container ensures that the content
+        // UIView respects the frame defined in the SwiftUI layer.
+        return UIStackView(arrangedSubviews: [makeView()])
     }
 
-    func updateUIView(_ view: UIView, context: Context) { }
+    func updateUIView(_ stackview: UIStackView, context: Context) {
+        // The UIStackView container needs to update its arranged subviews to cover
+        // the case where the makeview property now provides a different view.
+        stackview.arrangedSubviews.forEach({ subview in
+            subview.removeFromSuperview()
+        })
+
+        stackview.addArrangedSubview(makeView())
+    }
 }
--- a/ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
+++ b/ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
@@ -6,7 +6,7 @@
 import UIKit
 import SwiftUI
 
-/// This is a generic UIView wrapper to allow SwiftUI to use views from non-SwiftUI environments.
+/// This is a generic UIView wrapper that allows SwiftUI to use UIKit views in its hierarchy.
 struct UIViewAdapter: UIViewRepresentable {
 
     var makeView: () -> UIView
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -16,18 +16,14 @@ struct UIViewAdapter: UIViewRepresentable {
...
+        view.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
 
-        for view in stackView.arrangedSubviews {
-            view.removeFromSuperview()
-        }
-        stackView.addArrangedSubview(makeView())
+        return view
     }
+
+    func updateUIView(_ view: UIView, context: Context) { }
```
huwilkes, line 28: Does updateUIView still do anything? Or are we just not removing it in this review to reduce changes to the API?

huwilkes (approved):

huwilkes:
Also, should this be merged in to microsoft:vnext-prototype rather than microsoft:main?

ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
rdeassis: It does not do anything. It's meant to communicate changes back to the UIView when anything changes in the SwiftUI environment. The func needs to be here just because it's part of the requirements of the UIViewRepresentable protocol.

rdeassis:
> Also, should this be merged in to microsoft:vnext-prototype rather than microsoft:main?

Yes. It'll be merged to vnext-prototype on our next release version (or even earlier to update this change).
The reason I created this PR against main is because the UIViewAdapter is already in main and the flow of the merges is main -> vnext-prototype (instead of the other way around).

ios/FluentUI/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -16,18 +16,14 @@ struct UIViewAdapter: UIViewRepresentable {
...
-    func updateUIView(_ view: UIView, context: Context) {
-        // This logic should be removed wnce the "wrapping in a UIStackView" workaround is removed.
-        guard let stackView = view as? UIStackView else {
-            return
-        }
+        // Ensures that the UIView respects the frame defined in the SwiftUI layer.
+        view.setContentCompressionResistancePriority(.defaultLow, for: .vertical)
+        view.setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
 
-        for view in stackView.arrangedSubviews {
```
sophialee0416, line 28: This was introduced in #442, I was trying to brainstorm another solution for this without using the StackView. Will we work this out in a separate PR?

rdeassis: Thanks a lot for bringing this up @sophialee0416!
I have experimented some options and concluded we do need a "container" view.
The view we return in makeUIView isis never replaced/updated even when we do `UIViewAdapter(myUIview)` in the SwiftUI enviroment. So, we just replace the contents of that container view.
A UIStackView provides us the constraint and compression resistance functionality for free, which is why I think using it is probably the best option we have now.

sophialee0416 (approved):

PR #865, opened 2022-01-14
[vNext] Adopting protocols in List vNext

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The List vNext control is a bit outdated and needed some polishing. This PR adds public protocols to the control and reorganizes the file contents to match the other controls. 

### Verification

Tested on iOS 14/15, Dark/Light, verified that List works the same in List/LeftNav/ObjC demo controllers.

Demo:

https://user-images.githubusercontent.com/22566866/149566445-6b695909-b1b8-40b1-a938-19ba65de0125.mov

Theming validation:

https://user-images.githubusercontent.com/22566866/149566537-233e8051-d27e-4084-91c2-d3cc0f05854a.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/865)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
@@ -125,123 +125,7 @@ class LeftNavMenuViewController: UIViewController {
 
     private var statusCell = MSFListCellState()
 
-    private var leftNavMenuSections: [MSFListSectionState] {
-        let defaultMenuAction = {
-            guard let menuAction = self.menuAction else {
-                return
-            }
-            menuAction()
-        }
-
-        var statusCellChildren: [MSFListCellState] = []
-
-        for presence in LeftNavPresence.allCases {
-            let statusCellChild = MSFListCellState()
-            statusCellChild.leadingViewSize = .small
-            statusCellChild.title = presence.cellTitle()
-            statusCellChild.leadingUIView = presence.imageView()
-            statusCellChild.backgroundColor = .systemBackground
-            statusCellChild.onTapAction = {
-                self.setPresence(presence: presence)
-            }
-
-            statusCellChildren.append(statusCellChild)
-        }
-
-        let resetStatusCell = MSFListCellState()
-        resetStatusCell.title = "Reset status"
-        resetStatusCell.leadingViewSize = .small
-        resetStatusCell.backgroundColor = .systemBackground
-        let resetStatusImageView = UIImageView(image: UIImage(named: "ic_fluent_arrow_sync_24_regular"))
-        resetStatusImageView.tintColor = FluentUIThemeManager.S.Colors.Foreground.neutral4
-        resetStatusCell.leadingUIView = resetStatusImageView
-        resetStatusCell.onTapAction = {
-            self.setPresence(presence: .available)
-        }
-        statusCellChildren.append(resetStatusCell)
-
-        statusCell.title = LeftNavPresence.available.cellTitle()
-        statusCell.backgroundColor = .systemBackground
-        let statusImageView = LeftNavPresence.available.imageView()
-        statusImageView.tintColor = FluentUIThemeManager.S.Colors.Presence.available
-        statusCell.leadingUIView = statusImageView
-        statusCell.children = statusCellChildren
-
-        let statusMessageCell = MSFListCellState()
-        statusMessageCell.backgroundColor = .systemBackground
-        statusMessageCell.title = "Set Status Message"
-        let statusMessageImageView = UIImageView(image: UIImage(named: "ic_fluent_status_24_regular"))
-        statusMessageImageView.tintColor = FluentUIThemeManager.S.Colors.Foreground.neutral4
-        statusMessageCell.leadingUIView = statusMessageImageView
-        statusMessageCell.onTapAction = defaultMenuAction
-
-        let notificationsCell = MSFListCellState()
-        notificationsCell.backgroundColor = .systemBackground
-        notificationsCell.title = "Notifications"
-        notificationsCell.subtitle = "On"
-        notificationsCell.layoutType = .twoLines
-        let notificationsImageView = UIImageView(image: UIImage(named: "ic_fluent_alert_24_regular"))
-        notificationsImageView.tintColor = FluentUIThemeManager.S.Colors.Foreground.neutral4
-        notificationsCell.leadingUIView = notificationsImageView
-        notificationsCell.onTapAction = defaultMenuAction
-
-        let settingsCell = MSFListCellState()
-        settingsCell.backgroundColor = .systemBackground
-        settingsCell.title = "Settings"
-        let settingsImageView = UIImageView(image: UIImage(named: "ic_fluent_settings_24_regular"))
-        settingsImageView.tintColor = FluentUIThemeManager.S.Colors.Foreground.neutral4
-        settingsCell.leadingUIView = settingsImageView
-        settingsCell.onTapAction = defaultMenuAction
-
-        let whatsNewCell = MSFListCellState()
-        whatsNewCell.backgroundColor = .systemBackground
-        whatsNewCell.title = "What's new"
-        let whatsNewImageView = UIImageView(image: UIImage(named: "ic_fluent_lightbulb_24_regular"))
-        whatsNewImageView.tintColor = FluentUIThemeManager.S.Colors.Foreground.neutral4
-        whatsNewCell.leadingUIView = whatsNewImageView
-        whatsNewCell.onTapAction = defaultMenuAction
-
-        let menuSection = MSFListSectionState()
-        menuSection.cells = [statusCell, statusMessageCell, notificationsCell, settingsCell, whatsNewCell]
-
-        let microsoftAccountCell = MSFListCellState()
-        microsoftAccountCell.backgroundColor = .systemBackground
-        microsoftAccountCell.layoutType = .twoLines
-        microsoftAccountCell.title = "Contoso"
-        microsoftAccountCell.subtitle = "kat.larrson@contoso.com"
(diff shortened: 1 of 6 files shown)


rdeassis (changes requested):
These changes should follow the flow of vnext-prototype -> vnext-tokens.
Can you switch the target branch to vnext-prototype?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -97,41 +159,102 @@ public struct MSFListView: View {
...
-                      theme: FluentUIStyle? = nil) {
+    func removeCell(at index: Int) {
+        guard cells.indices.contains(index) else {
+            preconditionFailure("Index is out of bounds")
+        }
+        cells.remove(at: index)
+    }
+
+    var id = UUID()
+    @Published var cells: [MSFListCellState] = []
```
mischreiber, line 192: Opinion: since this isn't meant to be editable outside this class, how would you feel about a `private(set)`? So this line could become

```swift
@Published private(set) var cells: [MSFListCellState] = []
```

That would keep us from accidentally writing code later on that directly modifies this array.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -97,41 +159,102 @@ public struct MSFListView: View {
...
+    func removeSection(at index: Int) {
+        guard sections.indices.contains(index) else {
+            preconditionFailure("Index is out of bounds")
+        }
+        sections.remove(at: index)
     }
 
-    var window: UIWindow? {
-        return self.view.window
+    @Published var sections: [MSFListSectionStateImpl] = []
```
mischreiber, line 248: This would get (or not get) the same `private(set)` treatment as above.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -97,41 +159,102 @@ public struct MSFListView: View {
     }
 }
 
-/// UIKit wrapper that exposes the SwiftUI List implementation
-@objc open class MSFList: NSObject, FluentUIWindowProvider {
+/// Properties that make up section content
+class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, MSFListSectionState {
+    func createCell() -> MSFListCellState {
```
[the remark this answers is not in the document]
rdeassis: I agree with @mischreiber on the consistency.
My vote for the "create" term is because "insert" or "append" don't usually imply the creation of an object.
So I believe we can make explicit in the public documentation that this method creates a new cell and appends it to the section's collection of cells.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -97,16 +101,10 @@ - (void)viewDidLoad {
...
-    NSArray *sections = @[section];
-
-    MSFList *list = [[MSFList alloc] initWithSections:sections];
-
-    UIView *listView = [list view];
+    UIView *listView = [_testList view];
     listView.translatesAutoresizingMaskIntoConstraints = false;
 
-    [self.container addArrangedSubview:[list view]];
+    [self.container addArrangedSubview:[_testList view]];
```
rdeassis, line 107: Shouldn't this use listView directly?

`[self.container addArrangedSubview:listView];`

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListDemoController.swift
```diff
@@ -66,41 +64,38 @@ class ListDemoController: DemoController {
...
-        listSection.cells[1].onTapAction = {
+        collapsibleSection.getCellState(at: 0).children = children
+        collapsibleSection.getCellState(at: 0).isExpanded = true
+        collapsibleSection.getCellState(at: 1).onTapAction = {
             self.showAlertForAvatarTapped(name: samplePersonas[1].name)
         }
-        listData.append(listSection)
 
         /// TableViewCell Sample Data Sections
         for sectionIndex in 0...sections.count - 1 {
```
rdeassis, line 86: nit: opportunity for a more readable code:

`for sectionIndex in 0..<sections.count {`

(same on line 94 and other places you can find it)

ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
```diff
@@ -125,7 +125,42 @@ class LeftNavMenuViewController: UIViewController {
...
+        chevron.tintColor = Colors.textPrimary
+
+        persona.state.presence = .available
+        persona.state.primaryText = "Kat Larrson"
+        persona.state.secondaryText = "Designer"
+        persona.state.image = UIImage(named: "avatar_kat_larsson")
+        persona.state.titleTrailingAccessoryUIView = chevron
+        persona.state.backgroundColor = .systemBackground
+        persona.state.onTapAction = {
+            self.dismiss(animated: true, completion: {
```
rdeassis, line 143: Watch out for memory leaks when you capture self in a closure.
The closure will retain a strong reference to the container object (instance of LeftNavMenuViewController) and the instance of LeftNavMenuViewController will retain a strong reference to the closure.

This will prevent these 2 objects from being deallocated for the life of the app.

You can use the snippet below and for multiple operations on the self reference, you can do something like the code in this [link](https://github.com/microsoft/fluentui-apple/blob/b2f6b63cbfd77d6e8bde02e3d9ead05f8d08227d/ios/FluentUI/Bottom%20Commanding/BottomCommandingController.swift#L315):


 ```
    let defaultMenuAction = { [weak self] in
            guard let menuAction = self?.menuAction else {
                return
            }

            menuAction()
        }
```

File: ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift [hunk 7]
```diff
@@ -125,7 +125,42 @@ class LeftNavMenuViewController: UIViewController {
...
-    private var leftNavMenuSections: [MSFListSectionState] {
+    private var leftNavMenuList = MSFList()
+
+    private var leftNavAccountViewHeightConstraint: NSLayoutConstraint?
+
+    private lazy var leftNavAccountView: UIView = {
+        let chevron = UIImageView(image: UIImage(named: "ic_fluent_ios_chevron_right_20_filled"))
+        chevron.tintColor = Colors.textPrimary
+
+        persona.state.presence = .available
```
rdeassis, line 136: Can you store persona.state in a constant to use it below?

`let personaState = persona.state`

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -6,40 +6,100 @@
...
+
+    /// Remove a Cell from the Section.
+    /// - Parameter index: The zero-based index of the Cell that will be removed from the Section.
+    func removeCell(at index: Int)
+}
+
+/// Properties that can be used to customize the appearance of the List.
+@objc public protocol MSFListState {
+    /// The number of Sections in the List.
+    var count: Int { get }
```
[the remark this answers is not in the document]
rdeassis: I like the idea...
I think you can apply the same for the cell count in the MSFListSectionState above.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -97,41 +159,102 @@ public struct MSFListView: View {
...
+        return cells.count
+    }
+
+    var style: MSFHeaderFooterStyle {
+        get {
+            return headerTokens.style
+        }
+        set {
+            headerTokens.style = newValue
+            headerTokens.updateForCurrentTheme()
```
rdeassis, line 207: The call to updateForCurrentTheme() should be in the didSet handler of the tokens style property.

Examples:
- https://github.com/microsoft/fluentui-apple/blob/b2f6b63cbfd77d6e8bde02e3d9ead05f8d08227d/ios/FluentUI/Avatar/Avatar.swift#L486
- https://github.com/microsoft/fluentui-apple/blob/b2f6b63cbfd77d6e8bde02e3d9ead05f8d08227d/ios/FluentUI/Avatar/AvatarTokens.swift#L68

ios/FluentUI/Vnext/List/List.swift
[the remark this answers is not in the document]
sophialee0416: I think the explanation is definitely warranted! I think I will stick to using `create` for now and add further clarification.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -6,40 +6,95 @@
...
+    var style: MSFHeaderFooterStyle { get set }
+
+    /// The number of Cells in the Section.
+    var cellCount: Int { get }
+
+    /// Creates a new Cell and appends it to the array of cells in a Section.
+    func createCell() -> MSFListCellState
+
+    /// Creates a new Cell within the Section at a specific index.
+    func createCell(at index: Int) -> MSFListCellState
```
rdeassis, line 63: nit: add parameter documentation here too (like below)

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -6,40 +6,95 @@
...
+    /// The number of Sections in the List.
+    var sectionCount: Int { get }
+
+    /// Creates a new Section and appends it to the array of sections in a List.
+    func createSection() -> MSFListSectionState
 
-    public init(sections: [MSFListSectionState]) {
-        self.state = MSFListState()
+    /// Creates a new Section within the List at a specific index.
+    func createSection(at index: Int) -> MSFListSectionState
```
rdeassis, line 83: nit: add parameter documentation here

rdeassis (approved):
Just a few nits to address, but overall looks good! Thanks for addressing all the comments!

PR #868, opened 2022-01-14
Update padding, alignment, and font size of CardView

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Updated card view. Card view currently supports two variations, horizontal and vertical.
- Horizontal layout changes:
   - Changed the padding so the overall card height is larger.
   - Changed the width to be larger.
- Vertical layout changes:
   - Center aligned title, subtitle, and icon.
   - Decreased the title and subtitle font sizes
   - Increased padding to increase overall card size.
### Verification

Manually tested and checked the changes.

|Simulator Screen Shot - iPhone 13 Pro - 2022-01-14 at 10 11 48|Simulator Screen Shot - iPhone 13 - 2022-01-14 at 10 09 06 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/868)

### DIFF
--- a/ios/FluentUI/Card/CardView.swift
+++ b/ios/FluentUI/Card/CardView.swift
@@ -226,7 +226,7 @@ open class CardView: UIView {
         case .vertical:
             primaryLabel.textAlignment = .center
             primaryLabel.style = .caption1
-		}
+        }
         addSubview(primaryLabel)
 
         // Subtitle
--- a/ios/FluentUI/Card/CardView.swift
+++ b/ios/FluentUI/Card/CardView.swift
@@ -369,7 +369,7 @@ open class CardView: UIView {
                 primaryLabel.leadingAnchor.constraint(equalTo: iconView.trailingAnchor, constant: Constants.horizontalContentSpacing),
                 iconView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: Constants.paddingLeading),
                 primaryLabel.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -Constants.paddingTrailing)
-			])
+            ])
 
             // Center the title vertically if there is no subtitle
             if secondaryText == nil && !twoLineTitle {


huwilkes:
What's the reason for the change? The designs in the Fluent iOS Figma don't seem to have been changed.

huwilkes:
Maybe out of scope for this review, but sounds like we want to rename the class, also? Doesn't sound like we've decided on a name yet, though.

huwilkes:
Did you check the accessibility of the control? Dark mode, pointer support, high contrast?

ios/FluentUI/Card/CardView.swift
```diff
@@ -221,13 +219,27 @@ open class CardView: UIView {
...
+        case .horizontal:
+            primaryLabel.textAlignment = .natural
+            primaryLabel.style = .subhead
+        case .vertical:
+            primaryLabel.textAlignment = .center
+            primaryLabel.style = .caption1
+		}
         addSubview(primaryLabel)
 
         // Subtitle
```
huwilkes, line 232: Do we still need the subtitle? Doesn't look like the new design has one, just the title and icon.

ios/FluentUI/Card/CardView.swift
```diff
@@ -33,9 +33,9 @@ public enum CardStyle: Int, CaseIterable {
     var width: CGFloat {
         switch self {
         case .horizontal:
-            return 156
+            return 175
         case .vertical:
-            return 120
+            return 97
```
huwilkes, line 38: Were these intended to be final designs? 97 seems like a weird number.

huwilkes:
> Did you check the accessibility of the control? Dark mode, pointer support, high contrast?

Also different text sizes. Is the control supposed to grow as the text does, or do we need to add large text viewer support?

ios/FluentUI/Card/CardView.swift
alexanderboswell: Ill double check with design as this is what it was in the figma.

ios/FluentUI/Card/CardView.swift
alexanderboswell: Not really but also have no reason to remove it. The changes I made still support having a title, subtitle, or a two line title. 

My changes are being made so the component can be used now but that doesn't mean there isnt another iteration of the changes. I would assume for wider support they might want to keep both title and subtitle, so that's why I left it.

alexanderboswell:
@huwilkes 
Renaming the class
- There hasn't been a decision yet. There will be a vnext of the fluent card and we have until that point to rename. Because this component isnt used by anyone right now, I was not concerned on the rename until the next version is created.
"Did you check the accessibility of the control? Dark mode, pointer support, high contrast?"
- None of the changes I made affect how accessibility, darkmode, point support, or high contrast are currently supported in the component. My assumption was those have already been taken care. I did do a quick check and made sure that larger text size and dark mode still work as expected.
"Also different text sizes. Is the control supposed to grow as the text does, or do we need to add large text viewer support?"
- I kept with the current implementation which changes the height of the component as the text size gets larger.

ios/FluentUI/Card/CardView.swift
```diff
@@ -221,13 +219,27 @@ open class CardView: UIView {
...
         primaryLabel.numberOfLines = twoLineTitle ? Constants.twoLineTitle : Constants.defaultTitleNumberOfLines
-        primaryLabel.textAlignment = .natural
+        switch style {
+        case .horizontal:
+            primaryLabel.textAlignment = .natural
+            primaryLabel.style = .subhead
+        case .vertical:
+            primaryLabel.textAlignment = .center
+            primaryLabel.style = .caption1
+		}
```
sophialee0416, line 229: I think this needs to be changed into spaces (Xcode shows the tab indicator!)

alexanderboswell: good catch!

mischreiber (commented):
So this looks fine overall, but let's wait until we can engage with design after the holiday.

rdeassis:
> * darkmode, point support, or high contrast are currently supported in the component. My assumption was those have already been taken care. I did do a quick check and made sure that larger text size and dark mode still work as expected.
> 
> "Also different text sizes. Is the control supposed to grow as the text does, or do we ne

Did you happen to validate RTL scenarios too? (as you're touching constraints, it's a good sanity validation as well).
Do you mind filling the validation on the description of the PR with the validations you performed?

Thanks!

PR #874, opened 2022-01-20
[vNext Tokens] Convert `FluentDivider` 

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Update the Fluent Divider to use the new token system

### Verification
| Divider-Old-Tokens | Divider-New-Tokens |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/874)


ios/FluentUI/Vnext/Divider/Divider.swift
```diff
@@ -64,38 +63,38 @@ public struct FluentDivider: View {
...
-                          from: theme,
-                          with: windowProvider)
+            .resolveTokens(self)
+            .resolveTokenModifier(self, value: state.orientation)
     }
+
+    public func overrideTokens(_ tokens: DividerTokens?) -> FluentDivider {
+        state.overrideTokens = tokens
+        return self
+    }
```
huwilkes, line 73: Seems like this is usually in a *Modifiers.swift file, but I wasn't really sure what the point of putting it in a file all by itself was.

ios/FluentUI/Vnext/Divider/DividerTokens.swift
```diff
@@ -12,35 +12,41 @@ import UIKit
...
-
-    init(spacing: MSFDividerSpacing) {
+/// Design token set for the `FluentDivider` control.
+public class DividerTokens: ControlTokens {
+    /// Creates an instance of `MSFDividerTokens`with optional token value overrides.
+    /// - Parameters:
+    ///   - spacing: MSFDividerSpacing enumeration value that will define pre-defined value for the padding
+    ///   - padding: CGFloat that defines the padding around the Fluent Divider
+    ///   - color: The color of the Fluent Divider
+    public init(spacing: MSFDividerSpacing = .none,
```
huwilkes, line 22: Seems like most of our other controls don't have a default value like this in the init, but I figured it was fine in this case.

ios/FluentUI/Vnext/Divider/Divider.swift
mischreiber: I think it's probably still worth making the new file (A) for consistency, and (B) because there will more modifiers when the control is ready to be hosted in SwiftUI, specifically for `color` and `orientation`.

ios/FluentUI/Vnext/Divider/Divider.swift
```diff
@@ -64,38 +63,38 @@ public struct FluentDivider: View {
...
 }
 
 /// Properties available to customize the Divider.
-class MSFDividerStateImpl: NSObject, ObservableObject, MSFDividerState {
+class MSFDividerStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFDividerState {
+    @Published var overrideTokens: DividerTokens?
+    @Published var tokens: DividerTokens
+    var defaultTokens: DividerTokens {.init()}
+
     @Published var color: UIColor?
```
mischreiber, line 86: Now that the divider's color can be set via `DividerTokens`, do we still need a separate property here?

ios/FluentUI/Vnext/Divider/Divider.swift
```diff
@@ -64,38 +63,38 @@ public struct FluentDivider: View {
...
+    @Environment(\.fluentTheme) var fluentTheme: FluentTheme
+    @ObservedObject var state: MSFDividerStateImpl
 }
 
 /// Properties available to customize the Divider.
-class MSFDividerStateImpl: NSObject, ObservableObject, MSFDividerState {
+class MSFDividerStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFDividerState {
+    @Published var overrideTokens: DividerTokens?
+    @Published var tokens: DividerTokens
+    var defaultTokens: DividerTokens {.init()}
```
mischreiber, line 84: Nit: spacing around the curly braces.

mischreiber: Separate issue: `defaultTokens` should still take advantage of this `state` object's properties; `spacing` in this case. So the full line should read
```swift
var defaultTokens: DividerTokens { .init(spacing: self.spacing) }
```

ios/FluentUI/Vnext/Divider/DividerTokens.swift
mischreiber: My concern is that these token files may eventually be generated from JSON, which will make these unique elements tricky.

And once the above issue is resolved (wherein `spacing` is always passed in from `defaultTokens`), there should be no need for a default value.

ios/FluentUI/Extensions/UIColor+Extensions.swift
```diff
@@ -83,7 +83,7 @@ extension UIColor {
         }
     }
 
-    var dynamicColor: DynamicColor? {
+    public var dynamicColor: DynamicColor? {
```
mischreiber, line 86: Can you move this higher in the file to match the Microsoft Swift Guide?

https://microsoft.github.io/swift-guide/FileOrganization.html

ios/FluentUI/Vnext/Divider/DividerModifiers.swift
```diff
@@ -0,0 +1,15 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+import UIKit
+
+public extension FluentDivider {
+
+    func overrideTokens(_ tokens: DividerTokens?) -> FluentDivider {
```
mischreiber, line 11: Nit: comment for public function

ios/FluentUI/Vnext/Divider/MSFDivider.swift
```diff
@@ -38,23 +30,8 @@ import SwiftUI
     @objc public init(orientation: MSFDividerOrientation,
                       spacing: MSFDividerSpacing,
                       theme: FluentUIStyle?) {
```
mischreiber, line 32: This initializer can be merged with its convenience init; `FluentUIStyle` is part of the old theming system and should no longer be referenced in Fluent 2.0 controls.

ios/FluentUI/Extensions/UIColor+Extensions.swift
huwilkes: I think it might already be in the right place. It is below some internal functions, but they also happen to be inits and I'm not sure we want to break up the inits?

ios/FluentUI/Vnext/Divider/MSFDivider.swift
```diff
@@ -7,54 +7,18 @@ import UIKit
...
-    @objc open var view: UIView {
-        return hostingController.view
-    }
-
+@objc public class MSFDivider: ControlHostingContainer {
     /// The object that groups properties that allow control over the Divider appearance.
-    @objc open var state: MSFDividerState {
-        return divider.state
-    }
+    @objc public let state: MSFDividerState
```
mischreiber, line 12: Nit: variables should go below methods (see `MSFCardNudge.swift` for example).

mischreiber (approved):

PR #875, opened 2022-01-20
Add flexible height mode to `BottomSheetController`

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Currently our bottom sheet works in a "rigid" mode where its height is static and dragging it just translates the sheet view vertically on or off-screen.
Some clients need to pin views to the bottom of the sheet that are always visible (like a text field) so we need a flexible mode where dragging actually resizes the sheet, so bottom of the sheet content is always visible.

This PR adds the `isFlexibleHeight` property which enables the flexible height mode. The sheet will resize anywhere above collapsed state and stay rigid anywhere below it (to make sure we're not forcing the content to layout in small or zero heights.

### Verification
Default rigid mode:

https://user-images.githubusercontent.com/3610850/151105447-757a9359-69ae-47ba-9cff-f638f6f99851.mov

Flexible mode:

https://user-images.githubusercontent.com/3610850/151105461-199ebd9d-c123-49f5-b522-b3ad5dddaec4.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/875)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
@@ -21,9 +21,10 @@ class BottomSheetDemoController: UIViewController {
         view.addSubview(optionTableView)
         mainTableView = optionTableView
 
-        let bottomSheetViewController = BottomSheetController(headerContentView: headerView, expandedContentView: personaListView)
+        let bottomSheetViewController = BottomSheetController(headerContentView: headerView, expandedContentView: expandedContentView)
         bottomSheetViewController.hostedScrollView = personaListView
         bottomSheetViewController.headerContentHeight = BottomSheetDemoController.headerHeight
+        bottomSheetViewController.collapsedContentHeight = 150
         bottomSheetViewController.delegate = self
 
         self.bottomSheetViewController = bottomSheetViewController
@@ -66,6 +67,10 @@ class BottomSheetDemoController: UIViewController {
         bottomSheetViewController?.shouldHideCollapsedContent.toggle()
     }
 
+    @objc private func toggleFlexibleSheetHeight(_ sender: BooleanCell) {
+        bottomSheetViewController?.isFlexibleHeight = sender.isOn
+    }
+
     @objc private func toggleHandleUsingCustomAccessibilityLabel(_ sender: BooleanCell) {
         let isOn = sender.isOn
         bottomSheetViewController?.handleCollapseCustomAccessibilityLabel = isOn ? "Collapse Bottom Sheet" : nil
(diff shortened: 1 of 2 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -81,14 +86,37 @@ class BottomSheetDemoController: UIViewController {
...
         return personaListView
     }()
 
+    private lazy var expandedContentView: UIView = {
+        let view = UIView()
+        view.translatesAutoresizingMaskIntoConstraints = false
+        view.addSubview(personaListView)
+
+        let bottomView = UIView()
+        bottomView.backgroundColor = .systemTeal
```
rdeassis, line 103: not sure I understand the reason for setting the backgroundColor to .systemTeal here...

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -97,6 +97,23 @@ public class BottomSheetController: UIViewController {
...
+    ///
+    /// When set to `true`, moving the sheet beyond the `.collapsed` state will resize it.
+    ///
+    /// Use flexible height if you have views attached to the bottom of `expandedContentView` that should always be visible.
+    @objc open var isFlexibleHeight: Bool = false {
+        didSet {
+            guard isViewLoaded else {
+                return
+            }
+            view.setNeedsLayout()
```
anandrajeswaran, line 113: Why is this needed? I would have expected toggling this doesn't change anything about its current position and only affects what happens when interacted with? It sounds like this will also cause a one-time unanimated snap to allowed conditions that may be worth documenting?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -741,20 +768,33 @@ public class BottomSheetController: UIViewController {
...
+        return max(collapsedSheetHeight, height)
+    }
+
+    // Height of the sheet in collapsed state
+    private var collapsedSheetHeight: CGFloat {
+        let visibleContentHeight = (collapsedContentHeight > 0) ? collapsedContentHeight : headerContentHeight
+        return currentResizingHandleHeight + visibleContentHeight + view.safeAreaInsets.bottom
+    }
+
+    private var currentResizingHandleHeight: CGFloat {
```
anandrajeswaran, line 796: Minor: Do we have some convention on preferring computed vars vs. funcs? Or what the tradeoffs are?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: Yeah sorry, the demo app design should be a bit more intentional. I'll change it to something better.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: The sheet frame needs to be recalculated, because everywhere except fully expanded state the frame will be different in flexible mode. For example if we're in collapsed state, setting this to true will resize the sheet so that its bottom is aligned with the bottom of the screen. Its offset wouldn't change in this case.

This won't cause a snap if the sheet is being interacted with because it'll be in the `.transitioning` state. In that case the interaction will automatically switch to calculating the frame using the new flexible / rigid mode.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Not that I'm aware of. I'd be very surprised if there were actual perf differences. Do you prefer a func here?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
lcapkovic: I updated it, also added new screen recordings in the description.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
anandrajeswaran: I don't know of any differences or have an preferences - but it feels like we should since otherwise it's 2 identical ways of doing the same thing? Can talk about outside the scope of this PR though.

rdeassis: I think the undocumented pattern we're following is that we use computed properties when:
1. no parameters are needed
2. there is a return type
3. the implementation does not imply actions, but calculates a value to be returned

**Examples of computed properties:**
 - var leadingViewPadding: Int
 - var window: UIWindow?
 
 **Examples of funcs with no parameters:**
 - func updateForCurrentTheme()
 - func reload()
 - func setupDraggingWindow()

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -620,7 +648,7 @@ public class BottomSheetController: UIViewController {
             ? Constants.Spring.oscillatingDampingRatio
             : Constants.Spring.defaultDampingRatio
 
-        let springParams = UISpringTimingParameters(dampingRatio: damping, initialVelocity: CGVector(dx: 0.0, dy: springVelocity))
+        let springParams = UISpringTimingParameters(dampingRatio: damping, initialVelocity: CGVector(dx: springVelocity, dy: springVelocity))
```
rdeassis, line 651: nit: break this into multiple lines for better readability?

rdeassis (approved):

PR #882, opened 2022-01-26
Fix `BottomSheetController` not exiting .transitioning state

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

`BottomSheetController` has a `.transitioning` state used during animation and user interaction to prevent the sheet frame from resetting during a layout pass.
The sheet can get stuck in the transitioning state when an animation gets cancelled, causing it to stop updating its frame until a new animation happens.

This PR does a few things to prevent this:
- make sure we get out of `.transitioning` state in the animation completion handler, regardless of how the animation ended
- `layoutIfNeeded` before an animation starts so we're in a good state if the animation cancels
- because the completion handler will now always get us out of transitioning state, make sure we reset back to it when a pan begins

Unrelated, but fixing it here:
- forward correct interaction data to `stateChangeAnimator` to fix incorrect delegate callbacks.

### Verification

Sanity checks in test app in cases that previously reproduced. Also tested in product.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/882)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -596,7 +596,10 @@ public class BottomSheetController: UIViewController {
         completeAnimationsIfNeeded()
 
         if currentSheetVerticalOffset != offset(for: targetExpansionState) {
-            let animator = stateChangeAnimator(to: targetExpansionState, velocity: velocity, interaction: interaction, shouldNotifyDelegate: shouldNotifyDelegate)
+            let animator = stateChangeAnimator(to: targetExpansionState,
+                                               velocity: velocity,
+                                               interaction: interaction,
+                                               shouldNotifyDelegate: shouldNotifyDelegate)
             animator.addCompletion({ finalPosition in
                 completion?(finalPosition)
             })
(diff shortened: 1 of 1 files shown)


anandrajeswaran (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -596,7 +596,7 @@ public class BottomSheetController: UIViewController {
         completeAnimationsIfNeeded()
 
         if currentSheetVerticalOffset != offset(for: targetExpansionState) {
-            let animator = stateChangeAnimator(to: targetExpansionState, velocity: velocity)
+            let animator = stateChangeAnimator(to: targetExpansionState, velocity: velocity, interaction: interaction, shouldNotifyDelegate: shouldNotifyDelegate)
```
rdeassis, line 599: nit: do you mind breaking these params into multiple lines? (here and line 669)

rdeassis (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Done

mischreiber (approved):

PR #889, opened 2022-02-02
Support asynchronously loading of persona images

lingzlu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Introduce a new function to Persona protocol to allow the clients to fetch image asynchronously.
The change does not modify the existing behavior, and should have no impact on the existing consumers of the library. 

### Verification
The change was tested manually on Planner app for add group members feature, where the Persona images are fetched from Graph API. 

Added change in PeoplePickerDemoController to demonstrate the change. 
Simulator Screen Recording - iPhone 13 Pro - 2022-02-12 at 12 02 59

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/889)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
@@ -7,7 +7,7 @@ import FluentUI
 import UIKit
 
 let samplePersonas: [PersonaData] = [
-    PersonaData(name: "Kat Larrson", email: "kat.larrson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
+    PersonaData(name: "Kat Larsson", email: "kat.larrson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
     PersonaData(name: "Kristin Patterson", email: "kristin.patterson@contoso.com", subtitle: "Software Engineer"),
     PersonaData(name: "Ashley McCarthy", image: UIImage(named: "avatar_ashley_mccarthy")),
     PersonaData(name: "Allan Munger", email: "allan.munger@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_allan_munger")),
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
@@ -7,7 +7,7 @@ import FluentUI
 import UIKit
 
 let samplePersonas: [PersonaData] = [
-    PersonaData(name: "Kat Larsson", email: "kat.larrson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
+    PersonaData(name: "Kat Larsson", email: "kat.larsson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
     PersonaData(name: "Kristin Patterson", email: "kristin.patterson@contoso.com", subtitle: "Software Engineer"),
     PersonaData(name: "Ashley McCarthy", image: UIImage(named: "avatar_ashley_mccarthy")),
     PersonaData(name: "Allan Munger", email: "allan.munger@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_allan_munger")),
--- a/ios/FluentUI/Avatar/Persona.swift
+++ b/ios/FluentUI/Avatar/Persona.swift
@@ -42,6 +42,10 @@ public protocol Persona {
 
     /// The subtitle value for the persona.
     var subtitle: String { get }
+
+    /// Loads the image asynchronously. Implement this instead of use ``image``  property if it needs to be fetched asynchronously. Image caching should be handled in the implementation for optimial efficiency.
+    /// - Parameter completion: The completion block that returns the image
+    @objc optional func loadImageAsync(completion: @escaping (UIImage?) -> Void)
 }
 
 extension Persona {
--- a/ios/FluentUI/Avatar/Persona.swift
+++ b/ios/FluentUI/Avatar/Persona.swift
@@ -43,7 +43,8 @@ public protocol Persona {
     /// The subtitle value for the persona.
     var subtitle: String { get }
 
-    /// Loads the image asynchronously. Implement this instead of use ``image``  property if it needs to be fetched asynchronously. Image caching should be handled in the implementation for optimial efficiency.
+    /// Loads the image asynchronously. This is called when setting up Persona row for presentation. Image caching should be handled in the implementation for optimal efficiency.
+    /// If the ``image`` is provided, it will be overrided by the loaded image when completion block is called, acting like a custom placeholder.
     /// - Parameter completion: The completion block that returns the image
     @objc optional func loadImageAsync(completion: @escaping (UIImage?) -> Void)
 }
--- a/ios/FluentUI/Avatar/Persona.swift
+++ b/ios/FluentUI/Avatar/Persona.swift
@@ -43,10 +43,10 @@ public protocol Persona {
     /// The subtitle value for the persona.
     var subtitle: String { get }
 
-    /// Loads the image asynchronously. This is called when setting up Persona row for presentation. Image caching should be handled in the implementation for optimal efficiency.
-    /// If the ``image`` is provided, it will be overrided by the loaded image when completion block is called, acting like a custom placeholder.
+    /// Fetches the image asynchronously. This is called when setting up persona row for presentation. Image caching should be handled in the implementation for optimal efficiency.
+    /// If the ``image`` is provided, it will be overrided by the fetched image when completion block is called, acting like a custom placeholder.
     /// - Parameter completion: The completion block that returns the image
-    @objc optional func loadImageAsync(completion: @escaping (UIImage?) -> Void)
+    @objc optional func fetchImage(completion: @escaping (UIImage?) -> Void)
 }
 
 extension Persona {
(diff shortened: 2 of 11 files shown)


ios/FluentUI/People Picker/PersonaCell.swift
```diff
@@ -29,6 +29,22 @@ open class PersonaCell: TableViewCell {
...
         }
 
+        if let loadImageAsync = persona.loadImageAsync {
+            loadImageAsync { image in
+                guard let image = image else {
+                    return
+                }
+                
+                if Thread.isMainThread {
+                    avatar.state.image = image
```
harrieshin, line 39: we should still check if "avatar" is still valid for the personal cell on the async thread.

ios/FluentUI/Avatar/Persona.swift
```diff
@@ -42,6 +42,10 @@ public protocol Persona {
 
     /// The subtitle value for the persona.
     var subtitle: String { get }
+
+    /// Loads the image asynchronously. Implement this instead of use ``image``  property if it needs to be fetched asynchronously. Image caching should be handled in the implementation for optimial efficiency.
+    /// - Parameter completion: The completion block that returns the image
+    @objc optional func loadImageAsync(completion: @escaping (UIImage?) -> Void)
```
harrieshin, line 48: also wroth adding a test case for  demo controller of this usage.

markavitale, line 48: I think we should be explicit about how this interacts with the image property. What happens if you provide both? Are there potential race conditions? Should prevent both being implemented simultaneously (via `preconditionFailure`)?

I would lean towards allowing both as showing a static placeholder image that is replaced by an async loaded image seems like a normal workflow, but I think we should be explicit about it. The use of `instead of` in the comment above makes me think they are mutually exclusive, but the implementation looks like it would allow providing a placeholder image via `image` and the "real" image via `loadImageAsync`

lingzlu: I thought about it, but will need some design idea on this.
I was thinking have a toggle to switching between using the static image vs loading async, but unsure on how to present the toggle. Alternatively, we can do both at the same time, use the static image first, then do time delay to simulate background task, and set the same image after some delay?

ios/FluentUI/People Picker/PersonaCell.swift
lingzlu: Sorry, I'm not sure what you mean by "if avatar is still valid".
Are you suggesting capture `avatar` weakly in the `loadImageAsync `closure and verify that it's still in memory?

lingzlu: I think it make sense to capture avatar weakly here. Please let me know if the valid check you mentioned is refer to something else.  :)

ios/FluentUI/Avatar/Persona.swift
lingzlu: Updated comment. Grammar is definitely one of my weak areas. So please proofreading for me again and feel free to reword the sentence as you see fit. 😊

ios/FluentUI/People Picker/PersonaCell.swift
```diff
@@ -29,6 +29,23 @@ open class PersonaCell: TableViewCell {
...
+                guard let avatar = avatar,
+                      let image = image else {
+                    return
+                }
+
+                if Thread.isMainThread {
+                    avatar.state.image = image
+                } else {
+                    DispatchQueue.main.async {
+                        avatar.state.image = image
```
harrieshin, line 43: when we send over image and avatar to the next main queue, I still think we should weak capture here tool

lingzlu: I could include the weak capture in `DispatchQueue.main.async`, but logically, I don't see the need for it. The intend of weak capture in `loadImageAsync` closure is to prevent avatar living longer than needed, retain cycle shouldn't be an issue here. if `loadImageAsync` completed and `avatar` is still in memory, then `DispatchQueue.main.async` is not going to keep it for any longer since the DispatchQueue closure will get executed soon. Correct me if I missed something.

ios/FluentUI/Avatar/Persona.swift
rdeassis: One idea that came into my mind is quite similar to what you suggest, @lingzlu:
That idea is basically about adding a toggle switch at the top of the PeoplePicker demo controller. Something like  "Delay load images" or "Load persona images asynchronously".
Then if that toggle is on, the images are set asynchronously after a timeout (of maybe 1 second?). If not, they're just set directly as the current state of things.

Screen Shot 2022-02-07 at 9 31 29 AM

ios/FluentUI/Avatar/Persona.swift
```diff
@@ -42,6 +42,11 @@ public protocol Persona {
 
     /// The subtitle value for the persona.
     var subtitle: String { get }
+
+    /// Loads the image asynchronously. This is called when setting up Persona row for presentation. Image caching should be handled in the implementation for optimal efficiency.
+    /// If the ``image`` is provided, it will be overrided by the loaded image when completion block is called, acting like a custom placeholder.
+    /// - Parameter completion: The completion block that returns the image
+    @objc optional func loadImageAsync(completion: @escaping (UIImage?) -> Void)
```
rdeassis, line 49: nit: How about naming the func as "fetchImage" instead? (seems more aligned with the API naming convention that Apple is using)

lingzlu: Sounds good, updated!

ios/FluentUI/Avatar/Persona.swift
lingzlu: @rdeassis thanks for sharing the idea! I can mock something like you mentioned:

> "Load persona images asynchronously     [Toggle] ".

Another idea that just came to mind is that given there are two demo views for persona: `PeoplePicker` and `PersonaListView`, I can update `PeoplePickerDemoController` to use a new list of `samplePersonas` objects that are subclass of `PersonaData` with `fetchImage` implementation. From the scenario stand-point, it also make sense for people picker to load image asynchronously. It's also nice that we don't have to worry about UI mock up. Let me know what you think. 

Another request is that we (Planner iOS team) would like to have the change available as soon as possible, are you and @harrieshin okay with me doing a follow up change  for the demo?

rdeassis: Hey @lingzlu my personal opinion is that the toggle showcases the ability to set avatar images asynchronously better because it explicitly calls out that feature that can be turned on/off in the demo controller.

I'm holding on the [release of 0.4.4](https://github.com/microsoft/fluentui-apple/pull/893) so that this change can get in that version and be released this week.

Regarding the deferral of the Demo controller changes into another PR:
What would be the timeline for that followup change on the demo controller?
The reason I ask is because it seems to be a fairly simple change to make to the demo controller and ideally we'd like to have it all merged together.

rdeassis: Hey @lingzlu, just a heads up that we needed to release the version 0.4.4.
Your change will be included in version 0.4.5 and we can expedite it once you merge this PR.
Do you think you can include the demo controller changes in this PR still?

ios/FluentUI/Avatar/Persona.swift
```diff
@@ -42,6 +42,11 @@ public protocol Persona {
 
     /// The subtitle value for the persona.
     var subtitle: String { get }
+
+    /// Fetches the image asynchronously. This is called when setting up persona row for presentation. Image caching should be handled in the implementation for optimal efficiency.
+    /// If the ``image`` is provided, it will be overrided by the fetched image when completion block is called, acting like a custom placeholder.
+    /// - Parameter completion: The completion block that returns the image
+    @objc optional func fetchImage(completion: @escaping (UIImage?) -> Void)
```
rdeassis, line 49: Thanks for updating the naming here.
I left another message about versioning (we needed to release 0.4.4 and this change will be released in version 0.4.5).
Do you think you can include the demo controller changes in this PR too?

ios/FluentUI/Avatar/Persona.swift
lingzlu: hey @rdeassis, sorry, it has been a busy week for me. I probably won't get to it until early next week (likely next Monday). I know the change isn't big, the request for the deferral is out of the concern for the review process. What is your target date for the next release?

ios/FluentUI/Avatar/Persona.swift
lingzlu: Updated with the demo change. 
Simulator Screen Recording - iPhone 13 Pro - 2022-02-12 at 12 02 59

rdeassis (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
```diff
@@ -7,7 +7,7 @@ import FluentUI
 import UIKit
 
 let samplePersonas: [PersonaData] = [
-    PersonaData(name: "Kat Larrson", email: "kat.larrson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
+    PersonaData(name: "Kat Larsson", email: "kat.larrson@contoso.com", subtitle: "Designer", image: UIImage(named: "avatar_kat_larsson")),
```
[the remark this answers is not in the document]
lingzlu: Updated, also correct all other usages. Assuming `Kat Larsson` with two `s` is the correct name :)

lingzlu:
@rdeassis could you take another pass on this PR? let me know if you have other feedbacks. 
if all is good, could you run the workflow checks for me again, thanks for your time! :)

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaListViewDemoController.swift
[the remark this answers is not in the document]
rdeassis: > Updated, also correct all other usages. Assuming `Kat Larsson` with two `s` is the correct name :)

Thanks a lot for correcting the other ones too! I believe the image file is the source of truth as far as Kat's identity! :)

rdeassis:
Thank you @lingzlu for addressing all comments and also for the fixes beyond the scope of the changes!

PR #890, opened 2022-02-02
Cardview large content viewer support

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Updated the accessibility of CardView. For dynamic text, the control will increase in height. It does not grow in width also though and this is so that it can keep a consistent layout when in a collectionView. To be able to view the title and subtitle for accessibility text sizes large content viewer is being used to allow users to read the text. 

### Verification

Tested manually 

| No support was available, so the title text would be cut off with no way to view | Screen Shot 2022-02-03 at 10 28 59 AM
| No support was available, so all title and subtitle text would be cut off with no way to view | Screen Shot 2022-02-03 at 10 28 57 AM|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/890)

### DIFF
--- a/ios/FluentUI/Card/CardView.swift
+++ b/ios/FluentUI/Card/CardView.swift
@@ -489,7 +489,7 @@ open class CardView: UIView {
             largeContextText += "\n" + secondaryText
         }
         largeContentTitle = largeContextText
-	}
+    }
 
     @objc private func handleCardTapped(_ recognizer: UITapGestureRecognizer) {
         delegate?.didTapCard?(self)
--- a/ios/FluentUI/Card/CardView.swift
+++ b/ios/FluentUI/Card/CardView.swift
@@ -487,11 +487,13 @@ open class CardView: UIView {
     }
 
     private func updateLargeContentTitle() {
-        var largeContextText = primaryText
-        if let secondaryText = secondaryText, !twoLineTitle {
-            largeContextText += "\n" + secondaryText
-        }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Card/CardView.swift
```diff
@@ -475,6 +483,14 @@ open class CardView: UIView {
         NSLayoutConstraint.activate(layoutConstraints)
     }
 
+    private func setLargeContentTitle() {
```
harrieshin, line 486: "set" prefix is usually for property setter in apple. consider using "updateLargeContentTitle"? also should it be more generic to support icon as well? or is it by design to show only the title?

harrieshin:
more of a design question, should there be a max font size for the card if width is very much fixed?

ios/FluentUI/Card/CardView.swift
alexanderboswell: I did not realize I could add the icon, updated with the icon now! the pr description has new screen shots of what it looks like.

alexanderboswell: I decided to separate into two helper methods so we only update the icon or title when the text or icon changes.

harrieshin (approved):

rdeassis (approved):

PR #891, opened 2022-02-04
[vNext Tokens] Allow for dynamic style/size switching with custom tokens

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Sorry folks, this looks like a big change, but it's largely mechanical, and should be well broken down by commit.

First, the problem: until now, control configurations that affect tokens (such as size or style) have been baked into each instance of a token set, with the expectation that if one changes, a new token set can be created. This works for most cases, but fails when a custom instance of tokens is attached to a single control. It's a single instance created by our client, so we can't recreate it!

The TL;DR solution:
- Token objects are no longer recreated when e.g. `style` or `size` change.
- Token values are now computed variables, instead of being calculated at `init`.
- Token objects maintain copies of `style` and `size`, and their parent `state` object pushes updates to these values as needed.

Also I got rid of some now-unneeded code:
- No more `defaultTokens` now that all token classes share a common `required init` func.
- No more `resolveTokenModifier`, since individual properties no longer trigger token regeneration.

One of the consequences of this change is that token objects no longer have elaborate initializers with overrides. Instead, clients are expected to subclass them and override individual variables as needed. This does mean that every token value is publicly readable, which presumably adds some binary size. However, the switch from lazy stored properties to computed properties brings surprisingly large size benefits, meaning this change ultimately *reduces* binary size! See table below.

So what changes will you see here? Let's break it down.
- Each `XYZTokens.swift` file will drop their complex `init`, and convert all properties from `lazy internal` stored properties to `open` computed ones.
- Each `MSF***Impl` class will set the weak `state` callback after calling `super.init()`, and adds a `didSet` to the `tokens` property to keep that property set any time custom tokens get involved.

Oh, and size changes!

| Before | After | Difference |
|--------|-------|------------|
| 31,210,974 bytes | 31,197,310 bytes | 13,664 bytes, or 13.3 kB smaller |

Not bad, especially considering that size reduction was not the goal of this PR!

### Verification

For now, I went through and checked that all of the affected controls look and behave the same, including in their SwiftUI views (where applicable). This will be much more fun to demonstrate once it interacts with
- #879

In the meantime, here's a demo of buttons with a custom token set changing their state on press:

https://user-images.githubusercontent.com/4934719/152460709-f49157e7-ac8c-42e7-9497-5521bb05e187.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/891)


ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
```diff
@@ -16,40 +16,22 @@ import UIKit
...
-        if let side = side {
-            self.side = side
-        }
-        if let thickness = thickness {
-            self.thickness = thickness
-        }
+open class ActivityIndicatorTokens: ControlTokens {
+    /// MSFActivityIndicatorSize enumeration value that will define pre-defined values for side and thickness.
+    public var size: MSFActivityIndicatorSize {
+        guard let state = state else { preconditionFailure() }
```
huwilkes, line 22: This seems like a very odd way to treat an optional variable. I think I understand that state is weak to try and avoid retention cycles, but it doesn't seem like it should be a weak variable if we're going to try and crash when the value isn't there?

huwilkes: Maybe we could change one of the references to unowned? Seems like an unowned reference in the state impl should also work.

mischreiber: Ah hm I think I can make that work using a `!` reference. Thanks for the idea!

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
```diff
@@ -143,8 +145,10 @@ class MSFActivityIndicatorStateImpl: NSObject, ObservableObject, ControlConfigur
 
     init(size: MSFActivityIndicatorSize) {
         self.size = size
-        self.tokens = ActivityIndicatorTokens(size: size)
 
         super.init()
+
+        // Ensure `tokens` has a weak reference back to this object to fetch `size` and `style`.
+        self.tokens.state = self
```
huwilkes, line 152: Now that we're using unowned instead of weak, we should probably update the comments as well

ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBarTokens.swift
```diff
@@ -9,32 +9,14 @@ import UIKit
...
-    /// Creates a new instance of the IndeterminateProgressBarTokens.
-    /// - Parameters:
-    ///   - backgroundColor: Optional override value of the background color.
-    ///   - gradientColor: Optional override value of the gradient color.
-    ///   - height: Optional override value of the height.
-    public init(backgroundColor: DynamicColor? = nil,
-                gradientColor: DynamicColor? = nil,
-                height: CGFloat? = nil) {
-        super.init()
+    weak public var state: MSFIndeterminateProgressBarState?
```
huwilkes, line 12: unowned?

ios/FluentUI/PersonaButton/PersonaButtonTokens.swift
```diff
@@ -33,65 +33,16 @@ import UIKit
...
+        return state.buttonSize
     }
 
-    var size: MSFPersonaButtonSize
+    unowned var state: MSFPersonaButtonState!
 
     // MARK: - Design Tokens
 
+    /// The amount of space between the control's `Avatar` and text labels.
     lazy var avatarInterspace: CGFloat = {
```
huwilkes, line 46: open?

ios/FluentUI/PersonaButtonCarousel/PersonaButtonCarouselTokens.swift
```diff
@@ -9,25 +9,15 @@ import UIKit
...
+        return state.buttonSize
     }
 
-    var size: MSFPersonaButtonSize
+    unowned var state: MSFPersonaButtonCarouselState!
 
     // MARK: - Design Tokens
 
+    /// The background color for the `PersonaButtonCarousel`.
     lazy var backgroundColor: DynamicColor = aliasTokens.backgroundColors[.neutral1]
```
huwilkes, line 22: Should this also be open?

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
mischreiber: Will do! I'll also make sure all the comments reference the correct variables.

ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBarTokens.swift
mischreiber: Don't actually need this reference at all!

ios/FluentUI/PersonaButton/PersonaButtonTokens.swift
mischreiber: Good catch, missed a set

huwilkes (approved):

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
```diff
@@ -16,40 +16,21 @@ import UIKit
...
-            self.thickness = thickness
-        }
+open class ActivityIndicatorTokens: ControlTokens {
+    /// MSFActivityIndicatorSize enumeration value that will define pre-defined values for side and thickness.
+    public var size: MSFActivityIndicatorSize {
+        return state.size
     }
 
-    let size: MSFActivityIndicatorSize
+    unowned var state: MSFActivityIndicatorState!
```
rdeassis, line 25: Using and unowned var to point back to the state seems dangerous as it's tightly coupled to the lifetime of the state object.

Why can't we just store the style/size properties directly in the tokens object and query or set them through the state object?

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
```diff
@@ -16,40 +16,21 @@ import UIKit
...
-        if let side = side {
-            self.side = side
-        }
-        if let thickness = thickness {
-            self.thickness = thickness
-        }
+open class ActivityIndicatorTokens: ControlTokens {
+    /// MSFActivityIndicatorSize enumeration value that will define pre-defined values for side and thickness.
+    public var size: MSFActivityIndicatorSize {
+        return state.size
```
rdeassis, line 22: This points to ta property defined in the state class and it's of a type that is defined in the Tokens class.

I think it makes a better cohesion story to keep this all in the Tokens space and have the state use it to prevent the cross-dependency between them.

ios/FluentUI/Command Bar/CommandBarTokens.swift
```diff
@@ -16,67 +16,28 @@ public struct ButtonDynamicColors {
...
-                            groupBorderRadius: CGFloat? = nil,
-                            groupInterspace: CGFloat? = nil,
-                            itemBackgroundColor: ButtonDynamicColors? = nil,
-                            itemFixedIconColor: DynamicColor? = nil,
-                            itemIconColor: ButtonDynamicColors? = nil,
-                            itemInterspace: CGFloat? = nil) {
+    open var groupBorderRadius: CGFloat { globalTokens.borderRadius[.xLarge] }
 
-        self.init()
+    open var groupInterspace: CGFloat { globalTokens.spacing[.medium] }
```
rdeassis, line 23: What would be the story if the UIKit control like the CommandBar had a style or size enumeration?
Because it doesn't have a state object like the SwiftUI controls, are you proposing that the tokens object would have an unowned reference to the control?

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -76,8 +77,11 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
     @Published var style: MSFButtonStyle
 
     @Published var overrideTokens: ButtonTokens?
-    @Published var tokens: ButtonTokens
-    var defaultTokens: ButtonTokens { .init(style: self.style, size: self.size) }
+    @Published var tokens: ButtonTokens = .init() {
```
rdeassis, line 80: Have you tested switching the style in other way than the video in the description?

The reason I bring it up is because tapping the button causes the SwiftUI refresh as it needs to render state changes like the pressed state. 

It may not be updating because of the tokens/style changes, but because of the action of tapping it.

It's something worth validating just to make sure.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
```diff
@@ -105,7 +105,7 @@ class ThemingDemoController: DemoController {
 
         addRow(items: [customThemeAvatarAccent.view, customThemeAvatarDefault.view, customThemeAvatarOutlinedPrimary.view], itemSpacing: 20)
 
-        customCardNudge.state.overrideTokens = customCardNudgeTokens
+        customCardNudge.state.overrideTokens = CustomCardNudgeTokens()
```
rdeassis, line 108: Very likely that this PR didn't introduce the regression, but the buttons on the "Theme overriding" part are not purple and square anymore. Are you aware of what's causing it?

Screen Shot 2022-02-07 at 11 24 41 AM

ios/FluentUI/Vnext/Button/Button.swift
mischreiber: Great point! I validated with a fun new bit of code in the `ButtonDemoController`:

```swift
    private func dequeueDemoButton(indexPath: IndexPath,
                                   style: MSFButtonStyle,
                                   size: MSFButtonSize,
                                   disabled: Bool) -> MSFButton {
        let key = "\(indexPath)-\(disabled)"
        if let button = buttons[key] {
            return button
        } else {
            let button = MSFButton(style: style,
                                   size: size) { [weak self] _ in
                self?.buttons.values.forEach { aButton in
                    aButton.state.style = (aButton.state.style == .secondary) ? .primary : .secondary
                }
                self?.didPressButton()
            }
            button.state.isDisabled = disabled
            button.state.overrideTokens = self.overrideTokens
            buttons[key] = button

            return button
        }
    }
```

And here's the successful result:

https://user-images.githubusercontent.com/4934719/152887018-aa1e753c-8197-4fae-8d67-3074e91bba0e.mp4

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
mischreiber: I think I can get behind that argument, but I don't think we can just have `state` refer to the value in its `tokens`, as that would mean that `style` and `size` would reset any time a client provided custom tokens, either directly or through the current theme. So we'll need to maintain two sets of values and do a `didSet` dance in the `state` class to keep them in sync.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokens.swift
mischreiber: Done!

ios/FluentUI/Command Bar/CommandBarTokens.swift
mischreiber: With the new approach, it would just push values. No more callbacks!

ios/FluentUI.Demo/FluentUI.Demo/Demos/ThemingDemoController.swift
mischreiber: Looks like an oversight; it lost its custom theming in #846. I'll add it back now!

mischreiber: Update: the change will likely be a little more complex, so I'll delay for now and open a new PR after this one merges.

rdeassis (approved):

PR #894, opened 2022-02-04
Update Avatar and Avatar Group RTL checks

huwilkes:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

I was having trouble figuring out why RTL wasn't working for me when it was working for others. Turns out, they were using Arabic as their language, while I was using the Right-to-Left pseudolanguage. Upon further investigation, I found that the check we were using for if the view was RTL was only checking the direction of the characters, so it didn't work for the RTL pseudolanguage.

There's still an issue with the padding on the Avatar Group being flipped for border vs. no border, but I'll address that in a separate PR.

### Verification

| AvatarRTL_Before| AvatarRTL_After |
| AvatarGroupRTL_Before | AvatarGroupRTL_After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/894)


mischreiber (commented):
I notice that we have only one more control using the `isRightToLeftLayoutDirection()` extension to `Locale`: `IndeterminateProgressBar`. Should we just convert that one too and then delete the extension?

huwilkes:
@mischreiber 
Github doesn't seem to want to let me quote reply, but we should handle the IndeterminateProgressBar in a different PR. I tried to convert it for a little while, but the usage of isRightToLeftLayoutDirection() being in the private Constants struct complicated things.

rdeassis:
> @mischreiber Github doesn't seem to want to let me quote reply, but we should handle the IndeterminateProgressBar in a different PR. I tried to convert it for a little while, but the usage of isRightToLeftLayoutDirection() being in the private Constants struct complicated things.

I think you can create a struct inside the constants struct that takes a boolean param for "isRTL" and return the coordinate points based on that flag.

```
    private struct Constants {
        static let animationDuration: Double = 1.75

        struct Coordinates {
            var isRTLLanguage: Bool
            var initialStartPoint: UnitPoint { isRTLLanguage ? UnitPoint(x: 1, y: 0.5) : UnitPoint(x: -1, y: 0.5) }
            var initialEndPoint: UnitPoint { isRTLLanguage ? UnitPoint(x: 2, y: 0.5) : UnitPoint(x: 0, y: 0.5) }
            var finalStartPoint: UnitPoint { isRTLLanguage ? UnitPoint(x: -1, y: 0.5) : UnitPoint(x: 1, y: 0.5) }
            var finalEndPoint: UnitPoint { isRTLLanguage ? UnitPoint(x: 0, y: 0.5) : UnitPoint(x: 2, y: 0.5) }

            init(_ isRTLLanguage: Bool = false) {
                self.isRTLLanguage = isRTLLanguage
            }
        }
    }
```

Then just replace that throughout the SwiftUI implementation.

```
    @State var startPoint: UnitPoint = .zero
    @State var endPoint: UnitPoint = .zero
    var isRTLLanguage: Bool {
        return layoutDirection == .rightToLeft
    }

    /// Creates the Indeterminate Progress Bar.
    public init() {
        let state = MSFIndeterminateProgressBarStateImpl()
        self.state = state
        self.tokens = state.tokens
        startPoint = Constants.Coordinates(isRTLLanguage).initialStartPoint
        endPoint = Constants.Coordinates(isRTLLanguage).initialEndPoint
    }

...

    private func startAnimation() {
        stopAnimation()

        withAnimation(Animation.linear(duration: Constants.animationDuration)
                                .repeatForever(autoreverses: false)) {
            startPoint = Constants.Coordinates(isRTLLanguage).finalStartPoint
            endPoint = Constants.Coordinates(isRTLLanguage).finalEndPoint
        }
    }

    private func stopAnimation() {
        withAnimation(Animation.linear(duration: 0)) {
            startPoint = Constants.Coordinates(isRTLLanguage).initialStartPoint
            endPoint = Constants.Coordinates(isRTLLanguage).initialEndPoint
        }
    }
```

PR #895, opened 2022-02-08
Fix RTL Avatar Group spacing

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Update the RTL cutout origin to be based off the size of the cutout rather than the size of the current avatar

### Verification
| AvatarGroup_RTLBefore | AvatarGroup_RTLAfter |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/895)

### DIFF
--- a/ios/FluentUI/AvatarGroup/AvatarGroup.swift
+++ b/ios/FluentUI/AvatarGroup/AvatarGroup.swift
@@ -208,9 +208,12 @@ public struct AvatarGroup: View {
                 let stackPadding = (currentAvatarHasRing ? ringPaddingInterspace : noRingPaddingInterspace)
 
                 let cutoutSize = isLastDisplayed ? (ringOuterGap * 2) + imageSize : nextAvatarSize
-                let xPosition = currentAvatarHasRing ? x - ringOuterGap - ringOuterGap : x - ringOuterGap
-                let xPositionRTL = -cutoutSize - interspace + ringOuterGap + (currentAvatarHasRing ? ringOffset : 0)
-                let xOrigin = layoutDirection == .rightToLeft ? xPositionRTL : xPosition
+                let xOrigin: CGFloat = {
+                    if layoutDirection == .rightToLeft {
+                        return -cutoutSize - interspace + ringOuterGap + (currentAvatarHasRing ? ringOffset : 0)
+                    }
+                    return x - ringOuterGap - (currentAvatarHasRing ? ringOuterGap : 0)
+                }()
                 let yOrigin = sizeDiff / 2
 
                 // Hand the rendering of the avatar to a helper function to appease Swift's
--- a/ios/FluentUI/AvatarGroup/AvatarGroup.swift
+++ b/ios/FluentUI/AvatarGroup/AvatarGroup.swift
@@ -205,15 +205,13 @@ public struct AvatarGroup: View {
 
                 let ringPaddingInterspace = nextAvatarHasRing ? interspace - (ringOffset + ringOuterGap) : interspace - ringOffset
                 let noRingPaddingInterspace = nextAvatarHasRing ? interspace - ringOuterGap : interspace
-                let rtlRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOuterGap : -x + ringOffset)
-                let rtlNoRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOffset - ringOuterGap : -x)
                 let stackPadding = (currentAvatarHasRing ? ringPaddingInterspace : noRingPaddingInterspace)
 
+                let cutoutSize = isLastDisplayed ? (ringOuterGap * 2) + imageSize : nextAvatarSize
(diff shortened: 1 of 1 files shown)


ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -205,15 +205,13 @@ public struct AvatarGroup: View {
 
                 let ringPaddingInterspace = nextAvatarHasRing ? interspace - (ringOffset + ringOuterGap) : interspace - ringOffset
                 let noRingPaddingInterspace = nextAvatarHasRing ? interspace - ringOuterGap : interspace
-                let rtlRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOuterGap : -x + ringOffset)
-                let rtlNoRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOffset - ringOuterGap : -x)
                 let stackPadding = (currentAvatarHasRing ? ringPaddingInterspace : noRingPaddingInterspace)
 
+                let cutoutSize = isLastDisplayed ? (ringOuterGap * 2) + imageSize : nextAvatarSize
                 let xPosition = currentAvatarHasRing ? x - ringOuterGap - ringOuterGap : x - ringOuterGap
```
harrieshin, line 211: to be consistent with your changes in line 210, should it be ringOuterGap *2?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -205,15 +205,13 @@ public struct AvatarGroup: View {
...
-                let xPositionRTL = currentAvatarHasRing ? rtlRingPaddingInterspace : rtlNoRingPaddingInterspace
+                let xPositionRTL = -cutoutSize - interspace + ringOuterGap + (currentAvatarHasRing ? ringOffset : 0)
```
harrieshin, line 212: seems little weird we would have to compute xpostitionrtl even thought we might not need it at all? can we optimize how we calculate xOrigin?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
harrieshin: also for readability be consistent when you add -(currentavatarhasring? ringoutergap : 0 )?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
rdeassis: @huwilkes how about initializing the xOrigin with a closure like below:

```
let xOrigin = {
    if layoutDirection == .rightToLeft {
        return <calculated value for RTL>
    }

    return <calculated value for LTR>
}()
```

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -205,15 +205,13 @@ public struct AvatarGroup: View {
 
                 let ringPaddingInterspace = nextAvatarHasRing ? interspace - (ringOffset + ringOuterGap) : interspace - ringOffset
                 let noRingPaddingInterspace = nextAvatarHasRing ? interspace - ringOuterGap : interspace
-                let rtlRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOuterGap : -x + ringOffset)
-                let rtlNoRingPaddingInterspace = (nextAvatarHasRing ? -x - ringOffset - ringOuterGap : -x)
                 let stackPadding = (currentAvatarHasRing ? ringPaddingInterspace : noRingPaddingInterspace)
 
+                let cutoutSize = isLastDisplayed ? (ringOuterGap * 2) + imageSize : nextAvatarSize
```
rdeassis, line 210: This needs to consider 3 tokens for positioning the avatar:
- ringInnerGap
- ringThickness
- ringOuterGap

It may work today because the values are identical, but it should consider the tokens independently.

rdeassis: And I believe this needs to be fixed in the other lines too.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
huwilkes: I was trying to keep the changes for this PR minimal and come back with another PR to update the other lines of the body, but I can update this one now, especially if I update it with rdeassis' suggestion.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
huwilkes: I think for this line ringOuterGap is the value we want. I think the idea here is to say that the last avatar is the overflow, and so it should be ringOuterGap + imageSize + ringOuterGap

rdeassis (approved):

PR #896, opened 2022-02-09
[vNext Tokens] Update NotificationView

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

Converts `NotificationView` to use the new tokens.

Made background color clear by default to prevent a black rectangular border around rounded controls
- ios/FluentUI/Core/ControlHostingContainer.swift

Added extension files in order to get the window width and edge insets for the SwiftUI notification
- ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
- ios/FluentUI/Extensions/UIApplication+Extensions.swift

Update files to use new tokenization
- ios/FluentUI/Notification/NotificationTokens.swift
- ios/FluentUI/Notification/NotificationView.swift
- ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift

Update to conform to ControlHostingContainer which fixes a bug where the SwiftUI notification wouldn't update with the theme
- ios/FluentUI/Vnext/Notification/MSFNotificationView.swift

Deleted as they are not needed for new tokenization
- ios/FluentUI/Notification/MSFNotificationTokens.swift
- tools/sgen/output/MSFNotificationTokens.generated.swift
- ios/FluentUI/Notification/MSFNotificationTokens.generated.swift

Removed Notification tokens
- tools/sgen/input/FluentUIStyle.yml

### Verification

Made sure Notifications looked the same (or in this case, better, given the bug) before and after

|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-09 at 00 00 55|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-08 at 23 55 30|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-09 at 00 00 27| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-08 at 23 55 56|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-09 at 00 00 57| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-08 at 23 55 32|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-09 at 00 00 33|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-02-08 at 23 55 57|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/896)


ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -161,9 +164,9 @@ public struct NotificationViewSwiftUI: View {
     }
 
     public var body: some View {
-        // Note: we are using viewController's width because GeometryReader relies on the bounds of the hosting controller which cannot be set accurately at this point
-        if let windowWidth = windowProvider?.window?.bounds.width {
-            let width = windowWidth - safeAreaInsets.left - safeAreaInsets.right
+        // Note: we are using keyWindow's width because GeometryReader relies on the bounds of the hosting controller which cannot be set accurately at this point
+        if let windowWidth = UIApplication.shared.keyWindow?.bounds.width {
```
jeaniehuynh, line 168: Note: I know we spent time trying to move away from this, but since MSFNotification conforms to ControlHostingContainer now, the windowProvider can't be set (error: 'self' used before 'super.init' call). The pure SwiftUI implementation I'll be working on after this PR will be replacing this with GeometryReader, so it's just a temporary fix.

ios/FluentUI/Notification/NotificationView.swift
```diff
@@ -361,10 +356,38 @@ open class NotificationView: UIView, FluentUIWindowProvider {
...
+
+    var state: NotificationView { self }
+
+    /// Design token set for this control, to use in place of the control's default Fluent tokens.
+    var overrideTokens: NotificationTokens?
+
+    /// Style to draw the control.
+    public var style: MSFNotificationStyle {
+        didSet {
+            tokens.style = style
```
mischreiber, line 378: Do you need to call `updateForStyle()` here as well?

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -0,0 +1,24 @@
...
+import SwiftUI
+
+struct SafeAreaInsetsKey: EnvironmentKey {
+    static var defaultValue: EdgeInsets {
+        return UIApplication.shared.keyWindow?.safeAreaInsets.swiftUIInsets ?? EdgeInsets()
+    }
+}
+
+extension EnvironmentValues {
+  var swiftUIInsets: EdgeInsets {
```
mischreiber, line 14: Nit: whitespace is off (2 spaces instead of 4).

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -0,0 +1,24 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+import SwiftUI
```
mischreiber, line 5: Thinking about the name of this file. We have a few other places where we extend EnvironmentValues, and it's usually done in the context of what we're intending to add to the app (e.g. `FluentTheme.swift`).

Do you think we should move those all here to a single common `EnvironmentValues+Extensions.swift` file, or should this particular extension be renamed to something more explicit about its intent (e.g. `EnvironmentValues+SafeAreaInsets` or the like)?

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
```diff
@@ -0,0 +1,24 @@
...
+extension EnvironmentValues {
+  var swiftUIInsets: EdgeInsets {
+    get { self[SafeAreaInsetsKey.self] }
+    set { self[SafeAreaInsetsKey.self] = newValue }
+  }
+}
+
+private extension UIEdgeInsets {
+    var swiftUIInsets: EdgeInsets {
+        EdgeInsets(top: top, leading: left, bottom: bottom, trailing: right)
```
mischreiber, line 22: It's okay if this is a short-term patch, but long-term it should be pointed out that this is could cause issues in RTL languages. Leading should map to `right` and trailing to `left` in those cases.

Might be worth at least commenting to note this deficiency if we don't want to fix it right now.

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -0,0 +1,155 @@
...
+    private struct Constants {
+        static let animationDurationForShowToast: TimeInterval = 0.6
+        static let animationDurationForShowBar: TimeInterval = 0.3
+        static let animationDurationForHide: TimeInterval = 0.25
+        static let animationDampingRatioForToast: CGFloat = 0.5
+    }
+}
+
+public class NotificationTokens: ControlTokens {
+    public internal(set) var style: MSFNotificationStyle = .primaryToast
```
mischreiber, line 51: Could you provide quick one-line comments for each of the `public` and `open` vars here?

I know not all of the converted token classes have them, but we're definitely moving that direction. In the meantime check out `ActivityIndicatorTokens` for an example.

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -0,0 +1,155 @@
...
+
+    private struct Constants {
+        static let animationDurationForShowToast: TimeInterval = 0.6
+        static let animationDurationForShowBar: TimeInterval = 0.3
+        static let animationDurationForHide: TimeInterval = 0.25
+        static let animationDampingRatioForToast: CGFloat = 0.5
+    }
+}
+
+public class NotificationTokens: ControlTokens {
```
mischreiber, line 50: Class should be `open`, not `public`, so folks outside our module can subclass it.

Could you also add a one-liner of documentation here as well?

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -91,54 +91,52 @@ class NotificationViewDemoController: DemoController {
     }
 
     private func createNotificationView(forVariant variant: Variant) -> NotificationView {
-        let view = NotificationView()
         switch variant {
         case .primaryToast:
-            view.setup(style: .primaryToast,
+            return NotificationView.init(style: .primaryToast).setup(style: .primaryToast,
```
mischreiber, line 96: You can skip the `.init` part for each of these and just initialize with `NotificationView(style: .primaryToast)`.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -91,54 +91,52 @@ class NotificationViewDemoController: DemoController {
...
                               message: "Mail Archived",
```
mischreiber, line 97: Nit: could you re-align these lines? A quick `^I` should do the trick.

ios/FluentUI/Extensions/EnvironmentValues+Extensions.swift
mischreiber: Ah yes thanks for the reminder 🤠 I do still like the idea of it living separately! Just wondering about renaming it to be more specific to its contents.

ios/FluentUI/Notification/NotificationView.swift
jeaniehuynh: I think it actually only needs to be called in initialize()!

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -0,0 +1,182 @@
...
+        switch style.isToast {
+        case true:
+            return globalTokens.spacing[.medium]
+        case false:
+            return globalTokens.spacing[.none]
+        }
+    }
+
+    /// The value for the horizontal padding between the elements within a notification and its frame
+    open var horizontalPadding: CGFloat = 19.0
```
mischreiber, line 118: Ah sorry just caught these: all tokens have to be generated properties so they can be properly overridden. So this line should read
```swift
open var horizontalPadding: CGFloat { 19.0 }
```

Same for the values below.

mischreiber (approved):

PR #899, opened 2022-02-10
Improve readability of the AvatarGroup body

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Moved a bunch of variables around and tried to simplify some of the logic happening in the body of the AvatarGroup Body. Also removed some redundant logic from the avatar state impl totalSize method.

### Verification

| AvatarGroup_Body_LTRBefore | AvatarGroup_Body_LTRAfter |
| AvatarGroup_Body_RTLBefore | AvatarGroup_Body_RTLAfter |
| AvatarGroup_animation_before | AvatarGroup_animation_afterRebase |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/899)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -532,8 +532,8 @@ class MSFAvatarStateImpl: NSObject, ObservableObject, Identifiable, MSFAvatarSta
         if !isRingVisible {
             return avatarImageSize + (ringOuterGap * 2)
         } else {
-            let ringThickness: CGFloat = isRingVisible ? tokens.ringThickness : 0
-            let ringInnerGap: CGFloat = isRingVisible && hasRingInnerGap ? tokens.ringInnerGap : 0
+            let ringThickness: CGFloat = tokens.ringThickness
+            let ringInnerGap: CGFloat = hasRingInnerGap ? tokens.ringInnerGap : 0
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -519,8 +519,8 @@ class MSFAvatarStateImpl: NSObject, ObservableObject, MSFAvatarState {
         if !isRingVisible {
             return avatarImageSize + (ringOuterGap * 2)
```
huwilkes, line 520: Why isn't the avatar's size just image size when it doesn't have a ring? Also, I think it might just draw the avatars at this size, rather than the image size? If you draw a bunch of no ring avatars with 0 spacing, their edges touch, until the ring avatars where each one has an extra gap around it.

sophialee0416 (approved):

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: It was to align with when the avatars with ring have an extra gap, but these changes make sense!

huwilkes: Ah, makes sense

PR #903, opened 2022-02-11
[vNext Tokens] Convert PillButton to the new token system

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Convert the PillButton to the new token system. Also had a surprise change of adding an alternative TokenResolver/TokenizedControlInternal to deal with a collision with UIControl.state.

### Verification

| PillButton_before | PillButton_after |
| PillButton_before_dark | PillButton_after_dark |
| PillButton_toggle_before | PillButton_toggle_after |
| PillButton_toggle_before_dark | PillButton_toggle_after_dark |

### Pull request checklist

This PR has considered:


ios/FluentUI/Core/Theme/Tokens/UIControlTokenResolver.swift
```diff
@@ -0,0 +1,37 @@
...
+//
+
+import SwiftUI
+
+/// Internal extension to `TokenizedControl` that adds information about `configuration`.
+protocol TokenizedUIControlInternal: TokenizedControl {
+    associatedtype ConfigurationType: ControlConfiguration where ConfigurationType.TokenType == TokenType
+
+    /// Contains additional configuration information about the control.
+    var config: ConfigurationType { get }
```
huwilkes, line 13: Turns out that any control we make that inherits from UIControl isn't really allowed to override state.

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -9,7 +9,14 @@ import UIKit
...
+
+    open override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        updatePillButtonTokens()
+        updateAppearance()
+    }
 
     /// Set `backgroundColor` to customize background color of the pill button
     @objc open var customBackgroundColor: UIColor? {
```
huwilkes, line 22: Seems like maybe all of these custom*Colors should be moved in to the tokens?

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -111,25 +105,36 @@ open class PillButton: UIButton, FluentUIWindowProvider {
...
     }
 
-    public override func layoutSubviews() {
-        super.layoutSubviews()
-        updateUnreadDot()
+    @objc public static let cornerRadius: CGFloat = 16.0
+
+    @objc public let pillBarItem: PillButtonBarItem
+
+    @objc public let style: PillButtonStyle
```
mischreiber, line 112: I think you'll need to call `updatePillButtonTokens()` when this is set as well?

huwilkes: Even though it's a let? I don't think you can do didSet on a let. Should it change to a var? I don't think this is meant to change after initialization.

mischreiber: Ah good catch, totally missed that it's a `let`. Okay, if this can't change then it doesn't matter!

ios/FluentUI/Core/Theme/Tokens/UIControlTokenResolver.swift
mischreiber: Okay I'm prepping a PR with a workaround, but let's not block this work on that.

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -9,84 +9,43 @@ import UIKit
...
+        super.layoutSubviews()
+        updateUnreadDot()
+    }
+
+    public func overrideTokens(_ tokens: PillButtonTokens?) -> Self {
+        overrideTokens = tokens
+        return self
+    }
+
+    public typealias TokenType = PillButtonTokens
```
mischreiber, line 47: I don't believe this line should be necessary, even though Xcode sometimes suggests it is.

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -111,25 +70,36 @@ open class PillButton: UIButton, FluentUIWindowProvider {
...
+
+    @objc public let pillBarItem: PillButtonBarItem
+
+    @objc public let style: PillButtonStyle
+
+    var config: PillButton { self }
+    var tokens: PillButtonTokens = .init()
+    var overrideTokens: PillButtonTokens? {
+        didSet {
+            updatePillButtonTokens()
```
mischreiber, line 83: Should we also call `updateAppearance()` here as well?

ios/FluentUI/Pill Button Bar/PillButtonTokens.swift
```diff
@@ -5,101 +5,121 @@
...
-    var themeDidUpdate: (() -> Void)?
-
-    var style: PillButtonStyle
-
-    init(style: PillButtonStyle) {
-        self.style = style
+/// Represents the set of `DynamicColor` values for the various states of a `PillButton`
+public struct PillButtonDynamicColors {
+    public init (rest: DynamicColor,
+                 hover: DynamicColor? = nil,
```
mischreiber, line 11: Why are some of these nilable and some are not?

huwilkes: I think to have a functional customized pill button, you really only need to define the rest, selected, disabled, and selected disabled states. Hover, highlighted, and selectedHighlighted seem to just be copies of rest or selected, even in our definitions.

mischreiber: I see. Do you think it would be reasonable to just remove those colors entirely, especially since I don't see them called out in any specs? Or would that get too invasive?

mischreiber (approved):

ios/FluentUI/Pill Button Bar/PillButtonTokens.swift
huwilkes: I'm not opposed, but I think the fall out from that would more or less also remove the custom pointer interaction methods of the bar. Which I guess means they probably aren't doing too much at the moment. If it doesn't break pointer interaction, I can remove the extra colors.

huwilkes: Pointer interaction is still fine.

PR #906, opened 2022-02-15
Improve support for transient bottom sheets

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The bottom sheet was originally created to support a persistent scenario where the sheet controller is always installed in the hierarchy and the sheet is almost always visible. However we also want to support transient scenarios where the sheet controller is created and placed in the hierarchy on demand, and removed after the sheet is dismissed.

Fortunately there are only very minor changes necessary in `BottomSheetController`. Most of the work in this PR is to add an example of transient presentation to the demo controller. Note the bottom sheet can't use UIKit's presentation API because this is only suitable for modal scenarios, while our desired usage is better described as semi-modal - main content needs to be interactable and accessible. This can only be achieved by manually installing the sheet view and view controller in the hierarchy on demand.

Small fixes to improve transient sheet support:
- Pausing a `UIViewPropertyAnimator` was breaking the unhiding animation if it was started right after a layout pass (which is is likely to happen in a transient presentation). Looks like an Apple bug. Luckily we don't need to pause the animator at all.
- Transient sheets highlighted a bug when a pan starts during an unhiding animation - the sheet would disappear because the animation interruption would cause it to go back to the original state. Let's change the animation interruption logic to use the target expansion state as a fallback when an animation is interrupted.
- If we skip a state change animation because the sheet offset is already there, let's call the full completion handler to ensure delegates are notified and `.isHidden` state is where it should be

Demo controller changes:
- Add a button to show a transient sheet
- For a more modern look, switch the main table view to `.insetGrouped` style (also in `BottomCommandingDemoController` for consistency)

### Verification

https://user-images.githubusercontent.com/3610850/153987274-62464915-602e-4b0e-a3c4-88cd99ad9ea0.mov

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -11,7 +11,7 @@ class BottomCommandingDemoController: UIViewController {
     override func loadView() {
         view = UIView()
 
-        let optionTableViewController = UITableViewController(style: .plain)
+        let optionTableViewController = UITableViewController(style: .insetGrouped)
         mainTableViewController = optionTableViewController
 
         let optionTableView: UITableView = optionTableViewController.tableView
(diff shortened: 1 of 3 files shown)


noaMSFT (approved):
Looks good to me! Thanks! :)

mischreiber (approved):

PR #909, opened 2022-02-15
[vNext Tokens] Separate `TokenizedControlInternal` and `TokenizedControlConfigurable`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

PR #903 uncovered a problem with `TokenizedControlInternal`, the internal protocol used to describe tokenizable controls. We discovered that its mandatory `state` property conflicted with `UIControl.state`, which some of our controls inherit from.

Honestly I never really liked having that property be mandatory either, so this PR kills two birds with one stone by eliminating its need. Instead, `TokenizedControlInternal` now has two abstract, descriptive values: a readable `var overrideTokens`, and an actionable `func updateCurrentTokens()`.

In order to minimize code duplication, I also created an `extension` to this protocol, `TokenizedControlConfigurable`, which provides a common implementation for all SwiftUI controls to freely inherit.

### Verification

No real changes to appearance; this is just shuffling some methods around.

I did measure a small package size increase (~4kB), but considering how many controls are covered in one fell swoop, it seems like a reasonable price to pay.

In addition, while we're waiting for #879 to complete, I reused an old snippet of test code in `ButtonDemoController` to verify that token swapping was unaffected by this change:

https://github.com/microsoft/fluentui-apple/pull/891#discussion_r801133476

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/909)


ios/FluentUI/Core/Theme/Tokens/TokenizedControlConfigurable.swift
```diff
@@ -0,0 +1,36 @@
...
+
+    /// Custom design token set for this control, to use in place of the control's default Fluent tokens.
+    var overrideTokens: TokenType? { get set }
+
+    /// Common token lookup method, which should contain a lookup via the current environment's `fluentTheme`.
+    var tokens: TokenType { get set }
+}
+
+/// SwiftUI-specific extension to `TokenizedControl` that adds information about `configuration`.
+protocol TokenizedControlConfigurable: TokenizedControlInternal {
```
huwilkes, line 20: Maybe `ConfigurableTokenizedControl` might make more sense? I think Configurable being at the end makes it sound like a noun rather than an adjective. But then I think that also suggests that the other controls (UIKit) aren't configurable? Which isn't really the case. I guess a more boring but explicit name could be "TokenizedControlWithConfiguration", but I'm not a huge fan.

ios/FluentUI/Core/Theme/Tokens/TokenizedControlConfigurable.swift
```diff
@@ -0,0 +1,36 @@
...
+    associatedtype ConfigurationType: ControlConfiguration where ConfigurationType.TokenType == TokenType
+
+    /// Contains additional configuration information about the control.
+    var state: ConfigurationType { get }
```
harrieshin, line 24: should it still be "state"?

mischreiber: Fair question. I could switch it to `configuration` now, but that's gonna be a big, boring change that touches dozens of files across everything. I'm worried that the noise of that change will drown out the signal of what I want to be reviewed here: the splitting-up of the `TokenizedControlInternal` protocol.

So ideally I'd prefer to have a separate PR that is just that single, mechanical change-over.

ios/FluentUI/Core/Theme/Tokens/TokenizedControlConfigurable.swift
mischreiber: 🤔

huwilkes (approved):

huwilkes:
I merged #903, so we can update the pill button to this and remove the work around I added

ios/FluentUI/Core/Theme/Tokens/TokenizedControlConfigurable.swift
mischreiber: Going with `ConfigurableTokenizedControl`!

huwilkes (approved):

PR #910, opened 2022-02-16
[vNext-tokens] Updating List & ListHeader to use new token system

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated List & List Header controls to use the new design token system, along with Header style naming changes.

### Verification

| Simulator Screen Shot - iPhone 11 Pro - 2022-02-16 at 12 06 27 | Simulator Screen Shot - iPhone 11 Pro - 2022-02-16 at 12 05 46 |

Theme validation:

https://user-images.githubusercontent.com/22566866/154347536-676e9826-d5ec-45f2-8ba2-43f9ccf996f2.mov

### Pull request checklist

This PR has considered:


ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -737,9 +735,8 @@
 				53097D152702890900A6E4DC /* ListCell.swift */,
 				53097D132702890800A6E4DC /* ListHeaderFooter.swift */,
 				53097D112702890800A6E4DC /* ListTokens.swift */,
-				53097D122702890800A6E4DC /* MSFHeaderFooterTokens.generated.swift */,
 				53097D102702890800A6E4DC /* MSFListCellTokens.generated.swift */,
-				53097D142702890800A6E4DC /* MSFListTokens.generated.swift */,
+				0A3CD0B027BCA36000DBE728 /* HeaderFooterModifiers.swift */,
```
huwilkes, line 739: Could we move this up so it's in alphabetical order?

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,23 +6,24 @@
 import UIKit
 import SwiftUI
 
-struct Header: View {
-    @Environment(\.theme) var theme: FluentUIStyle
-    @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
-    @ObservedObject var tokens: MSFHeaderFooterTokens
-    @ObservedObject var state: MSFListSectionStateImpl
-
+struct Header: View, TokenizedControlInternal {
```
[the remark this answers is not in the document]
sophialee0416: Good point! Originally, the HeaderFooter control was all in one. We do not have a Footer control as of right now but the name stuck. Maybe I can use this as an opportunity to rename all HeaderFooters into just Header?

ios/FluentUI/Vnext/List/HeaderFooterTokens.swift
```diff
@@ -8,61 +8,58 @@ import SwiftUI
...
-    case headerSecondary
+    case standard
+    case subtle
 }
 
-class MSFHeaderFooterTokens: MSFTokensBase, ObservableObject {
-    @Published public var backgroundColor: UIColor!
-    @Published public var textColor: UIColor!
+open class MSFHeaderFooterTokens: ControlTokens {
+    /// The MSFHeaderFooter style. Currently we only have primary and secondary styles.
```
huwilkes, line 16: I'm not sure we need to list out which styles we have available, but we have standard and subtle, not primary and secondary.

ios/FluentUI/Vnext/List/HeaderFooterTokens.swift
```diff
@@ -8,61 +8,58 @@ import SwiftUI
...
 
-    @Published public var headerHeight: CGFloat!
-    @Published public var topPadding: CGFloat!
-    @Published public var leadingPadding: CGFloat!
-    @Published public var bottomPadding: CGFloat!
-    @Published public var trailingPadding: CGFloat!
+    // MARK: - Design Tokens
 
-    @Published public var textFont: UIFont!
+    ///  The background color of the List Header
```
huwilkes, line 21: Super nit: most of our comments have a period at the end

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -93,10 +72,9 @@ import SwiftUI
     func removeSection(at index: Int)
 }
 
-public struct MSFListView: View {
+public struct MSFListView: View, TokenizedControlInternal {
```
huwilkes, line 75: Maybe outside the scope of this PR, but usually our actual SwiftUI controls don't have the MSF prefix. If there's a collision with the SwiftUI List control, then the name should probably be FluentList instead, like we've done with the Button and Divider.

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
```diff
@@ -6,23 +6,24 @@
...
         self.state = state
-        self.tokens = state.headerTokens
     }
 
     var body: some View {
+        let backgroundColor: Color = {
+            guard let stateBackgroundColor = state.backgroundColor else {
+                return Color(dynamicColor: tokens.backgroundColor)
+            }
+            return Color(stateBackgroundColor)
```
huwilkes, line 19: Do we still need the background color in the state, or could we just move it to the tokens and get custom colors from custom tokens?

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -156,22 +130,36 @@ public struct MSFListView: View {
...
 }
 
 /// Properties that make up section content
-class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, MSFListSectionState {
-    var headerTokens: MSFHeaderFooterTokens
-    var id = UUID()
+class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, ControlConfiguration, MSFListSectionState {
+    init(style: MSFHeaderFooterStyle = .standard) {
+        let tokens = MSFHeaderFooterTokens()
+        tokens.style = style
```
huwilkes, line 144: I think you don't need `tokens.style = style` here since we reset `tokens.style` every time tokens or style gets updated.

mischreiber: Actually this is necessary, because this line defines the default starting point for `tokens`. It's possible that neither `tokens` nor `style` will be updated from this point on, so having `tokens` start with the correct `style` is crucial.

mischreiber:
The headers for "Single line cell" and "Double line cell" don't seem to match across screenshots (though they were working when we debugged earlier). Possibly a regression from any cleanup you've done since then? Or just an out-of-date screenshot?

ios/FluentUI/Vnext/List/List.swift
huwilkes: We set both tokens and style in the next three lines.

mischreiber: Oh I see what you are saying. So it's important to note that `didSet` is _not_ called during `init()`, since it is written with the assumption that the object in question has been fully initialized (which is not the case during `init`).

huwilkes: Ah, gotcha. That makes sense.

ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
```diff
@@ -128,9 +128,9 @@ class LeftNavMenuViewController: UIViewController {
 
     private var persona = MSFPersonaView()
 
-    private var leftNavMenuList = MSFList()
+    private var leftNavMenuList = FluentList()
 
-    private var statusCell: MSFListCellState?
+    private var statusCell: FluentListCellState?
```
huwilkes, line 133: I don't think MSFListCellState needed to change to FluentListCellState. If you look at the vNext button, we have FluentButton and MSFButtonStateImpl. We also have FluentButtonStyle, but maybe there was a naming collision on ButtonStyle? Anyway, I think most of the names were okay as MSF*

sophialee0416: Hm okay, so just the ones involving the big List, and excluding anything stemming from List? (Cell, Header, etc)

huwilkes: Yeah, that sounds right to me. The actual view / control shouldn't be MSF, but everything else is fine. So, in this case, I think we should have MSFList (currently FluentList) and FluentList (currently FluentListView), since what's currently names FluentListView is the actual SwiftUI control, while what's currently FluentList is just the hosting container. Relatedly, MSFList should get moved to its own file to match the other controls.

huwilkes (approved):

ios/FluentUI/Vnext/List/HeaderTokens.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+import SwiftUI
+
+/// HeaderFooter styles.
```
huwilkes, line 9: Header styles

ios/FluentUI/Vnext/List/HeaderTokens.swift
```diff
@@ -0,0 +1,65 @@
...
+@objc public enum HeaderStyle: Int, CaseIterable {
```
mischreiber, line 10: We need to be careful about how we rename some of these enums, particularly regarding how they are projected into Objective-C. We need to keep our three-letter `MSF` prefix for anything Objective-C compatible, because that's our only differentiation from other symbols (there is no proper namespacing in ObjC).

sophialee0416: Added them back in!

ios/FluentUI/Vnext/List/ListHeaderFooter.swift
sophialee0416: Resolved offline, will be removing backgroundColor in a future PR and it'll be controlled by custom tokens

mischreiber (approved):

PR #915, opened 2022-02-17
Move bottom sheet animator to active state before stopping it

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

TL;DR: App crashes during un-animated state changes because we are trying to stop an inactive animator.

In a previous change (2aaf64b8814a77a0ff21dd589dd4c7b24aa3577d) I removed code which paused the sheet state change animator before returning it. There are cases in which we skip to the end of an animation right after creating it, but `UIViewPropertyAnimator` is strict about needing to be in active state before being stopped. Pausing it used to move it to active state. To fix this, let's start the animator right before stopping it. This doesn't actually cause any animations to run, it just puts it in the right state to be stopped.

### Verification

Sanity checks that un-animated state changes work as expected.
Verify transient sheet animations still work as expected (which was the reason for the animator pause removal).

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/915)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -647,7 +647,7 @@ public class BottomSheetController: UIViewController {
                 }
                 animator.startAnimation()
             } else {
-                animator.startAnimation()
+                animator.startAnimation() // moves the animator into active state so it can be stopped
                 animator.stopAnimation(false)
                 animator.finishAnimation(at: .end)
             }
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -647,6 +647,7 @@ public class BottomSheetController: UIViewController {
                 }
                 animator.startAnimation()
             } else {
+                animator.startAnimation()
                 animator.stopAnimation(false)
                 animator.finishAnimation(at: .end)
             }


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -647,6 +647,7 @@ public class BottomSheetController: UIViewController {
                 }
                 animator.startAnimation()
             } else {
+                animator.startAnimation() // moves the animator into active state so it can be stopped
```
harrieshin, line 650: i am not sure i understand the bug. if the animation was nver started why can we check if it is running and only stop when we need to?

lcapkovic: The thing we want to achieve in this else branch is to skip to the end of the animation. Because of how `UIViewPropertyAnimator` is designed, to skip to the end you first have to move it to the active state (`startAnimation`), then stop it (`stopAnimation`) and then call finish to ensure all completion handlers are called.

harrieshin: do you mind sharing the video of the bug? i still having hard time understnading the issue.

lcapkovic: There used to be a `pauseAnimation` call at the end of our animator factory, which would move it to active state. That's why the bug didn't repro then.

lcapkovic: It just crashes when you try to stop an inactive animator. This happens when you for example change the `isExpandable` property after loadView happened, because that causes the sheet to reset to collapsed state.

lcapkovic: You can actually get a repro by just flipping the "Expandable" switch in the bottom sheet demo controller.

lcapkovic: Sorry, I should've been more clear in the description :D I added a tldr on top.

harrieshin (approved):

PR #924, opened 2022-02-23
[vNext Tokens] Remove `TokenResolver`, add `resolvedTokens` and restore `defaultTokens`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change updates how we fetch tokens in our controls one last time.

#### Problem: 
- Certain controls like `FluentButton` have multiple sub-structures that reference design tokens.
- These substructures do not conform to any specific external controlling protocol, so they are not directly hooked up to the `TokenResolver`.
- As such, they have not been receiving proper updates when tokens or themes change.

Result: the pure SwiftUI implementation of `FluentButton` did not properly update themes once they were injected via the `.fluentTheme(_:)` modifier.

#### Solution:
- Have all token resolution happen on the fly via a new function, `tokens(for:)`.
- Implement this function as part of an extension on `TokenizedControlInternal`.
- Remove `TokenResolver`, as this function supersedes all its previous behaviors.

Bonus benefits:
- There's far less boilerplate to deal with (e.g. no more `didSet` modifiers to update tokens! SwiftUI will deal with it for us!). 
- It's now easier to write correct code, and harder to write incorrect code 🎉
- Binary shrinks by more than 100 kB 🎉🎉

### Verification

This was forked off my work on #879. Using that branch, I was able to verify that all currently-supported controls dynamically respond to theme both in UIKit AND SwiftUI now.

Note: the following video requires some additional changes to pipe theming through to SwiftUI demos. That will come as part of #879, but observe the magic here:

https://user-images.githubusercontent.com/4934719/155615863-15315390-788b-46b9-a0ac-a39f5dc1d43b.mp4

Other than that, I just went through all of the Token 2.0 controls and made sure they turn green:

https://user-images.githubusercontent.com/4934719/155616110-4e88a734-2db3-4033-b089-07836104d5d6.mov

| 31,137,581 bytes (30,408 kB) | 31,001,293 bytes (30,215 kB) |

Improvement: **133 kB**

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/924)


ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
```diff
@@ -19,9 +19,20 @@ public protocol TokenizedControl {
...
 
     /// Fetches the current token override.
     var overrideTokens: TokenType? { get }
 }
+
+// MARK: - Extensions
+
+extension TokenizedControlInternal {
+    /// Returns the correct token set for a given tokenizable control.
+    func tokens(for fluentTheme: FluentTheme) -> TokenType {
```
huwilkes, line 33: Could we move the fluentTheme in to TokenizedControlInternal, rather than passing it in? Seems a bit odd to say "get me the tokens for this theme, please" and it replies "hey, I saw you have override tokens, have those instead".

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -220,9 +206,11 @@ struct FluentButtonBody: View {
 /// ButtonStyle which configures the Button View according to its state and design tokens.
 struct FluentButtonStyle: ButtonStyle {
     @ObservedObject var state: MSFButtonStateImpl
+    let tokensLookup: () -> ButtonTokens
```
huwilkes, line 209: I'm not really sure I understand what's happening here. tokensLookup is some sort of function that returns ButtonTokens? And it's just a closure around the tokens from FluentButton? Is that just so that the tokens in FluentButtonBody, FluentButtonStyle, and FluentButton stay in sync?

ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBar.swift
```diff
@@ -57,10 +57,13 @@ public struct IndeterminateProgressBar: View, ConfigurableTokenizedControl {
             .modifyIf(!state.isAnimating && state.hidesWhenStopped, { view in
                 view.hidden()
             })
-            .resolveTokens(self)
     }
 
-    var tokens: IndeterminateProgressBarTokens { state.tokens }
+    let defaultTokens: IndeterminateProgressBarTokens = .init()
+    var tokens: IndeterminateProgressBarTokens {
+        let tokens = tokens(for: fluentTheme)
```
rdeassis, line 64: nit: why not just `return tokens(for: fluentTheme)` ?

ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
rdeassis, line 33: Does this also resolves the issue of control overriding for the entire theme (that we discussed earlier this week)?

rdeassis:
`This was forked off my work on https://github.com/microsoft/fluentui-apple/pull/879. Using that branch, I was able to verify that all currently-supported controls dynamically respond to theme both in UIKit AND SwiftUI now.`

That's exciting! Shouldn't the demo videos in the description show that instead of the old Theming demo controller?

Also... do you mind updating the title of the PR to something that matches its description? (for our own sanity in future history searches)

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: That's a good question.
My question on top of that would be: why do we need the tokensLookup param? Can't we just make the tokens property of type `(() -> ButtonTokens)`?

ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
mischreiber: Great idea! Yes, I can do that. I'll also rename the function to `resolvedTokens()` to keep it from just being `tokens()`, which would be pretty confusing.

Backstory: I was originally focused on getting the `@Environment` variable stored into this extension, which is not allowed. It didn't occur to me that I could just put the d

ios/FluentUI/Vnext/Button/Button.swift
mischreiber: Exactly right! SwiftUI does some unusual cacheing of this button style, and I wasn't able to keep the tokens flowing through the system as I wanted without this fetching.

ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBar.swift
mischreiber: Ah no particular reason. Partly bulk changing of the property, partly some debugging I was doing. But I agree that that's cleaner, so I'll do that in all the cases where no props are changed.

ios/FluentUI/Core/Theme/Tokens/TokenizedControl.swift
mischreiber: Ughhhh I wish. Sadly that still doesn't force a render, so that'll have to be addressed independently still. 

Bleh.

mischreiber:
> That's exciting! Shouldn't the demo videos in the description show that instead of the old Theming demo controller?
Oh sure! Yeah the videos were taken from a weird combo build with _some_ of those changes in the demo app, but not the test view itself. Lemme spin up a separate branch with both bits of code and I can record new videos.

> Also... do you mind updating the title of the PR to something that matches its description? (for our own sanity in future history searches)
Oops! It had a good name when it was a single commit, but I added that second one before I finished making the PR and I got a horrible name out of it. Updated!

huwilkes:
> > That's exciting! Shouldn't the demo videos in the description show that instead of the old Theming demo controller?
> 
> Oh sure! Yeah the videos were taken from a weird combo build with _some_ of those changes in the demo app, but not the test view itself. Lemme spin up a separate branch with both bits of code and I can record new videos.
> 
> > Also... do you mind updating the title of the PR to something that matches its description? (for our own sanity in future history searches)
> 
> Oops! It had a good name when it was a single commit, but I added that second one before I finished making the PR and I got a horrible name out of it. Updated!

Your name is out of date now, though. Should be "add resolvedTokens()"

ios/FluentUI/Vnext/Button/Button.swift
mischreiber: @rdeassis sorry didn't see your comment when I replied. We could make all tokens be of type `(() -> ButtonTokens)`, but that would mean updating every reference to `tokens` throughout the file. I was hoping to minimize churn by keeping `tokens` as a convenience getter to call `tokensLookup()`.

ios/FluentUI/Core/Theme/Tokens/ConfigurableTokenizedControl.swift
```diff
@@ -29,8 +26,4 @@ protocol ConfigurableTokenizedControl: TokenizedControlInternal {
 /// This set of extensions implements the `TokenizedControlInternal` APIs for SwiftUI components.
 extension ConfigurableTokenizedControl {
     var overrideTokens: TokenType? { state.overrideTokens }
```
rdeassis, line 28: After all these changes is there any benefit on still keeping the override tokens in the state object?

ios/FluentUI/PersonaButton/PersonaButton.swift
```diff
@@ -77,13 +77,17 @@ public struct PersonaButton: View, ConfigurableTokenizedControl {
...
     }
 
     @Environment(\.fluentTheme) var fluentTheme: FluentTheme
     @Environment(\.sizeCategory) var sizeCategory: ContentSizeCategory
     @ObservedObject var state: MSFPersonaButtonStateImpl
-    var tokens: PersonaButtonTokens { state.tokens }
+    let defaultTokens: PersonaButtonTokens = .init()
+    var tokens: PersonaButtonTokens {
+        let tokens = resolvedTokens()
+        tokens.size = state.buttonSize
```
rdeassis, line 88: Related to the comment I made in another file:
`After all these changes is there any benefit on still keeping the override tokens in the state object?`

Since we're propagating the state object's properties here, is there any reason why the override tokens should be in the state object? It could eventually be another property in the SwiftUI control just like the UIkit ones have it?

ios/FluentUI/Core/Theme/Tokens/ConfigurableTokenizedControl.swift
mischreiber: Yeah, the main reason is for clients to be able to specify override tokens by adding them to the state object either directly (using our UIKit wrapper) or through the view modifier.

ios/FluentUI/PersonaButton/PersonaButton.swift
mischreiber: We can't store a mutable object reference in a SwiftUI view because the struct is immutable and prone to recreation. The only way we could keep `overrideTokens` in the view is it if was part of the designated initializer for every control, which doesn't feel quite right to me.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -312,9 +311,28 @@ public struct Avatar: View, ConfigurableTokenizedControl {
         state = avatarState
     }
 
+    /// Calculates the size of the avatar, including ring spacing
+    func totalSize() -> CGFloat {
```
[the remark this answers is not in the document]
mischreiber: Could be! I just copy/pasted it from the state to the view, but yeah it can definitely be a computed property instead. I'll switch it!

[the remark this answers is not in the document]
mischreiber: Also switched  `resolvedTokens()` -> `resolvedTokens`.

rdeassis (approved):

ios/FluentUI/Core/Theme/Tokens/ConfigurableTokenizedControl.swift
rdeassis: I'm wondering if there's a way to keep all the logic of token management in the state object (through what the TokenizedControlInternal functionality accomplishes).

ios/FluentUI/PersonaButton/PersonaButton.swift
mischreiber: Discussed offline: we'll continue to think about how to bring tokens together, but for now we'll merge this PR to unblock other work.

ios/FluentUI/Core/Theme/Tokens/ConfigurableTokenizedControl.swift
mischreiber: Discussed offline: we'll continue to think about how to bring tokens together, but for now we'll merge this PR to unblock other work.

PR #925, opened 2022-02-25
[vNext Tokens] Convert Drawer controller

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Convert the Drawer controller to the new tokens system. 

### Verification

| Drawer_popover_before | Drawer_popover_after |
| Drawer_TopClear_before | Drawer_TopClear_after2 |
| Drawer_TopDimmed_before | Drawer_TopDimmed_after2 |
| Drawer_Leading_before | Drawer_Leading_after2 | 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/925)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -407,7 +405,7 @@ open class DrawerController: UIViewController, FluentUIWindowProvider {
     /// for iPad split mode, navigation bar has a different dark elevated color, and if it is a `.down` presentation style, match `Colors.NavigationBar.background` elevated color
     private var useNavigationBarBackgroundColor: Bool = false
     /// Drawer controller design tokens
-    private let drawerTokens = MSFDrawerTokens()
+//    private let drawerTokens = DrawerTokens()
```
mischreiber, line 408: Commented code

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -566,6 +560,22 @@ open class DrawerController: UIViewController, FluentUIWindowProvider {
         return dismissPresentingViewController(animated: true)
     }
 
+    func updateCurrentTokens(_ tokens: DrawerTokens) {
+        self.tokens = tokens
+    }
+
+    var tokens: DrawerTokens = .init()
+    var overrideTokens: DrawerTokens? {
```
mischreiber, line 568: Is this the only place that this should be called?

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -566,6 +560,22 @@ open class DrawerController: UIViewController, FluentUIWindowProvider {
         return dismissPresentingViewController(animated: true)
     }
 
+    func updateCurrentTokens(_ tokens: DrawerTokens) {
+        self.tokens = tokens
+    }
+
+    var tokens: DrawerTokens = .init()
```
mischreiber, line 567: FYI there's a small update to the tokens branch from #924 that'll require a small change to how this is stored. Check out some of the examples in that PR for guidance, or ping me offline if you have more questions!

mischreiber (changes requested):
I notice that there are a number of changes in appearance between the sgen tokens and our new ones, at least based on your screenshots. In particular, the drop shadow and dimming view do not match the previous tokenized control.

The goal is for these transitions to be 100% seamless, at least from a consumer perspective. We shouldn't be updating any appearance unless it's already been pre-cleared by design before this work started.

I know we discussed the shadow offline, but ultimately I think we need to make sure that our control is matching whatever odd behavior we saw previously; we can file bugs on the design tracker for any open questions we have for design going forward, but we should not block this transition waiting for those.

mischreiber (approved):

PR #919, opened 2022-02-18
[vNext Tokens] Convert Avatar to the new token system

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Update Avatar to use the new token system.

### Verification

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Avatar_before | Avatar_after2 |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/919)

PR #535, opened 2021-04-21
Replacing AvatarView with Avatar Vnext throughout FluentUI library.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change replaces all uses of the current AvatarView (written on UIKit) with the SwiftUI Avatar Vnext throughout the Library.
The AvatarGroupView is not being updated as it'll be re-written in SwiftUI.
#517, #527 and #528 unblocked key scenarios to make this update possible.

### Verification

https://user-images.githubusercontent.com/68076145/115629482-45e0b000-a2b7-11eb-9602-15406ea8d552.mov

Demo app scenarios on the following controls:

**- ContactCollectionView**

| iPhone                                       | iPad                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 11 Pro - 2021-04-21 at 15 06 08 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-21 at 15 07 30 |

**- NavigationController**

| iPhone                                       | iPad                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 12 - 2021-04-21 at 15 09 12 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-21 at 15 09 34 |

**- PeoplePicker**

| iPhone                                       | iPad                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 12 - 2021-04-21 at 15 11 30 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-21 at 15 10 33 |

**- PersonaListView**

| iPhone                                       | iPad                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 11 Pro - 2021-04-21 at 15 12 29 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-21 at 15 12 52 |

**- SideTabBar**

| iPhone                                       | iPad                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 11 Pro - 2021-04-21 at 15 14 08 | Simulator Screen Shot - iPad Pro (12 9-inch) (4th generation) - 2021-04-21 at 15 14 24 |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
[...]

PR #934, opened 2022-03-02
Move Avatar size constants out of MSFAvatarSize

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Moving the size out of the enum as requested in https://github.com/microsoft/fluentui-apple/pull/919#discussion_r810401567

There were a few places where we would try to get the size of an Avatar by going through the MSFAvatarSize rather than the Avatar itself. 
One of those places was in some constraints for the LargeTitleView (added in https://github.com/microsoft/fluentui-apple/pull/535), but that seems to have no problems resizing without them.

### Verification

| AvatarSize_Group_before | AvatarSize_Group_after |
| AvatarSize_PersonaButton_before | AvatarSize_PersonaButton_after |
| AvatarSize_LargeTitle_before | AvatarSize_LargeTitle_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/934)

### DIFF
--- a/ios/FluentUI/AvatarGroup/AvatarGroup.swift
+++ b/ios/FluentUI/AvatarGroup/AvatarGroup.swift
@@ -132,7 +132,7 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
         let isStackStyle = tokens.style == .stack
 
         let interspace: CGFloat = tokens.interspace
-        let imageSize: CGFloat = tokens.size.size
+        let imageSize: CGFloat = avatarViews[0].contentSize
         let ringOuterGap: CGFloat = tokens.ringOuterGap
         let ringGapOffset: CGFloat = ringOuterGap * 2
         let ringOffset: CGFloat = tokens.ringThickness + tokens.ringInnerGap + tokens.ringOuterGap
--- a/ios/FluentUI/AvatarGroup/AvatarGroup.swift
+++ b/ios/FluentUI/AvatarGroup/AvatarGroup.swift
@@ -132,11 +132,12 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
         let isStackStyle = tokens.style == .stack
 
         let interspace: CGFloat = tokens.interspace
-        let imageSize: CGFloat = avatarViews[0].contentSize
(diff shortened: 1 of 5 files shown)


ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -132,7 +132,7 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
         let isStackStyle = tokens.style == .stack
 
         let interspace: CGFloat = tokens.interspace
-        let imageSize: CGFloat = tokens.size.size
+        let imageSize: CGFloat = avatarViews[0].contentSize
```
sophialee0416, line 135: Is it ok to access the size like this? What if the avatarViews was empty?

huwilkes: That's fair. I've changed it to make the overflow avatar earlier so we can get the content size from that. With the tokens just being stored in the actual control now, there's not really a good way to get the size without going through an actual avatar.

sophialee0416 (approved):

ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -58,9 +58,6 @@ class LargeTitleView: UIView {
         }
     }
 
-    var avatarHeightConstraint: NSLayoutConstraint?
```
harrieshin, line 61: what was the original change/commit/pr? what was the bug?

harrieshin: and actually mention it in your pr and commit message.

huwilkes: https://github.com/microsoft/fluentui-apple/pull/919#discussion_r810401567
There wasn't really a good way to get the size to the LargeTitleView, so I checked if it worked without the constraints, and it did.

PR #939, opened 2022-03-05
[vNext Tokens] `state` -> `configuration`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

As part of our effort to consolidate and update our approach to our new Fluent controls, we've decided to formally swap the name for our SwiftUI controls' internal data structure from `state` to `configuration`.

The reasons for this change:
* `state` is a highly overloaded term in Apple-land, from `@State` in SwiftUI to `UIControl.State` for the interaction state of the control (e.g. highlighted, disabled), there is potential for confusion and conflict.
* On the other hand, `configuration` is only lightly used in either SwiftUI or UIKit, and only for buttons so far.
* `configuration` also more precisely describes the goal of the class/property: to describe how the control is "configured". This can include titles or images in buttons, or the sub-items in an aggregate control like `AvatarGroup`, along with custom token overrides for every other control. 

As for why now: this branch will result in a fair bit of churn to our clients when it merges, so might as well pull the bandage off now!

### Verification

Since this is a fairly wide-ranging change, the testing approach was more "bug bash" style than particularly focused on one area. My verification flow:
* Ensure every existing vNext control still displays in its demo controller.
* Verify the SwiftUI versions of vNext controls with support for that hosting environment.
* Verify custom tokens for controls with this view hooked up.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/939)

### DIFF
--- a/ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
+++ b/ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
@@ -6,8 +6,8 @@
 import SwiftUI
 import UIKit
 
-/// Properties available to customize the state of the Activity Indicator state
-@objc public protocol MSFActivityIndicatorState {
+/// Properties available to customize the configuration of the Activity Indicator configuration
+@objc public protocol MSFActivityIndicatorConfiguration {
 
     /// Sets the accessibility label for the Activity Indicator.
     var accessibilityLabel: String? { get set }
@@ -34,25 +34,25 @@ public struct ActivityIndicator: View, ConfigurableTokenizedControl {
     /// Creates the ActivityIndicator.
     /// - Parameter size: The MSFActivityIndicatorSize value used by the Activity Indicator.
     public init(size: MSFActivityIndicatorSize) {
-        let state = MSFActivityIndicatorStateImpl(size: size)
-        self.state = state
+        let configuration = MSFActivityIndicatorConfigurationImpl(size: size)
+        self.configuration = configuration
     }
 
     public var body: some View {
         let side = tokens.side
         let color: Color = {
-            guard let stateUIColor = state.color else {
+            guard let configurationUIColor = configuration.color else {
                 return Color(dynamicColor: tokens.defaultColor)
             }
 
-            return Color(stateUIColor)
+            return Color(configurationUIColor)
         }()
         let accessibilityLabel: String = {
-            if let overriddenAccessibilityLabel = state.accessibilityLabel {
+            if let overriddenAccessibilityLabel = configuration.accessibilityLabel {
                 return overriddenAccessibilityLabel
             }
 
-            return state.isAnimating ?
+            return configuration.isAnimating ?
                 "Accessibility.ActivityIndicator.Animating.label".localized
                 :
                 "Accessibility.ActivityIndicator.Stopped.label".localized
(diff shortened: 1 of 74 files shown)


ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
```diff
@@ -61,20 +61,20 @@ public struct ActivityIndicator: View, ConfigurableTokenizedControl {
...
-            .modifyIf(state.isAnimating, { animatedView in
+            .modifyIf(configuration.isAnimating, { animatedView in
                 animatedView
                     .rotationEffect(.degrees(rotationAngle), anchor: .center)
                     .onAppear {
                         startAnimation()
                     }
             })
-            .modifyIf(!state.isAnimating) { staticView in
+            .modifyIf(!configuration.isAnimating) { staticView in
```
huwilkes, line 71: It seems like this is one case where configuration makes less sense as a name. Not sure if we should split variables that actually represent the state the control is in and variables that decided how the control are configured in to different classes, though.

harrieshin: either a separate bug to go clean up but I do agree it seems weird that this is configuration rather than the actual state of the control like isHidden

harrieshin (approved):

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
```diff
@@ -6,8 +6,8 @@
 import SwiftUI
 import UIKit
 
-/// Properties available to customize the state of the Activity Indicator state
-@objc public protocol MSFActivityIndicatorState {
+/// Properties available to customize the configuration of the Activity Indicator configuration
+@objc public protocol MSFActivityIndicatorConfiguration {
```
rdeassis, line 10: Beyond the discussion behind the naming of these classes and objects, we need be aware that this will cause an extra disruption on the integration of the vnext-tokens branch to main which will likely make it an even longer process.

In the devmain environment this will mean a coordinated update of all indirect dependencies along with the code changes in devmain itself.

We can talk about this offline too, but I believe this change could go in a separate smaller and focused integration to prevent multiple potential issues in only a single big integration.

rdeassis (commented):
Beyond the

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
mischreiber: The merge to `main`, ignoring devmain, won't be disrupted by this change, since we've done a good job of FIing from `main` into `vnext-*`.

Your point about devmain, however, is fair! That said, we may yet need to coordinate downstream dependencies since there have been a few minor public API changes along with this. We won't know for sure until we test, however.

If we did break this into a separate integration, presumably it would have to wait for another minor version change, right? So tokens are 0.5.x, and then this change would create 0.6.x?

ios/FluentUI/ActivityIndicator/ActivityIndicator.swift
mischreiber: Yeah this definitely isn't 100% perfect. I wonder if properties like that would make more sense living on the SwiftUI control itself, rather than as part of a state/configuration object?

mischreiber:
As discussed offline, we will table this work for a future release.

PR #940, opened 2022-03-09
[vNext] Tokenize Segmented Control

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Convert the Segmented Control to a tokenized control. Also update the demo controller to support the new theme/token overrides.

### Verification

| SegmentedControl_before_light | SegmentedControl_after_light |
| SegmentedControl_before_dark | SegmentedControl_after_dark |
| SegmentedControl_before_color | SegmentedControl_after_color |
| SegmentedControl_rtl_before | SegmentedControl_rtl_after |
|  | SegmentedControl_after_themes |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/940)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
@@ -147,7 +147,7 @@ extension SegmentedControlDemoController: DemoAppearanceDelegate {
     }
 
     func isThemeWideOverrideApplied() -> Bool {
-        return self.view.window?.fluentTheme.tokenOverride(for: FluentButton.self) != nil
+        return self.view.window?.fluentTheme.tokenOverride(for: SegmentedControl.self) != nil
     }
 
     // MARK: - Custom tokens
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
@@ -55,6 +55,11 @@ class SegmentedControlDemoController: DemoController {
         addTitle(text: "Disabled On Brand Pill")
 
         addPillControl(items: Array(segmentItems.prefix(2)), style: .onBrandPill, enabled: false)
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
     }
 
     @objc func updateLabel(forControl control: SegmentedControl) {
@@ -71,8 +76,10 @@ class SegmentedControlDemoController: DemoController {
         if style == .primaryPill {
             backgroundView.backgroundColor = Colors.navigationBarBackground
         } else {
-            backgroundView.backgroundColor = UIColor(light: Colors.communicationBlue, dark: Colors.navigationBarBackground)
+            setOnBrandBackgroundColorFor(backgroundView)
+            onBrandSegmentedControlViews.append(backgroundView)
         }
+        segmentedControls.append(pillControl)
 
         backgroundView.translatesAutoresizingMaskIntoConstraints = false
         pillControl.translatesAutoresizingMaskIntoConstraints = false
@@ -90,11 +97,30 @@ class SegmentedControlDemoController: DemoController {
         if enabled {
             controlLabels[pillControl] = addDescription(text: "", textAlignment: .center)
         }
+    }
 
-        segmentedControls.append(pillControl)
+    @objc func themeDidChange(_ notification: Notification) {
+        for background in onBrandSegmentedControlViews {
+            setOnBrandBackgroundColorFor(background)
+        }
+    }
+
+    func setOnBrandBackgroundColorFor(_ backgroundView: UIView) {
+        var lightColor: UIColor
+        var darkColor: UIColor
+        if let window = self.view.window {
+            lightColor = Colors.primary(for: window)
+            darkColor = UIColor(colorValue: window.fluentTheme.globalTokens.neutralColors[.grey12])
+        } else {
+            lightColor = Colors.communicationBlue
+            darkColor = Colors.navigationBarBackground
+        }
+
+        backgroundView.backgroundColor = UIColor(light: lightColor, dark: darkColor)
     }
 
     private var segmentedControls: [SegmentedControl] = []
+    private var onBrandSegmentedControlViews: [UIView] = []
 }
 
 extension SegmentedControlDemoController: DemoAppearanceDelegate {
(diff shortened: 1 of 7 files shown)


harrieshin:
Screen Shot 2022-03-08 at 8 10 08 PM
what is this weird line in the segmented control in your gif?

ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
```diff
@@ -91,4 +87,101 @@ class SegmentedControlDemoController: DemoController {
...
+
+    func isThemeWideOverrideApplied() -> Bool {
+        return self.view.window?.fluentTheme.tokenOverride(for: SegmentedControl.self) != nil
+    }
+
+    // MARK: - Custom tokens
+
+    private class ThemeWideOverrideSegmentedControlTokens: SegmentedControlTokens {
+        override var font: FontInfo {
+            return FontInfo(name: "Times", size: 20.0, weight: .regular)
```
harrieshin, line 125: can we not passin the size so even with Times font the accessibility text will still layout correctly?

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
```diff
@@ -58,12 +70,15 @@ class SegmentPillButton: UIButton {
 
     private let item: SegmentItem
 
-    private let unreadDotLayer: CALayer = {
+    private func initUnreadDotLayer() -> CALayer {
```
harrieshin, line 73: nit: should we be using "init" in this function name?

harrieshin:
tested RTL?

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
huwilkes: That's what we named it in the Pill Button.

huwilkes:
> Screen Shot 2022-03-08 at 8 10 08 PM what is this weird line in the segmented control in your gif?

I'm not sure, but it was there before my changes. It doesn't really show up in the view hierarchy, and it goes away if you scroll at all. Could just be a simulator rendering issue? I can check an actual device.

harrieshin:
> > Screen Shot 2022-03-08 at 8 10 08 PM what is this weird line in the segmented control in your gif?
> 
> I'm not sure, but it was there before my changes. It doesn't really show up in the view hierarchy, and it goes away if you scroll at all. Could just be a simulator rendering issue? I can check an actual device.

that's concerning. try simulator slow animation? something is not calculated right?

ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
huwilkes: FontInfo needs all three.

huwilkes: I updated the font to not scale in the SegmentPillButton

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -494,6 +340,36 @@ open class SegmentedControl: UIControl {
...
+    private func updateSegmentedControlTokens() {
+        let tokens = resolvedTokens
+        tokens.style = style
+        self.tokens = tokens
+    }
+
+    private func updateButtons() {
+        for (index, button) in buttons.enumerated() {
+            button.tokens = tokens
+            if isEnabled {
```
rdeassis, line 364: nit: since the 2 calls are identical with only the different param, it could be cleaner to just create a `let labelColor` and assign it with the correct color based on the isEnabled and just have one call in the code to set the title color on the button.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -580,6 +456,15 @@ open class SegmentedControl: UIControl {
         button.isSelected = isSelected
     }
 
+    @objc private func themeDidChange(_ notification: Notification) {
+        if let window = self.window,
```
rdeassis, line 460: Can you change this to a guard let to conform to the swift guid on early return?

ios/FluentUI/SegmentedControl/SegmentedControlTokens.swift
```diff
@@ -0,0 +1,129 @@
...
+        switch style {
+        case .primaryPill:
+            return DynamicColor(light: globalTokens.neutralColors[.grey94], dark: globalTokens.neutralColors[.grey44])
+        case .onBrandPill:
+            return DynamicColor(light: globalTokens.neutralColors[.grey74], dark: globalTokens.neutralColors[.grey44])
+        }
+    }
+
+    /// The color of the unread dot when the `SegmentedControl` is enabled.
+    open var enabledUnreadDotColor: DynamicColor {
```
rdeassis, line 94: Although the disabledUnreadDotColor is just the same as the label color, I'm wondering if we shouldn't have it here for completeness (and returning the disabledLabelColor).
It will make the disabledUnreadDotColor configurable from the token system instead of hard coded.
The tradeoff is obviously an extra property that will add on the binary size.

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
mischreiber: I don't think we can use a closure initialized property because it refers to elements in `self`, so I think we need to go for a `lazy var` instead.

ios/FluentUI/SegmentedControl/SegmentedControlTokens.swift
```diff
@@ -0,0 +1,129 @@
...
+/// Design token set for the `SegmentedControl`.
+open class SegmentedControlTokens: ControlTokens {
+    /// Defines the style of the `SegmentedControl`.
+    public internal(set) var style: SegmentedControlStyle = .primaryPill
+
+    /// Defines the background color of the unselected segments of the `SegmentedControl`.
+    open var restTabColor: DynamicColor {
+        switch style {
+        case .primaryPill:
+            return DynamicColor(light: aliasTokens.backgroundColors[.neutral4].light, dark: aliasTokens.backgroundColors[.neutral3].dark)
```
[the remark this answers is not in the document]
huwilkes: I thought the guidance for that was if there were three or more on one line.

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
rdeassis: @mischreiber lazy vars can't be computed properties.
I think you're referring to the fact that non-lazy vars can't refer to self in their initialization.

But this is what I'm suggesting:

```
    private lazy var unreadDotLayer: CALayer = {
        let unreadDotLayer = CALayer()
        let unreadDotSize = tokens.unreadDotSize
        unreadDotLayer.bounds.size = CGSize(width: unreadDotSize, height: unreadDotSize)
        unreadDotLayer.cornerRadius = unreadDotSize / 2
        return unreadDotLayer
    }()
```

rdeassis (approved):

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
mischreiber: Ah yep that makes sense. Dang, this stuff confuses me.

mischreiber (approved):

PR #941, opened 2022-03-10
[vNext Tokens] Update AvatarGroup to get Avatar tokens from the overflow Avatar

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Get the Avatar Tokens from the already initialized overflow avatar so we can remove the duplicate tokens from the AvatarGroup Tokens

### Verification

| AvatarGroup_duplicateTokens_before | AvatarGroup_duplicateTokens_after |
| AvatarGroup_duplicateTokens_rtl_before | AvatarGroup_duplicateTokens_rtl_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/941)

### DIFF
--- a/ios/FluentUI/AvatarGroup/AvatarGroup.swift
+++ b/ios/FluentUI/AvatarGroup/AvatarGroup.swift
@@ -115,14 +115,23 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
         let isStackStyle = tokens.style == .stack
 
         let overflowAvatar = createOverflow(count: overflowCount)
-        let imageSize: CGFloat = overflowAvatar.contentSize
 
         let interspace: CGFloat = tokens.interspace
-        let avatarTokens = overflowAvatar.tokens
-        let ringOuterGap: CGFloat = avatarTokens.ringOuterGap
-        let ringGapOffset: CGFloat = ringOuterGap * 2
-        let ringOffset: CGFloat = avatarTokens.ringThickness + avatarTokens.ringInnerGap + avatarTokens.ringOuterGap
-        let groupHeight: CGFloat = imageSize + (ringOffset * 2)
+
+        let groupHeight: CGFloat = {
+            let avatarMaxHeight: CGFloat
+            if let avatar = avatarViews.first {
+                let avatarTokens = avatar.tokens
+                avatarMaxHeight = avatarTokens.avatarSize + 2 * (avatarTokens.ringThickness + avatarTokens.ringInnerGap + avatarTokens.ringOuterGap)
+            } else {
+                avatarMaxHeight = 0
+            }
+
+            let overflowTokens = overflowAvatar.tokens
+            let overflowMaxHeight = overflowTokens.avatarSize + 2 * (overflowTokens.ringThickness + overflowTokens.ringInnerGap + overflowTokens.ringOuterGap)
+
+            return max(avatarMaxHeight, overflowMaxHeight)
+        }()
 
         @ViewBuilder
         var avatarGroupContent: some View {
@@ -142,10 +151,14 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
                     let nextAvatarHasRing = !isLastDisplayed ? avatars[nextIndex].isRingVisible : false
 
                     // Calculating the different interspace scenarios considering rings.
+                    let currentTokens = avatarView.tokens
+                    let ringOuterGap = currentTokens.ringOuterGap
+                    let ringOffset = currentTokens.ringInnerGap + currentTokens.ringThickness + ringOuterGap
                     let stackPadding = interspace - (currentAvatarHasRing ? ringOffset : 0) - (nextAvatarHasRing ? ringOuterGap : 0)
 
                     // Finalized calculations for x and y coordinates of the Avatar if it needs a cutout, including RTL.
-                    let cutoutSize = isLastDisplayed ? ringGapOffset + imageSize : nextAvatarSize
+                    let cutoutSize = isLastDisplayed ? overflowAvatar.totalSize : nextAvatarSize
+                    let ringGapOffset = 2 * ringOuterGap
                     let xOrigin: CGFloat = {
                         if layoutDirection == .rightToLeft {
                             return -cutoutSize - interspace + ringOuterGap + (currentAvatarHasRing ? ringOffset : 0)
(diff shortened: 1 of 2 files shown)


ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -114,13 +114,14 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
...
-        let interspace: CGFloat = tokens.interspace
-        let ringOuterGap: CGFloat = tokens.ringOuterGap
-        let ringGapOffset: CGFloat = ringOuterGap * 2
-        let ringOffset: CGFloat = tokens.ringThickness + tokens.ringInnerGap + tokens.ringOuterGap
-
         let overflowAvatar = createOverflow(count: overflowCount)
         let imageSize: CGFloat = overflowAvatar.contentSize
+
+        let interspace: CGFloat = tokens.interspace
+        let avatarTokens = overflowAvatar.tokens
```
rdeassis, line 121: These tokens might vary depending on the style of the avatars.
Overflow is supposed to be the last one, but the other ones should could get different tokens if those are overridden and we're retrieving the resolved ones for each avatar.

huwilkes: Do we let avatars have different styles and sizes in the Avatar Group? Either way, the old way we got the tokens would've had that same problem.

rdeassis: We don't have any way of setting different avatars through the group with different styles.
But the overflow avatar is an exception to that. It's a different style from the remainder of the avatars.
And I agree that this was gap we currently have. Probably not exposed so far as tokens were not overridden differently across styles.

mischreiber: Interestingly, we shouldn't need any of these values unless `avatarViews.count > 0`, so we could always use the tokens from `avatarViews.first`, right? No need to leverage the overflow avatar which, as mentioned by @rdeassis, could easily be different than the rest at some point.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -114,14 +114,24 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
...
-        let ringOuterGap: CGFloat = tokens.ringOuterGap
-        let ringGapOffset: CGFloat = ringOuterGap * 2
-        let ringOffset: CGFloat = tokens.ringThickness + tokens.ringInnerGap + tokens.ringOuterGap
 
-        let overflowAvatar = createOverflow(count: overflowCount)
-        let imageSize: CGFloat = overflowAvatar.contentSize
-        let groupHeight: CGFloat = imageSize + (ringOffset * 2)
+        let groupHeight: CGFloat = {
+            let avatarMaxHeight: CGFloat
+            if let avatar = avatarViews.first {
```
mischreiber, line 123: Nit: should be a `guard let`

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -114,14 +114,24 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
...
+                let avatarTokens = avatar.tokens
+                avatarMaxHeight = avatarTokens.avatarSize + 2 * (avatarTokens.ringThickness + avatarTokens.ringInnerGap + avatarTokens.ringOuterGap)
+            } else {
+                avatarMaxHeight = 0
+            }
+
+            let overflowTokens = overflowAvatar.tokens
+            let overflowMaxHeight = overflowTokens.avatarSize + 2 * (overflowTokens.ringThickness + overflowTokens.ringInnerGap + overflowTokens.ringOuterGap)
+
+            return max(avatarMaxHeight, overflowMaxHeight)
```
rdeassis, line 133: If **hasOverflow** is false, we shouldn't consider it in this calculation, right?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -141,10 +151,14 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
...
 
                     // Calculating the different interspace scenarios considering rings.
+                    let currentTokens = avatarView.tokens
+                    let ringOuterGap = currentTokens.ringOuterGap
+                    let ringOffset = currentTokens.ringInnerGap + currentTokens.ringThickness + ringOuterGap
                     let stackPadding = interspace - (currentAvatarHasRing ? ringOffset : 0) - (nextAvatarHasRing ? ringOuterGap : 0)
 
                     // Finalized calculations for x and y coordinates of the Avatar if it needs a cutout, including RTL.
-                    let cutoutSize = isLastDisplayed ? ringGapOffset + imageSize : nextAvatarSize
+                    let cutoutSize = isLastDisplayed ? overflowAvatar.totalSize : nextAvatarSize
```
rdeassis, line 160: Is it possible to get rid of he cutoutSize constant and use the nextAvatarSize only?
It can have either the next avatar's totalSize or just the overflow avatar's totalSize depending on the scenario.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
huwilkes: I think we still want to? In case the overflow shows up later, we wouldn't want the size of the whole control to change. Also, if there's not overflow and no avatars, then the size would be zero, and I don't think we want that, either.

rdeassis: If the overflow shows up later this will be recalculated as it's part of the body property of the view.
The scenario of no avatars (regular or overflow) is unlikely, but I agree it'd be zero. I don't think we have a clear way of defining a constant height in those scenarios. Because the tokens could be overridden too. So we don't know whether the overflow could be larger or smaller than the other avatars.

rdeassis: Nevermind this comment... 
The overflow avatar is the default case here as we talked offline (in case there are no other avatars).
I think we can stick with that now!
Thanks!

rdeassis (approved):

PR #946, opened 2022-03-12
[vNext Tokens] Update PillButtonBar to update colors on theme change

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added an observer for themeDidChange so that the PillButtonBar would reload its colors when the theme changes. Also moved the BackgroundView from the SegmentedControlDemo to a new file and renamed it ColoredBackgroundView so I could use it in the PillButtonBar's demo also.

### Verification

| PillButtonBar_before | PillButtonBar_after |
| SegmentedControl_before | SegmentedControl_after |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/ColoredBackgroundView.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/ColoredBackgroundView.swift
@@ -1,0 +1,65 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+public enum ColoredBackgroundStyle: Int {
+    case neutral
+    case brand
+}
+
+class ColoredBackgroundView: UIView {
+    init(style: ColoredBackgroundStyle) {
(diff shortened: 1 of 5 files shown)


ios/FluentUI.Demo/FluentUI.Demo/ColoredBackgroundView.swift
```diff
@@ -0,0 +1,65 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+public enum ColoredBackgroundStyle: Int {
```
harrieshin, line 9: feels like the naming is too generic when only specific case would use this class and style.

huwilkes: That's fair. It probably also doesn't need to be public. I'll see if I can come up with a better name.

harrieshin (approved):

mischreiber (approved):

PR #947, opened 2022-03-17
Update the Fluent Button to change background colors when focus changes

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Since the collection view doesn't want to share the focus ring during keyboard navigation, we'll do some focus changes ourselves to make it more clear that the a primary button has focus.

### Verification

| ButtonFocus_keyboard_before | ButtonFocus_keyboard_after |
| ButtonFocus_voiceover_before | ButtonFocus_voiceover_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/947)

### DIFF
--- a/ios/FluentUI/Button/Button.swift
+++ b/ios/FluentUI/Button/Button.swift
@@ -225,16 +225,11 @@ open class Button: UIButton {
     }
 
     open override func didUpdateFocus(in context: UIFocusUpdateContext, with coordinator: UIFocusAnimationCoordinator) {
-        guard let window = window, style == .primaryFilled else {
+        guard style.isFilledStyle, (self == context.nextFocusedView || self == context.previouslyFocusedView) else {
             return
         }
-        let highlightedColor = UIColor(light: Colors.primaryTint10(for: window),
-                                     dark: Colors.primaryTint20(for: window))
-        if context.nextFocusedView == self {
-            backgroundColor = highlightedColor
-        } else if context.previouslyFocusedView == self {
-            backgroundColor = isHighlighted ? highlightedColor : Colors.primary(for: window)
-        }
+
+        updateBackgroundColor()
     }
 
     public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Button/Button.swift
```diff
@@ -224,6 +224,19 @@ open class Button: UIButton {
         update()
     }
 
+    open override func didUpdateFocus(in context: UIFocusUpdateContext, with coordinator: UIFocusAnimationCoordinator) {
+        guard let window = window, style == .primaryFilled else {
+            return
+        }
+        let highlightedColor = UIColor(light: Colors.primaryTint10(for: window),
```
harrieshin, line 231: wouldn't a better fix be utilizing updatebackgroundcolor() because dangerFill has similar problems and also one place to grab all the right colors?

rdeassis: I like the idea. That approach would probably consist of create another property like `isFocused` that would be set here and used in the `updateBackgroundColor()` method as @harrieshin suggested.

huwilkes: Since button already has isFocused, I didn't add a new redundant hasFocus, but it looks like it does get updated before the call to didUpdateFocus (which makes sense), so we're able to use that in updateBackgroundColor.

rdeassis (approved):

mischreiber:
Do we have a plan to implement this for Fluent 2 `Button` as well? That way we don't immediately regress when that merges to `main`.

huwilkes:
> Do we have a plan to implement this for Fluent 2 `Button` as well? That way we don't immediately regress when that merges to `main`.

I'm working on the SwiftUI button separately

harrieshin:
it would have been great to tested light and dark mode and updated the gif with danger button case as well

PR #949, opened 2022-03-23
[FHL] [vNext Tokens] Make SwiftUI control host be a `UIView` subclass, part 1

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### Problem: 
Our SwiftUI-backed Fluent 2 controls are hosted in a raw `NSObject` subclass (`ControlHostingContainer`), while our "legacy" Fluent 1 controls, as well as any UIKit-backed Fluent 2 controls, are all some form of `UIView` subclass. This means that different Fluent controls have very different ownership expectations:

* `UIView` subclasses can be added directly to the view hierarchy, which will manage their ownership lifetime on your behalf.
* `ControlHostingContainer` subclasses, on the other hand, must be stored independently as a property somewhere, and their `UIView` must be accessed via a `view` property.

This divergence approach puts a significant burden on developers, as failure to properly store a `ControlHostingContainer`-backed control will result in unexpected and unsupported behaviors. In particular, these controls will _often_ (but not _always_) continue to appear as expected, but may not respond to tap events (depending on the control!), and will _never_ support custom theming or color schemes.

#### Solution:
This solution consolidates the two styles of FluentUI control by converting `ControlHostingContainer` into `ControlHostingView`, a `UIView` subclass that can be safely stored in the view hierarchy and forgotten by the developer.

This PR is part 1, which keeps the `view` property (as a reference to `self`). Part two removes this property, and is a much larger, but mechanically uninteresting, change.

#### Ownership:
This approach introduces no strong reference cycles, as ownership is a pure digraph:

```
                 +-------------+
                 | Parent View |
                 +-------------+
                        | (via subview)
                        V
              +--------------------+
              | ControlHostingView |
              +--------------------+
 (via strong ref) |              | (via subview)
                  V              V
+---------------------+      +--------------+
| UIHostingController |  ->  | SwiftUI view |
+---------------------+      +--------------+
```

### Verification

Tested every tokenized Fluent control that I changed! That means the following:
* `ActivityIndicator`
* `Avatar`
* `AvatarGroup`
* `Button`
* `CardNudge`
* `Divider`
* `IndeterminateProgressBar`
* `List`
* `NotificationView`
* `PersonaButton[Carousel]`
* `PersonaView` (via `LeftNav`)

Video is attached that shows some of the most interesting controls, including ones that resize dynamically such as `CardNudge`.

https://user-images.githubusercontent.com/4934719/159597065-20a1c27d-e8c5-4a63-9a09-530230e4cdad.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/949)

### DIFF
--- a/ios/FluentUI/Core/ControlHostingView.swift
+++ b/ios/FluentUI/Core/ControlHostingView.swift
@@ -14,6 +14,13 @@ open class ControlHostingView: UIView {
         return self
     }
 
+    @objc public override var intrinsicContentSize: CGSize {
+        guard let hostedView = hostingController.view else {
+            return super.intrinsicContentSize
+        }
+        return hostedView.intrinsicContentSize
+    }
+
     /// Initializes and returns an instance of `ControlHostingContainer` that wraps `controlView`.
     ///
     /// Unfortunately this class can't use Swift generics, which are incompatible with Objective-C interop. Instead we have to wrap
--- a/ios/FluentUI/Core/ControlHostingView.swift
+++ b/ios/FluentUI/Core/ControlHostingView.swift
@@ -7,7 +7,7 @@ import SwiftUI
 import UIKit
 
 /// Common wrapper for hosting and exposing SwiftUI components to UIKit-based clients.
-open class ControlHostingContainer: UIView {
+open class ControlHostingView: UIView {
 
     /// The UIView representing the wrapped SwiftUI view.
     @objc public var view: UIView {
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/LeftNavDemoController.swift
@@ -103,14 +103,6 @@ class LeftNavMenuViewController: UIViewController {
         view = leftNavView
     }
 
-    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
-        super.traitCollectionDidChange(previousTraitCollection)
-        if let sizeCategory = previousTraitCollection?.preferredContentSizeCategory,
-            sizeCategory != traitCollection.preferredContentSizeCategory {
-            leftNavAccountViewHeightConstraint?.constant = leftNavAccountView.intrinsicContentSize.height
-        }
-    }
-
     var menuAction: (() -> Void)?
 
     private func setPresence(presence: LeftNavPresence) {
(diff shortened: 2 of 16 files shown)


mischreiber:
> There are a few of the UIKit controls that aren't UIViews, still. There are a few that are UIViewControllers (like Drawer) or UIControls (like Pill Button Bar and Segmented Control). If the goal of this PR is to try and standardize how our controls are used, should those also be updated? Or is this only for the SwiftUI backed controls?

Ultimately a `UIControl` is still a `UIView` subclass, so I'd call that good enough. The `UIViewController` is a good point, but having two types of control is still better than three, especially when both are standard iOS mechanisms.

harrieshin:
long term thinking: if we want our swiftuicode tobe used in macOS too... are we ifdefing uiview to nsview in controlhosting?

ios/FluentUI/Core/ControlHostingView.swift
```diff
@@ -36,7 +36,30 @@ open class ControlHostingContainer: NSObject {
...
+    private func configureHostedView() {
+        guard let hostedView = hostingController.view else {
+            return
+        }
+        hostedView.backgroundColor = UIColor.clear
+
+        addSubview(hostedView)
+        hostedView.translatesAutoresizingMaskIntoConstraints = false
+
+        let requiredConstraints = [
```
rdeassis, line 56: Have you tested scenarios where the SwiftUI View is set with the frame maxWidth/Height to .infinity and constraints are (or are not) set from the outer SwiftUI view?
i believe my main concern is about constraint priorities and how the SwiftUI View dictates the container UIView dimensions.

mischreiber: That's a good question! I _believe_ that `AvatarGroup` should meet that requirement:

https://github.com/mischreiber/fluentui-apple/blob/6eca8a1ab1deb04c77a6bedee3436338e107e43a/ios/FluentUI/AvatarGroup/AvatarGroup.swift#L197-L200

And the layout/overflow behavior matches the existing demo app:

https://user-images.githubusercontent.com/4934719/159806575-1bee69e8-3e32-4297-9c3a-e602463c0485.mov

huwilkes (approved):

rdeassis (approved):

PR #951, opened 2022-03-25
[FHL] Introducing the SwiftUI PresentationModifier.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

SwiftUI controls that will need to be presented on top of Views will leverage from this modifier to overlay their contents on top of the SwiftUI View that is modified with it.

It allows optionally blocking the underlying view and providing a backdrop to it as well.

The Heads-up Display (HUD) and NotificationView are the first controls that will use that modifier.

### Verification

Tested it as part of the ongoing HUD implementation in the purely SwiftUI environment:

https://user-images.githubusercontent.com/68076145/160185951-0a88f7db-ac7c-4f93-87f2-41c1401a7315.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/951)

### DIFF
--- a/ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
+++ b/ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
@@ -29,14 +29,14 @@ struct PresentationModifier<PresentedView: View>: ViewModifier {
 
     var hasDimmedBackground: Bool
     var dimmedBackgroundColor: Color
-    var dimmedBackgroundColroOpacity: Double
+    var dimmedBackgroundColorOpacity: Double
     var isBlocking: Bool
     @Binding var isPresented: Bool
     @ViewBuilder var presentedView: PresentedView
 
     var backgroundColor: Color {
         if hasDimmedBackground {
-            return dimmedBackgroundColor.opacity(dimmedBackgroundColroOpacity)
+            return dimmedBackgroundColor.opacity(dimmedBackgroundColorOpacity)
         } else if isBlocking {
             // A clear color won't allow the background
             // view to block the underlying view
@@ -52,6 +52,8 @@ extension View {
     /// Presents a View as a layer on top of this View.
     /// - Parameters:
     ///   - isPresented: Binding boolean value that determines whehter the `presentedView` is presented on top of this view..
+    ///   - dimmedBackgroundColor: Defines the color of the dimmed background presented with the `presentedView`. The default color is black.
+    ///   - dimmedBackgroundColorOpacity: Defines the opacity color of the dimmed background presented with the `presentedView`. The default color is 0.1.
     ///   - hasDimmedBackground: Defines whether the `presentedView` has a dimmed backdrop. The default value is false which renders a transparent backdrop.
     ///   - isBlocking: Defines whether the user can interact with this view while the `presentedView` is being presented on top of it.
     ///   - presentedView: The SwiftUI View that will be presented on top of this view.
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
```diff
@@ -0,0 +1,73 @@
...
+
+        return Color.clear
+    }
+}
+
+extension View {
+
+    /// Presents a View as a layer on top of this View.
+    /// - Parameters:
+    ///   - isPresented: Binding boolean value that determines whehter the `presentedView` is presented on top of this view..
```
jeaniehuynh, line 54: nit: missing the dimmed background parameters here

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
```diff
@@ -0,0 +1,73 @@
...
+            content
+
+            // Backdrop for the presented view which overlays the
+            // content where it's being presented from.
+            Rectangle()
+            .frame(maxWidth: .infinity,
+                   maxHeight: .infinity,
+                   alignment: .center)
+            .foregroundColor(isPresented ? backgroundColor : Color.clear)
+            .allowsHitTesting(isPresented && isBlocking)
```
jeaniehuynh, line 21: Do we want to allow interaction with dimmed background? (i.e. should we add `&& !hasDimmedBackground` here or leave that up to implementer to decide)

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
```diff
@@ -0,0 +1,73 @@
...
+    @Binding var isPresented: Bool
+    @ViewBuilder var presentedView: PresentedView
+
+    var backgroundColor: Color {
+        if hasDimmedBackground {
+            return dimmedBackgroundColor.opacity(dimmedBackgroundColroOpacity)
+        } else if isBlocking {
+            // A clear color won't allow the background
+            // view to block the underlying view
+            return Color.black.opacity(Double.ulpOfOne)
```
jeaniehuynh, line 43: Why ulpOfOne here? If that's the lowest possible double that's greater than 1.0, wouldn't 0.1 suffice/be more opaque?

rdeassis: ulpOfOne is "The positive difference between 1.0 and the next greater representable number."
It's smaller than 0.1 and it's the best "clear" view that we can have if we don't use Color.clear.
The problem with Color.clear is that it seems to be "absence" of a View, which won't work to block the interaction.

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
rdeassis: I thought about that and decided to leave it very flexible so that the implementer would decide (as you mentioned).
I'm thinking of a scenario where the dimmedBackgroundColor would be very transparent to even make sense... 
Do you have any concerns or a scenario that this could lead to undesired/unintended consequences?

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
```diff
@@ -0,0 +1,73 @@
...
+            presentedView
+        }
+        .frame(maxWidth: .infinity,
+               maxHeight: .infinity,
+               alignment: .center)
+    }
+
+    var hasDimmedBackground: Bool
+    var dimmedBackgroundColor: Color
+    var dimmedBackgroundColroOpacity: Double
```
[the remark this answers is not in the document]
rdeassis: Thanks for catching!

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
rdeassis: Thanks for catching!

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
jeaniehuynh: No concerns! Now that I'm thinking about it, I actually think it may be good for future controls like Teaching Callout where tapping on the background can dismiss the control entirely like below:

image

jeaniehuynh (approved):

PR #952, opened 2022-03-25
[FHL] Adding the AnimationCompletionModifier.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

As part of the Heads-up Display (HUD) implementation in SwiftUI, a completion block for animations was needed to ensure the view would be set up to an initial state after it was dismissed, so that it'd be ready for a new presentation.

This modifier allows a SwiftUI View to observe animatable properties and fires a "callback" func that executes when the animation to a target value completes.

### Verification

The size (scale factor) of the HUD in the video below is animated from bigger to normal in the presentation, normal to smaller when it's dismissed.
Once it's dismissed we need to set its size to the bigger state again.

The slow animations mode illustrates that better than words:

https://user-images.githubusercontent.com/68076145/160211550-b711ad95-63d4-4e33-9e7f-82af8be10024.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/952)

### DIFF
--- a/ios/FluentUI/Core/SwiftUI+ViewAnimation.swift
+++ b/ios/FluentUI/Core/SwiftUI+ViewAnimation.swift
@@ -1,0 +1,53 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// Modifier that provides a completion closure for an animation on a given animatable value.
+struct AnimationCompletionModifier<Value>: AnimatableModifier where Value: VectorArithmetic {
+
+    func body(content: Content) -> some View {
+        content
+    }
+
+    init(animatableValue: Value, onComplete: @escaping () -> Void) {
+        self.onComplete = onComplete
+        self.animatableData = animatableValue
+        targetValue = animatableValue
+    }
+
+    var animatableData: Value {
+        didSet {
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Core/SwiftUI+ViewAnimation.swift
```diff
@@ -0,0 +1,53 @@
...
+/// Modifier that provides a completion closure for an animation on a given animatable value.
+struct AnimationCompletionModifier<Value>: AnimatableModifier where Value: VectorArithmetic {
+
+    func body(content: Content) -> some View {
+        content
+    }
+
+    init(animatableValue: Value, onComplete: @escaping () -> Void) {
+        self.onComplete = onComplete
+        self.animatableData = animatableValue
```
jeaniehuynh, line 17: nit: self not needed here

jeaniehuynh (approved):

ios/FluentUI/Core/SwiftUI+ViewAnimation.swift
rdeassis: Good catch! Thanks!

PR #954, opened 2022-03-28
[FHL] Introducing the SquareShapedViewModifier.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

One of the requirements of the HUD is that the view needs to remain with an aspect ratio of a square (equal values for its height and width).
It is also a requirement that the size of the HUD needs to adjust based on the text of its label with constraints of minimum and maximum constraints.

The **SquareShapedViewModifier** addresses these requirements by leveraging from the alignmentGuide modifier that provides the SwiftUI View dimensions once it's resized based on its contents.

### Verification

Note from the video below that the HUD has a minimum value that is used when no label is provided to it, and it grows to a limit as the label text becomes longer. Always keeping the width and height equal.

The view also does not shrink to avoid "jumps" on the size when the label is updated while the HUD is presented.

https://user-images.githubusercontent.com/68076145/160493083-7333727e-7830-4346-a751-1e3a35ac7cc4.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/954)

### DIFF
--- a/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
+++ b/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
@@ -5,6 +5,38 @@
 
 import SwiftUI
 
+struct SquareShapedViewModifier: ViewModifier {
+    func body(content: Content) -> some View {
+        let modifiedContent = HStack {
+            content
+                .alignmentGuide(HorizontalAlignment.center) { (viewDimensions) -> CGFloat in
+                    DispatchQueue.main.async {
+                        self.size = max(size, // ensures the View does not shrink
+                                        min(maxSize,
+                                            max(minSize,
+                                                max(viewDimensions.height,
+                                                    viewDimensions.width))))
+                    }
+
+                    return viewDimensions[HorizontalAlignment.center]
+                }
+        }
+
+        modifiedContent
+            .frame(width: size, height: size)
+    }
+
+    init(minSize: CGFloat, maxSize: CGFloat) {
+        self.size = minSize
+        self.minSize = minSize
+        self.maxSize = maxSize
+    }
+
+    @State var size: CGFloat
+    var minSize: CGFloat
+    var maxSize: CGFloat
+}
+
 extension View {
     /// Applies modifiers defined in a closure if a condition is met.
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


huwilkes:
Is this supposed to be applied to the HUD? There aren't any changes to the HUD in this PR yet.

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -31,4 +63,16 @@ extension View {
...
     }
+
+    /// Ensures that the modified View has equal height and width values based on the
+    /// maximum of either dimension considering its dynamic content size.
+    /// Minimum and maximum cap values need to be specified.
+    /// - Parameters:
+    ///   - minSize: Minimum heigh/width value of the view.
+    ///   - maxSize: Maximum heigh/width value of the view.
+    /// - Returns: Modified view in a square shaped format.
+    func squareShaped(minSize: CGFloat, maxSize: CGFloat) -> some View {
```
huwilkes, line 74: Do we have plans to apply this to multiple views? If not, should it just be part of the HUD?

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -5,6 +5,38 @@
...
+    func body(content: Content) -> some View {
+        let modifiedContent = HStack {
+            content
+                .alignmentGuide(HorizontalAlignment.center) { (viewDimensions) -> CGFloat in
+                    DispatchQueue.main.async {
+                        self.size = max(size, // ensures the View does not shrink
+                                        min(maxSize,
+                                            max(minSize,
+                                                max(viewDimensions.height,
+                                                    viewDimensions.width))))
```
huwilkes, line 18: I'm not sure if it's just me, but this chain of max/min/max/max is a bit confusing. I understand what's going on after thinking about it for a bit, but it might help if we had a comment with a longer explanation.

rdeassis:
> Is this supposed to be applied to the HUD? There aren't any changes to the HUD in this PR yet.

Correct. I have the HUD changes in my local branches and it will be out for review as soon as I have it tokenized.
The HUD changes depend on these modifiers, so I'm sending separate PRs to make them easier to review.

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
rdeassis: Yeah! That's a good point. So far I believe the HUD is the only control with this requirement. So your suggestion makes a lot of sense even though the addition of this modifier will not make a significant size impact.
I can move it to the HUD folder when I send the PR for the new SwiftUI HUD control.
So far all of these new modifiers being introduced are not yet being used and the goal is just to make them easier to review.

harrieshin:
can we come up with different name than squareshapedview? it seems very retrictive to the future design changes... backplate?

rdeassis:
> can we come up with different name than squareshapedview? it seems very retrictive to the future design changes... backplate?

I think we need to have the word "square" in its name as this is the point of this modifier: restrict the shape of the view to be a square instead of any other rectangle.
This can be used for any view. The HUD just happens to be used as a backplate.

Thoughts or suggestions for other names?

huwilkes (approved):

PR #958, opened 2022-03-30
Update MSFButton to change with keyboard focus

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

I couldn't figure out a way to do this with just SwiftUI yet, so the pure SwiftUI FluentButton still doesn't support this, but at least UIKit usage of our FluentButton does. Looks like this approach only changes colors for keyboard focus, so VoiceOver is unchanged.

### Verification

| MSFButtonFocus_before_light | MSFButtonFocus_after_light_pressed |
| MSFButtonFocus_before_light_vo | MSFButtonFocus_after_light_vo |
| MSFButtonFocus_before_dark | MSFButtonFocus_after_dark_pressed |
| MSFButtonFocus_before_dark_vo | MSFButtonFocus_after_dark_vo |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/958)

### DIFF
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -76,7 +76,7 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
     var action: () -> Void
     @Published var image: UIImage?
     @Published var disabled: Bool?
-    @Published var focused: Bool?
+    @Published var isFocused: Bool = false
     @Published var text: String?
     @Published var size: MSFButtonSize
     @Published var style: MSFButtonStyle
@@ -89,15 +89,6 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
         }
         set {
             disabled = newValue
-        }
-    }
-
-    var isFocused: Bool {
-        get {
-            return focused ?? false
-        }
-        set {
-            focused = newValue
         }
     }
 
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -125,16 +125,11 @@ struct FluentButtonBody: View {
             textColor = tokens.textColor.disabled
             borderColor = tokens.borderColor.disabled
             backgroundColor = tokens.backgroundColor.disabled
-        } else if isPressed {
+        } else if isPressed || isFocused {
             iconColor = tokens.iconColor.pressed
             textColor = tokens.textColor.pressed
             borderColor = tokens.borderColor.pressed
             backgroundColor = tokens.backgroundColor.pressed
-        } else if isFocused {
-            iconColor = tokens.iconColor.hover
-            textColor = tokens.textColor.hover
-            borderColor = tokens.borderColor.hover
-            backgroundColor = tokens.backgroundColor.hover
         } else {
             iconColor = tokens.iconColor.rest
             textColor = tokens.textColor.rest
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -76,6 +76,7 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
     var action: () -> Void
     @Published var image: UIImage?
     @Published var disabled: Bool?
+    @Published var focused: Bool?
     @Published var text: String?
     @Published var size: MSFButtonSize
     @Published var style: MSFButtonStyle
@@ -88,6 +89,15 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
         }
         set {
             disabled = newValue
+        }
+    }
+
+    var isFocused: Bool {
+        get {
+            return focused ?? false
+        }
+        set {
+            focused = newValue
         }
     }
 
@@ -112,6 +122,7 @@ struct FluentButtonBody: View {
 
     var body: some View {
         let isDisabled = !isEnabled
+        let isFocused = state.isFocused
         let isFloatingStyle = tokens.style.isFloatingStyle
         let shouldUsePressedShadow = isDisabled || isPressed
         let iconColor: DynamicColor
@@ -128,6 +139,11 @@ struct FluentButtonBody: View {
             textColor = tokens.textColor.pressed
             borderColor = tokens.borderColor.pressed
             backgroundColor = tokens.backgroundColor.pressed
+        } else if isFocused {
+            iconColor = tokens.iconColor.hover
+            textColor = tokens.textColor.hover
+            borderColor = tokens.borderColor.hover
+            backgroundColor = tokens.backgroundColor.hover
         } else {
             iconColor = tokens.iconColor.rest
             textColor = tokens.textColor.rest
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -11,6 +11,15 @@ import UIKit
 @objc open class MSFButton: ControlHostingView,
                             UIGestureRecognizerDelegate {
 
+    open override func didUpdateFocus(in context: UIFocusUpdateContext, with coordinator: UIFocusAnimationCoordinator) {
+        let style = state.style
+        guard style == .primary || style == .accentFloating, let stateImpl = state as? MSFButtonStateImpl else {
+            return
+        }
+
+        stateImpl.isFocused = !stateImpl.isFocused
```
huwilkes, line 20: I originally tried `self == context.nextFocusedView` like with the other button, but `self` was never the next or previous focused view.

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -128,6 +139,11 @@ struct FluentButtonBody: View {
             textColor = tokens.textColor.pressed
             borderColor = tokens.borderColor.pressed
             backgroundColor = tokens.backgroundColor.pressed
+        } else if isFocused {
+            iconColor = tokens.iconColor.hover
+            textColor = tokens.textColor.hover
+            borderColor = tokens.borderColor.hover
+            backgroundColor = tokens.backgroundColor.hover
```
huwilkes, line 146: I think hover isn't quite different enough, it can be a bit hard still to tell if the focus moved or where it is.

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -11,6 +11,15 @@ import UIKit
 @objc open class MSFButton: ControlHostingView,
                             UIGestureRecognizerDelegate {
 
+    open override func didUpdateFocus(in context: UIFocusUpdateContext, with coordinator: UIFocusAnimationCoordinator) {
+        let style = state.style
+        guard style == .primary || style == .accentFloating, let stateImpl = state as? MSFButtonStateImpl else {
```
rdeassis, line 16: Instead of doing this type cast, you can create an internal constant of type MSFButtonStateImpl to store that reference directly:

`let stateImpl: MSFButtonStateImpl`

```
    @objc public init(style: MSFButtonStyle = .secondary,
                      size: MSFButtonSize = .large,
                      action: ((_ sender: MSFButton) -> Void)?) {
        self.action = action
        let buttonView = FluentButton(style: style,
                                      size: size,
                                      action: {})
        stateImpl = buttonView.state
...
    }
```

And make the state property a computed one that return this reference as the MSFButtonState protocol:

```
    @objc public var state: MSFButtonState {
        return stateImpl
    }
```

ios/FluentUI/Vnext/Button/MSFButton.swift
rdeassis: I looked into this and that's because the view being focused is the first subview of self.

Here are some dumps of my debugger output:

**First event:**
------------------------------

(lldb) po context.previouslyFocusedView
▿ Optional<UIView>
  ▿ some : <_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: **0x61800004d880**; frame = (0 0; 60 60); gestureRecognizers = <NSArray: 0x604000218650>; layer = <CALayer: 0x6030001c96f0>>

(lldb) po context.nextFocusedView
▿ Optional<UIView>
  ▿ some : <_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: 0x618000044880; frame = (0 0; 93.6667 60.3333); focused = YES; gestureRecognizers = <NSArray: 0x604000241a50>; layer = <CALayer: 0x6030001a1580>>

(lldb) po self.subviews[0]
<_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: **0x61800004d880**; frame = (0 0; 60 60); gestureRecognizers = <NSArray: 0x604000218650>; layer = <CALayer: 0x6030001c96f0>>


**Second event...**
------------------------------

(lldb) po context.previouslyFocusedView
▿ Optional<UIView>
  ▿ some : <_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: 0x61800004d880; frame = (0 0; 60 60); gestureRecognizers = <NSArray: 0x604000218650>; layer = <CALayer: 0x6030001c96f0>>

(lldb) po context.nextFocusedView
▿ Optional<UIView>
  ▿ some : <_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: **0x618000044880**; frame = (0 0; 93.6667 60.3333); focused = YES; gestureRecognizers = <NSArray: 0x604000241a50>; layer = <CALayer: 0x6030001a1580>>

(lldb) po self.subviews[0]
<_TtGC7SwiftUI14_UIHostingViewVS_7AnyView_: **0x618000044880**; frame = (0 0; 93.6667 60.3333); focused = YES; gestureRecognizers = <NSArray: 0x604000241a50>; layer = <CALayer: 0x6030001a1580>>

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -76,6 +76,7 @@ class MSFButtonStateImpl: NSObject, ObservableObject, ControlConfiguration, MSFB
     var action: () -> Void
     @Published var image: UIImage?
     @Published var disabled: Bool?
+    @Published var focused: Bool?
```
rdeassis, line 79: Can this be of type Bool and have a default to false?

`@Published var focused: Bool = false`

Then we could discard the computed property (as we're not exposing them in the protocol for ObjC consumption).

ios/FluentUI/Vnext/Button/Button.swift
rdeassis: I think it's worth confirming these colors with the design team. Including the other question I asked about why just some styles that are getting focus colors.

ios/FluentUI/Vnext/Button/MSFButton.swift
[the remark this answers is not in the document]
huwilkes: The other styles don't have any issue showing keyboard focus, just the filled ones. Same as the UIKit button.

ios/FluentUI/Vnext/Button/MSFButton.swift
huwilkes: Looks like the focus is going to the hostingController's view, then. I've updated it to check if the *focusedView is a subview. Do we want it to check the hostingController's view directly, instead? We'd have to expose the hostingController as internal if we did.

ios/FluentUI/Vnext/Button/Button.swift
huwilkes: I've reached out, we'll see what they say

rdeassis (approved):

ios/FluentUI/Vnext/Button/Button.swift
huwilkes: They agreed with pressed for now, possibly revisit in the future.

harrieshin: is there a github ticket for design team to make sure it is part of the fluent 2 design spec?

PR #965, opened 2022-04-13
Implement NotificationView in Pure SwiftUI

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR implements the pure SwiftUI implementation of the Notification View leveraging the new SwiftUI PresentationModifier.

### Verification

https://user-images.githubusercontent.com/31874971/163114136-cd8070f5-adb0-41ea-ad6b-82e3fe7c4047.mov

### Pull request checklist

This PR has considered:
- [ ] VoiceOver and Keyboard Accessibility - working through some issues now! will address accessibility in a future CR

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/965)


ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -78,6 +96,18 @@ public struct NotificationViewSwiftUI: View, ConfigurableTokenizedControl {
...
     }
 
+    private func presentAnimated() {
+        withAnimation(.spring(response: tokens.style.animationDurationForShow, dampingFraction: tokens.style.animationDampingRatio, blendDuration: 0)) {
+            yOffsetFromBottom = 0
+        }
+    }
+
+    private func dismissAnimated() {
+        withAnimation(.spring(response: tokens.style.animationDurationForHide, dampingFraction: tokens.style.animationDampingRatio, blendDuration: 0)) {
```
rdeassis, line 106: If this animation is going to be the same in multiple places, it could be a computed property to prevent duplication:

```
    private var presentationAnimation: Animation {
        .spring(response: tokens.style.animationDurationForShow,
                dampingFraction: tokens.style.animationDampingRatio,
                blendDuration: 0)
    }
```

ios/FluentUI/Extensions/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,24 @@
...
+    weak var value: UIViewController?
+}
+
+struct ViewControllerKey: EnvironmentKey {
+    static var defaultValue: ViewControllerHolder {
+        return ViewControllerHolder(value: UIApplication.shared.windows.first?.rootViewController)
+    }
+}
+
+public extension EnvironmentValues {
```
harrieshin, line 19: why is the file name uiviewcontroller+extension when the actual externsion is evnironment values?

ios/FluentUI/Vnext/Notification/NotificationModifiers.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+public extension View {
```
harrieshin, line 8: it shouldn't be a public extension of every view to add the ability to present notifications, right?

jeaniehuynh: can you elaborate? do you mean that we should have a protocol that views would have to conform to in order to be able to present notifications and instead extend off that protocol?

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
```diff
@@ -21,6 +21,7 @@ struct PresentationModifier<PresentedView: View>: ViewModifier {
             .allowsHitTesting(isPresented && isBlocking)
 
             presentedView
+                .clipped()
```
rdeassis, line 24: Shouldn't this modifier be applied after the frame is defined? (after line 28)

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -31,4 +31,25 @@ extension View {
 
         return AnyView(self)
     }
+
+    /// Measures the size of a view, monitors when its size is updated, and takes a closure to be called when it does
+    /// - Parameter onChange: Block to be performed on size change
+    /// - Returns The modified view.
+    func measureSize(onChange: @escaping (CGSize) -> Void) -> some View {
+      background(
```
rdeassis, line 39: Can you create a struct that implements ViewModifier and just apply the it in the logic of this func? (and name it OnSizeChangeViewModifier as per the comment above)

Reference:
https://github.com/microsoft/fluentui-apple/blob/57bbca9b3a48078b1bab6b46907548aa439110bd/ios/FluentUI/Vnext/HUD/HUDModifiers.swift#L52

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
 /// View that represents the Notification.
 public struct NotificationViewSwiftUI: View, ConfigurableTokenizedControl {
     @Environment(\.horizontalSizeClass) var horizontalSizeClass: UserInterfaceSizeClass?
-    @Environment(\.swiftUIInsets) private var safeAreaInsets: EdgeInsets
     @Environment(\.fluentTheme) var fluentTheme: FluentTheme
     @ObservedObject var state: MSFNotificationStateImpl
+    @Binding var isPresented: Bool
+    @State private var dismissedOffset: CGFloat = 0
+    @State private var yOffsetFromBottom: CGFloat = 0
```
rdeassis, line 49: nit: I'm thinking that **bottomOffset** is a simpler name for this state var (since bottom implies that it's the Y coordinate).

On the same logic, I think the variable below could be something like **bottomOffsetForDismissedState** or something like that?

Thoughts?

ios/FluentUI/Vnext/Notification/NotificationModifiers.swift
rdeassis: I don't think we should have a specific protocol to limit the presentation of the NotificationView unless there is a requirement from that view.
Creating a public extension for View is how we ensure the NotificationView can be displayed on top of any SwiftUI view.

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
...
+                .onTapGesture {
+                    if let messageAction = state.messageButtonAction, let dismissAction = state.dismissAction {
+                        isPresented = false
+                        dismissAction()
+                        messageAction()
+                    }
+                }
+                .frame(width: state.style.isToast && horizontalSizeClass == .regular ? width / 2 : width - (2 * tokens.presentationOffset), alignment: .bottom)
+                .background(
+                    RoundedRectangle(cornerRadius: tokens.cornerRadius)
```
rdeassis, line 100: Can you store properties that are used more than once in a constant?

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
...
+            let width = geometryReader.size.width
+            innerContents
+                .onTapGesture {
+                    if let messageAction = state.messageButtonAction, let dismissAction = state.dismissAction {
+                        isPresented = false
+                        dismissAction()
+                        messageAction()
+                    }
+                }
+                .frame(width: state.style.isToast && horizontalSizeClass == .regular ? width / 2 : width - (2 * tokens.presentationOffset), alignment: .bottom)
```
rdeassis, line 98: Is there a way to make the width param more concise and easier to read?

I'm one way I can think of is to initialize the width constant in line 89 with a closure and have all this logic there.

```
            let width: CGFloat = {
                ...
            }()
```

Thoughts?

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
...
+                )
+                .onChange(of: isPresented, perform: { present in
+                    if present {
+                        presentAnimated()
+                    } else {
+                        dismissAnimated()
+                    }
+                })
+                .padding(.bottom, tokens.bottomPresentationPadding)
+                .measureSize { viewSize in
```
rdeassis, line 123: Once this is changed to **.onSizeChanged**, I believe **newSize** is a good name for the parameter of the closure (instead of viewSize).

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
...
+                    dismissedOffset = viewSize.height + (tokens.ambientShadowOffsetY / 2)
+                    if !isPresented {
```
rdeassis, line 125: I think it's worth a comment here explaining why we update the offset only if the View is not presenting. (to compensate for the the increased height of the notification view, etc...)

Thoughts?

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
```diff
@@ -45,41 +42,98 @@ import SwiftUI
...
+                        yOffsetFromBottom = dismissedOffset
+                    }
+                }
+                .frame(maxHeight: .infinity, alignment: .bottom)
+                .offset(y: yOffsetFromBottom)
+                .position(x: geometryReader.frame(in: .local).midX, y: geometryReader.frame(in: .local).midY)
```
rdeassis, line 131: nit: break the params into multiple lines

rdeassis (commented):
Added a few more comments (a lot of them are nits or minor changes).
This is looking great!
Thanks a lot for this work!

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
jeaniehuynh: Sure! Though I'm curious as to why, is it just easier to read or is there a performance difference?

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
jeaniehuynh: Sounds good to me!

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
jeaniehuynh: Looks good to me!

ios/FluentUI/Core/SwiftUI+ViewPresentation.swift
jeaniehuynh: Hm, I thought it was the presentedView (the notification in this case) that we wanted to be clipped if it was outside of the ZStack versus the whole ZStack being clipped (though after testing both ways, I don't see any differences)

rdeassis: I think it's a bit confusing and I ended up going to the [clipped(antialiased:) documentation](https://developer.apple.com/documentation/swiftui/path/clipped(antialiased:)) again:

My understanding is that it will make the ZStack to clip everything out of its frame:
`A view that clips this view to its bounding frame.`

The counterpart API in UIKit ([clipsToBounds](https://developer.apple.com/documentation/uikit/uiview/1622415-clipstobounds)) has a clearer explanation:
`Setting this value to true causes subviews to be clipped to the bounds of the receiver. If set to false, subviews whose frames extend beyond the visible bounds of the receiver are not clipped. The default value is false.`

I did test it myself and observed the same as you did (no difference). I believe that is probably because SwiftUI understands that the bounding frame of the **presentedView** is the ZStack frame.

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
rdeassis: Not for any perf difference as far as I'm aware. Although I've seen clipping animations not working property in the past when using a func returning the clip shape as opposed to just creating the shape in a View struct.

I believe SwiftUI needs to keep track of the structs to decide what to refresh/redraw or animate.

Not too long ago we decided to follow the standard which is creating these modifiers as structs that implement the protocol and leaving the func just as an easy way to implement it.

ios/FluentUI/Vnext/Notification/NotificationView_SwiftUI.swift
jeaniehuynh: Will never say no to more documentation 😄

ios/FluentUI/Vnext/Notification/NotificationModifiers.swift
harrieshin: seems like a unnecessary taxing on all swiftuiview to have this capability. when in reality they can just call the present notificatonview themselves? what's the overall size impact on this?

rdeassis: @harrieshin creating a func in a View extension is the standard way of creating a SwiftUI modifier. We have to make an extension to the SwiftUI View to ensure that client apps can present the notification on top of the View they want.

The size impact is not significant as it's only about one new public func being introduced.

rdeassis (approved):

PR #967, opened 2022-04-14
Improve bottom sheet handling of low available height

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There can be cases where the height available to the bottom sheet is less than even the collapsed height. We're not handling this gracefully right now, letting the sheet go off screen and not dimming the content behind it.

The system sheet bypasses this issue by just going full screen whenever we're in compact height. Our sheet is more flexible and might not always have the full screen height available to it, so we can't rely on this metric (client could restrict our height arbitrarily even in regular screen height scenario).

To solve this:
- When the collapsed sheet height is sufficiently close to the max height, go into restricted height mode - the threshold for this is a constant and we might tweak it
- In this mode, collapsed state is effectively skipped and goes straight to expanded (if the sheet is expandable), because it doesn't make sense to have states that are just a few pts from each other.
- At the same time, ensure that both expanded and collapsed state never exceed the max available height.

While we're here - simplify our state animator to use the same dimming view / content alpha updating functions as everything else.

### Verification

Before

https://user-images.githubusercontent.com/3610850/163442271-0d7be647-fedd-43db-96da-24a60e4197e2.mov

After

https://user-images.githubusercontent.com/3610850/163442390-26023ea3-72c3-4610-966b-df771ec377d1.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/967)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -526,7 +526,7 @@ public class BottomSheetController: UIViewController {
             // -------- fully undimmed (0.0)-----------
             // ----------------------------------------
             // ---------------------------------------- <--- high offset
-            
+
             let currentOffset = currentSheetVerticalOffset
             let highestDimmedOffset = offset(for: .expanded)
             let lowestUndimmedOffset = isHeightRestricted ? offset(for: .hidden) : offset(for: .collapsed)
@@ -704,7 +704,7 @@ public class BottomSheetController: UIViewController {
             }
             strongSelf.bottomSheetView.frame = strongSelf.sheetFrame(offset: targetVerticalOffset)
             strongSelf.sheetLayoutGuideTopConstraint.constant = targetVerticalOffset
-            
+
             strongSelf.updateDimmingViewAlpha()
             strongSelf.updateExpandedContentAlpha()
             strongSelf.view.layoutIfNeeded()
@@ -825,16 +825,16 @@ public class BottomSheetController: UIViewController {
         let idealHeight = currentResizingHandleHeight + visibleContentHeight + view.safeAreaInsets.bottom
         return min(idealHeight, maxSheetHeight)
     }
-    
+
     private var maxSheetHeight: CGFloat {
         let maxHeight: CGFloat
-        
+
         if view.frame != .zero {
             maxHeight = view.frame.height - view.safeAreaInsets.top - Constants.minimumTopExpandedPadding
         } else {
             maxHeight = Constants.defaultMaxSheetHeight
         }
-        
+
         return maxHeight
     }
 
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -858,6 +886,11 @@ public class BottomSheetController: UIViewController {
 
         static let maxSheetWidth: CGFloat = 610
         static let minSheetWidth: CGFloat = 300
+        static let defaultMaxSheetHeight: CGFloat = 600
+        
+        // When the difference in collapsed height and max sheet height is less than this,
+        // we go into height restricted mode which skips the collapsed state and adjusts dimming.
+        static let heightRestrictedThreshold: CGFloat = 50
```
huwilkes, line 893: Where did 50 come from?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -510,19 +508,37 @@ public class BottomSheetController: UIViewController {
...
+            // ----------------------------------------
+            // -------- fully undimmed (0.0)-----------
+            // ----------------------------------------
+            // ---------------------------------------- <--- high offset
+            
             let currentOffset = currentSheetVerticalOffset
-            let collapsedOffset = offset(for: .collapsed)
-            let expandedOffset = offset(for: .expanded)
+            let highestDimmedOffset = offset(for: .expanded)
+            let lowestUndimmedOffset = isHeightRestricted ? offset(for: .hidden) : offset(for: .collapsed)
```
huwilkes, line 532: Would it be better to make the alpha independent of the offset when the height is restricted?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: Magic number. The threshold has to be some number and this is what I came up with from testing of what distance feels sufficiently close to where a transition doesn't really make sense anymore.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
lcapkovic: This is in preparation for my next PR, which will allow swiping further down to go to hidden state. So we still need to accommodate interactive transition of alpha between .expanded and .hidden.

huwilkes (approved):

huwilkes:
Looks like you have some whitespace lint error

lcapkovic:
>

lcapkovic:
> Looks like you have some whitespace lint error

Yeah... new machine, XCode not configured properly.

huwilkes (approved):

PR #968, opened 2022-04-15
[vNext Tokens] Update List selection style to match PopupMenu selection

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated the List to turn the text and trailing item Brand rest colored on selection and toggle a checkmark, rather than just toggle the highlighted background color.

### Verification

| ListSelectionStyle_before | ListSelectionStyle_after2 |

Also, it does turn green when the theme is set to green:
ListSelectionStyle_after_green

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/968)

### DIFF
--- a/ios/FluentUI/Vnext/List/List.swift
+++ b/ios/FluentUI/Vnext/List/List.swift
@@ -165,7 +165,7 @@ class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, Control
             preconditionFailure("Index is out of bounds")
         }
         let cell = MSFListCellStateImpl()
-        cells.insert(cell, at: index)
+        cell.selectionStyle = selectionStyle
         cell.onSelectAction = { [weak self] (selectedCell) in
             guard let strongSelf = self,
                   let onSelectAction = strongSelf.onSelectAction else {
@@ -173,6 +173,7 @@ class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, Control
             }
             onSelectAction(selectedCell)
         }
+        cells.insert(cell, at: index)
         return cell
     }
 
@@ -231,6 +232,7 @@ class MSFListStateImpl: NSObject, ObservableObject, MSFListState {
             preconditionFailure("Index is out of bounds")
         }
         let section = MSFListSectionStateImpl()
+        section.selectionStyle = selectionStyle
         section.onSelectAction = { [weak self] (selectedCell) in
             guard let strongSelf = self, strongSelf.allowsSelection else {
                 return
--- a/ios/FluentUI/Vnext/List/List.swift
+++ b/ios/FluentUI/Vnext/List/List.swift
@@ -46,7 +46,7 @@ import SwiftUI
     var sectionCount: Int { get }
 
     /// Configures if the List allows selection.
-    var isSelectable: Bool { get set }
+    var allowsSelection: Bool { get set }
 
     /// Configures the selection style of the List.
     var selectionStyle: MSFListSelectionStyle { get set }
@@ -201,9 +201,9 @@ class MSFListStateImpl: NSObject, ObservableObject, MSFListState {
         return sections.count
     }
 
-    @Published var isSelectable: Bool = false {
+    @Published var allowsSelection: Bool = false {
         didSet {
-            if !isSelectable {
+            if !allowsSelection {
                 if let selectedCell = selectedCellState {
                     selectedCell.isSelected = false
                     selectedCellState = nil
@@ -232,7 +232,7 @@ class MSFListStateImpl: NSObject, ObservableObject, MSFListState {
         }
         let section = MSFListSectionStateImpl()
         section.onSelectAction = { [weak self] (selectedCell) in
-            guard let strongSelf = self, strongSelf.isSelectable else {
+            guard let strongSelf = self, strongSelf.allowsSelection else {
                 return
             }
 
--- a/ios/FluentUI/Vnext/List/List.swift
+++ b/ios/FluentUI/Vnext/List/List.swift
@@ -45,8 +45,11 @@ import SwiftUI
     /// The number of Sections in the List.
     var sectionCount: Int { get }
 
-    /// Configures if the list allows selection
+    /// Configures if the List allows selection.
     var isSelectable: Bool { get set }
+
+    /// Configures the selection style of the List.
+    var selectionStyle: MSFListSelectionStyle { get set }
 
     /// Creates a new Section and appends it to the array of sections in a List.
     func createSection() -> MSFListSectionState
@@ -145,6 +148,14 @@ class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, Control
 
     var style: MSFHeaderStyle
 
+    var selectionStyle: MSFListSelectionStyle = .trailingCheckmark {
+        didSet {
+            for cell in cells {
+                cell.selectionStyle = selectionStyle
+            }
+        }
+    }
+
     func createCell() -> MSFListCellState {
         return createCell(at: cells.endIndex)
     }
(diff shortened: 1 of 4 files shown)


harrieshin:
i am bit confused. for the list item cell, the selection state is usually the left side with the arounded background colored circle with a checkmark?  the accessoryView selection is maybe only applicable for popupmenu case and if clients wants to change the accessory view because they are using the listcell like a dropdown menu?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -268,12 +276,24 @@ struct MSFListCellView: View, ConfigurableTokenizedControl {
     }
 
     var body: some View {
+        let labelColor: Color
+        let sublabelColor: Color
+        let trailingItemColor: DynamicColor
+        if state.isSelected {
+            labelColor = Color(dynamicColor: tokens.labelSelectedColor)
```
edjamesmsft, line 283: I assume it's "because that's what the figma defines", but is there a reason the label and sublabel colors aren't dynamic?

harrieshin:
in the video it seems like the selection checkmark is lagging in disappearing after the alert has been dismissed?

ios/FluentUI/Vnext/List/ListTokens.swift
```diff
@@ -39,40 +39,67 @@ import SwiftUI
...
+    open var labelSelectedColor: DynamicColor { aliasTokens.foregroundColors[.brandRest] }
+
+    /// Defines the color of the leading view of the Cell.
     open var leadingViewColor: DynamicColor { aliasTokens.foregroundColors[.neutral1] }
 
+    /// Defines the color of the sublabel of the Cell.
     open var sublabelColor: DynamicColor { aliasTokens.foregroundColors[.neutral3] }
 
+    /// Defines the color of the sublabel when the Cell is selected.
+    open var sublabelSelectedColor: DynamicColor { aliasTokens.foregroundColors[.brandRest] }
```
edjamesmsft, line 61: Oh interesting. Again for my information mostly, but why are these defined as DynamicColors in the Tokens class, when they're used as just Colors in code?

ios/FluentUI/Vnext/List/ListCell.swift
huwilkes: The label color didn't really need to change, but we were calling `Color(dynamicColor: tokens.sublabelColor)` twice, so I just changed them to be `Color` here instead. `DynamicColor` is mostly an internal class we use to hold multiple different related colors, but if we want to actually use it in the UI we need to change it to either `Color` (for Swift) or `UIColor` (for UIKit (on iOS, Mac would need NSColor, but we don't do that yet)).

huwilkes: Though, I guess technically, `DynamicColor` is a public struct, so it's neither internal nor a class. But, what I meant was it's something we added to group colors together.

ios/FluentUI/Vnext/List/ListTokens.swift
huwilkes: I think my answer to your other question should cover this, but all of the tokens defined as `DynamicColor` need to eventually end up as either `Color` or `UIColor`. I guess a lot of the `UIColor` needs to then also change to `CGColor`, too. If you look at the usage, you should see a bunch of `Color(dynamicColor: )` or `UIColor(dynamicColor: )`

huwilkes:
> i am bit confused. for the list item cell, the selection state is usually the left side with the arounded background colored circle with a checkmark? the accessoryView selection is maybe only applicable for popupmenu case and if clients wants to change the accessory view because they are using the listcell like a dropdown menu?

This is how the popup menu is styled. I think the leading circle for selection is for multi select scenarios?

huwilkes:
> in the video it seems like the selection checkmark is lagging in disappearing after the alert has been dismissed?

I think it's just a combination of me clicking too fast and saving the gif at 10 fps. Here's a gif at 20 fps, and with slow animations.
ListSelectionStyle_slow_after

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -223,6 +223,14 @@ class MSFListCellStateImpl: NSObject, ObservableObject, Identifiable, ControlCon
         return children.count
     }
 
+    var trailingItem: MSFListAccessoryType {
+        if isSelected {
```
rdeassis, line 227: nit: can you use guard !isSelected to return the .checkmark and remove the else block by just leaving the `return accessoryType` below it?

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -223,6 +223,14 @@ class MSFListCellStateImpl: NSObject, ObservableObject, Identifiable, ControlCon
         return children.count
     }
 
+    var trailingItem: MSFListAccessoryType {
```
rdeassis, line 226: Do we need to have this stored in the state object?
What I had in mind is to calculate this on the rendering of the view body instead and store this in a let constant.

ios/FluentUI/Vnext/List/ListCell.swift
```diff
@@ -380,8 +400,6 @@ struct MSFListCellView: View, ConfigurableTokenizedControl {
         var cellContent: some View {
             let children: [MSFListCellStateImpl] = state.children
             let hasChildren: Bool = children.count > 0
-            let horizontalCellPadding: CGFloat = tokens.horizontalCellPadding
-            let leadingViewAreaSize: CGFloat = tokens.leadingViewAreaSize
 
             Button(action: {
                 if hasChildren {
```
rdeassis, line 405: This seems like an unintended consequence of the 2 scenarios:
 - expanding children
 - selecting a cell

https://user-images.githubusercontent.com/68076145/163840211-a8fab788-f765-475a-9ad7-1bba8a33e1b0.mov


I believe this should be addressed in a way that we don't allow selection of parent cells as they're just expanding/collapsing their children.

Thoughts?

ios/FluentUI/Vnext/List/ListTokens.swift
mischreiber: Most of the color classes we deal with in iOS are either toolkit-specific (`UIColor` in UIKit on iOS, `NSColor` in AppKit on macOS, `Color` in SwiftUI on both), or too simple (`CGColor` has no concept of multiple color values for different UI states). So as part of the new token work we created a new platform-agnostic struct called `DynamicColor` that can act as a superset of all needed functionality for all platforms/toolkits. Then we have specific converters that convert `DynamicColor` values to the color type expected in our toolkit of choice.

ios/FluentUI/Vnext/List/ListCell.swift
huwilkes: I can update it for now, but selecting folders with subfolders in Outlook is an example of where exanding / selecting should both be doable.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -200,6 +211,15 @@ class MSFListStateImpl: NSObject, ObservableObject, MSFListState {
             }
         }
     }
+
+    @Published var selectionStyle: MSFListSelectionStyle = .trailingCheckmark {
```
rdeassis, line 215: Does this property and the `isSelectable` above need to be @Published?
It seems to me the only one that needs the property wrapper is the one in the MSFListCellStateImpl because it's directly used to draw/layout the cell View.

ios/FluentUI/Vnext/List/List.swift
```diff
@@ -145,6 +148,14 @@ class MSFListSectionStateImpl: NSObject, ObservableObject, Identifiable, Control
 
     var style: MSFHeaderStyle
 
+    var selectionStyle: MSFListSelectionStyle = .trailingCheckmark {
```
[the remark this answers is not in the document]
huwilkes: The one in the section is mostly there to get it to the cells. The section's cells are private, so either this or a method was needed. I can switch to a method, though.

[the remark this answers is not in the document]
huwilkes: Ah, actually, I think we need to keep this, since an update method wouldn't really work to let us create cells with the same selection style.

rdeassis (approved):

PR #971, opened 2022-04-20
[vNext Tokens] `ControlHostingView` should have an initial frame that matches its subview

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Previously, all views from `ControlHostingContainer` would be properly sized immediately upon initialization; clients often use this size info to inform their layouts. I did not account for this when switching to `ControlHostingView`, so I introduced the need for an additional layout pass before size information would be accurate. This is no problem in our demo app, since we rely heavily on autolayout, but any client that manually inspected our controls' frames would find that their layout would break.

Simple fix: match `ControlHostingView` initial size to the initial size of its contents, instead of waiting for a layout pass to complete.

### Verification

Verified in one of our clients that a broken `Avatar` layout now appears correct. Also did a quick sanity pass inside the test app to ensure no visible change.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/971)

### DIFF
--- a/ios/FluentUI/Core/ControlHostingView.swift
+++ b/ios/FluentUI/Core/ControlHostingView.swift
@@ -56,6 +56,11 @@ open class ControlHostingView: UIView {
         addSubview(hostedView)
         hostedView.translatesAutoresizingMaskIntoConstraints = false
 
+        // Initialize our frame with the hosted view's initial bounds so we don't start as a (0,0,0,0) view.
+        // Future updates will be handled by the autolayout constraints below.
+        hostedView.sizeToFit()
+        self.frame = hostedView.bounds
+
         let requiredConstraints = [
             hostedView.leadingAnchor.constraint(equalTo: leadingAnchor),
             hostedView.trailingAnchor.constraint(equalTo: trailingAnchor),


ios/FluentUI/Core/ControlHostingView.swift
```diff
@@ -56,6 +56,11 @@ open class ControlHostingView: UIView {
         addSubview(hostedView)
         hostedView.translatesAutoresizingMaskIntoConstraints = false
 
+        // Initialize our frame with the hosted view's initial bounds so we don't start as a (0,0,0,0) view.
+        // Future updates will be handled by the autolayout constraints below.
+        hostedView.sizeToFit()
+        self.frame = hostedView.bounds
```
huwilkes, line 62: Nit: seems like this is more about initializing the hosting view than configuring the hosted view, so maybe the call to set the frame should be in the init instead? Also, I can't leave a comment up there very easily through GitHub, but is there any chance of the intrinsic content size getting out of sync?

huwilkes (approved):

ios/FluentUI/Core/ControlHostingView.swift
mischreiber: > maybe the call to set the frame should be in the init instead?

I originally decided to lump all stuff about `hostingController.hostedView` view sizing and hierarchy in our `init` method, after calling `super.init()`. But that bloated the method quite a bit, so I intentionally broke that code out into this separate helper method. It's only called during `init`, so the effect is similar.

> is there any chance of the intrinsic content size getting out of sync?

Hm I don't believe so--the `intrinsicContentSize` is a generated var, so it will always recalculate whenever it's requested.

rdeassis (approved):

PR #978, opened 2022-04-26
[vNext-tokens] Tokenizing TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change introduces the first step of the TableViewCell tokenization process. Not all TableViewCell tokens have been fully replaced due to the breadth of the control, which will be followed up with subsequent PRs. To detail, this PR includes:

- Adding an extension to TableViewCellDemoController that conforms to DemoAppearanceDelegate for theme demoing
- Adding new tokens to TableViewTokens
- Implementing tokens to TableViewCell by conforming to TokenizedControlInternal (note: not all tokenizable properties have been fully removed for the sake of maintaining scope. They will be removed in the next PR)
- Note: there has been a slight change to the new tokenized TableViewCell due to the updated font types for subtitle and footer properties. This resulted in a font appearance change, which also slightly affected the height of the cell. To resolve this, the Fluent tokens for fonts have been included, but for now have been overridden with the original UIFonts.

What to expect in the next PR:
- Tokenizing subclasses of TableViewCell

### Verification

Running on iPhone 12 Pro, iOS 15.5

| Simulator Screen Shot - iPhone 12 Pro - 2022-05-27 at 11 17 14 | Simulator Screen Shot - iPhone 12 Pro - 2022-05-27 at 11 17 25 |
| Simulator Screen Shot - iPhone 12 Pro - 2022-05-27 at 11 17 51 | Simulator Screen Shot - iPhone 12 Pro - 2022-05-27 at 11 18 19 |

Theme Verification:

https://user-images.githubusercontent.com/22566866/170768545-6f10c29a-3ff8-43be-ab57-83f9062762db.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/978)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -150,8 +150,11 @@ extension TableViewCellDemoController: DemoAppearanceDelegate {
             return DynamicColor(light: GlobalTokens().sharedColors[.brass][.tint50],
                                 dark: GlobalTokens().sharedColors[.brass][.shade40])
         }
-        override var titleFont: FontInfo {
-            return FontInfo(name: "Papyrus", size: 20.0, weight: .regular)
+
+        override var accessoryDisclosureIndicatorColor: DynamicColor {
+            // "Forest"
+            return DynamicColor(light: GlobalTokens().sharedColors[.forest][.tint10],
+                                dark: GlobalTokens().sharedColors[.forest][.shade40])
         }
     }
 }
@@ -207,9 +210,7 @@ extension TableViewCellDemoController {
 
         cell.isInSelectionMode = section.allowsMultipleSelection ? isInSelectionMode : false
 
-        if let overrideTokens = overrideTokens {
-            cell.tableViewCellOverrideTokens = overrideTokens
-        }
+        cell.tableViewCellOverrideTokens = overrideTokens
 
         return cell
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -56,6 +56,8 @@ class TableViewCellDemoController: DemoTableViewController {
 
     private var perControlOverride: Bool = false
 
+    private var overrideTokens: TableViewCellTokens?
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -111,10 +113,6 @@ class TableViewCellDemoController: DemoTableViewController {
 }
 
 extension TableViewCellDemoController: DemoAppearanceDelegate {
-    override func scrollViewDidScroll(_ scrollView: UIScrollView) {
-        perControlOverrideDidChange(isOverrideEnabled: perControlOverride)
-    }
-
     func themeWideOverrideDidChange(isOverrideEnabled: Bool) {
         guard let fluentTheme = self.view.window?.fluentTheme else {
             return
@@ -131,12 +129,8 @@ extension TableViewCellDemoController: DemoAppearanceDelegate {
     }
 
     func perControlOverrideDidChange(isOverrideEnabled: Bool) {
-        perControlOverride = isOverrideEnabled
-        let cells = self.tableView.visibleCells as! [TableViewCell]
-        for cell in cells {
-            let tokens = isOverrideEnabled ? PerControlOverrideTableViewCellTokens() : nil
-            cell.tableViewCellOverrideTokens = tokens
-        }
+        overrideTokens = isOverrideEnabled ? PerControlOverrideTableViewCellTokens() : nil
+        self.tableView.reloadData()
     }
 
     func isThemeWideOverrideApplied() -> Bool {
@@ -211,6 +205,10 @@ extension TableViewCellDemoController {
         cell.bottomSeparatorType = isLastInSection ? .full : .inset
 
         cell.isInSelectionMode = section.allowsMultipleSelection ? isInSelectionMode : false
+
+        if let overrideTokens = overrideTokens {
+            cell.tableViewCellOverrideTokens = overrideTokens
+        }
 
         return cell
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -54,6 +54,8 @@ class TableViewCellDemoController: DemoTableViewController {
 
     private var editButton: UIBarButtonItem?
 
+    private var perControlOverride: Bool = false
+
     override func viewDidLoad() {
         super.viewDidLoad()
 
@@ -109,6 +111,10 @@ class TableViewCellDemoController: DemoTableViewController {
 }
 
 extension TableViewCellDemoController: DemoAppearanceDelegate {
+    override func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        perControlOverrideDidChange(isOverrideEnabled: perControlOverride)
+    }
+
     func themeWideOverrideDidChange(isOverrideEnabled: Bool) {
         guard let fluentTheme = self.view.window?.fluentTheme else {
             return
@@ -125,6 +131,7 @@ extension TableViewCellDemoController: DemoAppearanceDelegate {
     }
 
     func perControlOverrideDidChange(isOverrideEnabled: Bool) {
+        perControlOverride = isOverrideEnabled
         let cells = self.tableView.visibleCells as! [TableViewCell]
         for cell in cells {
             let tokens = isOverrideEnabled ? PerControlOverrideTableViewCellTokens() : nil
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -54,8 +54,6 @@ class TableViewCellDemoController: DemoTableViewController {
 
     private var editButton: UIBarButtonItem?
 
-    private var perControlOverride: Bool = false
-
     private var overrideTokens: TableViewCellTokens?
 
     override func viewDidLoad() {
@@ -151,6 +149,9 @@ extension TableViewCellDemoController: DemoAppearanceDelegate {
             // "Brass"
             return DynamicColor(light: GlobalTokens().sharedColors[.brass][.tint50],
                                 dark: GlobalTokens().sharedColors[.brass][.shade40])
+        }
+        override var titleFont: FontInfo {
+            return FontInfo(name: "Papyrus", size: 20.0, weight: .regular)
         }
     }
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -8,7 +8,7 @@ import UIKit
 
 // MARK: TableViewCellDemoController
 
-class TableViewCellDemoController: UITableViewController {
+class TableViewCellDemoController: DemoTableViewController {
     let sections: [TableViewSampleData.Section] = TableViewCellSampleData.sections
 
     override init(nibName nibNameOrNil: String?, bundle nibBundleOrNil: Bundle?) {
@@ -44,13 +44,15 @@ class TableViewCellDemoController: UITableViewController {
             }
 
             updateNavigationTitle()
-            navigationItem.rightBarButtonItem?.title = isInSelectionMode ? "Done" : "Select"
+            editButton?.title = isInSelectionMode ? "Done" : "Select"
         }
     }
 
     private var styleButtonTitle: String {
         return isGrouped ? "Switch to Plain style" : "Switch to Grouped style"
     }
+
+    private var editButton: UIBarButtonItem?
 
     override func viewDidLoad() {
         super.viewDidLoad()
@@ -61,7 +63,9 @@ class TableViewCellDemoController: UITableViewController {
         tableView.sectionFooterHeight = 0
         updateTableView()
 
-        navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Select", style: .plain, target: self, action: #selector(selectionBarButtonTapped))
+        let editButton = UIBarButtonItem(title: "Select", style: .plain, target: self, action: #selector(selectionBarButtonTapped))
+        navigationItem.rightBarButtonItems?.append(editButton)
+        self.editButton = editButton
 
         toolbarItems = [
             UIBarButtonItem(barButtonSystemItem: .flexibleSpace, target: nil, action: nil),
@@ -101,6 +105,52 @@ class TableViewCellDemoController: UITableViewController {
     private func updateTableView() {
         tableView.backgroundColor = isGrouped ? Colors.tableBackgroundGrouped : Colors.tableBackground
         tableView.reloadData()
+    }
+}
+
+extension TableViewCellDemoController: DemoAppearanceDelegate {
+    func themeWideOverrideDidChange(isOverrideEnabled: Bool) {
+        guard let fluentTheme = self.view.window?.fluentTheme else {
+            return
+        }
+
+        var tokensClosure: ((TableViewCell) -> TableViewCellTokens)?
+        if isOverrideEnabled {
+            tokensClosure = { _ in
+                return ThemeWideOverrideTableViewCellTokens()
+            }
+        }
+
+        fluentTheme.register(controlType: TableViewCell.self, tokens: tokensClosure)
+    }
+
+    func perControlOverrideDidChange(isOverrideEnabled: Bool) {
+        let cells = self.tableView.visibleCells as! [TableViewCell]
+        for cell in cells {
+            let tokens = isOverrideEnabled ? PerControlOverrideTableViewCellTokens() : nil
+            cell.tableViewCellOverrideTokens = tokens
+        }
+    }
+
+    func isThemeWideOverrideApplied() -> Bool {
+        return self.view.window?.fluentTheme.tokenOverride(for: TableViewCell.self) != nil
+    }
+
+    // MARK: - Custom tokens
+    private class ThemeWideOverrideTableViewCellTokens: TableViewCellTokens {
+        override var cellBackgroundColor: DynamicColor {
+            // "Berry"
+            return DynamicColor(light: GlobalTokens().sharedColors[.berry][.tint50],
+                                dark: GlobalTokens().sharedColors[.berry][.shade40])
+        }
+    }
+
+    private class PerControlOverrideTableViewCellTokens: TableViewCellTokens {
+        override var cellBackgroundColor: DynamicColor {
+            // "Brass"
+            return DynamicColor(light: GlobalTokens().sharedColors[.brass][.tint50],
+                                dark: GlobalTokens().sharedColors[.brass][.shade40])
+        }
     }
 }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
@@ -519,8 +519,9 @@ class TableViewCellFileAccessoryViewDemoController: DemoTableViewController {
 
         for accessoryView in topAccessoryViews + bottomAccessoryViews {
             if let cell = accessoryView.tableViewCell {
-                cell.paddingLeading = TableViewCell().defaultPaddingLeading + extraPadding
-                cell.paddingTrailing = TableViewCell().defaultPaddingTrailing + extraPadding
+                let tokens = TableViewCellTokens()
+                cell.paddingLeading = tokens.paddingLeading + extraPadding
+                cell.paddingTrailing = tokens.paddingTrailing + extraPadding
             }
         }
     }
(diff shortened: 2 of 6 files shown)


ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -0,0 +1,90 @@
...
+
+     open var labelVerticalMarginForTwoLines: CGFloat { globalTokens.spacing[.small] }
+
+     open var labelVerticalSpacing: CGFloat { globalTokens.spacing[.none] }
+
+     open var minHeight: CGFloat { globalTokens.iconSize[.xxxLarge] }
+
+     open var selectionImageMarginTrailing: CGFloat { globalTokens.spacing[.medium] }
+
+    open var selectionImageSize: CGSize { CGSize(width: globalTokens.iconSize[.medium], height: globalTokens.iconSize[.medium]) }
```
rdeassis, line 67: nit: it seems this one is misaligned

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+     //TODO: Update backgroundGrouped for light/dark/darkElevated
+    open var backgroundGrouped: DynamicColor {
+        .init(light: aliasTokens.backgroundColors[.neutral2].light,
+              dark: aliasTokens.backgroundColors[.neutral1].dark)
+    }
+}
+
+ open class TableViewCellTokens: TableViewTokens {
```
rdeassis, line 21: Worth adding documentation for all of these public facing token definitions. For both tableview and tableviewcell.

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -0,0 +1,90 @@
...
+    open var cellBackgroundColor: DynamicColor { aliasTokens.backgroundColors[.neutral1] }
+
+     //TODO: Replace cellBackgroundGrouped with 2.0 token
```
rdeassis, line 24: Is this part of the fluent2 color updates? I think this should be tracked in ADO instead of TODO items in the code so we can manage it in an easier way. (same question applies for the other TODO items in this file)

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1147,7 +1181,7 @@ open class TableViewCell: UITableViewCell {
         if let trailingAccessoryView = trailingAccessoryView {
             let yOffset = UIScreen.main.roundToDevicePixels(topOffset + (labelSize.height - trailingAccessoryViewSize.height) / 2)
             let availableWidth = textAreaWidth - labelSize.width - leadingAccessoryAreaWidth
-            let leadingMargin = TableViewCell.labelTrailingAccessoryMarginLeading(text: text)
+            let leadingMargin = TableViewCell.labelTrailingAccessoryMarginLeading(text: text, labelAccessoryViewMarginLeading: tokens.labelAccessoryViewMarginLeading)
```
rdeassis, line 1184: nit: can you break these long lines into multiple ones (one for each param)?
(Here and throughout the changes being made.)

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -0,0 +1,90 @@
...
+    open var horizontalSpacing: CGFloat { globalTokens.spacing[.medium] }
+
+     /// The  leading padding in the cell.
+    open var paddingLeading: CGFloat { globalTokens.spacing[.medium] }
+
+     /// The  trailing padding in the cell.
+    open var paddingTrailing: CGFloat { globalTokens.spacing[.medium] }
+}
+
+ open class TableViewCellAccessoryTokens: ControlTokens {
```
rdeassis, line 87: Is this class being used anywhere?

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -0,0 +1,90 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+ import UIKit
+
+ open class TableViewTokens: ControlTokens {
```
rdeassis, line 7: Is there any scenario that will use these tokens as a base class type?
I only see the usage of the TableViewCellTokens subclass throughout the code.
What's the rationale for making this a base class?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1307,7 +1341,15 @@ open class TableViewCell: UITableViewCell {
 
     open override func didMoveToWindow() {
         super.didMoveToWindow()
+        updateTokens()
+    }
+
+    private func updateTokens() {
+        self.tokens = resolvedTokens
```
rdeassis, line 1348: nit: is `self.` redundant here?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -932,11 +953,17 @@ open class TableViewCell: UITableViewCell {
...
     @objc public override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
+        let tokens = TableViewCellTokens()
+        self.tokens = tokens
+
         super.init(style: .default, reuseIdentifier: reuseIdentifier)
         initialize()
     }
 
     public required init?(coder aDecoder: NSCoder) {
+        let tokens = TableViewCellTokens()
```
rdeassis, line 964: shouldn't this be the preconditionFailure?

`preconditionFailure("init(coder:) has not been implemented")`

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -242,18 +213,17 @@ open class TableViewCell: UITableViewCell {
...
-    @objc public static let horizontalSpacing: CGFloat = 16
+    @objc public var horizontalSpacing: CGFloat { tokens.horizontalSpacing }
 
     /// The default leading padding in the cell.
-    @objc public static let defaultPaddingLeading: CGFloat = horizontalSpacing
+    @objc public var defaultPaddingLeading: CGFloat { tokens.paddingLeading }
 
     /// The default trailing padding in the cell.
-    @objc public static let defaultPaddingTrailing: CGFloat = horizontalSpacing
+    @objc public var defaultPaddingTrailing: CGFloat { tokens.paddingTrailing }
```
rdeassis, line 222: Changing this from static to an instance property will break the consumers anyways.
What is the rationale behind it? Is it about retrieving the default values of the tokens of the window associated theme?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
```diff
@@ -519,8 +519,8 @@ class TableViewCellFileAccessoryViewDemoController: DemoTableViewController {
 
         for accessoryView in topAccessoryViews + bottomAccessoryViews {
             if let cell = accessoryView.tableViewCell {
-                cell.paddingLeading = TableViewCell.defaultPaddingLeading + extraPadding
-                cell.paddingTrailing = TableViewCell.defaultPaddingTrailing + extraPadding
+                cell.paddingLeading = TableViewCell().defaultPaddingLeading + extraPadding
+                cell.paddingTrailing = TableViewCell().defaultPaddingTrailing + extraPadding
```
rdeassis, line 523: Couldn't these defaults be retrieved by just creating a tokens class a returning the properties in it? (since it's not being added to a window and will use the default implementation)

ios/FluentUI/Table View/TableViewTokens.swift
sophialee0416: Not right now, but I believe PersonaListView calls for the table background but not the other colors. I thought of isolating the tableview this way so that when I create the tokens for the other cell types (ex. ActionCell) they could also access the tokens this way. I can definitely change these though

ios/FluentUI/Table View/TableViewTokens.swift
sophialee0416: I was planning on using these when tokenizing the TableViewCellFileAccessoryView, which calls for these values. Should I redo how I would call the tokens?

ios/FluentUI/Table View/TableViewTokens.swift
sophialee0416: This is indeed part of the new color updates, I will add tracking info to them

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
mischreiber: Yeah, I would prefer we avoid creating throwaway instances of any view whenever possible (especially two throwaway instances like here!).

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -1070,6 +1072,7 @@
 				B498141321E424920077B48D /* TableViewCell.swift */,
 				B4EF66502294A664007FEAB0 /* TableViewHeaderFooterView.swift */,
 				B444D6B02181403C0002B4D4 /* UITableViewCell+Extension.swift */,
+				2A9745DD281733D700E1A1FD /* TableViewTokens.swift */,
```
mischreiber, line 1075: Nit: alphabetical order

ios/FluentUI/Table View/TableViewTokens.swift
mischreiber: Still looks misaligned in the latest commit?

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: The `static` vars were useful for consumers who wanted to build their own custom UIs that looked like a Fluent `TableViewCell` but was not in fact one. Now that these values are token-backed, I see less value in having them available as public APIs at all, when callers could just as easily query these properties from a token instance.

Thoughts?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -961,7 +989,14 @@ open class TableViewCell: UITableViewCell {
...
 
-        NotificationCenter.default.addObserver(self, selector: #selector(handleContentSizeCategoryDidChange), name: UIContentSizeCategory.didChangeNotification, object: nil)
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(handleContentSizeCategoryDidChange),
+                                               name: UIContentSizeCategory.didChangeNotification, object: nil)
```
mischreiber, line 999: Nit: `object: nil` should be on its own line.

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Looks like this control has historically supported `NSCoder` inits, necessary for nib/storyboard initialization. I'd rather not remove that support if we've already shipped it, since it's unclear who might be relying on it.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -421,47 +420,51 @@ open class TableViewCell: UITableViewCell {
...
     }
 
     private static func customViewSize(from size: CustomViewSize, layoutType: LayoutType) -> CustomViewSize {
         return size == .default ? layoutType.customViewSize : size
     }
 
-    private static func customViewLeadingOffset(isInSelectionMode: Bool, paddingLeading: CGFloat) -> CGFloat {
-        return paddingLeading + selectionModeAreaWidth(isInSelectionMode: isInSelectionMode)
+    private static func customViewLeadingOffset(isInSelectionMode: Bool, tokens: TableViewCellTokens) -> CGFloat {
+            return tokens.paddingLeading + selectionModeAreaWidth(isInSelectionMode: isInSelectionMode, selectionImageMarginTrailing: tokens.selectionImageMarginTrailing, selectionImageSize: tokens.selectionImageSize.width)
```
mischreiber, line 432: Nit: extra padding

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Why do some controls have the preconditionFailure, and others continue with the initialization? @mischreiber @rdeassis  (am a little unfamiliar/confused as I looked through the other controls!)

mischreiber: Honestly, I assume the answer is historical. We haven't prioritized supporting `NSCoder` initialization for new controls, but presumably some older controls did in fact support that.

I don't have a strong preference on what we _should_ do here, but I definitely see the value in consistency one way or another.

ios/FluentUI/Table View/TableViewCell.swift
rdeassis: I agree that these values could be queried directly from the tokens. But in this case they're not necessarily the "default" options, right? It'll return whatever the tokens are for that specific TableViewCell.

Are these vars still here just for the ObjC availability?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -242,18 +213,17 @@ open class TableViewCell: UITableViewCell {
...
 
     /// The default leading padding in the cell.
-    @objc public static let defaultPaddingLeading: CGFloat = horizontalSpacing
+    @objc public var defaultPaddingLeading: CGFloat { tokens.paddingLeading }
 
     /// The default trailing padding in the cell.
-    @objc public static let defaultPaddingTrailing: CGFloat = horizontalSpacing
+    @objc public var defaultPaddingTrailing: CGFloat { tokens.paddingTrailing }
 
+    //TODO: Remove labelVerticalMarginForOneAndThreeLines
```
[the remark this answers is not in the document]
rdeassis: If we're tracking these in ADO, it's probably better to remove the TODO comments and reference this code in the ADO task.

ios/FluentUI/Table View/TableViewCell.swift
rdeassis: Was this one updated? I still see all in one line. (some other instances in this PR as well)

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1396,9 +1568,7 @@ open class TableViewCell: UITableViewCell {
     }
 
     private func updateSelectionImageColor() {
-        if let window = window {
-            selectionImageView.tintColor = isSelected ? Colors.primary(for: window) : Colors.Table.Cell.selectionIndicatorOff
-        }
+        selectionImageView.tintColor = isSelected ? UIColor(dynamicColor: GlobalTokens().brandColors[.primary]) : UIColor(dynamicColor: tokens.selectionIndicatorOffColor)
```
rdeassis, line 1571: You can have just one UIColor init call here by doing this:

```
let tintColor = isSelected ? tokens.globalTokens.brandColors[.primary] : tokens.selectionIndicatorOffColor
selectionImageView.tintColor = UIColor(dynamicColor: tintColor)
```

rdeassis, line 1571: shouldn't the global tokens reference come from the tokens class?

**tokens.globalTokens**.brandColors[.primary] instead of using **GlobalTokens()**

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -13,6 +13,7 @@
 		0AC8A8B227C72D4600D9FAF5 /* MSFPersonaView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AC8A8B127C72D4600D9FAF5 /* MSFPersonaView.swift */; };
 		0ACD82192620453B0035CD9F /* PersonaViewTokens.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0ACD82172620453B0035CD9F /* PersonaViewTokens.swift */; };
 		0AD70F5B26E80A5D008774EC /* CommandBarTokens.swift in Sources */ = {isa = PBXBuildFile; fileRef = 0AD70F5926E80A5D008774EC /* CommandBarTokens.swift */; };
+		2A9745DE281733D700E1A1FD /* TableViewTokens.swift in Sources */ = {isa = PBXBuildFile; fileRef = 2A9745DD281733D700E1A1FD /* TableViewTokens.swift */; };
```
rdeassis, line 16: Seems like there's a conflict in this file.

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
```diff
@@ -519,8 +519,9 @@ class TableViewCellFileAccessoryViewDemoController: DemoTableViewController {
 
         for accessoryView in topAccessoryViews + bottomAccessoryViews {
             if let cell = accessoryView.tableViewCell {
-                cell.paddingLeading = TableViewCell.defaultPaddingLeading + extraPadding
-                cell.paddingTrailing = TableViewCell.defaultPaddingTrailing + extraPadding
+                let tokens = TableViewCellTokens()
+                cell.paddingLeading = tokens.paddingLeading + extraPadding
```
rdeassis, line 523: If we're not using the property mentioned in one of the comments below, shouldn't we just remove it?

`@objc public var defaultPaddingLeading: CGFloat { tokens.paddingLeading }`

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Yes they are only available for static use in ObjC, but unfortunately that makes it difficult to work with tokens

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -201,33 +180,15 @@ open class TableViewCell: UITableViewCell {
...
-        static let textAreaMinWidth: CGFloat = 100
-
-        static let enabledAlpha: CGFloat = 1
-        static let disabledAlpha: CGFloat = 0.35
     }
 
     /**
      The height for the cell based on the text provided. Useful when `numberOfLines` of `title`, `subtitle`, `footer` is 1.
-
+
```
rdeassis, line 191: nit: remove?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -96,35 +93,35 @@ public extension Colors {
     }
 
     // Objective-C support
-    @objc static var tableBackground: UIColor { return Table.background }
-    @objc static var tableBackgroundGrouped: UIColor { return Table.backgroundGrouped }
-    @objc static var tableCellBackground: UIColor { return Table.Cell.background }
-    @objc static var tableCellBackgroundGrouped: UIColor { return Table.Cell.backgroundGrouped }
-    @objc static var tableCellBackgroundSelected: UIColor { return Table.Cell.backgroundSelected }
-    @objc static var tableCellImage: UIColor { return Table.Cell.image }
+    @objc static var tableBackground: UIColor { return surfacePrimary }
```
rdeassis, line 96: With tokens in place, these properties should probably be deprecated.
@mischreiber going forward what's the best practice for the client code that wants to retrieve these token values? Instantiate a default TableViewCellTokens and query the desired token property?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -64,12 +64,9 @@ public extension Colors {
     internal struct Table {
```
rdeassis, line 64: Isn't this entire internal struct "Table" supposed to be replaced by tokens?

Looks like only the following are still used in code:

- Colors.Table.Cell.accessoryDisclosureIndicator
- Colors.Table.Cell.accessoryDetailButton

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
```diff
@@ -8,7 +8,7 @@ import UIKit
 
 // MARK: TableViewCellDemoController
 
-class TableViewCellDemoController: UITableViewController {
+class TableViewCellDemoController: DemoTableViewController {
```
rdeassis, line 11: nit: The entry point for **TableViewCell** and **TableViewCellFileAccesorryView** should be moved up to the "Vnext Control" section of the demo app.

ios/FluentUI/Table View/TableViewTokens.swift
sophialee0416: Resolved offline: moving tokens under CellTokens

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Resolved offline: users should no longer need the static variables as they are accessible from the tokens.

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Makes sense. @rdeassis going forward I think you're probably correct, though that's a pretty significant change (especially as it pertains to all the stuff in `Colors.swift`).

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -265,9 +214,39 @@ open class TableViewCell: UITableViewCell {
         return paddingLeading
     }
 
+    // MARK: - TableViewCell TokenizedControl
+    public typealias TokenType = TableViewCellTokens
```
mischreiber, line 218: Nit: this line should not be necessary.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -96,35 +93,35 @@ public extension Colors {
...
+    @objc static var tableCellBackgroundSelected: UIColor { return surfaceTertiary }
+    @objc static var tableCellImage: UIColor { return iconSecondary }
 }
 
 // MARK: - TableViewCell
 
 /**
  `TableViewCell` is used to present a cell with one, two, or three lines of text with an optional custom view and an accessory.
-
+
```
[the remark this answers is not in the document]
sophialee0416: Just checked, for some reason ctrl-i reformatted it to this. Removing all those spaces.

PR #982, opened 2022-05-06
Fix bug in calculation of available height for titleView in TableViewHeaderFooterView

adarsh-sridhar:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When the style is set to `.footer`, the `layoutSubviews` method uses a shorter top margin for the `titleYOffset` variable. However, when calculating the available height for the `titleView`, the code assumes that the default top and bottom margins are used regardless of the style. This change fixes the calculation to use the `titleYOffset` value which is set to the correct top margin instead.

### Verification

- Manual verification on iOS Simulator
- All unit tests were run.

| Screen Shot 2022-05-05 at 8 41 43 PM | Screen Shot 2022-05-05 at 8 38 07 PM|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/982)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -396,7 +396,7 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
         switch style {
         case .header, .footer, .headerPrimary:
             titleYOffset = style == .footer ? Constants.titleDefaultBottomMargin : Constants.titleDefaultTopMargin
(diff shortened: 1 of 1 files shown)


mischreiber (approved):
Looks good, but I'd like @sophialee0416, who is currently looking at some Table View stuff, to take a peek as well.

rdeassis:
@adarsh-sridhar Thank you for contributing to FluentUI!

I noticed you're not part of the FluentUI Native group and therefore won't have permissions to merge the PR. Could you reach out on Teams so we can get you onboarded on the group?

PR #983, opened 2022-05-12
Removing tokenization of TabBarItems

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

After lengthy discussions on how we want to tokenize compound controls through the entire codebase, it was determined that non-public controls, i.e., ones not explicitly defined in the Figma file and not exposed through public interfaces, should not be tokenized. As such, this change removes the public tokens for the TabBarItems, leaving only the three TabBar tokens remaining.

### Verification

Manually tested via Simulator.

| TokenizeTabBarAfter | UnTokenizeTabBarItemsDemo |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/983)


ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -174,6 +154,25 @@ class TabBarItemView: UIControl, TokenizedControlInternal {
         updateLayout()
     }
 
+    private struct Constants {
```
huwilkes, line 157: I think this isn't quite what we had in mind. I think that rather than having all of these constants, we still want these values to come from tokens. But they should be included in the parent's tokens, then set on the child control by that parent.

edjamesmsft: Better than my interpretation by a long shot!

rdeassis: Yes. My understanding of the decision we made offline aligns with what @huwilkes mentioned:

We won't have the tokenization infrastructure in each of the TabBarItemView (such as the defaultTokens and overrideTokens). But the tokens for the items will still be defined in the TabBarViewTokens class.

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -255,7 +262,6 @@ class TabBarItemView: UIControl, TokenizedControlInternal {
         imageView.highlightedImage = item.selectedImage(isInPortraitMode: isInPortraitMode, labelIsHidden: titleLabel.isHidden)
 
         if isInPortraitMode {
-            container.axis = .vertical
```
edjamesmsft, line 258: This was a mistake - fixing shortly

ios/FluentUI/Tab Bar/TabBarItemTokens.swift
```diff
@@ -0,0 +1,73 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Design token set for the `TabBarItem`.
+open class TabBarItemTokens: ControlTokens {
```
rdeassis, line 9: Do we need this class to be open?
There's currently no need for the client apps to subclass it as the TabBarItemView is not public and we have not public property of this type.

It seems more lightweight to have these properties in the TabBarViewTokens as it was before.

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -6,15 +6,15 @@
...
         overrideTokens = tokens
         return self
     }
     let item: TabBarItem
 
-    var defaultTokens: TabBarTokens = .init()
-    var tokens: TabBarTokens = .init()
-    var overrideTokens: TabBarTokens? {
+    var defaultTokens: TabBarItemTokens = .init()
+    var tokens: TabBarItemTokens = .init()
```
rdeassis, line 16: I think the decision on private subcontrols is that they would not be tokenized and the tokens that define their appearance would be defined in the container public control (TabBarView in this case).

It seems to be a more lightweight approach than having each private control support the tokens infrastructure.

edjamesmsft: I've made a quick change to demonstrate why I feel this is a problematic approach, especially for this control. To me, it it ends up being a lot more cumbersome than a nicely encapsulated Tokens class that is reset automatically on token and theme changes.

PR #984, opened 2022-05-13
Add Fluent Notification Attributed String Support

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds supports for attributed strings in SwiftUI. This is done by adding optional parameters for an attributed message/title to the init of the Fluent Notification. If provided, the notification will use a UILabel hosted by a UIViewRepresentable instead of a SwiftUI Text object. This change also adds ways to showcase this support in the demo controllers.

| image | image |
| image | image |
| image | image |
| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/984)


ios/FluentUI/Vnext/Notification/FluentNotification.swift
```diff
@@ -274,11 +274,11 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
 
     private var hasSecondTextRow: Bool {
         let isToast = state.style.isToast
-        guard let attributedTitle = state.attributedTitle, attributedTitle.string != "" else {
+        guard state.attributedTitle != nil else {
             return isToast && state.title != ""
```
rdeassis, line 278: nit: `!isEmpty` instead `!= ""`

ios/FluentUI/Vnext/Notification/FluentNotification.swift
```diff
@@ -42,24 +48,31 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
...
     ///   - title: Optional text to draw above the message area.
+    ///   - attributedTitle: Optional attributed text to draw above the message area. If set, it will override the title parameter.
     ///   - image: Optional icon to draw at the leading edge of the control.
     ///   - actionButtonTitle:Title to display in the action button on the trailing edge of the control.
     ///   - actionButtonAction: Action to be dispatched by the action button on the trailing edge of the control.
     ///   - messageButtonAction: Action to be dispatched by tapping on the toast/bar notification.
     public init(style: MSFNotificationStyle,
                 shouldSelfPresent: Bool = true,
-                message: String,
+                message: String = "",
```
rdeassis, line 62: If you're changing the message property to be optional, I think this should follow and the default value should be nil like the other ones.

Thoughts?

rdeassis (approved):

PR #986, opened 2022-05-18
[PreviewCard] Placeholder image over PreviewCard

lewist13:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Positioned a placeholder image over flat and elevated Preview Cards
Adjusted height of PreviewCard for 16:9 adherence

### Verification

| flatelevatedcards_resized | prevcard_w_placeholderbordered_resized | 

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/986)


ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/PreviewCard/W11_Placeholder.imageset/Contents.json
```diff
@@ -0,0 +1,21 @@
+{
+  "images" : [
+    {
+      "filename" : "Windows_11.pdf",
```
edjamesmsft, line 4: Is there a reason this is a .pdf file, rather than another file format?

ios/FluentUI/PreviewCard/PreviewCard.swift
```diff
@@ -8,16 +8,18 @@ import SwiftUI
...
+    /// The UIView representing the PreviewCard.
     @objc open var view: UIView {
         return hostingController.view
     }
     /// Creates a new MSFPreviewCard instance.
     /// - Parameters:
     ///   - theme: The FluentUIStyle instance representing the theme to be overriden for this PreviewCard.
-    @objc public init(theme: FluentUIStyle?, isElevated: Bool = false) {
+    ///   - isElevated: Boolean determining if PreviewCard is flat or elevated.`
+    ///   - placeholderImage: The UIImage assigning an image over PreviewCard
```
edjamesmsft, line 19: I may just not understand the assignment, but is it really a "placeholder" image? To me "placeholder" implies temporary or not the real image. Is that the intent?

ios/FluentUI/PreviewCard/PreviewCard.swift
```diff
@@ -31,13 +33,16 @@ import SwiftUI
...
+    var placeholderImage: UIImage
     @Environment(\.theme) var theme: FluentUIStyle
     @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
     /// Creates inner PreviewCard view.
     @ViewBuilder
     var innerContents: some View {
         HStack {
             Spacer()
+                /// Creates placeholder image over PreviewCard.
+            Image(uiImage: placeholderImage)
```
huwilkes, line 45: Do we want to add some sort of clipping shape to this so that it'll also have rounded corners / be the same shape as the card?

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/PreviewCard/W11_Placeholder.imageset/Contents.json
huwilkes: Generally we prefer having pdfs and having xcode generate the different sizes for us.
I think this one isn't quite set up correctly, but I'll have to look at another file to be sure. Less of an issue since it's just in our demo, though.
https://microsoft.github.io/apple-ux-guide/iOSImageFileFormat.html

lewist13: > Is there a reason this is a .pdf file, rather than another file format?

The SVG image would show as half of the image when exported. Noticed other images were uploaded in assets using the .pdf file format - attempted it and it worked without any issues . Should I retry SVG exporting?

mischreiber: Actually do you think it could be a JPEG or PNG? The PDF is ~5.8 MB, but it's ultimately a raster (non-vector) image so one of those formats would probably be a lot smaller.

ios/FluentUI/PreviewCard/PreviewCard.swift
mischreiber: Makes sense from the demo app's perspective, but the `PreviewCard` itself doesn't know the difference. So with that in mind it may make more sense to just call it `image` and then it can be called a placeholder in the demo app.

mischreiber (approved):

ios/FluentUI/PreviewCard/PreviewCard.swift
```diff
@@ -31,16 +33,21 @@ import SwiftUI
...
+    var image: UIImage
     @Environment(\.theme) var theme: FluentUIStyle
     @Environment(\.windowProvider) var windowProvider: FluentUIWindowProvider?
     /// Creates inner PreviewCard view.
     @ViewBuilder
     var innerContents: some View {
         HStack {
             Spacer()
-                /// Sets the height and width of the Preview Card.
+                /// Creates placeholder image over PreviewCard.
```
edjamesmsft, line 44: Nit: still says placeholder here. All three of these comments may not be fully necessary, either, but not a big deal.

PR #987, opened 2022-05-18
Make CommandBar mutable and allow for multiple leading and trailing items.

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change aims to address a few different limitations of the existing CommandBar implementation:
1. The existing CommandBar implementation does not allow for items to be added or removed after initialization, making it impossible to update the items shown without creating an entirely new bar. To fix this, the itemGroups property was made public and a didSet property observer was added which triggers a refresh of the views shown in the stackView. The maps holding the items and their corresponding buttons have to be refreshed when this happens.
2. Any changes made to CommandBarItems will only be reflected in the UI if CommandBar.updateButtonState() is called. Calling this iterates over and updates all of the buttons shown. Ideally, buttons can be updated individually without having to update all of the buttons. To solve this, I've added CommandBarItemDelegate protocol that is implemented by the CommandBar and allows for buttons to be updated individually after properties on the their corresponding CommandBarItem changes.
3. Expand the leadingButton and trailingButton areas to allow for multiple items. To do this, UIStackViews were used and implemented in a similar manner to that of the UIStackView used for the center items in itemGroups. New item-to-button mappings are created to store the items and their corresponding buttons. The CommandBar.leadingItemGroups and CommandBar.trailingItemGroups properties are exposed and work in the same way that itemsGroup property works. The biggest downside that I see to this approach is that there are now 3 mappings being maintained for the CommandBar (leading items, center items, trailing items).

### Verification

Before         
Before

After

https://user-images.githubusercontent.com/10938746/169106813-ed886694-f422-41a0-8cd8-15614d9d6327.mov
The video above demos of how the items in the the CommandBar can be refreshed after initialization.

https://user-images.githubusercontent.com/10938746/169105547-6983c781-c982-4b14-96c2-dc766910819f.mov
The video above shows how the the new CommandBarItemDelegate can be used to update individual items without doing a refresh of all the items in the bar.

https://user-images.githubusercontent.com/10938746/169106904-dbb01896-04e1-4031-ab74-79c4cade848d.mov
The video above shows how the items in the leading and trailing sections can be refreshed after initialization.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/987)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
@@ -141,6 +141,8 @@ class CommandBarDemoController: DemoController {
         }
     }
 
+    var defaultCommandBar: CommandBar?
+
     let textField: UITextField = {
         let textField = UITextField()
         textField.translatesAutoresizingMaskIntoConstraints = false
@@ -156,6 +158,84 @@ class CommandBarDemoController: DemoController {
         container.layoutMargins.left = 0
         view.backgroundColor = Colors.surfaceSecondary
 
+        container.addArrangedSubview(createLabelWithText("Default"))
+
+        defaultCommandBar = CommandBar(itemGroups: createItemGroups(), leadingItemGroups: [[newItem(for: .keyboard)]])
+        defaultCommandBar?.backgroundColor = Colors.navigationBarBackground
+        container.addArrangedSubview(defaultCommandBar!)
+
+        let itemCustomizationContainer = UIStackView()
+        itemCustomizationContainer.spacing = 8.0
+        itemCustomizationContainer.axis = .vertical
+        itemCustomizationContainer.backgroundColor = Colors.navigationBarBackground
+
+        itemCustomizationContainer.addArrangedSubview(UIView()) //Spacer
+
+        let refreshButton = Button(style: .tertiaryOutline)
+        refreshButton.setTitle("Refresh 'Default' Bar", for: .normal)
+        refreshButton.addTarget(self, action: #selector(refreshDefaultBarItems), for: .touchUpInside)
+        itemCustomizationContainer.addArrangedSubview(refreshButton)
+
+        let removeTrailingItemButton = Button(style: .tertiaryOutline)
+        removeTrailingItemButton.setTitle("Remove Trailing Button", for: .normal)
+        removeTrailingItemButton.addTarget(self, action: #selector(removeDefaultTrailingBarItems), for: .touchUpInside)
+        itemCustomizationContainer.addArrangedSubview(removeTrailingItemButton)
+
+        let refreshTrailingItemButton = Button(style: .tertiaryOutline)
+        refreshTrailingItemButton.setTitle("Refresh Trailing Button", for: .normal)
+        refreshTrailingItemButton.addTarget(self, action: #selector(refreshDefaultTrailingBarItems), for: .touchUpInside)
+        itemCustomizationContainer.addArrangedSubview(refreshTrailingItemButton)
+
+        let removeLeadingItemButton = Button(style: .tertiaryOutline)
+        removeLeadingItemButton.setTitle("Remove Leading Button", for: .normal)
+        removeLeadingItemButton.addTarget(self, action: #selector(removeDefaultLeadingBarItems), for: .touchUpInside)
+        itemCustomizationContainer.addArrangedSubview(removeLeadingItemButton)
+
+        let refreshLeadingItemButton = Button(style: .tertiaryOutline)
+        refreshLeadingItemButton.setTitle("Refresh Leading Button", for: .normal)
+        refreshLeadingItemButton.addTarget(self, action: #selector(refreshDefaultLeadingBarItems), for: .touchUpInside)
+        itemCustomizationContainer.addArrangedSubview(refreshLeadingItemButton)
+
+        let customizationStackView = UIStackView()
+        customizationStackView.axis = .horizontal
+        customizationStackView.alignment = .fill
+        customizationStackView.distribution = .fillProportionally
+        customizationStackView.addArrangedSubview(createLabelWithText("'+' Item Enabled"))
+        let itemEnabledSwitch: UISwitch = UISwitch()
+        itemEnabledSwitch.isOn = true
+        itemEnabledSwitch.addTarget(self, action: #selector(itemEnabledValueChanged), for: .valueChanged)
+        customizationStackView.addArrangedSubview(itemEnabledSwitch)
+        itemCustomizationContainer.addArrangedSubview(customizationStackView)
+
+        itemCustomizationContainer.addArrangedSubview(UIView()) //Spacer
+
+        container.addArrangedSubview(itemCustomizationContainer)
+
+        container.addArrangedSubview(createLabelWithText("With Fixed Button"))
+
+		let fixedButtonCommandBar = CommandBar(itemGroups: createItemGroups(), leadingItemGroups: [[newItem(for: .copy)]], trailingItemGroups: [[newItem(for: .keyboard)]])
+        fixedButtonCommandBar.backgroundColor = Colors.navigationBarBackground
+        container.addArrangedSubview(fixedButtonCommandBar)
+
+        container.addArrangedSubview(createLabelWithText("In Input Accessory View"))
+
+        let textFieldContainer = UIView()
+        textFieldContainer.backgroundColor = Colors.navigationBarBackground
+        textFieldContainer.addSubview(textField)
+        NSLayoutConstraint.activate([
+            textField.topAnchor.constraint(equalTo: textFieldContainer.topAnchor, constant: 16.0),
+            textField.leadingAnchor.constraint(equalTo: textFieldContainer.leadingAnchor, constant: 16.0),
+            textFieldContainer.bottomAnchor.constraint(equalTo: textField.bottomAnchor, constant: 16.0),
+            textFieldContainer.trailingAnchor.constraint(equalTo: textField.trailingAnchor, constant: 16.0)
+        ])
+
+        container.addArrangedSubview(textFieldContainer)
+
+        let accessoryCommandBar = CommandBar(itemGroups: createItemGroups(), trailingItemGroups: [[newItem(for: .keyboard)]])
+        textField.inputAccessoryView = accessoryCommandBar
+    }
+
+    func createItemGroups() -> [CommandBarItemGroup] {
         let commandGroups: [[Command]] = [
             [
                 .add,
@@ -204,32 +284,7 @@ class CommandBarDemoController: DemoController {
                                           UIAction(title: "Copy Text", image: UIImage(named: "text24Regular"), handler: { _ in })])
         copyItem.showsMenuAsPrimaryAction = true
 
-        container.addArrangedSubview(createLabelWithText("Default"))
-
-        let defaultCommandBar = CommandBar(itemGroups: itemGroups)
-        container.addArrangedSubview(defaultCommandBar)
-
-        container.addArrangedSubview(createLabelWithText("With Fixed Button"))
-
-        let fixedButtonCommandBar = CommandBar(itemGroups: itemGroups, leadingItem: newItem(for: .copy), trailingItem: newItem(for: .keyboard))
-        container.addArrangedSubview(fixedButtonCommandBar)
-
-        container.addArrangedSubview(createLabelWithText("In Input Accessory View"))
-
-        let textFieldContainer = UIView()
-        textFieldContainer.backgroundColor = Colors.navigationBarBackground
-        textFieldContainer.addSubview(textField)
-        NSLayoutConstraint.activate([
-            textField.topAnchor.constraint(equalTo: textFieldContainer.topAnchor, constant: 16.0),
-            textField.leadingAnchor.constraint(equalTo: textFieldContainer.leadingAnchor, constant: 16.0),
-            textFieldContainer.bottomAnchor.constraint(equalTo: textField.bottomAnchor, constant: 16.0),
-            textFieldContainer.trailingAnchor.constraint(equalTo: textField.trailingAnchor, constant: 16.0)
-        ])
-
-        container.addArrangedSubview(textFieldContainer)
-
-        let accessoryCommandBar = CommandBar(itemGroups: itemGroups, trailingItem: newItem(for: .keyboard))
-        textField.inputAccessoryView = accessoryCommandBar
+        return itemGroups
     }
 
     func createLabelWithText(_ text: String = "") -> Label {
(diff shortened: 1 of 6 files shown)


ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -26,17 +26,30 @@ open class CommandBar: UIView {
 
     // MARK: - Public methods
 
+    @available(*, deprecated, renamed: "init(itemGroups:leadingItemGroups:trailingItemGroups:)")
```
nodes11, line 29: In retrospect, it's probably best to not mark this as deprecated quite yet. Maybe just mark it as "renamed" to allow for users to update and not break code.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -69,16 +82,97 @@ open class CommandBar: UIView {
...
+            for view in trailingButtonGroupViews {
+                trailingStackView.addArrangedSubview(view)
+            }
+        }
+    }
+
     // MARK: - Private properties
 
-    private let itemGroups: [CommandBarItemGroup]
+    private var _itemsToButtonsMap: [CommandBarItem: CommandBarButton]?
```
edjamesmsft, line 129: FYI, we don't do leading underscores for internal variable names!

edjamesmsft: Oh, and it's the same name as itemsToButtonsMap below, with only the underscore as the difference... The intent of the two variables should be made more distinct and clear.

cy255: Per offline discussions, _itemsToButtonsMap does not need to have an underscore in the naming and the two private itemsToButtonsMap properties can be combined into one.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -252,3 +429,38 @@ extension CommandBar: UIScrollViewDelegate {
         updateShadow()
     }
 }
+
+// MARK: - CommandBarItemDelegate
+
+extension CommandBar: CommandBarItemDelegate {
```
cy255, line 435: Any state change will result in the button to update the state so will it be easier to have one function that handles all stage changes? Or could we have separate functions to update the button's enabled, selected, title, image, etc. properties?

cy255 (changes requested):

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Introduced the CommandBarCommandGroupsStackView that encapsulates most of the duplicate logic I introduced with the first commit. This work removes the use of the ivars with the prefixed variable names.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -26,17 +26,30 @@ open class CommandBar: UIView {
...
+        translatesAutoresizingMaskIntoConstraints = false
+
+        configureHierarchy()
+        updateButtonsState()
+    }
+
+    @objc public init(itemGroups: [CommandBarItemGroup], leadingItemGroups: [CommandBarItemGroup]? = nil, trailingItemGroups: [CommandBarItemGroup]? = nil) {
+        self.itemGroups = itemGroups
+
+        self.leadingItemGroups = leadingItemGroups != nil ? leadingItemGroups! : []
```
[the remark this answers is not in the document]
nodes11: Updated this with the second commit. No specific reason for the properties not being optional in the initial commit.

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Hmm, I'm going to give this a little more thought and then update this PR accordingly.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -7,37 +7,62 @@ import UIKit
...
+            self.leadingItemGroups = [[leadingItem]]
         }
+
         if let trailingItem = trailingItem {
-            self.trailingButton = button(forItem: trailingItem, isPersistSelection: false)
+            self.trailingItemGroups = [[trailingItem]]
         }
 
+        self.leadingCommandGroupsStackView = CommandBarCommandGroupsStackView(itemGroups: self.leadingItemGroups)
+        self.centerCommandGroupsStackView = CommandBarCommandGroupsStackView(itemGroups: self.itemGroups)
```
nodes11, line 44: The term "center" feels off, open to naming suggestions. I feel that providing some sort of position in the naming is important.

cy255: How about "middle" or "main"? I agree that center is a little bit off because it's not necessarily centered.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -69,16 +94,53 @@ open class CommandBar: UIView {
...
+            trailingCommandGroupsStackView.itemGroups = trailingItemGroups
+            constrainTrailingCommandGroupsStackView()
+        }
+    }
+
     // MARK: - Private properties
 
-    private let itemGroups: [CommandBarItemGroup]
+    /// UIStackView holding the items pinned to the leading end of the CommandBar
+	private var leadingCommandGroupsStackView: CommandBarCommandGroupsStackView
```
cy255, line 131: minor: this line and line 134's indentation looks off.

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: "main" was at the top of my list so I'll go with that.

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Probably just having one delegate method that all `CommandBarItem` properties call would be the best and easiest right now.

ios/FluentUI/Command Bar/CommandBarCommandGroupsStackView.swift
```diff
@@ -0,0 +1,120 @@
...
+
+        translatesAutoresizingMaskIntoConstraints = false
+        axis = .horizontal
+        spacing = CommandBarCommandGroupsStackView.buttonGroupSpacing
+
+        refreshButtonLayout()
+    }
+
+    required init(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
```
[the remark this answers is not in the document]
nodes11: Thanks for providing that link. Very helpful.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -51,9 +76,9 @@ open class CommandBar: UIView {
 
     /// Apply `isEnabled` and `isSelected` state from `CommandBarItem` to the buttons
     @objc public func updateButtonsState() {
```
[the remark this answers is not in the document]
nodes11: That is the plan. My initial thought was to leave this here to allow for an easier transition, however, if all item properties are automatically propagating to their button then I guess this code is useless. Will remove.

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -90,4 +120,19 @@ open class CommandBarItem: NSObject {
     func handleTapped(_ sender: CommandBarButton) {
         itemTappedHandler(sender, self)
     }
+
+    weak var delegate: CommandBarItemDelegate?
+}
+
+protocol CommandBarItemDelegate: AnyObject {
```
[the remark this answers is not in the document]
nodes11: Yes, a similar delegate pattern was used for the CommandingItem: https://github.com/microsoft/fluentui-apple/blob/4c9e779e1cebb71ab05bf5d3c19d3329574f7d99/ios/FluentUI/Bottom%20Commanding/CommandingItem.swift#L96-L113

I can give this a little more thought to see if there is a better alternative pattern that we can follow.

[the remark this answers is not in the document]
nodes11: The initial thought that comes to mind that is maybe a little more Swift-y is to expose a closure property on `CommandBarItem` that can be set during initialization of the  `CommandBarButton`. When a property changes, the closure can be called and the corresponding button will update.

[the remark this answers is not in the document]
nodes11: Updated.

nodes11:
This is no longer considered a draft PR. I hope for these changes to be checked in after review.

ios/FluentUI/Command Bar/CommandBarItem.swift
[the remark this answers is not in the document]
lcapkovic: I think I went through a similar progression with the CommandingItem. I wanted to use closures, but I wanted per-property mutation, so I would either need a closure per property, or pass the property keypath that changed. That ended up being messy for reasons I don't remember, so I ended up with a delegate pattern at the end 😄

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -103,53 +149,28 @@ open class CommandBar: UIView {
         scrollView.translatesAutoresizingMaskIntoConstraints = false
         scrollView.contentInset = UIEdgeInsets(
```
harrieshin, line 150: in the RTL scenarios shouldn't the spacing be flipped?

ios/FluentUI/Command Bar/CommandBarCommandGroupsStackView.swift
```diff
@@ -0,0 +1,91 @@
...
+        }
+    }
+
+    /// Refreshes the `buttonGroupViews` array of `CommandBarButtonGroupView`s that are displayed in the view
+    private func updateButtonGroupViews() {
+        updateItemsToButtonsMap()
+        buttonGroupViews = itemGroups.map { items in
+                CommandBarButtonGroupView(buttons: items.compactMap { item in
+                    guard let button = itemsToButtonsMap[item] else {
+                        preconditionFailure("Button is not initialized in commandsToButtons")
```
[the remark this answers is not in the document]
nodes11: Good catch. I believe this is referring to the creation of the `itemsToButtonsMap` map. `commandsToButtons` does not exist at the tip of main either. This must have slipped through the cracks in a previous commit. Will update.

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: I believe the transform used on the `scrollView` in `configureHierarchy()` takes care of flipping the content insets. I tested this locally using a larger value for `CommandBar.fixedButtonSpacing` I was able to see the shift of the content inset from one side to the other as I changed from an LTR language to an RTL language.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -74,6 +85,12 @@ open class CommandBar: UIView {
     public override func layoutSubviews() {
         super.layoutSubviews()
 
+        scrollView.contentInset = scrollViewContentInset()
```
lcapkovic, line 88: This definitely shouldn't be necessary. We shouldn't be triggering redundant layout passes during a layout pass.
The content inset setting should also only happen when necessary, not during every layout pass.

lcapkovic: Clarification - this applies to this line and all the layoutIfNeeded calls below.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -156,6 +158,84 @@ class CommandBarDemoController: DemoController {
         container.layoutMargins.left = 0
         view.backgroundColor = Colors.surfaceSecondary
 
+        container.addArrangedSubview(createLabelWithText("Default"))
+
+        defaultCommandBar = CommandBar(itemGroups: createItemGroups(), leadingItemGroups: [[newItem(for: .keyboard)]])
+        defaultCommandBar?.backgroundColor = Colors.navigationBarBackground
+        container.addArrangedSubview(defaultCommandBar!)
```
edjamesmsft, line 165: We should be avoiding force unwraps wherever possible. Better to guard/let here, perhaps?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -156,6 +158,84 @@ class CommandBarDemoController: DemoController {
...
+        customizationStackView.addArrangedSubview(itemEnabledSwitch)
+        itemCustomizationContainer.addArrangedSubview(customizationStackView)
+
+        itemCustomizationContainer.addArrangedSubview(UIView()) //Spacer
+
+        container.addArrangedSubview(itemCustomizationContainer)
+
+        container.addArrangedSubview(createLabelWithText("With Fixed Button"))
+
+		let fixedButtonCommandBar = CommandBar(itemGroups: createItemGroups(), leadingItemGroups: [[newItem(for: .copy)]], trailingItemGroups: [[newItem(for: .keyboard)]])
```
edjamesmsft, line 216: Nit: whitespace looks odd here. Is this tabs instead of spaces?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -7,41 +7,60 @@ import UIKit
...
 
     // MARK: - Public methods
 
-    @objc public init(itemGroups: [CommandBarItemGroup], leadingItem: CommandBarItem? = nil, trailingItem: CommandBarItem? = nil) {
-        self.itemGroups = itemGroups
-
-        super.init(frame: .zero)
+    @available(*, renamed: "init(itemGroups:leadingItemGroups:trailingItemGroups:)")
+    @objc public convenience init(itemGroups: [CommandBarItemGroup], leadingItem: CommandBarItem? = nil, trailingItem: CommandBarItem? = nil) {
+        var leadingItems: [CommandBarItemGroup] = []
```
edjamesmsft, line 33: Should these two be optionals, then you wouldn't have to check for nil below and could just pass through?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -65,20 +85,63 @@ open class CommandBar: UIView {
...
-        return Dictionary(uniqueKeysWithValues: allButtons.map { ($0.item, $0) })
-    }()
+    /// UIStackView holding the items pinned to the trailing end of the CommandBar
+    private var trailingCommandGroupsStackView: CommandBarCommandGroupsStackView
+
+    /// Leading constraint between the `containerView` and either the `leadingCommandGroupsStackView` or the parent view
+    private var leadingConstraint: NSLayoutConstraint?
+
+    /// Trailing constraint between the `containerView` and either the `trailingCommandGroupsStackView` or the parent view
+    private var trailingConstraint: NSLayoutConstraint?
```
huwilkes, line 144: Maybe a bit late to be saying this, but would it be possible to just stick the leading, container, and trailing views in a horizontal stack and just hide the leading/trailing views when they're empty? Then you wouldn't need to dance around different constraints depending on if they're empty or not.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
sophialee0416: Looks like the indentation is off

lcapkovic (approved):

PR #990, opened 2022-05-23
[vNext Tokens] Re-tokenizing TabBar

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change demonstrates the new strategy for handling tokenization of compound controls, where the "contained" control(s) is/are internal only. We expose only a small subset of tokens of the internal control to the public, via special nil-able values in the containing control's token set, which then sets the internal, private control's token set via a custom derived token class. This allows the public to change a subset of tokens, while still allowing our token system, and hence designers, to still have full control and access to all tokens of the internal class.

### Verification

Manually tested via Simulator.

| TokenizeTabBarAfter | ReTokenizeTabBar |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/990)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
@@ -237,11 +237,11 @@ extension TabBarViewDemoController: DemoAppearanceDelegate {
     }
 
     private class PerControlOverrideTabBarItemTokens: TabBarTokens {
-        override var tabBarItemTitleLabelPortrait: FontInfo? {
+        override var tabBarItemTitleLabelFontPortrait: FontInfo? {
             return .init(size: 15, weight: .bold)
         }
 
-        override var tabBarItemTitleLabelLandscape: FontInfo? {
+        override var tabBarItemTitleLabelFontLandscape: FontInfo? {
             return .init(size: 15, weight: .bold)
         }
     }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
@@ -222,13 +222,13 @@ extension TabBarViewDemoController: DemoAppearanceDelegate {
 
     // MARK: - Custom tokens
     private class ThemeWideOverrideTabBarTokens: TabBarTokens {
-        override var selectedColor: DynamicColor {
+        override var tabBarItemSelectedColor: DynamicColor {
             return .init(light: globalTokens.sharedColors[.burgundy][.tint10],
                          lightHighContrast: globalTokens.sharedColors[.pumpkin][.tint10],
                          dark: globalTokens.sharedColors[.darkTeal][.tint40],
                          darkHighContrast: globalTokens.sharedColors[.teal][.tint40])
         }
-        override var unselectedColor: DynamicColor {
+        override var tabBarItemUnselectedColor: DynamicColor {
             return .init(light: globalTokens.sharedColors[.darkTeal][.tint20],
                          lightHighContrast: globalTokens.sharedColors[.teal][.tint40],
                          dark: globalTokens.sharedColors[.pumpkin][.tint40],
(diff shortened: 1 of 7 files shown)


ios/FluentUI.xcworkspace/xcshareddata/swiftpm/Package.resolved
```diff
@@ -1,23 +1,25 @@
 {
-  "pins" : [
-    {
-      "identity" : "appcenter-sdk-apple",
-      "kind" : "remoteSourceControl",
-      "location" : "https://github.com/microsoft/appcenter-sdk-apple.git",
-      "state" : {
-        "revision" : "25f64229373de97ff3920941cd52203193e5d8be",
-        "version" : "4.3.0"
+  "object": {
```
huwilkes, line 2: I've also had this file changing a bunch of me recently. Not sure if we need to update it, but if we do it should probably be in a different PR.

ios/FluentUI/Tab Bar/TabBarTokens.swift
```diff
@@ -17,66 +17,14 @@ open class TabBarTokens: ControlTokens {
...
-    open var badgeCornerRadii: CGFloat { 10.0 }
+    open var tabBarItemUnselectedColor: DynamicColor? { nil }
 
     /// Font info for the title label when in portrait view
-    open var titleLabelPortrait: FontInfo { return .init(size: 10, weight: .medium) }
+    open var tabBarItemTitleLabelPortrait: FontInfo? { nil }
 
     /// Font info for the title label when in landscape view
-    open var titleLabelLandscape: FontInfo { return .init(size: 13, weight: .medium) }
+    open var tabBarItemTitleLabelLandscape: FontInfo? { nil }
```
huwilkes, line 29: Can we add the word "font" to `tabBarItemTitleLabelPortrait` and `tabBarItemTitleLabelLandscape`? Those names sound like the actual strings/labels rather than their fonts to me.

ios/FluentUI/Tab Bar/TabBarItemTokens.swift
```diff
@@ -0,0 +1,73 @@
...
+    open var badgeBorderWidth: CGFloat { globalTokens.borderSize[.thick] }
+
+    /// The horizontal padding for the `BadgeLabel`
+    open var badgeHorizontalPadding: CGFloat { 10.0 }
+
+    /// The radii of the four corners of the `BadgeLabel`
+    open var badgeCornerRadii: CGFloat { 10.0 }
+
+    /// Font info for the title label when in portrait view
+    open var titleLabelPortrait: FontInfo { return .init(size: 10, weight: .medium) }
```
mischreiber, line 69: Same question about scaling or non-scaling for these values.

ios/FluentUI/Tab Bar/TabBarView.swift
```diff
@@ -191,13 +191,43 @@ open class TabBarView: UIView, TokenizedControlInternal {
         }
     }
 
+    private class CustomTabBarItemTokens: TabBarItemTokens {
+        var tabBarTokens: TabBarTokens = .init()
```
mischreiber, line 195: I don't think we should be initializing this to a default value here. Presumably that will cause the `required init()` below to fail, but we can work around that by having it throw an error (since no one should be calling it anyway):

```swift
@available(*, unavailable)
required init() {
    preconditionFailure("init() has not been implemented")
}
```

ios/FluentUI.xcworkspace/xcshareddata/swiftpm/Package.resolved
edjamesmsft: I can remove it from this pr at the very least

ios/FluentUI/Tab Bar/TabBarTokens.swift
edjamesmsft: Will do!

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -237,27 +237,11 @@ extension TabBarViewDemoController: DemoAppearanceDelegate {
...
-        }
-
-        override var phonePortraitHeight: CGFloat {
-            return 35.0
-        }
-
-        override var phoneLandscapeHeight: CGFloat {
-            return 35.0
+        override var tabBarItemTitleLabelPortrait: FontInfo? {
+            return .init(size: 15, weight: .bold)
```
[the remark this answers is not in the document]
edjamesmsft: Who would I verify with? Design? PM?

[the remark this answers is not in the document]
harrieshin: what does apple native app do for text? i hunch would be it will increase upto somepoint but stop and use larger context viewer for really big fonts.

mischreiber (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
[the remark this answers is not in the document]
mischreiber: Followed up offline: these should probably be non-scaling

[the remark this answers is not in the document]
edjamesmsft: These are already shouldScale: false where the FontInfo is actually consumed in TabBarItemView.swift, line 261 and 269. Is that sufficient?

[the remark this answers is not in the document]
edjamesmsft: Followed up offline, yes it is sufficient!

PR #991, opened 2022-05-24
[PreviewCard] Created multiple PreviewCards with variations in height

lewist13:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Introduced multiple aspect ratio variations a PreviewCard could have. Those sizes include: 

- Ultrawide (21:9) 
- Landscape (16:9)
- Old Landscape (4:3)
-  Square Format (1:1)

PreviewCardDemoController is inheriting DemoController instead of DemoTableViewController to reduce code redundancy and allows the entire PreviewCard view to fully scroll down the list of PreviewCards.

### Verification

| prevcard_w_placeholderbordered_resized | scrolling_previewcard_5-24 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/991)


ios/FluentUI/PreviewCard/PreviewCardTokens.swift
```diff
@@ -0,0 +1,27 @@
...
+/// Pre-defined sizes of the PreviewCard.
+@objc public enum MSFPreviewCardSize: Int, CaseIterable {
+    case ultrawideHeight
+    case landscapeHeight
+    case oldLandscapeHeight
+    case squareHeight
+    var cardHeight: CGFloat {
+        switch self {
+        case .ultrawideHeight:
+        return 147
```
huwilkes, line 18: I assume these are just based on the aspect ratio and minWidth, but it might be worth adding a comment to explain where these numbers come from.

lewist13: Yup, only the height of the cards since the width never changes. I'll add a comment, good note!

ios/FluentUI/PreviewCard/PreviewCardTokens.swift
```diff
@@ -0,0 +1,27 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+import UIKit
+
```
huwilkes, line 8: Does this file use either of these?

ios/FluentUI/PreviewCard/PreviewCardTokens.swift
harrieshin: where is the terminology "oldlandscape" come from?

ios/FluentUI/PreviewCard/PreviewCardTokens.swift
lewist13: Hmm good question. There is a build error if neither import is included but builds successfully if UIKit or SwiftUI is imported by itself. I can leave SwiftUI and take out UIKit just so it builds

ios/FluentUI/PreviewCard/PreviewCardTokens.swift
lewist13: Old landscape was the standard aspect ratio before modern TVs switched to 16:9 with the rise of HD. Certain devices still use 4:3 like Apple's iPad

edjamesmsft: Is the term "old landscape" commonly used elsewhere?

lewist13: No idea. Old landscape popped up as 4:3 a lot while I researched it, but I now believe fullscreen would be more appropriate. Apparently, that's the official name for 4:3 aspect ratio

ios/FluentUI.Demo/FluentUI.Demo/Demos/PreviewCardDemoController.swift
```diff
@@ -19,21 +19,36 @@ class PreviewCardDemoController: DemoTableViewController {
         guard let placeholderImage = UIImage.init(named: "W11_Placeholder") else {
             return
         }
-        let previewFlatCard = MSFPreviewCard(theme: nil, image: placeholderImage)
-        let previewElevatedCard = MSFPreviewCard(theme: nil, isElevated: true, image: placeholderImage)
-        let previewFlatCardView = previewFlatCard.view
-        let previewElevatedCardView = previewElevatedCard.view
+        let previewUltrawideFlatCard = MSFPreviewCard(theme: nil, image: placeholderImage, cardHeight: .ultrawideHeight)
```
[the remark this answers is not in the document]
lewist13: The compiler or linter doesn't usually like when there's whitespace. Would it be okay to add whitespace anyway or maybe adding a couple comments would suffice?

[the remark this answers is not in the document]
huwilkes: I think it'll usually complain if you have two blank lines in a row, or if you have trailing whitespace. You should be okay if you just have a single, empty new line.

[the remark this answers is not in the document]
lewist13: I updated it. Let me know if the added whitespace is better or needs changes.

ios/FluentUI/PreviewCard/PreviewCardTokens.swift
```diff
@@ -0,0 +1,31 @@
...
+/// Pre-defined sizes of the PreviewCard.
+@objc public enum MSFPreviewCardSize: Int, CaseIterable {
+        /// 21:9 aspect ratio
+    case ultrawideHeight
+        /// 16:9 aspect ratio
+    case landscapeHeight
+        /// 4:3 aspect ratio
+    case fullscreenHeight
+        /// 1:1 aspect ratio
+    case squareHeight
```
huwilkes, line 17: nit: I don't think we need the comments to be indented

huwilkes (approved):

PR #992, opened 2022-05-25
Update SegmentedControl to support images

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Added an optional image to the SegmentedItem. If the image is set, we'll only display the image and not the text, but we'll still use the text for accessibility.
Changed pillMaskedLabelsContainerView's name to pillMaskedContentContainerView, since now it holds more than just labels.
Changed the maskedLabels array to be an array of optionals, since I'm also adding an array for the maskedImages, and we don't need the label/image that isn't being used.
Also updated the logic in sizeThatFits, since I was getting weird results where the window's layout guide was nil or something and just returning zero. (the before iPhone portrait picture shows what I'm talking about)
Moved some logic out of for loops, since there were a few places we were doing things every loop when it was never going to change during the loop.

### Verification
| SegmentedControl_images_iPad_before | SegmentedControl_images_iPad_after |
| SegmentedControl_images_iPhone_before | SegmentedControl_images_iPhone_after |
| SegmentedControl_images_iPhone_dark_before | SegmentedControl_images_iPhone_dark_after |
| SegmentedControl_images_iPhone_landscape_before | SegmentedControl_images_iPhone_landscape_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/992)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentItem.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentItem.swift
@@ -22,7 +22,7 @@ public class SegmentItem: NSObject {
    }
 
     @objc public convenience init(title: String, isUnread: Bool = false) {
-        self.init(title: title, isUnread: isUnread)
+        self.init(title: title, image: nil, isUnread: isUnread)
     }
 
     @objc public init(title: String,
--- a/ios/FluentUI/SegmentedControl/SegmentItem.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentItem.swift
@@ -3,11 +3,14 @@
 //  Licensed under the MIT License.
 //
 import Foundation
+import UIKit
 
 /// Used for SegmentedControl array of views
 @objc(MSFSegmentItem)
 public class SegmentItem: NSObject {
     public let title: String
+    public let image: UIImage?
+    public let isTemplateImage: Bool
 
     /// This value will determine whether or not to show dot next to the pill button label
     public var isUnread: Bool {
@@ -18,8 +21,13 @@ public class SegmentItem: NSObject {
        }
    }
 
-    @objc public init(title: String, isUnread: Bool = false) {
+    @objc public init(title: String,
+                      image: UIImage? = nil,
+                      isTemplateImage: Bool = true,
+                      isUnread: Bool = false) {
         self.title = title
+        self.image = image
+        self.isTemplateImage = isTemplateImage
         self.isUnread = isUnread
         super.init()
     }
--- a/ios/FluentUI/SegmentedControl/SegmentItem.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentItem.swift
@@ -20,12 +20,21 @@ public class SegmentItem: NSObject {
        }
    }
 
+    /// Creates a new instance of a `SegmentItem` that holds data used to create a segment in the `SegmentedControl`.
+    /// - Parameters
+    ///   - title: Title that will be displayed by the segment and used as the accessibility label and large content viewer title.
+    ///   - isUnread:  Whether the segment shows the mark that represents the "unread" state.
     @objc public convenience init(title: String, isUnread: Bool = false) {
         self.init(title: title,
                   image: nil,
                   isUnread: isUnread)
     }
 
+    /// Creates a new instance of a `SegmentItem` that holds data used to create a segment in the `SegmentedControl`.
+    /// - Parameters
+    ///   - title: Title that will be displayed by the segment if the image is nil, and used as the accessibility label and large content viewer title.
+    ///   - image: Image that will display instead of the title if not nil.
+    ///   - isUnread:  Whether the segment shows the mark that represents the "unread" state.
     @objc public init(title: String,
                       image: UIImage? = nil,
                       isUnread: Bool = false) {
--- a/ios/FluentUI/SegmentedControl/SegmentItem.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentItem.swift
@@ -21,6 +21,10 @@ public class SegmentItem: NSObject {
        }
    }
 
+    @objc public convenience init(title: String, isUnread: Bool = false) {
+        self.init(title: title, isUnread: isUnread)
+    }
+
     @objc public init(title: String,
                       image: UIImage? = nil,
                       isTemplateImage: Bool = true,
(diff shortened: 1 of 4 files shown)


ios/FluentUI/SegmentedControl/SegmentItem.swift
```diff
@@ -18,8 +21,13 @@ public class SegmentItem: NSObject {
        }
    }
 
-    @objc public init(title: String, isUnread: Bool = false) {
+    @objc public init(title: String,
```
harrieshin, line 24: would we create convenience method with deprecation on the "old" one for objc clients to migrate better?

ios/FluentUI/SegmentedControl/SegmentItem.swift
```diff
@@ -3,11 +3,14 @@
...
 //
 import Foundation
+import UIKit
 
 /// Used for SegmentedControl array of views
 @objc(MSFSegmentItem)
 public class SegmentItem: NSObject {
     public let title: String
+    public let image: UIImage?
+    public let isTemplateImage: Bool
```
harrieshin, line 13: we can't make this optional too?

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
```diff
@@ -35,9 +35,13 @@ class SegmentPillButton: UIButton {
         self.contentEdgeInsets = Constants.insets
 
         let title = item.title
-        self.setTitle(title, for: .normal)
         self.accessibilityLabel = title
         self.largeContentTitle = title
+        if let image = item.image {
+            self.setImage(image.withRenderingMode(item.isTemplateImage ? .alwaysTemplate : .automatic), for: .normal)
```
harrieshin, line 41: we can't trust the clients would do the right thing?

huwilkes: Do you mean for the rendering mode? We don't even do the right thing, our placeholder_20 isn't a template image. We do this in a the popup menu, also.

huwilkes: Looks like we do have that image as a template, but if we don't change the rendering mode we just draw the circle in blue.

huwilkes: Never mind, that must've been before I was setting the tint color, or something. I removed it and it works fine now.

ios/FluentUI/SegmentedControl/SegmentItem.swift
```diff
@@ -18,8 +20,17 @@ public class SegmentItem: NSObject {
...
    }
 
-    @objc public init(title: String, isUnread: Bool = false) {
+    @objc public convenience init(title: String, isUnread: Bool = false) {
+        self.init(title: title,
+                  image: nil,
+                  isUnread: isUnread)
+    }
+
+    @objc public init(title: String,
```
harrieshin, line 29: documentation?

harrieshin:
tested darkmode and accessibility as well?

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
```diff
@@ -35,9 +35,13 @@ class SegmentPillButton: UIButton {
         self.contentEdgeInsets = Constants.insets
 
         let title = item.title
-        self.setTitle(title, for: .normal)
         self.accessibilityLabel = title
         self.largeContentTitle = title
+        if let image = item.image {
+            self.setImage(image, for: .normal)
```
harrieshin, line 41: do we need https://developer.apple.com/documentation/uikit/uilargecontentvieweritem/3183932-largecontentimage?language=objc ?

huwilkes: Large content viewer works, shows the image and the title

huwilkes:
> tested darkmode and accessibility as well?

There's an issue with the disabled dark mode colors, but I haven't had a chance to test voice over or keyboard navigation yet.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -461,12 +466,12 @@ open class SegmentedControl: UIControl {
...
-            if let windowWidth = window?.safeAreaLayoutGuide.layoutFrame.width {
-                if traitCollection.userInterfaceIdiom == .pad {
-                    maxButtonWidth = max(windowWidth / 2, 375.0)
-                } else {
-                    maxButtonWidth = windowWidth
-                }
+            let windowSafeAreaInsets = window.safeAreaInsets
+            let windowWidth = window.bounds.width - windowSafeAreaInsets.left - windowSafeAreaInsets.right
+            if traitCollection.userInterfaceIdiom == .pad {
+                maxButtonWidth = max(windowWidth / 2, 375.0)
```
edjamesmsft, line 472: Nit: should there be a comment or a const somewhere explaining why you're potentially padding up to the magic number of 375?

edjamesmsft (approved):

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -438,14 +441,16 @@ open class SegmentedControl: UIControl {
     }
 
     open override func sizeThatFits(_ size: CGSize) -> CGSize {
+        guard let window = window else {
+            return CGSize.zero
+        }
+        let screen = window.windowScene?.screen ?? UIScreen.main
```
sophialee0416, line 447: I could be wrong, but didn't we want to deprecate our use of UIScreen.main?

huwilkes: Could be. I don't remember hearing that guidance. But, we're still using it all over the codebase, and I think trying to get the screen from the window, then using UIScreen.main as a fallback is probably alright.

harrieshin: yes, uiscreen.main is wrong and we haven't had time to remove it all over the place.

harrieshin (approved):

PR #994, opened 2022-05-27
[vNext Tokens] Update Button to use new iOS 15 large content viewer

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Removed our old UIKit large content viewer work around and added a new SwiftUI View modifier that will add a large content viewer with the provided text and image

### Verification

I was really struggling to get the large content viewer to even show up without the changes, so this change also made the large content viewer more reliable (in iOS 15.0+)

| Button_largeContentViewer_before | Button_largeContentViewer_after |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
+++ b/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
@@ -40,6 +40,30 @@ extension View {
             .onPreferenceChange(SizePreferenceKey.self,
                                 perform: action)
     }
+
+    /// Adds a large content viewer for the view
+    /// - Parameters
+    ///  - text: Optional String to display in the large content viewer.
+    ///  - image: Optional UIImage to display in the large content viewer.
+    /// - Returns: The modified view.
+    func showsLargeContentViewer(text: String?, image: UIImage?) -> AnyView {
+        if #available(iOS 15.0, *) {
+            if text != nil || image != nil {
+                return AnyView(self.accessibilityShowsLargeContentViewer({
+                    if let image = image {
+                        Image(uiImage: image)
+                    }
+                    if let text = text {
+                        Text(text)
+                    }
+                }))
+            }
+
+            return AnyView(self.accessibilityShowsLargeContentViewer())
+        }
+
+        return AnyView(self)
+    }
 }
 
 /// PreferenceKey that will store the measured size of the view
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -76,14 +76,7 @@ public struct FluentButton: View, ConfigurableTokenizedControl {
         @ViewBuilder
         var largeContentButton: some View {
             if #available(iOS 15.0, *), state.style.isFloatingStyle {
-                button.accessibilityShowsLargeContentViewer({
-                    if let image = state.image {
-                        Image(uiImage: image)
-                    }
-                    if let text = state.text {
-                        Text(text)
-                    }
-                })
+                button.showsLargeContentViewer(text: state.text, image: state.image)
             } else {
                 button
             }
--- a/ios/FluentUI/Vnext/Button/Button.swift
+++ b/ios/FluentUI/Vnext/Button/Button.swift
@@ -64,11 +64,31 @@ public struct FluentButton: View, ConfigurableTokenizedControl {
     }
 
     public var body: some View {
-        Button(action: state.action, label: {})
-            .buttonStyle(FluentButtonStyle(state: state,
-                                           tokensLookup: { tokens }))
-            .disabled(state.disabled ?? false)
-            .frame(maxWidth: .infinity)
+        @ViewBuilder
+        var button: some View {
+            Button(action: state.action, label: {})
+                .buttonStyle(FluentButtonStyle(state: state,
+                                               tokensLookup: { tokens }))
+                .disabled(state.disabled ?? false)
+                .frame(maxWidth: .infinity)
+        }
+
+        @ViewBuilder
+        var largeContentButton: some View {
+            if #available(iOS 15.0, *), state.style.isFloatingStyle {
+                button.accessibilityShowsLargeContentViewer({
+                    if let image = state.image {
+                        Image(uiImage: image)
+                    }
+                    if let text = state.text {
+                        Text(text)
+                    }
+                })
+            } else {
+                button
+            }
+        }
+        return largeContentButton
     }
 }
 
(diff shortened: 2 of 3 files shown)


edjamesmsft (approved):

harrieshin:
genuine question : is it worth adding it into the viewmodifier? or should it be done individually per control?

huwilkes:
> genuine question : is it worth adding it into the viewmodifier? or should it be done individually per control?

I figured it'd at least make the available check clean up easier. And we have a few other controls that'll need that large content viewer. Before iOS 14 was the minimum, we would've needed to do it this way because Swift wouldn't let us do an if #available check in a view builder, but I guess we could just do it per control now.

ios/FluentUI/Vnext/Button/MSFButton.swift
```diff
@@ -58,29 +58,12 @@ import UIKit
...
             strongSelf.action?(strongSelf)
         }
-        setupLargeContentViewer()
     }
 
     required public init?(coder: NSCoder) {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
-    private func setupLargeContentViewer() {
```
rdeassis, line 68: We should keep this while we support iOS 14.

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -64,11 +64,24 @@ public struct FluentButton: View, ConfigurableTokenizedControl {
     }
 
     public var body: some View {
-        Button(action: state.action, label: {})
-            .buttonStyle(FluentButtonStyle(state: state,
-                                           tokensLookup: { tokens }))
-            .disabled(state.disabled ?? false)
-            .frame(maxWidth: .infinity)
+        @ViewBuilder
```
rdeassis, line 67: You can do this with modifyIf to have a simpler code:

```
    Button(action: state.action, label: {})
                .buttonStyle(FluentButtonStyle(state: state,
                                               tokensLookup: { tokens }))
                .disabled(state.disabled ?? false)
                .frame(maxWidth: .infinity)
                .modifyIf(state.style.isFloatingStyle, { button in
                    button.showsLargeContentViewer(text: state.text, image: state.image)
                })
```

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -40,6 +40,30 @@ extension View {
             .onPreferenceChange(SizePreferenceKey.self,
                                 perform: action)
     }
+
+    /// Adds a large content viewer for the view
+    /// - Parameters
+    ///  - text: Optional String to display in the large content viewer.
+    ///  - image: Optional UIImage to display in the large content viewer.
+    /// - Returns: The modified view.
+    func showsLargeContentViewer(text: String?, image: UIImage?) -> AnyView {
```
rdeassis, line 49: You can avoid the AnyView return here and use some View instead.

rdeassis, line 49: Can you create a modifier struct for this?

```

    func showsLargeContentViewer(text: String?, image: UIImage?) -> some View {
        modifier(LargeContentViewerModifier(text: text,
                                            image: image))
    }

```

```
struct LargeContentViewerModifier: ViewModifier {

    init(text: String?, image: UIImage?) {
        self.text = text
        self.image = image
    }

    var text: String?
    var image: UIImage?

    func body(content: Content) -> some View {
        if #available(iOS 15.0, *) {
            content.accessibilityShowsLargeContentViewer({
                    if let image = image {
                        Image(uiImage: image)
                    }
                    if let text = text {
                        Text(text)
                    }
                })
        } else {
            content
        }
    }
}

```

rdeassis (changes requested):

ios/FluentUI/Vnext/Button/MSFButton.swift
huwilkes: @harrieshin Thoughts?

huwilkes: I think when we were talking about it, we were saying we should just switch to the new API, but I could be misremembering.

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
huwilkes: It doesn't build with some View, since it can return 3 different types of views.

mischreiber: If you use a `ViewModifier` as mentioned by @rdeassis then I _think_ you can get away with `some View`.

ios/FluentUI/Vnext/Button/MSFButton.swift
mischreiber: @huwilkes That's correct. Consensus was that we will support full a11y on the latest version of iOS, which is currently iOS 15.

ios/FluentUI/Vnext/Button/Button.swift
```diff
@@ -69,6 +69,9 @@ public struct FluentButton: View, ConfigurableTokenizedControl {
                                            tokensLookup: { tokens }))
             .disabled(state.disabled ?? false)
             .frame(maxWidth: .infinity)
+            .modifyIf(state.style.isFloatingStyle, {button in
```
mischreiber, line 72: Space before `button` pls :)

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -57,3 +67,34 @@ struct OnSizeChangeViewModifier: ViewModifier {
         })
     }
 }
+
+/// ViewModifier for showing the large content viewer with optional text and optional image.
+/// If both the text and image are bil, the default large content viewer will be used.
```
sophialee0416, line 72: Nit: misspelling on nil?

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -57,3 +67,34 @@ struct OnSizeChangeViewModifier: ViewModifier {
...
+struct LargeContentViewerModifier: ViewModifier {
+    init (text: String?, image: UIImage?) {
```
mischreiber, line 74: Nit: no space after `init`

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -57,3 +67,34 @@ struct OnSizeChangeViewModifier: ViewModifier {
...
+        self.text = text
+        self.image = image
+    }
+
+    var text: String?
```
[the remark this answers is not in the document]
mischreiber: They should probably also be private.

rdeassis (approved):

PR #998, opened 2022-06-03
Adding CustomView values to TableViewCell tokens

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Tokenizing CustomViewSize enum in TableViewCell. This includes the dimensions and trailing margin of the customView.

**Note: This PR brings breaking changes to the TableViewCell API: the CustomViewSize enum is being moved over to tokens and renamed as MSFTableViewCellCustomViewSize.

### Verification

Tested on iOS 15.5 iPhone 12 Pro. There should be no visible difference in the before/after.

| Simulator Screen Shot - iPhone 12 Pro - 2022-06-02 at 17 24 16 | Simulator Screen Shot - iPhone 12 Pro - 2022-06-02 at 17 25 00 |
| Simulator Screen Shot - iPhone 12 - 2022-06-08 at 13 38 37 | Simulator Screen Shot - iPhone 12 - 2022-06-08 at 13 45 28 |
| Simulator Screen Shot - iPhone 12 - 2022-06-08 at 13 39 15 | Simulator Screen Shot - iPhone 12 - 2022-06-08 at 13 45 09 |

Per-control theme validation:

https://user-images.githubusercontent.com/22566866/171761485-01ad55a8-4aac-4a98-9ed9-791402355e06.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/998)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -134,7 +134,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         case twoLines
         case threeLines
 
-        var customViewSize: MSFTableViewCellCustomViewSize { return self == .oneLine ? .small : .medium }
+        var customViewSizeType: MSFTableViewCellCustomViewSize { return self == .oneLine ? .small : .medium }
     }
 
     struct TextStyles {
@@ -246,6 +246,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                                          subtitleTrailingAccessoryView: subtitleTrailingAccessoryView,
                                          footerLeadingAccessoryView: footerLeadingAccessoryView,
                                          footerTrailingAccessoryView: footerTrailingAccessoryView)
+        // Layout type should accommodate for the customViewSize, even if it is only one line.
         if customViewSize == .medium && layoutType == .oneLine {
             layoutType = .twoLines
         }
@@ -443,7 +444,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     }
 
     private static func customViewSize(from size: MSFTableViewCellCustomViewSize, layoutType: LayoutType) -> MSFTableViewCellCustomViewSize {
-        return size == .default ? layoutType.customViewSize : size
+        return size == .default ? layoutType.customViewSizeType : size
     }
 
     private static func customViewLeadingOffset(isInSelectionMode: Bool,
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -291,7 +246,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                                          subtitleTrailingAccessoryView: subtitleTrailingAccessoryView,
                                          footerLeadingAccessoryView: footerLeadingAccessoryView,
                                          footerTrailingAccessoryView: footerTrailingAccessoryView)
-        customViewSize.validateLayoutTypeForHeightCalculation(&layoutType)
+        if customViewSize == .medium && layoutType == .oneLine {
+            layoutType = .twoLines
```
mischreiber, line 250: Can you leave some comments explaining this logic? Seems odd to me that there's a check that says `if layoutType is oneLine then make it twoLines` without context.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -736,24 +693,26 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
-    /// Override to set a specific `CustomViewSize` on the `customView`
-    @objc open var customViewSize: CustomViewSize {
+    /// Override to set a specific `MSFTableViewCellCustomViewSize` on the `customView`
+    @objc open var customViewSize: MSFTableViewCellCustomViewSize {
         get {
             if customView == nil {
                 return .zero
             }
-            return _customViewSize == .default ? layoutType.customViewSize : _customViewSize
+            tokens.customViewSize = _customViewSize == .default ? layoutType.customViewSize : _customViewSize
```
mischreiber, line 702: Won't this mean that an overriden token set with a custom value for `customViewSize` will never be respected?

mischreiber: Ah I understand, that's a state prop rather than a measurement. I was expecting something of type `CGSize`. I wonder if maybe this value should be named something else?

harrieshin:
nit: can we put a clear note on this PR which includes "breaking" change of api?
another thought: we might need to have a checkbox like RN pr  checklist "does this include breaking changes?" so that we can easily capture them in release notes later.

harrieshin:
the pr description screenshots should include before and after of persona cell and popupmenu cell?

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Actually I don't think adding `Type` is great here because that word is never used anywhere else. I think we should probably stick with the original naming for now; consistency is more important here.

mischreiber (approved):

PR #1006, opened 2022-06-14
[vNext Tokens] Convert SegmentedControl from UIControl to UIView

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The change from UIControl to UIView will be breaking, as clients will now need to set the onSelectAction rather than adding a target.

The current SegmentedControl is a UIControl, and that means that keyboard navigation will only select the entire control. By changing it to a UIView, we can now select each individual button. I recorded the keyboard navigation videos with increased contrast to show where the focus was. We don't have focus colors from design right now, so I'll come back and update the colors to change once we have those. With this change, though, the control is at least usable with keyboard navigation.

### Verification

| SegmentedControl_UIView_before | SegmentedControl_UIView_after |
| SegmentedControl_Color_before | SegmentedControl_Color_after |
| Crash? | SegmentedControl_Table_after |
| SegmentedControl_Date_before | SegmentedControl_Date_after |
| SegmentedControl_Time_before | SegmentedControl_Time_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1006)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -541,3 +541,11 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
         }
     }
 }
+
+// MARK: SegmentedControlDelegate
+
+@objc(MSFSegmentedControlDelegate)
+public protocol SegmentedControlDelegate {
+    /// Called after the button representing the item is tapped in the UI.
+    @objc optional func segmentedControl(_ segmentedControl: SegmentedControl, didSelectItem item: SegmentItem, atIndex index: Int)
+}
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -20,7 +20,7 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
         static let iPadMinimumWidth: CGFloat = 375
     }
 
-    open override var isEnabled: Bool {
+    open var isEnabled: Bool = true {
         didSet {
             for button in buttons {
                 button.isEnabled = isEnabled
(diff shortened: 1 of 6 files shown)


ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -47,6 +47,8 @@ open class SegmentedControl: UIControl, TokenizedControlInternal {
             invalidateIntrinsicContentSize()
         }
     }
+    /// Action called when a segment is selected.
+    @objc public var onSelectAction: ((SegmentedControl, SegmentItem, Int) -> Void)?
```
harrieshin, line 51: better documentation what int means? but i don't quite understand why we need segmentedcoontrol as a param?

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -7,7 +7,7 @@ import UIKit
 // MARK: SegmentedControl
 /// A styled segmented control that should be used instead of UISegmentedControl. It is designed to flex the button width proportionally to the control's width.
 @objc(MSFSegmentedControl)
-open class SegmentedControl: UIControl, TokenizedControlInternal {
+open class SegmentedControl: UIView, TokenizedControlInternal {
```
harrieshin, line 10: any voiceover code changes we need now that we are uiview rather than uicontrol?

huwilkes: VoiceOver is still working fine, says "First, button, 1 of 3"

ios/FluentUI/SegmentedControl/SegmentedControl.swift
huwilkes: The SegmentedControl is in there so it could be used in the same ways as sender used to. We could probably remove it, but our demo control uses the SegmentedControls to access the labels below them.

edjamesmsft (approved):

harrieshin:
> Please document some details about the public API change.

also make note about in the PR how this is breaking change so when we merge and bring release notes easier to capture all the breaking changes together.

huwilkes:
> > Please document some details about the public API change.
> 
> also make note about in the PR how this is breaking change so when we merge and bring release notes easier to capture all the breaking changes together.

Updated the description and added a API break label

PR #1007, opened 2022-06-14
Add ability to set a reference start/end date for a specific datepicker

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A client asked for the ability to set a custom reference start/end date for datepicker in FURN. The initial fix for this (in the FURN repo) allowed the default reference start/end dates to be overridden, but this fix caused some build issues that are not straightforward to resolve.

This is the second attempt to add the ability to set a reference start/end date. It differs from the first in that instead of updating the default reference start/end dates, the reference start/end dates can be set for a specific date picker (which seems more useful anyway)

This is part 1 of 2 of this fix. The second half of the change will be in the FURN repo.

### Verification

Added a toggle in the DemoController date picker. When the toggle is on, all the date pickers should conform to the custom calendar configuration.

Screen Shot 2022-06-16 at 12 33 38 PM

### Notes

A lot of method headers were changed to add referenceStart/EndDate parameters. The parameters are optional and also have a default value of nil. Not sure what the best practices are for this but maybe the requirements for the parameters in some of the methods could be stricter.

### Known issues

1. For certain date pickers, a date outside of the given reference start/end date can be picked.
   - There are two "date picker types", and the latter does not restrict the available dates

"Calendar" date picker | "Components" date picker
--- | ---
Screen Shot 2022-06-16 at 2 53 28 PM | Screen Shot 2022-06-16 at 2 53 41 PM
This PR's addition of a custom calendar configuration works for all the date pickers shown in the demo controller, in that dates outside the range of referenceStartDate/referenceEndDate are not shown. | This PR does **not** restrict the range of some of the components. Components include month, day, year, date (not pictured in the screenshot), and others. In this PR, only the date and year component responds to the referenceStartDate and referenceEndDate (i.e. will restrict the dates/years outside of the given range of referenceDates). The day and month components do not respond to the referenceStart/EndDate, but maybe they should. The day and month components were built to be an infinite scrolling list, and additional work will be needed to restrict the possible options. Since the original issue that was opened was more related to expanding the available range of dates, rather than restricting the range of available dates, I think we can leave this as is in this PR and document this somewhere as a known issue. 

2. Edge case specific to the date range picker:
    - The date range picker is two date pickers, one for the start date (shown above) and one for the end date. 
   - Whatever range of dates was selected in the start date picker, will be kept when selecting the end date. So, if the range of selected days is 7 days, and then the selected start date gets moved to the last possible date, the end date will be out of the range of possible dates
   - Fixing this will involve adding logic to restrict the possible selected end date to the restricted reference range. I plan to also document this as a known issue.

Screen Shot 2022-06-16 at 3 17 31 PM

3. The creator of the control can set the reference start date to be after the reference end date. This will show a blank calendar, which seems reasonable.

4.  The only valid values for 'firstWeekday' are between 1 and 7, inclusive. The creator of the control can pass in an integer outside of this valid range. This will hit an assert when the firstWeekday is getting converted into a heading for the days of the week

5. There is a date picker type in code that cannot be created by the exposed datepicker API. There are three "DateTimePickerViewModes" (not to be confused with "DateTimePickerMode"): date, dateTime, dayOfMonth. The dayOfMonth view mode is never used anywhere (forced to show in code below), maybe can be removed?

Screen Shot 2022-06-16 at 3 55 12 PM

Testing done:
- there are 4 "modes" of date picker (date, datetime, daterange, datetimerange), and 2 "types" of date picker (calendar, components). All 8 combinations were tested (and the results documented above)
- confirmed that when the new custom calendarconfiguration switch is turned off in the democontroller, the default calendar configuration is used

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1007)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
@@ -25,15 +25,16 @@ class DateTimePickerDemoController: DemoController {
 
     private let customCalendarConfigurationSwitch = UISwitch()
     private var calendarConfiguration: CalendarConfiguration? {
-        if customCalendarConfigurationSwitch.isOn {
-            let customCalendarConfiguration = CalendarConfiguration()
-            customCalendarConfiguration.referenceStartDate = Date()
-            customCalendarConfiguration.referenceEndDate = Calendar.current.date(byAdding: .month, value: 1, to: Date())!
-            customCalendarConfiguration.firstWeekday = 2
-            return customCalendarConfiguration
-        } else {
+        guard customCalendarConfigurationSwitch.isOn else {
             return nil
         }
+
+        let customCalendarConfiguration = CalendarConfiguration()
+        customCalendarConfiguration.referenceStartDate = Date()
+        customCalendarConfiguration.referenceEndDate = Calendar.current.date(byAdding: .month, value: 1, to: Date())!
+        customCalendarConfiguration.firstWeekday = 2
+
+        return customCalendarConfiguration
     }
 
     private let validationSwitch = UISwitch()
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
@@ -65,44 +65,97 @@ class DateTimePickerDemoController: DemoController {
     }
 
     @objc func presentDatePicker() {
-        dateTimePicker.present(from: self, with: .date, startDate: startDate ?? Date(), calendarConfiguration: calendarConfiguration, datePickerType: datePickerType)
+        dateTimePicker.present(
+            from: self,
+            with: .date,
+            startDate: startDate ?? Date(),
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType)
     }
 
     @objc func presentDateTimePicker() {
-        dateTimePicker.present(from: self, with: .dateTime, startDate: startDate ?? Date(), calendarConfiguration: calendarConfiguration, datePickerType: datePickerType)
+        dateTimePicker.present(
+            from: self,
+            with: .dateTime,
+            startDate: startDate ?? Date(),
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType)
     }
 
     @objc func presentDateRangePicker() {
         let (startDate, endDate, _) = calcDatePickerParams()
-        dateTimePicker.present(from: self, with: .dateRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType)
+        dateTimePicker.present(
+            from: self,
+            with: .dateRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType)
     }
 
     @objc func presentTabbedDateRangePicker() {
         let (startDate, endDate, _) = calcDatePickerParams()
-        dateTimePicker.present(from: self, with: .dateRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType, dateRangePresentation: .tabbed)
+        dateTimePicker.present(
+            from: self,
+            with: .dateRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType,
+            dateRangePresentation: .tabbed)
     }
 
     @objc func presentDateTimeRangePicker() {
         let (startDate, endDate, _) = calcDatePickerParams()
-        dateTimePicker.present(from: self, with: .dateTimeRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType)
+        dateTimePicker.present(
+            from: self,
+            with: .dateTimeRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType)
     }
 
     @objc func presentCustomSubtitlePicker() {
         let (startDate, endDate, titles) = calcDatePickerParams()
-        dateTimePicker.present(from: self, with: .dateRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType, titles: titles)
+        dateTimePicker.present(
+            from: self,
+            with: .dateRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType,
+            titles: titles)
     }
 
     @objc func presentLeftBarButtonPicker() {
         let (startDate, endDate, titles) = calcDatePickerParams()
         let leftBarButtonItem = cancelButton(target: self, action: #selector(handleDidTapCancel))
-        dateTimePicker.present(from: self, with: .dateRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType, titles: titles, leftBarButtonItem: leftBarButtonItem)
+        dateTimePicker.present(
+            from: self,
+            with: .dateRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType,
+            titles: titles,
+            leftBarButtonItem: leftBarButtonItem)
     }
 
     @objc func presentRightBarButtonPicker() {
         let (startDate, endDate, titles) = calcDatePickerParams()
         let leftBarButtonItem = confirmButton(target: self, action: #selector(handleDidTapDone))
         let rightBarButtonItem = cancelButton(target: self, action: #selector(handleDidTapCancel)) // or simply assign UIBarButtonItem() to hide the default confirm button
-        dateTimePicker.present(from: self, with: .dateRange, startDate: startDate, endDate: endDate, calendarConfiguration: calendarConfiguration, datePickerType: datePickerType, titles: titles, leftBarButtonItem: leftBarButtonItem, rightBarButtonItem: rightBarButtonItem)
+        dateTimePicker.present(
+            from: self,
+            with: .dateRange,
+            startDate: startDate,
+            endDate: endDate,
+            calendarConfiguration: calendarConfiguration,
+            datePickerType: datePickerType,
+            titles: titles,
+            leftBarButtonItem: leftBarButtonItem,
+            rightBarButtonItem: rightBarButtonItem)
     }
 
     func createDatePickerTypeUI() -> UIStackView {
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
@@ -22,18 +22,16 @@ class DateTimePickerDemoController: DemoController {
             preconditionFailure("Unknown date picker type index")
         }
     }
-    
+
     private let customCalendarConfigurationSwitch = UISwitch()
     private var calendarConfiguration: CalendarConfiguration? {
-        if (customCalendarConfigurationSwitch.isOn)
-        {
+        if customCalendarConfigurationSwitch.isOn {
             let customCalendarConfiguration = CalendarConfiguration()
             customCalendarConfiguration.referenceStartDate = Date()
             customCalendarConfiguration.referenceEndDate = Calendar.current.date(byAdding: .month, value: 1, to: Date())!
             customCalendarConfiguration.firstWeekday = 2
             return customCalendarConfiguration
-        }
-        else {
+        } else {
             return nil
         }
     }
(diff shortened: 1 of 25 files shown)


harrieshin:
please update democontroller to test this case as well. make sure you add the screenshots or video of this change for before and after

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
```diff
@@ -99,7 +99,7 @@ class DatePickerController: UIViewController, GenericDateTimePicker {
...
     ///   - leftBarButtonItem: optional UIBarButtonItem to be presented as left bar-button.
     ///   - rightBarButtonItem: optional UIBarButtonItem to be presented oas right bar-button. Note that if this view is presented, the Confirm button is not generated automatically.
-    init(startDate: Date, endDate: Date, mode: DateTimePickerMode, selectionMode: DatePickerSelectionManager.SelectionMode = .start, rangePresentation: DateTimePicker.DateRangePresentation, titles: DateTimePicker.Titles?, leftBarButtonItem: UIBarButtonItem?, rightBarButtonItem: UIBarButtonItem?) {
+    init(startDate: Date, endDate: Date, referenceStartDate: Date? = nil, referenceEndDate: Date? = nil, mode: DateTimePickerMode, selectionMode: DatePickerSelectionManager.SelectionMode = .start, rangePresentation: DateTimePicker.DateRangePresentation, titles: DateTimePicker.Titles?, leftBarButtonItem: UIBarButtonItem?, rightBarButtonItem: UIBarButtonItem?) {
```
harrieshin, line 102: this should be a new convenience methods so existing objc callers won't be broken?

ios/FluentUI/Date Time Pickers/DateTimePicker.swift
```diff
@@ -99,11 +99,13 @@ public class DateTimePicker: NSObject {
...
     ///   - titles: A `Titles` object that holds strings for use in overriding the default picker titles, subtitles, and tab titles. If a string is provided for a property that does not apply to the current mode, it will be ignored.
-    @objc public func present(from presentingController: UIViewController, with mode: DateTimePickerMode, startDate: Date = Date(), endDate: Date? = nil, datePickerType: DatePickerType = .calendar, dateRangePresentation: DateRangePresentation = .paged, titles: Titles? = nil) {
-        present(from: presentingController, with: mode, startDate: startDate, endDate: endDate, datePickerType: datePickerType, dateRangePresentation: dateRangePresentation, leftBarButtonItem: nil, rightBarButtonItem: nil)
+    @objc public func present(from presentingController: UIViewController, with mode: DateTimePickerMode, startDate: Date = Date(), endDate: Date? = nil, referenceStartDate: Date? = nil, referenceEndDate: Date? = nil, datePickerType: DatePickerType = .calendar, dateRangePresentation: DateRangePresentation = .paged, titles: Titles? = nil) {
```
harrieshin, line 107: we should add new convenience objc public function so that existing objc callers remains unbroken.

lyzhan7: I assumed that FURN was the only place calling this, and that there's a way to get both the FURN side fix and the FUA side fix in at the same time so nothing is broken. Are there potentially other places that are calling this? Does that mean that methods like this cannot be modified at all after they've been made?

lyzhan7: I see okay. I think it makes the most sense to pass the referenceStartDate/referenceEndDate into the present method along with all of the other parameters needed to get the picker set up, but if it's not possible to modify the present function at all I'll need to rethink this PR.

lyzhan7: Never mind - understand what you mean now and will update

ios/FluentUI/Calendar/CalendarConfiguration.swift
```diff
@@ -24,7 +24,7 @@ open class CalendarConfiguration: NSObject {
     /// By default, this is the default 'referenceStartDate' plus 10 years.  If overridden, make sure it's after (greater than) the 'referenceStartDate'
     @objc open var referenceEndDate: Date
 
-    @objc init(calendar: Calendar = .current) {
+    @objc public init(calendar: Calendar = .current) {
```
lyzhan7, line 27: I think I needed to add this in order to make a CalendarConfiguration object in the demoController

lyzhan7:
As of commit 5: added convenience methods + added a test to the demo controller.

Screenshot before | Screenshots after
---|---
image | image

Still to do:
- fix datetimecontroller (maybe also needs a test)
- there are some other datepickers (ex. pick year) that might need to get the calendarConfiguration set as well - will look into

ios/FluentUI/Calendar/CalendarConfiguration.swift
harrieshin: now that this is public we need to add documentation on this class

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
```diff
@@ -114,15 +117,16 @@ class DatePickerController: UIViewController, GenericDateTimePicker {
...
         super.init(nibName: nil, bundle: nil)
 
         defer {
             self.startDate = startDate
             self.endDate = endDate
         }
-
-        calendarViewDataSource = CalendarViewDataSource(styleDataSource: self)
+        
+        calendarViewDataSource = CalendarViewDataSource(styleDataSource: self, calendarConfiguration: self.calendarConfiguration)
```
lyzhan7, line 129: will fix the spaces

sophialee0416: (Having invisibles on should help with this if you don't have that on already!)

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
```diff
@@ -44,6 +44,8 @@ class DateTimePickerDemoController: DemoController {
         container.addArrangedSubview(createButton(title: "Show picker with custom subtitles or tabs", action: #selector(presentCustomSubtitlePicker)))
         container.addArrangedSubview(createButton(title: "Show picker with left bar-button", action: #selector(presentLeftBarButtonPicker)))
         container.addArrangedSubview(createButton(title: "Show picker with left and right bar-buttons", action: #selector(presentRightBarButtonPicker)))
+        container.addArrangedSubview(createButton(title: "Show date picker with custom calendar configuration (First weekday: Monday, Range of reference dates: a month from today", action: #selector(presentCustomCalendarConfigurationDatePicker)))
+        container.addArrangedSubview(createButton(title: "Show date time picker with custom calendar configuration (First weekday: Monday, Range of reference dates: a month from today", action: #selector(presentCustomCalendarConfigurationDateTimePicker)))
```
Saadnajmi, line 48: To be more terse, "Range of reference dates" -> "Reference date range"
Maybe we can also add newlines, but that's a nit?

lyzhan7: Just changed the demo controller to use a switch for the calendar configuration so that the custom config can be applied to all the test pickers - will add screenshots

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
```diff
@@ -165,6 +167,24 @@ class DateTimePickerDemoController: DemoController {
         alert.addAction(UIAlertAction(title: "Ok", style: .default, handler: nil))
         present(alert, animated: true)
     }
+    
+    @objc private func presentCustomCalendarConfigurationDatePicker() {
+        let customCalendarConfiguration = CalendarConfiguration()
+        customCalendarConfiguration.referenceStartDate = Date()
+        customCalendarConfiguration.referenceEndDate = Calendar.current.date(byAdding: .month, value: 1, to: Date())!
```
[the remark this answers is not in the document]
lyzhan7: Yes I was a little unsure about this, but seemed safe enough to do and it's also just in the democontroller. Will generally avoid, thanks for letting me know!

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
```diff
@@ -93,13 +95,14 @@ class DatePickerController: UIViewController, GenericDateTimePicker {
...
     ///   - leftBarButtonItem: optional UIBarButtonItem to be presented as left bar-button.
     ///   - rightBarButtonItem: optional UIBarButtonItem to be presented oas right bar-button. Note that if this view is presented, the Confirm button is not generated automatically.
-    init(startDate: Date, endDate: Date, mode: DateTimePickerMode, selectionMode: DatePickerSelectionManager.SelectionMode = .start, rangePresentation: DateTimePicker.DateRangePresentation, titles: DateTimePicker.Titles?, leftBarButtonItem: UIBarButtonItem?, rightBarButtonItem: UIBarButtonItem?) {
+    init(startDate: Date, endDate: Date, calendarConfiguration: CalendarConfiguration? = nil, mode: DateTimePickerMode, selectionMode: DatePickerSelectionManager.SelectionMode = .start, rangePresentation: DateTimePicker.DateRangePresentation, titles: DateTimePicker.Titles?, leftBarButtonItem: UIBarButtonItem?, rightBarButtonItem: UIBarButtonItem?) {
```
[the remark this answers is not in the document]
lyzhan7: Yes that's definitely an option. I have this "calendarConfiguration: CalendarConfiguration? = nil" in a lot of places. Do we want to change ALL of them to calendarConfiguration: CalendarConfiguration? = CalendarConfiguration.default? Or maybe just one of the ones further up the stack can be the default, and all the calls further down the stack can just use "calendarConfiguration: CalendarConfiguration". Not sure what best practice is here

lyzhan7:
Update screenshots as of commit 9:

Screen Shot 2022-06-16 at 12 33 38 PM

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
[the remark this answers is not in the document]
Saadnajmi: All of them was my suggestion, but feel free to disagree if you think it makes sense to keep it as is

ios/FluentUI/Calendar/CalendarConfiguration.swift
harrieshin: or comand + option + / in xcode shortcut to follow the latest syntax of documentation

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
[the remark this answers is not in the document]
lyzhan7: Fine with all three (all nil, or all default, or some default/some required parameters). Are there any advantages to switching them from all nil to all default? 

I think a possible benefit of keeping it as nil until the very end is that we can differentiate between whether the creator of the date picker decided not to pass in anything at all, or actually passed in the default value. This could be useful for debugging, or maybe for new features in the future (although I can also see an argument that these two scenarios should always be treated the same way). 

Not a strong opinion though,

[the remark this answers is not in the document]
Saadnajmi: My reasoning was I mostly try to avoid the keyword "nil" because null pointer exceptions scare me, and it felt like the default value of an optional value should be.. the default configuration.

I think those two aren't strong reasons though, and I couldn't really find any guidance on aka.ms/swift-guide or docs.swift.org/swift-book/LanguageGuide/Functions.html saying otherwise. As such, I think this is mostly a stylistic choice, so I'll leave it up to you what you prefer.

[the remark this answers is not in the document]
rdeassis: I think that if all the instances we'd be using the calendar configuration are something like below, it makes sense to make the parameter type non-optional and just provide the CalendarConfiguration.default as the default parameter.

`calendarConfiguration?.referenceStartDate ?? CalendarConfiguration.default.referenceStartDate`

It'd be just simpler because it'd avoid these checks everywhere same code above would be replaced by just:

`calendarConfiguration.referenceStartDate`

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
```diff
@@ -23,6 +23,19 @@ class DateTimePickerDemoController: DemoController {
         }
     }
 
+    private let customCalendarConfigurationSwitch = UISwitch()
+    private var calendarConfiguration: CalendarConfiguration? {
+        if customCalendarConfigurationSwitch.isOn {
```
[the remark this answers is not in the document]
lyzhan7: Just updated - haven't used this before so if this is different from what you were expecting let me know!

ios/FluentUI.Demo/FluentUI.Demo/Demos/DateTimePickerDemoController.swift
```diff
@@ -46,10 +59,52 @@ class DateTimePickerDemoController: DemoController {
         container.addArrangedSubview(createButton(title: "Show picker with left and right bar-buttons", action: #selector(presentRightBarButtonPicker)))
         container.addArrangedSubview(UIView())
         container.addArrangedSubview(createDatePickerTypeUI())
+        container.addArrangedSubview(createCustomCalendarConfigurationUI())
         container.addArrangedSubview(createValidationUI())
         container.addArrangedSubview(createButton(title: "Reset selected dates", action: #selector(resetDates)))
     }
 
+    @objc func presentDatePicker() {
```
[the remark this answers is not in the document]
lyzhan7: Just updated in a bunch of places I saw in a couple files. Looks like neither swiftlint nor xcode (through CTRL + I) has an opinion on the how the parameters are formatted, let me know if this is what you were expecting/if there are other examples I can follow

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerController.swift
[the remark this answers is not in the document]
lyzhan7: Okay just pushed a commit that replaces a bunch of places where I had `calendarConfiguration: CalendarConfiguration? = nil` with just `calendarConfiguration: CalendarConfiguration`. I realized while doing this that  there were two variables startYear and endYear in `case date(startYear: Int, endYear: Int)` in the DateTimePickerViewMode enum that I think are redundant now after introducing calendarConfiguration. The code is kind of confusing but as far as I can tell, before this PR these variables always ended up as the default calendar configuration. I also went ahead and removed static let defaultStartYear and defaultEndYear in DateTimePickerViewMode enum. 

The two changes in this commit (making calendarConfiguration parameters non-optional/no default) and the changes to the DateTimePickerViewMode enum are related. If the above change to the DateTimePickerViewMode enum seems out of scope for this PR, I can revert this whole commit.

harrieshin (approved):

ios/FluentUI/Calendar/CalendarConfiguration.swift
lyzhan7: @harrieshin the documentation I added is in this commit: https://github.com/microsoft/fluentui-apple/pull/1007/commits/f13ea87307f045b225f5724866f8acccb9384f22. I added a comment above the class declaration. Were you also expecting documentation on the init method?

Saadnajmi (approved):

ios/FluentUI/Calendar/CalendarConfiguration.swift
lyzhan7: Never mind actually - looked at other @objc public init methods and noticed they all have comments. Will add

rdeassis (approved):

PR #1013, opened 2022-06-15
Adding NSAttributedString support to TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added support to use `NSAttributedString` for labels in TableViewCell. The attributed text is optional, so if it is not set, the cell will default to the original `String` text.

### Verification

Example NSAttributedString usage for title (NSAttributedString takes precedence over String):

Simulator Screen Shot - iPhone 12 - 2022-06-15 at 16 07 51

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1013)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1036,20 +1036,38 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     ///
     /// - Parameters:
     ///   - title: Text that appears as the first line of text
+    ///   - attributedTitle: Optional attributed text for the first line of text. If this is not set, the title will be used
     ///   - subtitle: Text that appears as the second line of text
+    ///   - attributedSubtitle: Optional attributed text for the second line of text. If this is not set, the subtitle will be used
     ///   - footer: Text that appears as the third line of text
+    ///   - attributedFooter: Optional attributed text for the third line of text. If this is not set, the footer will be used
     ///   - customView: The custom view that appears near the leading edge next to the text
     ///   - customAccessoryView: The view acting as an accessory view that appears on the trailing edge, next to the accessory type if provided
     ///   - accessoryType: The type of accessory that appears on the trailing edge: a disclosure indicator or a details button with an ellipsis icon
-    @objc open func setup(title: String,
+    @objc open func setup(title: String = "",
+                          attributedTitle: NSAttributedString? = nil,
                           subtitle: String = "",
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1036,20 +1036,38 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
+    ///   - attributedTitle: Optional attributed text for the first line of text. If this is not set, the title will be used
     ///   - subtitle: Text that appears as the second line of text
+    ///   - attributedSubtitle: Optional attributed text for the second line of text. If this is not set, the subtitle will be used
     ///   - footer: Text that appears as the third line of text
+    ///   - attributedFooter: Optional attributed text for the third line of text. If this is not set, the footer will be used
     ///   - customView: The custom view that appears near the leading edge next to the text
     ///   - customAccessoryView: The view acting as an accessory view that appears on the trailing edge, next to the accessory type if provided
     ///   - accessoryType: The type of accessory that appears on the trailing edge: a disclosure indicator or a details button with an ellipsis icon
-    @objc open func setup(title: String,
+    @objc open func setup(title: String = "",
```
huwilkes, line 1047: We probably need to add a convenience init so we don't break the objc clients, right?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1036,20 +1036,38 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
+                          attributedTitle: NSAttributedString? = nil,
```
huwilkes, line 1048: Seems a bit odd to me to have both the title and attributed title. Would it be worth passing in the string and attributed separately? So that way you're using title either way, and then only sometimes adding a bunch of attributes.

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: I can add another setup method!

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: I think with the comment in the first thread, I want to leave the original setup method alone. Either I can create another setup method that basically implements this, or maybe I can add support for adding NSAttributedString as a property (kinda like you mentioned)

PR #1015, opened 2022-06-16
Adding NSAttributedString support to TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added new API support to use `NSAttributedString` for labels in TableViewCell. The attributed text is optional, so if it is not set, the cell will default to the original `String` text.

Referencing #1013.

### Verification

Example NSAttributedString usage for title (NSAttributedString takes precedence over String):

Simulator Screen Shot - iPhone 12 Pro - 2022-06-21 at 03 10 12

Simulator Screen Shot - iPhone 12 - 2022-06-15 at 16 07 51

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1015)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -970,9 +970,54 @@ open class TableViewCell: UITableViewCell {
     ///   - customAccessoryView: The view acting as an accessory view that appears on the trailing edge, next to the accessory type if provided
     ///   - accessoryType: The type of accessory that appears on the trailing edge: a disclosure indicator or a details button with an ellipsis icon
     @objc open func setup(title: String, subtitle: String = "", footer: String = "", customView: UIView? = nil, customAccessoryView: UIView? = nil, accessoryType: TableViewCellAccessoryType = .none) {
-        titleLabel.text = title
-        subtitleLabel.text = subtitle
-        footerLabel.text = footer
+        setup(title: title,
+              attributedTitle: nil,
+              subtitle: subtitle,
+              attributedSubtitle: nil,
+              footer: footer,
+              attributedFooter: nil,
+              customView: customView,
+              customAccessoryView: customAccessoryView,
+              accessoryType: accessoryType)
+    }
+
+    /// Sets up the cell with text, a custom view, a custom accessory view, and an accessory type
+    ///
+    /// - Parameters:
+    ///   - title: Text that appears as the first line of text
+    ///   - attributedTitle: Optional attributed text for the first line of text. If this is not set, the title will be used
+    ///   - subtitle: Text that appears as the second line of text
+    ///   - attributedSubtitle: Optional attributed text for the second line of text. If this is not set, the subtitle will be used
+    ///   - footer: Text that appears as the third line of text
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -970,9 +970,54 @@ open class TableViewCell: UITableViewCell {
...
+    @objc open func setup(title: String = "",
+                          attributedTitle: NSAttributedString? = nil,
+                          subtitle: String = "",
+                          attributedSubtitle: NSAttributedString? = nil,
+                          footer: String = "",
+                          attributedFooter: NSAttributedString? = nil,
+                          customView: UIView? = nil,
+                          customAccessoryView: UIView? = nil,
+                          accessoryType: TableViewCellAccessoryType = .none) {
+        if let attributedTitle = attributedTitle {
```
harrieshin, line 1005: could this block of code be one line something like  attributedTitle?:title

sophialee0416: titleLabel.attributedText = attributedTitle != nil ? attributedTitle : NSAttributedString(string: title)

Does that look okay? That would make any text into an attributed string though

harrieshin: ahhh you are right. actually we should jsut use https://developer.apple.com/documentation/uikit/uilistcontentconfiguration ?

sophialee0416: This looks like a good idea. I would have to override the `defaultContentConfiguration` and add our own properties though, right? Which would include updating the demo? Should I do that in this PR?

harrieshin: so it would be something like self.defaultContentConfiguration().text = "foo" in this function?

harrieshin:
i think we should also start puting deprecation warning in "setTitleLabelTextColor" and "setSubTitleLabelTextColor" as well so people can move over to your new function?

ios/FluentUI/Table View/TableViewCell.swift
harrieshin: oh wait... sorry... do our cell actually use the OS tableviewcell's label? :)

sophialee0416: Well there's 3 different text labels so I think I'd have to override the default config so that it knows the difference between the 3 labels?

harrieshin: i don't think overiding defaultcontentconfiguration would work. looking at the tableviewcell code we created our OWN labels. so your current approach is correct.

sophialee0416: Sorry your previous message about using the OS labels didn't load early so I didn't see that before my reply :P Okay yes sounds good, but this discussion was productive still!

rdeassis (approved):

PR #1019, opened 2022-06-20
Fixing warning in TableViewHeaderFooterViewDemoController.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The following warning is currently being issued by Xcode in the TableViewHeaderFooterViewDemoController:

warning_self_initialisation

I found out that Xcode's suggestion (and fix) is not correct and ends up crashing the app when the segmented control is used to switch tabs.

**Cause:**
The use of `self` should be available once the class is fully initialized.
The warning is being issued because there is a reference to `self` in the initialisation closure of the property of the class which is not yet ready at the time it's executed.

**Fix:**
The fix for the warning is to make the property a lazy var so that it's only initialized when it's called for use, which will be after the class initialisation. At this point `self` will be available.

### Verification

1. No warnings are issued anymore.
2. The demo controller works as expected when switching tabs.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1019)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewHeaderFooterViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewHeaderFooterViewDemoController.swift
@@ -12,9 +12,13 @@ class TableViewHeaderFooterViewDemoController: DemoController {
     private let groupedSections: [TableViewHeaderFooterSampleData.Section] = TableViewHeaderFooterSampleData.groupedSections
     private let plainSections: [TableViewHeaderFooterSampleData.Section] = TableViewHeaderFooterSampleData.plainSections
 
-    private let segmentedControl: SegmentedControl = {
-        let segmentedControl = SegmentedControl(items: TableViewHeaderFooterSampleData.tabTitles.map({return SegmentItem(title: $0)}), style: .primaryPill)
-        segmentedControl.addTarget(self, action: #selector(updateActiveTabContent), for: .valueChanged)
+    private lazy var segmentedControl: SegmentedControl = {
+        let tabTitles = TableViewHeaderFooterSampleData.tabTitles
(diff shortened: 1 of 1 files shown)


huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewHeaderFooterViewDemoController.swift
```diff
@@ -12,9 +12,13 @@ class TableViewHeaderFooterViewDemoController: DemoController {
...
     private let plainSections: [TableViewHeaderFooterSampleData.Section] = TableViewHeaderFooterSampleData.plainSections
 
-    private let segmentedControl: SegmentedControl = {
-        let segmentedControl = SegmentedControl(items: TableViewHeaderFooterSampleData.tabTitles.map({return SegmentItem(title: $0)}), style: .primaryPill)
-        segmentedControl.addTarget(self, action: #selector(updateActiveTabContent), for: .valueChanged)
+    private lazy var segmentedControl: SegmentedControl = {
+        let tabTitles = TableViewHeaderFooterSampleData.tabTitles
+        let segmentedControl = SegmentedControl(items: tabTitles.map({ return SegmentItem(title: $0) }),
+                                                style: .primaryPill)
+        segmentedControl.addTarget(self,
```
huwilkes, line 19: Since I removed this in #1006, we'll need to just remove this when we merge to fluent2-tokens

rdeassis: I agree. This is mainly to remove the warning while that particular change is not yet merged back into main.

PR #1025, opened 2022-06-22
Fix badge on/off for tab bar in demo

vickprat:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes
While running the demo, if we switch ON the badge value for the tab bar view items then for the home item, the badge value never gets hidden even if we switch off the badge value. The reason for this was that we were reusing the same home item instead of creating a new one as we are doing with the other items, so this PR is fixing that by creating a new home item.

### Verification
|Badge Bug Tab Bar|BadgeOnTabBar|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1025)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1081,7 +1081,7 @@ open class TableViewCell: UITableViewCell {
         }
     }
 
-    @available(*, deprecated, message: "Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString.")
+    // Any color or stylistic changes on TableViewCell labels should be done through attributedTitle
     /// To set color for title label
     /// - Parameter color: UIColor to set
     @objc public func setTitleLabelTextColor(color: UIColor) {
@@ -1089,7 +1089,7 @@ open class TableViewCell: UITableViewCell {
         isUsingCustomTextColors = true
     }
 
-    @available(*, deprecated, message: "Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString.")
+    // Any color or stylistic changes on TableViewCell labels should be done through attributedSubtitle
     /// To set color for subTitle label
     /// - Parameter color: UIColor to set
     @objc public func setSubTitleLabelTextColor(color: UIColor) {
@@ -1097,6 +1097,7 @@ open class TableViewCell: UITableViewCell {
         isUsingCustomTextColors = true
     }
 
+    // Any color or stylistic changes on TableViewCell labels should be done through attributedFooter
     /// To set color for footer label
     /// - Parameter color: UIColor to set
     public func setFooterLabelTextColor(color: UIColor) {


ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -74,12 +74,14 @@ class TabBarViewDemoController: DemoController {
         updatedTabBarView.delegate = self
 
         if showsItemTitles {
+            homeItem = TabBarItem(title: "Home", image: UIImage(named: "Home_24")!, selectedImage: UIImage(named: "Home_Selected_24")!)
```
rdeassis, line 77: The initialization of the lazy var homeItem is a bit different than these 2 calls being introduced:

```
    private lazy var homeItem: TabBarItem = {
        return TabBarItem(title: "Home", 
                          image: UIImage(named: "Home_28")!, 
                          selectedImage: UIImage(named: "Home_Selected_28")!,
                          landscapeImage: UIImage(named: "Home_24")!,
                          landscapeSelectedImage: UIImage(named: "Home_Selected_24")!)
    }()
```

Have you tested it in landscape mode?

rdeassis, line 77: Probably better to have a func that creates the items based on a parameter (show titles) and call it here and in the initalization of the homeItem variable?

vickprat: yes I have tested and I did this intentionally because now it is correct and earlier it was wrong because when we show titles then we should be using _24 asset otherwise _28 asset and in landscape it should always be _24 asset.

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -103,6 +105,20 @@ class TabBarViewDemoController: DemoController {
         updateBadgeNumbers()
     }
 
+    private func homeItem(shouldShowTitle: Bool) -> TabBarItem {
+        if shouldShowTitle {
+            return TabBarItem(title: "Home",
+                              image: UIImage(named: "Home_24")!,
+                              selectedImage: UIImage(named: "Home_Selected_24")!)
+        } else {
```
rdeassis, line 113: nit: no need for the else here if you're just returning directly in the if clause.

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -103,6 +105,20 @@ class TabBarViewDemoController: DemoController {
         updateBadgeNumbers()
     }
 
+    private func homeItem(shouldShowTitle: Bool) -> TabBarItem {
```
rdeassis, line 108: You can use this for the initialization of the property too. That way you're avoiding duplicated code.

rdeassis (approved):

mischreiber (approved):

PR #1026, opened 2022-06-22
Removing deprecation messages in TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Removing deprecation messages and replacing with comments to prevent breaking downstream dependencies.

### Verification

Build + pod lib lint

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1026)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1081,7 +1081,7 @@ open class TableViewCell: UITableViewCell {
         }
     }
 
-    @available(*, deprecated, message: "Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString.")
+    //Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString
     /// To set color for title label
     /// - Parameter color: UIColor to set
     @objc public func setTitleLabelTextColor(color: UIColor) {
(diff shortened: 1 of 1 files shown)


huwilkes:
Why are the deprecation messages being turned in to comments?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1081,22 +1081,23 @@ open class TableViewCell: UITableViewCell {
         }
     }
 
-    @available(*, deprecated, message: "Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString.")
+    // Any color or stylistic changes on TableViewCell labels should be done through NSAttributedString
```
huwilkes, line 1084: Should we name the actual property instead of its type?

huwilkes, line 1084: Should we just be taking the color they give here and just applying it to the correct attributed string?

sophialee0416: I don't think so, I think we want to deter clients from using this in the future so eventually, and hopefully, we'll be able to remove this?

rdeassis (approved):

sophialee0416:
> Why are the deprecation messages being turned in to comments?

Resolved offline + updated description

PR #1032, opened 2022-06-23
[Fluent2 iOS] Moved new color tokens into a consolidated enum

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

After offline discussions, we have decided to consolidate the new tokens under a single "color" enum/structure. Only one reference to the new tokens has been checked in (ActivityIndicator).

### Verification

Built and ran the demo app, verifying that the ActivityIndicator demo is behaving and looks as expected.

|  Simulator Screen Shot - iPhone 13 Pro - 2022-06-23 at 16 12 23 | Simulator Screen Shot - iPhone 13 Pro - 2022-06-23 at 16 15 47 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1032)

### DIFF
--- a/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
@@ -7,278 +7,6 @@ import SwiftUI
 
 public final class AliasTokens {
 
-    // MARK: Colors
-
-    public enum ColorsTokens: CaseIterable {
-        // Foreground colors
-        case foreground1
-        case foreground2
-        case foreground3
-        case foregroundDisabled1
-        case foregroundDisabled2
-        case foregroundContrast
-        case foregroundOnColor
-        case foregroundInverted1
-        case foregroundInverted2
-        case brandForeground1
-        case brandForeground1Pressed
-        case brandForeground1Selected
-        case brandForeground2
-        case brandForeground3
-        case brandForeground4
-        case brandForeground5
-        case brandForegroundInverted
-        case brandForegroundDisabled
-        // Background colors
-        case background1
-        case background1Pressed
-        case background1Selected
-        case background2
-        case background2Pressed
-        case background2Selected
-        case background3
-        case background3Pressed
-        case background3Selected
-        case background4
-        case background4Pressed
-        case background4Selected
-        case background5
-        case background5Pressed
-        case background5Selected
-        case background5SelectedBrandFilled
-        case background5BrandTint
-        case background6
-        case background6Pressed
-        case background6Selected
-        case backgroundInverted
-        case backgroundDisabled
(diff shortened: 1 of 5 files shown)


huwilkes (approved):

laminesm:
What were the main reasons behind this change?

rdeassis:
> fline discussions, we have decided to consolidate the new tokens under a single "color" enum/structure. Only one reference to the new tokens has been checked in (ActivityIndicator).

Good question!
I think it'd be nice to have the reasoning behind consolidating all the color tokens in a single enum.
Does that mean that any other alias token that is a color will be added to this enum?

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -40,56 +29,61 @@ public final class AliasTokens {
         case brandForeground5
         case brandForegroundInverted
         case brandForegroundDisabled
-
+        // Background colors
+        case background1
+        case background1Pressed
```
rdeassis, line 34: The [ButtonDynamicColors](https://github.com/microsoft/fluentui-apple/blob/a65fb6224a789f562ca51c5f12b264b2f4a37f97/ios/FluentUI/Core/Theme/Tokens/ButtonDynamicColors.swift#L7) has a structured way to define different colors for different states.
This enumeration is just concatenating the states to the type of color.

Can we have a way to reuse that structure or a similar idea so we don't have a combination matrix of all **colors x states**?

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -7,21 +7,10 @@ import SwiftUI
 
 public final class AliasTokens {
 
-    // MARK: ForegroundColors
+    // MARK: Colors
```
huwilkes, line 10: nit: I think the git diff would be easier to read if we move colors to the bottom, so it's more clear that all the new values were removed and put in one group.

edjamesmsft:
The goal of consolidating into a single enum is to 1. align exactly with the design specification in figma 2. align with FURN token organization and 3. make it easier for the new tokens to live alongside the old tokens while we transition between the two.

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
edjamesmsft: See response to conversation - organizing it in the style of ButtonDynamicColors would defeat purposes 1 and 2, and I believe add unnecessary complication to what should be a very straightforward set of enums.

Issue #985, opened 2022-05-17
`overrideTokens` replace theme-wide tokens, instead of cascading


mischreiber:
### Environment Information

- **Platform**:
  - [x] iOS
  - [ ] macOS
- **Package version(s)**: 0.5.0, all package managers
- **Xcode and OS versions**: Xcode 13.3.1, iOS 15.x

### Please provide a reproduction of the bug:

* Open a demo view that supports both per-control override and theme-wide override tokens (e.g. `CardNudge`)
* Update theme-wide override to include additional properties beyond what are covered by per-control override tokens (e.g. `accentColor`)
* Enable theme-wide override, then per-control override.

#### Actual behavior:

Theme-wide custom `accentColor` is replaced by default value.

#### Expected behavior:

Theme-wide custom `accentColor` is maintained, as per-control override doesn't override that value.

### Priorities and help requested:

Are you willing to submit a PR to fix? (**Yes**, No)

Requested priority: (Blocking, **High**, Normal, Low)

Products/applications affected: (if applicable)

mischreiber:
The fix for this will be to implement some form of cascading token resolution. I have a WIP in my personal fork; PR expected in the coming days.

mischreiber:
This is now fixed in the `fluent2-tokens` branch. Next step will be merging to `main`.

mischreiber:
Resolved back with https://github.com/microsoft/fluentui-apple/pull/1219, which is now released in the 0.8 branch!

PR #1033, opened 2022-06-23
[Fluent2 Tokens] Introducing the `TokenResolver`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In response to issue #985, this change brings a new internal system for fetching tokens: the `TokenResolver`.

This new class is responsible for the logic that was previously handled by the `resolvedTokens` property. That property would resolve between `overrideTokens`, `themeTokens`, and `defaultTokens` on a per-control basis, but we actually need per-token granular resolution. This new class provides that!

The per-value lookup is implemented via a method called `value` that takes a `KeyPath` of the control's `TokenType` and returns the first value in the `overrideTokens -> themeTokens -> defaultTokens` chain that _differs_ from the default value (or, naturally, the default value if no such difference exists). This allows for both per-control and theme-wide token overrides to modify different values. 

The API for fetching a token has changed. Here's a simple example:

Old:
```swift
var tokens: CardNudgeTokens { return resolvedTokens } // Special getter that does resolution

var body: some View {
    Rectangle()
        .foregroundColor(Color(dynamicColor: tokens.foregroundColor))
}
```

New:
```swift
@ObservedObject var tokenResolver: TokenResolver<CardNudge> = .init()

var body: some View {
    Rectangle()
        .foregroundColor(Color(dynamicColor: tokenResolver.value(\.foregroundColor)))
}
```

### Verification

In these examples, the `CardNudgeDemoController` is providing overrides in both `overrideTokens` and `themeTokens` -- both overrides have a custom background color, but only theme-wide tokens provide custom border width and color values. In the old system, this custom  border information is lost, but with this change we now successfully present both sets of tokens.

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1033)


ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -250,6 +251,7 @@
 		1168630222E131CF0088B302 /* TabBarItemView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TabBarItemView.swift; sourceTree = "<group>"; };
 		1168630322E131CF0088B302 /* TabBarView.swift */ = {isa = PBXFileReference; fileEncoding = 4; lastKnownFileType = sourcecode.swift; path = TabBarView.swift; sourceTree = "<group>"; };
 		118D9847230BBA2300BC0B72 /* TabBarItem.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TabBarItem.swift; sourceTree = "<group>"; };
+		2A9745DD281733D700E1A1FD /* TableViewTokens.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = TableViewTokens.swift; sourceTree = "<group>"; };
```
huwilkes, line 254: Did you mean to move the table view tokens? Seems like it should stay un-alphabetical for this review.

huwilkes (approved):

ios/FluentUI.xcodeproj/project.pbxproj
mischreiber: That is not actually an alphabetization; it looks like this file was merged improperly at some point and the reference to `TableViewTokens` was incorrectly placed in the list of files. This move was automatically done by Xcode to sanitize the project file.

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
```diff
@@ -0,0 +1,72 @@
...
+        } else if let themeValue = themeTokens?[keyPath: keyPath], themeValue != defaultValue {
+            return themeValue
+        } else {
+            return defaultValue
+        }
+    }
+
+    /// Prepares all three token sets by installing the current `FluentTheme` and then allowing the control
+    /// to perform additional configuration (e.g. setting `style` or `size` properties) via this class's `tokenConfig` property.
+    func configure(_ tokenConfig: ((_ tokens: TokenType) -> Void)? = nil) {
```
laminesm: Why do we pass in TokenType for additional configuration on the tokens? Would it work with ControlTokens instead for example?

mischreiber: > Would it work with ControlTokens instead for example?

Yep! In fact, if you look at line 45, you'll see that `TokenType` is simply an alias to the current control's token type, which itself is required to be of type `ControlTokens`.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CardNudgeDemoController.swift
```diff
@@ -260,6 +260,15 @@ extension CardNudgeDemoController: DemoAppearanceDelegate {
     // MARK: - Custom tokens
 
     private class ThemeWideOverrideCardNudgeTokens: CardNudgeTokens {
+        override var outlineWidth: CGFloat {
+            return 10.0
+        }
+
+        override var outlineColor: DynamicColor {
+            return DynamicColor(light: GlobalTokens().sharedColors[.darkRed][.tint50],
```
rdeassis, line 268: Can you store the GlobalTokens in a constant instead of instantiating it twice? 
I don't think it'll be a big deal in this case, but it's a good way to set the example for the client apps on how to retrieve these tokens.

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -177,7 +174,7 @@ public struct CardNudge: View, ConfigurableTokenizedControl {
     }
 }
 
-class MSFCardNudgeStateImpl: NSObject, ControlConfiguration, MSFCardNudgeState {
+class MSFCardNudgeStateImpl: NSObject, ObservableObject, Identifiable, MSFCardNudgeState {
```
rdeassis, line 177: Are these protocols always going to be implemented by the State classes? Worth creating a typealias to combine both of them?

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -197,9 +194,6 @@ class MSFCardNudgeStateImpl: NSObject, ControlConfiguration, MSFCardNudgeState {
     /// Action to be dispatched by the dismiss ("close") button on the trailing edge of the control.
     @Published var dismissButtonAction: CardNudgeButtonAction?
 
-    /// Design token set for this control, to use in place of the control's default Fluent tokens.
-    @Published var overrideTokens: CardNudgeTokens?
```
rdeassis, line 201: Just to help me understand better: What's the reasoning behind moving the "override tokens" (now token resolver) out of the state object?

ios/FluentUI/Core/Theme/Tokens/TokenizedControlView.swift
```diff
@@ -0,0 +1,30 @@
...
+
+// MARK: - Extensions
+
+extension TokenizedControlView {
+    /// Modifier function that updates the design tokens for a given control.
+    ///
+    /// - Parameter tokens: The tokens to apply to this control.
+    ///
+    /// - Returns: A version of this control with these overridden tokens applied.
+    func overrideTokens(_ overrideTokens: TokenType?) -> Self {
```
rdeassis, line 26: I like this! 
It provides the default implementation to save all SwiftUI controls from implementing the same modifier while not requiring UIKit controls to have to implement a func that looks like a modifier.

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
```diff
@@ -0,0 +1,68 @@
...
+/// Contains all possible token overrides for a given control, and returns the appropriate token value for a given property.
+public class TokenResolver<ControlType: TokenizedControl>: NSObject, ObservableObject {
+
+    /// Returns the appropriate token value for a given key path on this control's token set.
+    ///
+    /// This method looks for the first value that does not match the default value provided by `defaultTokens`.
+    /// Priority order for tokens are `overrideTokens`, `themeTokens`, then finally `defaultTokens`.
+    func value<ValueType: Equatable>(_ keyPath: KeyPath<TokenType, ValueType>) -> ValueType {
+        let defaultValue = defaultTokens[keyPath: keyPath]
+        if let overrideValue = overrideTokens?[keyPath: keyPath], overrideValue != defaultValue {
```
rdeassis, line 17: Do we have to compare it with the default value?

I'm thinking about this scenario:
- The default value of a given token is the blue color.
- There is a theme override changing that token to a green color.
- A specific instance of the control wants to have that token set back to the original blue value.

Wouldn't this scenario end up incorrectly returning green?

ios/FluentUI/Card Nudge/MSFCardNudge.swift
```diff
@@ -18,6 +18,7 @@ import UIKit
                       title: String) {
         let cardNudge = CardNudge(style: style, title: title)
         state = cardNudge.state
+        tokenResolver = cardNudge.tokenResolver
```
rdeassis, line 21: I think it'd be nice to see this in a composite in UIKit too.
It'll probably help us validate the token resolver in that scenario.
Maybe **DrawerController** is a good one?

ios/FluentUI/Core/Theme/Tokens/TokenizedControlView.swift
mischreiber: Yep! This has been bothering me since the first version of "new" tokens, so I figured now was the time to fix it.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: Simplicity. We have to introduce a new class that does token stuff (`TokenResolver`), so it's simplest from a conceptual point to just let that class also own the truth of the current `overrideTokens`, instead of trying to keep the two in sync.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: I thought about it, but don't ultimately have a strong preference. If you'd like then I can do that (probably a typealias of `NSObject & ObservableObject & Identifiable` all at once).

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
mischreiber: Yep--this is totally a limitation that @huwilkes, @edjamesmsft, and I discovered. We weren't able to come up with a solution, so we decided to push forward with this approach anyway (it was the best we had), but I would LOVE to find out that you have an idea of how to solve this?

Things we considered:
* Separate out the token base class from the "default" tokens, so that the base can return a "null" value to be compared against
  * Problem: we double the number of class instances we have to ship for all controls
* Switch tokens from overriding a base class to implementing an `@objc` protocol with optional methods
  * Problem: we still need to have both a protocol and an implementing base class, which may introduce developer error in keeping them in sync (as well as still some binary bloat)

@huwilkes, @edjamesmsft, any others?

ios/FluentUI/Card Nudge/MSFCardNudge.swift
mischreiber: Sure! I was originally going to save that for a different PR, but I agree that having one example of each will be helpful. Let me finalize that now.

ios/FluentUI/Core/Theme/Tokens/TokenizedControlView.swift
```diff
@@ -28,3 +28,6 @@ extension TokenizedControlView {
         return self
     }
 }
+
+/// Common base type alias for all `state` objects.
+typealias ControlState = NSObject & ObservableObject & Identifiable
```
huwilkes, line 33: Didn't we want to change the name from state?

mischreiber: I think that ship may have sailed at this point, but even if we eventually do I think I'd rather we focus on consistency first and foremost here.

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
rdeassis: I think we could potentially explore a hybrid of the 2 options that you mentioned:

What if instead of Base classes, we provide protocols with nullable properties and provide default implementations for the protocols that return nil?

That way we wouldn't need to expose the "default value class" publicly. Instead we would just expose the protocol (which could be even cheaper in terms of binary size).

That way the TokenResolver will be solely responsible for the order of precedence of the token values (instead of relying on inheritance).

mischreiber: I like the overall approach you're suggesting, but unfortunately there are a few problems with presenting only a pure protocol to clients. In particular, that leaves them with no access to the current context's global and alias tokens, which are currently made available via the base `ControlTokens` class. Though I guess if we required `overrideTokens` to both conform to the given control's tokens, _and_ inherit from the base `ControlTokens` class, then I guess we could make it work...!

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -40,35 +40,29 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
...
     @objc var dismissButtonAction: CardNudgeButtonAction? { get set }
-
-    /// Custom design token set for this control, to use in place of the control's default Fluent tokens.
-    @objc var overrideTokens: CardNudgeTokens? { get set }
 }
 
 /// View that represents the CardNudge.
-public struct CardNudge: View, ConfigurableTokenizedControl {
+public struct CardNudge: View, TokenizedControlView {
+    public typealias TokenType = CardNudgeTokens
```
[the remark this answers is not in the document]
mischreiber: If this ends up merging then yes, it becomes necessary. The new, simpler definition for the `TokenizedControl` protocol doesn't include enough information for Swift to accurately infer what the `TokenType` is meant to be, so we'll need to explicitly inform it.

ios/FluentUI/Core/Theme/Tokens/TokenResolver.swift
mischreiber: Update: After verifying that this approach is technically possible, we still run into binary size issues--my initial conversion for `CardNudge` added another ~40kB, which is not feasible. So that approach is out.

mischreiber:
Closing this PR ahead of new version with similar intent.

PR #1046, opened 2022-06-29
CommandBar implementation improvements

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Make some additional improvements to the CommandBar implementation
--> The usages of `CommandBar` in `CommandBarDemoController` were not setting `translatesAutoResizingMaskIntoConstraints` to `false`
--> Move implementation from `configureHierarchy()` in `CommandBarCommandGroupsView` to be inline to reduce future confusion around use. The method should only be called once on init.
--> Update the `CommandBar` `*ItemGroups` properties to get and set the `itemGroups` property on the corresponding `CommandBarCommandGroupsView`. This will make it so that there is only 1 source of truth for each of the views.
--> Call `setNeedsLayout` after updating the value of `leadingItemGroups` and `trailingItemGroups`. In some cases, I found that the size of the `containerMaskLayer` was not updating as the leading and trailing views hide. Calling `setNeedsLayout` on the `CommandBar` marks the view as needing a layout and causes the `containerMaskLayer` frame to be updated.

### Verification

- Verified in devmain necessary devmain apps.
- Verified in the CommandBarDemoController.
- Some screenshots can be found in the original PR [here](https://github.com/microsoft/fluentui-apple/pull/987).

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -33,15 +33,21 @@ open class CommandBar: UIView {
     @objc public convenience init(itemGroups: [CommandBarItemGroup],
                                   leadingItem: CommandBarItem? = nil,
                                   trailingItem: CommandBarItem? = nil) {
-        var leadingItems: [CommandBarItemGroup]?
-        var trailingItems: [CommandBarItemGroup]?
+        let leadingItems: [CommandBarItemGroup]? = {
+            guard let leadingItem = leadingItem else {
+                return nil
+            }
 
-        if let leadingItem = leadingItem {
-            leadingItems = [[leadingItem]]
-        }
-        if let trailingItem = trailingItem {
-            trailingItems = [[trailingItem]]
-        }
+            return [[leadingItem]]
+        }()
+
+        let trailingItems: [CommandBarItemGroup]? = {
+            guard let trailingItem = trailingItem else {
+                return nil
+            }
+
+            return [[trailingItem]]
+        }()
 
         self.init(itemGroups: itemGroups,
                   leadingItemGroups: leadingItems,
@@ -113,7 +119,7 @@ open class CommandBar: UIView {
             leadingCommandGroupsView.itemGroups
         }
         set {
-            set(leadingCommandGroupsView, with: newValue)
+            setupGroupsView(leadingCommandGroupsView, with: newValue)
         }
     }
 
@@ -123,7 +129,7 @@ open class CommandBar: UIView {
             trailingCommandGroupsView.itemGroups
         }
         set {
-            set(trailingCommandGroupsView, with: newValue)
+            setupGroupsView(trailingCommandGroupsView, with: newValue)
         }
     }
 
@@ -248,12 +254,8 @@ open class CommandBar: UIView {
     }
 
     /// Updates the provided `CommandBarCommandGroupsView` with the `items` array and marks the view as needing a layout
-    private func set(_ commandGroupsView: CommandBarCommandGroupsView, with items: [CommandBarItemGroup]?) {
-        if let items = items {
-            commandGroupsView.itemGroups = items
-        } else {
-            commandGroupsView.itemGroups = []
-        }
+    private func setupGroupsView(_ commandGroupsView: CommandBarCommandGroupsView, with items: [CommandBarItemGroup]?) {
+        commandGroupsView.itemGroups = items ?? []
 
         commandGroupsView.isHidden = commandGroupsView.itemGroups.isEmpty
         scrollView.contentInset = scrollViewContentInset()
@@ -263,7 +265,10 @@ open class CommandBar: UIView {
     private struct LayoutConstants {
         static let fadeViewWidth: CGFloat = 16.0
         static let fixedButtonSpacing: CGFloat = 2.0
-        static let insets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+        static let insets = UIEdgeInsets(top: 8.0,
+                                         left: 8.0,
+                                         bottom: 8.0,
+                                         right: 8.0)
     }
 }
 
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -46,15 +46,11 @@ open class CommandBar: UIView {
     }
 
     @objc public init(itemGroups: [CommandBarItemGroup], leadingItemGroups: [CommandBarItemGroup]? = nil, trailingItemGroups: [CommandBarItemGroup]? = nil) {
-        self.itemGroups = itemGroups
-        self.leadingItemGroups = leadingItemGroups
-        self.trailingItemGroups = trailingItemGroups
-
-        leadingCommandGroupsView = CommandBarCommandGroupsView(itemGroups: self.leadingItemGroups, buttonsPersistSelection: false)
+        leadingCommandGroupsView = CommandBarCommandGroupsView(itemGroups: leadingItemGroups, buttonsPersistSelection: false)
         leadingCommandGroupsView.translatesAutoresizingMaskIntoConstraints = false
-        mainCommandGroupsView = CommandBarCommandGroupsView(itemGroups: self.itemGroups)
+        mainCommandGroupsView = CommandBarCommandGroupsView(itemGroups: itemGroups)
         mainCommandGroupsView.translatesAutoresizingMaskIntoConstraints = false
-        trailingCommandGroupsView = CommandBarCommandGroupsView(itemGroups: self.trailingItemGroups, buttonsPersistSelection: false)
+        trailingCommandGroupsView = CommandBarCommandGroupsView(itemGroups: trailingItemGroups, buttonsPersistSelection: false)
         trailingCommandGroupsView.translatesAutoresizingMaskIntoConstraints = false
 
         commandBarContainerStackView = UIStackView()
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -7,41 +7,59 @@ import UIKit
...
         if let trailingItem = trailingItem {
-            self.trailingButton = button(forItem: trailingItem, isPersistSelection: false)
+            trailingItems = [[trailingItem]]
         }
 
-        translatesAutoresizingMaskIntoConstraints = false
+        self.init(itemGroups: itemGroups, leadingItemGroups: leadingItems, trailingItemGroups: trailingItems)
+    }
+
+    @objc public init(itemGroups: [CommandBarItemGroup], leadingItemGroups: [CommandBarItemGroup]? = nil, trailingItemGroups: [CommandBarItemGroup]? = nil) {
```
rdeassis, line 48: nit: can you break the parameters into multiple lines for better readability? (here and all the other func calls or definitions of the PR)

rdeassis (commented):
I started a thread offline about validation in our main client apps (devmain).
Let's make sure that the scenarios that were broken by the previous merge are validated before we merge this one.

Thank you for the quick turnaround on addressing these issues!

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Updated where it seemed necessary.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -65,20 +84,71 @@ open class CommandBar: UIView {
...
+        get {
+            mainCommandGroupsView.itemGroups
+        }
+        set {
+            mainCommandGroupsView.itemGroups = newValue
+        }
+    }
+
+    /// Items pinned to the leading end of the CommandBar
+    public var leadingItemGroups: [CommandBarItemGroup]? {
```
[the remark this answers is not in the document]
nodes11: Not necessarily. However, I feel like having a tiny bit of additional handling logic here and being able to do `commandBar. leadingItemGroups = nil` to clear the items is better than the alternative of making consumers of the `CommandBar` call `commandBar.leadingItemGroups.removeAll()`, `commandBar.leadingItemGroups = []` or exposing a `CommandBar.removeLeadingItemGroups()` instance method. I guess having to set an array to an empty array for a public API seems clunky to me. I'm open to a further discussion on this. I may be missing something.

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -0,0 +1,109 @@
...
+
+        return button
+    }
+
+    @objc private func handleCommandButtonTapped(_ sender: CommandBarButton) {
+        sender.item.handleTapped(sender)
+        sender.updateState()
+    }
+
+    private static let buttonGroupSpacing: CGFloat = 16
```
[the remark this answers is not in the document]
nodes11: Great to know. I went ahead and updated this for each of the CommandBar* classes.

ios/FluentUI/Command Bar/CommandBar.swift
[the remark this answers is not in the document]
nodes11: Actually, after playing around with this a bit more it's not so bad. I will update these to be non-optional.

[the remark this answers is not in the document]
anandrajeswaran: Fwiw I am for using nil instead of an empty object wherever possible. There might not be a distinction now, but in these cases there often arises a distinction where empty means something different than nil from the consumer's perspective. And I think storing no object technically should use less memory than storing an empty object.

If the if let is causing code to seem overly nested, I would rather focus on other factoring we can do to reduce some of the other layers of nesting. If there is other rationale I'm not thinking of, would love to hear that (for my education).

anandrajeswaran:
Sorry if this is obvious, but would it be better to separate into multiple PRs? One which presumably will be undoing the revert (it sounds like additional changes on top are not needed and are instead needed in the consumer?) and then one for the other improvements which sound like they are unrelated?

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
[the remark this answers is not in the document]
nodes11: Updated structs to be split into different groups as suggested by the style guide.

nodes11:
The revert of the revert commit was put into a separate PR and merged. This PR is now on rebased on top of that commit (the tip on main).

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -100,15 +100,20 @@ class CommandBarButton: UIButton {
...
             }
         }
     }
 
-    private static let contentEdgeInsets = UIEdgeInsets(top: 8.0, left: 10.0, bottom: 8.0, right: 10.0)
-    private static let normalTintColor: UIColor = Colors.textPrimary
-    private static let normalBackgroundColor = UIColor(light: Colors.gray50, dark: Colors.gray600)
-    private static let highlightedBackgroundColor = UIColor(light: Colors.gray100, dark: Colors.gray900)
+    private struct LayoutConstants {
+        static let contentEdgeInsets = UIEdgeInsets(top: 8.0, left: 10.0, bottom: 8.0, right: 10.0)
```
rdeassis, line 111: nit: can you make this indentation consistent with the other constants in the CommandBarCommandGroupsView.swift file?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -30,31 +30,34 @@ open class CommandBar: UIView {
...
         if let leadingItem = leadingItem {
             leadingItems = [[leadingItem]]
         }
-
         if let trailingItem = trailingItem {
             trailingItems = [[trailingItem]]
         }
 
-        self.init(itemGroups: itemGroups, leadingItemGroups: leadingItems, trailingItemGroups: trailingItems)
+        self.init(itemGroups: itemGroups,
```
rdeassis, line 46: Small suggestion for making this code a bit more Swifty:

You can declare constants (let) and initialize them with closures. Something like below:

```
    @available(*, renamed: "init(itemGroups:leadingItemGroups:trailingItemGroups:)")
    @objc public convenience init(itemGroups: [CommandBarItemGroup],
                                  leadingItem: CommandBarItem? = nil,
                                  trailingItem: CommandBarItem? = nil) {
        let leadingItems: [CommandBarItemGroup]? = {
            guard let leadingItem = leadingItem else {
                return nil
            }

            return [[leadingItem]]
        }()

        let trailingItems: [CommandBarItemGroup]? = {
            guard let trailingItem = trailingItem else {
                return nil
            }

            return [[trailingItem]]
        }()

        self.init(itemGroups: itemGroups,
                  leadingItemGroups: leadingItems,
                  trailingItemGroups: trailingItems)
    }
```

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -237,21 +235,36 @@ open class CommandBar: UIView {
...
         }
 
         containerMaskLayer.locations = locations.map { NSNumber(value: Float($0)) }
     }
 
-    private static let fadeViewWidth: CGFloat = 16.0
-    private static let fixedButtonSpacing: CGFloat = 2.0
-    private static let insets = UIEdgeInsets(top: 8.0, left: 8.0, bottom: 8.0, right: 8.0)
+    /// Updates the provided `CommandBarCommandGroupsView` with the `items` array and marks the view as needing a layout
+    private func set(_ commandGroupsView: CommandBarCommandGroupsView, with items: [CommandBarItemGroup]?) {
```
rdeassis, line 251: nit: The name "set" seems a bit too generic.
How about something like: 

`private func setupGroupsView(_ commandGroupsView: CommandBarCommandGroupsView, with items: [CommandBarItemGroup]?)`

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -100,15 +100,20 @@ class CommandBarButton: UIButton {
...
+    }
+
+    private struct ColorConstants {
+        static let normalTintColor: UIColor = Colors.textPrimary
+        static let normalBackgroundColor = UIColor(light: Colors.gray50, dark: Colors.gray600)
```
rdeassis, line 116: nit: break the dark: param into its own line.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -237,21 +235,36 @@ open class CommandBar: UIView {
...
+        if let items = items {
```
rdeassis, line 252: you can  do this in a single line:

`commandGroupsView.itemGroups = items ?? []`

ios/FluentUI/Command Bar/CommandBar.swift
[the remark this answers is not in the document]
rdeassis: I guess my main point was to keep the 2 APIs consistent (either both are optionals or both are non-optionals).

It seems like the **CommandBarCommandGroupsView** class could have a nullable **itemGroups** property.

rdeassis (approved):

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: This is great!

PR #1051, opened 2022-07-01
Tooltip Colors Update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The tooltip's colors were updated to match Fluent 2 colors.

### Verification

The color changes were tested on the Demo app.

| before_dark | after_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1051)

### DIFF
--- a/ios/FluentUI/Tooltip/TooltipView.swift
+++ b/ios/FluentUI/Tooltip/TooltipView.swift
@@ -88,6 +88,11 @@ class TooltipView: UIView {
 
         super.init(frame: .zero)
 
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+
         isAccessibilityElement = true
 
         addSubview(backgroundView)
@@ -144,6 +149,11 @@ class TooltipView: UIView {
         updateWindowSpecificColors()
     }
 
+    @objc private func themeDidChange(_ notification: Notification) {
+        updateMessageLabelColor()
+        updateWindowSpecificColors()
+    }
+
     private func transformForArrowImageView() -> CGAffineTransform {
         switch positionController.arrowDirection {
         case .up:
@@ -163,6 +173,10 @@ class TooltipView: UIView {
             backgroundView.backgroundColor = backgroundColor
             arrowImageView.image = arrowImageViewBaseImage?.withTintColor(backgroundColor, renderingMode: .alwaysOriginal)
         }
+    }
+
+    private func updateMessageLabelColor() {
+        messageLabel.textColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foregroundInverted1])
     }
 
     // MARK: - Accessibility
--- a/ios/FluentUI/Tooltip/TooltipView.swift
+++ b/ios/FluentUI/Tooltip/TooltipView.swift
@@ -5,14 +5,6 @@
 
 import UIKit
 
-// MARK: Tooltip Colors
-
-private extension Colors {
-    struct Tooltip {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -74,7 +74,7 @@ class TooltipView: UIView {
 
     private let messageLabel: UILabel = {
         let label = Label(style: Constants.messageLabelTextStyle)
-        label.textColor = Colors.Tooltip.text
+        label.textColor = UIColor(dynamicColor: label.fluentTheme.aliasTokens.colors[.foregroundInverted1])
```
harrieshin, line 77: isn't this kind of duplicating updateMessageLabelColor ?

harrieshin:
wow, not having branding color in darkmode is pretty drastic user facing change. we should flag this and make sure it is communicated well to our partners in our release notes

ios/FluentUI/Tooltip/TooltipView.swift
laminesm: Kind of. I can change `updateMessageLabelColor` and use it here as well.

harrieshin: why is it needed? by the time label is shown in the window technically it should be already called? have you tried debugging line by line to see the view life cycle?

laminesm:
> wow, not having branding color in darkmode is pretty drastic user facing change. we should flag this and make sure it is communicated well to our partners in our release notes

Noted! I'll make sure to keep an eye out for other similarly drastic changes

ios/FluentUI/Tooltip/TooltipView.swift
laminesm: Ah I think I see what you mean. `updateMessageLabelColor` is only getting called when there is a theme change.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -159,12 +161,16 @@ class TooltipView: UIView {
 
     private func updateWindowSpecificColors() {
         if let window = window {
```
harrieshin, line 163: but if we don't need branding color, we don't need to check window anymore? there is no use?

ios/FluentUI/Tooltip/TooltipView.swift
harrieshin: instead of here then... should be it below to line 141 similar to how you handle background color update of the tooltip?

laminesm: Sounds good, I'll do that!

harrieshin: yeah it seems like all the color updating logic can be group to one function that is used by init and themechange notification.

ios/FluentUI/Tooltip/TooltipView.swift
laminesm: Why do we need to check window with branding colors?

harrieshin: great question. our framework is setup so that branding color is associated with UIWindow. Clients can set different branding color per UIWindow. https://github.com/microsoft/fluentui-apple/blob/e0d2f34d07d7ccb04d11af9ccc36a6b0db824d2e/ios/FluentUI/Core/Colors.swift#:~:text=public%20protocol-,ColorProviding,-%7B

sophialee0416 (approved):

PR #1054, opened 2022-07-05
Shimmer Colors Update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The colors for Shimmer were updated to match Fluent 2 tokens.

### Verification

The changes were tested on the Demo app.

| before_dark | after_dark |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Shimmer/ShimmerView.swift
+++ b/ios/FluentUI/Shimmer/ShimmerView.swift
@@ -153,10 +153,16 @@ open class ShimmerView: UIView {
         self.shimmerStyle = shimmerStyle
         self.shimmerAlpha = shimmerStyle.defaultAlphaValue
         super.init(frame: CGRect(origin: .zero, size: containerView?.bounds.size ?? .zero))
-        self.viewTintColor = shimmerStyle.defaultTintColor(view: self)
+
+        updateViewTintColor()
+
         NotificationCenter.default.addObserver(self,
                                                selector: #selector(syncAnimation),
                                                name: UIAccessibility.reduceMotionStatusDidChangeNotification,
+                                               object: nil)
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
                                                object: nil)
     }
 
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -186,6 +191,14 @@ open class ShimmerView: UIView {
         updateShimmeringAnimation()
     }
 
+    @objc private func themeDidChange(_ notification: Notification) {
+        updateViewTintColor()
+    }
+    
+    private func updateViewTintColor(){
```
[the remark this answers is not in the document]
sophialee0416: Running pod lib lint before pushing up a PR should help you catch these errors as well if your linter doesn't catch them for any reason :)

sophialee0416 (approved):

ios/FluentUI/Shimmer/ShimmerView.swift
```diff
@@ -24,12 +24,12 @@ public enum ShimmerStyle: Int, CaseIterable {
         }
     }
 
-    var defaultTintColor: UIColor {
+    func defaultTintColor(view: UIView) -> UIColor {
```
mischreiber, line 27: Why do you need a `view` argument here? Can you just access `fluentTheme` directly below?

mischreiber (changes requested):

ios/FluentUI/Shimmer/ShimmerView.swift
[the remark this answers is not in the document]
mischreiber: Does `pod lib lint` actually run `swiftlint`? Learned a new thing today!

ios/FluentUI/Shimmer/ShimmerView.swift
laminesm: I can only access `fluentTheme` from a UIView and this color is defined in one,.

mischreiber: Oh I see, and this control isn't tokenized yet in this branch. In that case, could you change the API to take the `fluentTheme` instead of the view to minimize the dependency between the two?

laminesm: Makes sense, I'll make the change!

mischreiber (approved):

PR #1055, opened 2022-07-06
[Fluent2 Tokens] Add an optional trailing image to the Notification

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds the option to set the image the notification shows for the action button if no button title is set. Currently, if there's a button action set but no button title set, we'll show the X icon shown in the before image. With this change, the user can change the icon that's shown in the button when there is a button action set but no button title.

### Verification
| Notification_trailingImage_before | Notification_trailingImage_after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1055)

### DIFF
--- a/ios/FluentUI/Notification/FluentNotification.swift
+++ b/ios/FluentUI/Notification/FluentNotification.swift
@@ -58,6 +58,7 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
     ///   - title: Optional text to draw above the message area.
     ///   - attributedTitle: Optional attributed text to draw above the message area. If set, it will override the title parameter.
     ///   - image: Optional icon to draw at the leading edge of the control.
+    ///   - trailingImage: Optional icon to show in the action button if no button title is provided.
     ///   - actionButtonTitle:Title to display in the action button on the trailing edge of the control.
     ///   - actionButtonAction: Action to be dispatched by the action button on the trailing edge of the control.
     ///   - messageButtonAction: Action to be dispatched by tapping on the toast/bar notification.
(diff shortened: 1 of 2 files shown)


sophialee0416:
Confused about the title part in the description did you mean the trailing image? Or did I miss something

huwilkes:
> Confused about the title part in the description did you mean the trailing image? Or did I miss something

I'm not entirely sure what you're confused about, but the title I meant in the description is the title of the button. The trailing image isn't just an image, it's part of the action button. I can update the description to make that a bit clearer. The behavior right now is if there's a button action and a button title, it'll show a button with that title. If there's no button title, it'll by default show a X icon. Now, if there's no button title and there is a trailing image set, it'll show that image instead of the X. The comments in the code also specify that the image is only shown in the button, and only when no button title is set.

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -67,6 +71,7 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
                 title: String? = nil,
                 attributedTitle: NSAttributedString? = nil,
                 image: UIImage? = nil,
+                trailingImage: UIImage? = nil,
```
edjamesmsft, line 74: Mostly for my knowledge, but does it matter where you add new optional params? Should this be added at the end of the list? It would need to be in other languages, like C#, to keep this from being a breaking change.

huwilkes: I think that's a reasonable concern, but I think this should be fine in this case. Since it's in Swift, and not available in Objective C, you can just leave out arguments. If this were available in Objective C, I would've needed to add a convenience init.

sophialee0416 (approved):

edjamesmsft (approved):

PR #1056, opened 2022-07-06
Fixing text color update in TableViewCell for rich text support

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Rich text support does not update the labels in TableViewCell due to the `updateTextColors()` method. This fix takes the rich text into consideration when updating text colors.

Note: This fix does not include the fonts update with rich text support, which will come in a later PR.

### Verification

| Simulator Screen Shot - iPhone 12 Pro - 2022-07-06 at 14 54 24 | Simulator Screen Shot - iPhone 12 Pro - 2022-07-06 at 14 53 41 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1056)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1075,25 +1075,22 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                           customAccessoryView: UIView? = nil,
                           accessoryType: TableViewCellAccessoryType = .none) {
         if let attributedTitle = attributedTitle {
-            titleLabel.attributedText = attributedTitle
-            isUsingAttributedTitle = true
+            self.attributedTitle = attributedTitle
         } else {
+            self.attributedTitle = nil
             titleLabel.text = title
-            isUsingAttributedTitle = false
         }
         if let attributedSubtitle = attributedSubtitle {
-            subtitleLabel.attributedText = attributedSubtitle
-            isUsingAttributedSubtitle = true
+            self.attributedSubtitle = attributedSubtitle
         } else {
+            self.attributedSubtitle = nil
             subtitleLabel.text = subtitle
-            isUsingAttributedSubtitle = false
         }
         if let attributedFooter = attributedFooter {
-            footerLabel.attributedText = attributedFooter
-            isUsingAttributedFooter = true
+            self.attributedFooter = attributedFooter
         } else {
+            self.attributedFooter = nil
             footerLabel.text = footer
-            isUsingAttributedFooter = false
         }
 
         self.customView = customView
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1076,18 +1076,24 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                           accessoryType: TableViewCellAccessoryType = .none) {
         if let attributedTitle = attributedTitle {
             titleLabel.attributedText = attributedTitle
+            isUsingAttributedTitle = true
         } else {
             titleLabel.text = title
+            isUsingAttributedTitle = false
```
sophialee0416, line 1082: For clarification: the additional reaffirmation of assigning the `isUsingAttributed` text variables `false` is necessary because the cells seem to keep the old values and makes it show into the other cells. This change is essentially a double-check safety.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1148,12 +1154,22 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
 
     private var isUsingCustomTextColors: Bool = false
 
+    private var isUsingAttributedTitle: Bool = false
```
rdeassis, line 1157: I think it's better to keep a nullable property of the NSAttributedString for each of the labels and compare against nil before assigning it to the label. This way you don't have to keep this boolean in sync with the state of the labels.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1148,12 +1151,45 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
 
     private var isUsingCustomTextColors: Bool = false
```
huwilkes, line 1152: Probably outside the scope of this PR, but would it be possible to remove this and instead just also use the attributed strings internally if someone tries to set a custom color? Not sure it's entirely possible, but it might be possible to make a new attributed string that has all of the old attributes (which could be none) plus the new custom color.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1075,18 +1075,21 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                           customAccessoryView: UIView? = nil,
                           accessoryType: TableViewCellAccessoryType = .none) {
         if let attributedTitle = attributedTitle {
-            titleLabel.attributedText = attributedTitle
+            self.attributedTitle = attributedTitle
```
rdeassis, line 1078: In the attributed string scenario the text property is not set.
There are other places using this property for layout sizing calculation:

https://github.com/microsoft/fluentui-apple/blob/261f2b0fe5721ff6e5d3ffa26a64de755422dde7/ios/FluentUI/Table%20View/TableViewCell.swift#L770

Have you tested an attributed string with a larger font to see how this could impact the height and width of the cell? If the height is not correctly calculated, we need to have a bug filed for that.

ios/FluentUI/Table View/TableViewCell.swift
rdeassis: I think it's worth the try. But I believe it warrants its own PR. The Label control in the experiment might also be involved because of this:
https://github.com/microsoft/fluentui-apple/blob/1e8d6806062c12aee072caff2b10cf8254e9a51b/ios/FluentUI/Label/Label.swift#L114

@sople could you log a work item that so we can remove the isUsingCustomTextColors var and experiment using NSAttributedStrings even when only the text is set?

rdeassis (approved):

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Screenshot below is with larger text on and set to max
Simulator Screen Shot - iPhone 12 Pro - 2022-07-06 at 16 58 43

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Surprised that someone else out there has my alias 😆 but yes will log the work item and experiment!

rdeassis: Just edited the comment tagging the right one! 
:|

PR #1064, opened 2022-07-08
Fixing slow NSAttributedString update in TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When setting an attributed string to the cell, the cells would not automatically update, even with calling `setNeedsLayout()` being called in the cell setup. What was more interesting was that the fonts would all be updated, but not the color, although debugging showed that all attributes were set correctly. This change makes sure that the attributed strings are fully and accurately updating.

### Verification

Colors are changed to better demonstrate the bug.

| ezgif com-gif-maker | ezgif com-gif-maker-2 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1064)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
@@ -122,18 +122,20 @@ extension TableViewCellDemoController {
         let section = sections[indexPath.section]
         let item = section.item
         if section.title == "Inverted double line cell" {
-            cell.setup(
-                attributedTitle: NSAttributedString(string: item.text1,
-                                                    attributes: [.font: TextStyle.footnote.font,
-                                                                 .foregroundColor: Colors.Table.Cell.footer]),
-                attributedSubtitle: NSAttributedString(string: item.text2,
-                                                       attributes: [.font: TextStyle.body.font,
-                                                                    .foregroundColor: Colors.Table.Cell.title]),
-                footer: TableViewCellSampleData.hasFullLengthLabelAccessoryView(at: indexPath) ? "" : item.text3,
-                customView: TableViewSampleData.createCustomView(imageName: item.image),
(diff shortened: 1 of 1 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellDemoController.swift
```diff
@@ -122,18 +122,20 @@ extension TableViewCellDemoController {
...
-                                                                 .foregroundColor: Colors.Table.Cell.footer]),
-                attributedSubtitle: NSAttributedString(string: item.text2,
-                                                       attributes: [.font: TextStyle.body.font,
-                                                                    .foregroundColor: Colors.Table.Cell.title]),
-                footer: TableViewCellSampleData.hasFullLengthLabelAccessoryView(at: indexPath) ? "" : item.text3,
-                customView: TableViewSampleData.createCustomView(imageName: item.image),
-                customAccessoryView: section.hasAccessory ? TableViewCellSampleData.customAccessoryView : nil,
-                accessoryType: TableViewCellSampleData.accessoryType(for: indexPath)
-            )
+            DispatchQueue.main.async {
```
huwilkes, line 125: Was this a bug in only the demo, and not a bug with the actual cell?

sophialee0416: I don't believe so - the attributes are all loaded correctly, but they are only partially updating. The cells display correctly after scrolling, which is why I interpreted as a demo controller issue

sophialee0416: Updating this thread, it's an issue with the Label component, so I'll be opening up a new PR for that so I can also merge into fluent2-tokens!

PR #1065, opened 2022-07-11
[Fluent 2 iOS] Card Colors Update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The Card colors were updated to match fluent 2 specs.

### Verification

The cards were tested on the demo app.

| before_dark | after_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1065)

### DIFF
--- a/ios/FluentUI/Card/CardView.swift
+++ b/ios/FluentUI/Card/CardView.swift
@@ -329,7 +329,7 @@ open class CardView: UIView {
         iconView = UIImageView(image: icon)
 
         super.init(frame: .zero)
-
+        setupColors()
         NotificationCenter.default.addObserver(self,
                                                selector: #selector(themeDidChange),
                                                name: .didChangeTheme,
@@ -380,11 +380,6 @@ open class CardView: UIView {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
-    open override func didMoveToWindow() {
-        super.didMoveToWindow()
-        setupColors()
-    }
-
     @objc private func themeDidChange(_ notification: Notification) {
         setupColors()
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Card/CardView.swift
```diff
@@ -409,19 +418,19 @@ open class CardView: UIView {
     private func setupColors() {
         switch colorStyle {
         case .appColor:
```
huwilkes, line 420: Looks like we lost the variation that had the brand colors. Can you follow up with design, and if that was intentional, create a work item to track the removal of the unused styles? Or at least to condense the shared logic. I guess there's still a variation that uses brand color for the icon.

huwilkes (approved):

ios/FluentUI/Card/CardView.swift
laminesm: The brand colors were updated on the last commit. Design omitted them on the Figma but they gave me the right tokens this morning.

ios/FluentUI/Card/CardView.swift
```diff
@@ -409,19 +418,19 @@ open class CardView: UIView {
...
-            primaryLabel.textColor = UIColor(light: .black, dark: Colors.gray100)
-            secondaryLabel.textColor = UIColor(light: Colors.gray600, dark: Colors.gray400)
-            iconView.tintColor = UIColor(light: Colors.gray600, dark: Colors.gray500)
+            primaryLabel.textColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foreground1])
+            secondaryLabel.textColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foreground2])
+            iconView.tintColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.brandForeground1])
             if let window = window {
```
huwilkes, line 424: Also, since you're pulling the tokens from the theme, you don't need the window anymore

edjamesmsft:
Nit: could you update the title with  [Fluent 2 iOS] at the beginning? No huge deal though!

edjamesmsft:
Is this actually how we're showing a card is selected, just updating the color of the icon but not the card itself? That seems like a big step back for accessibility. Could you double check with design?

laminesm:
> Is this actually how we're showing a card is selected, just updating the color of the icon but not the card itself? That seems like a big step back for accessibility. Could you double check with design?

I think those are the cards with brand colors, not selected cards.  They weren't initially part of the design but Ian gave me these tokens today with brand colors only for the icon.

PR #1066, opened 2022-07-11
Adding custom ring color for AvatarGroup in demo app

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Custom ring color was added as a toggle for AvatarGroup demo
- Made sure `maxDisplayedAvatars` cannot exceed `avatarCount`
- Made sure newly displayed avatars update to their custom ring color if the toggle is on 

### Verification

The tests were run on the demo app.

| options_before | option_after |
| before_dark | after_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1066)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
@@ -75,6 +75,19 @@ class AvatarGroupDemoController: DemoTableViewController {
 
             return cell
 
+        case .customRingColor:
+            guard let cell = tableView.dequeueReusableCell(withIdentifier: BooleanCell.identifier) as? BooleanCell else {
+                return UITableViewCell()
+            }
+
+            cell.setup(title: row.title, isOn: self.isUsingImageBasedCustomColor)
+            cell.titleNumberOfLines = 0
+            cell.onValueChanged = { [weak self, weak cell] in
+                self?.isUsingImageBasedCustomColor = cell?.isOn ?? true
+            }
+
+            return cell
+
         case .maxDisplayedAvatars,
              .overflow:
             guard let cell = tableView.dequeueReusableCell(withIdentifier: TableViewCell.identifier) as? TableViewCell else {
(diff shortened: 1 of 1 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -383,8 +402,15 @@ class AvatarGroupDemoController: DemoTableViewController {
     }()
 
     @objc private func setMaxAvatarCount() {
-        if let text = maxAvatarsTextField.text, let count = Int(text) {
-            maxDisplayedAvatars = count
+        let oldMax = maxDisplayedAvatars
+
+        if let text = maxAvatarsTextField.text, let newMax = Int(text) {
```
edjamesmsft, line 407: What happens if you cast the text to an int and the behavior is unexpected? Is it possible for that text to be anything but an integer?

huwilkes: Since it's in an if, I think it should just skip that case. So if the text ends up failing to cast to int, then the newMax would be unchanged.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarGroupDemoController.swift
```diff
@@ -496,6 +527,25 @@ class AvatarGroupDemoController: DemoTableViewController {
         }
     }
 
+    private var isUsingImageBasedCustomColor: Bool = false {
+        didSet {
+            updateAvatarsCustomColorRing(index1: 0, index2: avatarCount - 1)
+        }
+    }
+
+    private func updateAvatarsCustomColorRing(index1: Int, index2: Int) {
```
huwilkes, line 536: Can you update this to something like `updateAvatarsCustomColorRing(for range: Range<Int>)` and call it something like `updateAvatarsCustomColorRing(for: index1..<index2)`? I think this would be cleaner if we used ranges and the half open range operator.

huwilkes (approved):

PR #1076, opened 2022-07-18
[Fluent2 Tokens] Add MSFNotification convenience init with just MSFNotificationStyle

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added a new convenience init so that all the Objective C clients wouldn't have to specify if they wanted a flexible width toast or not on every notification.

### Verification

Objective C can access an init with just the MSFNotificationStyle.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1076)

### DIFF
--- a/ios/FluentUI/Notification/FluentNotification.swift
+++ b/ios/FluentUI/Notification/FluentNotification.swift
@@ -108,7 +108,6 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
                                height: imageSize.height,
                                alignment: .center)
                         .foregroundColor(Color(dynamicColor: tokens.imageColor))
-                        .accessibilityHidden(true)
                 }
             }
         }
--- a/ios/FluentUI/Notification/FluentNotification.swift
+++ b/ios/FluentUI/Notification/FluentNotification.swift
@@ -108,6 +108,7 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
                                height: imageSize.height,
                                alignment: .center)
                         .foregroundColor(Color(dynamicColor: tokens.imageColor))
+                        .accessibilityHidden(true)
                 }
             }
         }
--- a/ios/FluentUI/Notification/MSFNotificationView.swift
+++ b/ios/FluentUI/Notification/MSFNotificationView.swift
@@ -12,6 +12,13 @@ import UIKit
     /// Creates a new MSFNotification instance.
     /// - Parameters:
     ///   - style: The MSFNotification value used by the Notification.
+    @objc public convenience init(style: MSFNotificationStyle) {
+        self.init(style: style, isFlexibleWidthToast: false)
+    }
+
+    /// Creates a new MSFNotification instance.
+    /// - Parameters:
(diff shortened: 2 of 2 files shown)


ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -108,6 +108,7 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
                                height: imageSize.height,
                                alignment: .center)
                         .foregroundColor(Color(dynamicColor: tokens.imageColor))
+                        .accessibilityHidden(true)
```
harrieshin, line 111: we can't use https://developer.apple.com/documentation/swiftui/form/accessibilityelement(children:) .contain or combine to make the notificationview more consistent container behavior for voice over users? it should act somewhat like a dialog right?

harrieshin:
"No visual changes, but now VoiceOver will only read the title, message, and action of the Notification, and Objective C can access an init with just the MSFNotificationStyle." would you mind sharing a recording of the voiceover focus flow?

ios/FluentUI/Notification/FluentNotification.swift
huwilkes: That'll put the scope of the work out of this PR, so I'll revert this for now and come back with a more VoiceOver focused PR.

huwilkes:
> "No visual changes, but now VoiceOver will only read the title, message, and action of the Notification, and Objective C can access an init with just the MSFNotificationStyle." would you mind sharing a recording of the voiceover focus flow?



Before:  

https://user-images.githubusercontent.com/67026548/179627095-5775e57b-0ec5-4764-af89-ba1e07ac6db6.mov 

After:

https://user-images.githubusercontent.com/67026548/179626981-edb63aed-727c-420b-adc7-b4831cf97b1f.mov

harrieshin (approved):

harrieshin:
why do you have api break label on this?

huwilkes:
> why do you have api break label on this?

@mischreiber was saying that any API change should count, even additions. Unless I misunderstood what they were saying.

mischreiber:
> > why do you have api break label on this?
> 
> @mischreiber was saying that any API change should count, even additions. Unless I misunderstood what they were saying.

@huwilkes I agree that this’ll need a version bump, but since it’s technically different than an API change, how do you feel about a separate tag? Maybe an “API Added” tag?

huwilkes:
> > > why do you have api break label on this?
> > 
> > 
> > @mischreiber was saying that any API change should count, even additions. Unless I misunderstood what they were saying.
> 
> @huwilkes I agree that this’ll need a version bump, but since it’s technically different than an API change, how do you feel about a separate tag? Maybe an “API Added” tag?

I don't mind removing the tag from this one, and adding that one is probably a good idea. We just need to be better about tracking when we're actually changing things. Since the change that added the new init without a convenience init would actually be a breaking change.

PR #1082, opened 2022-07-19
Updating + improving TableViewCell fonts

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change brings fixes to the following:

- Using token-defined fonts for labels. The line height was off, so we had used the old Textstyles, but with this change, that has been removed and we set our fonts using the tokens and also set the label height so that it matches design requirements (we cannot edit lineHeight without using NSMutableParagraphStyle, which we want to avoid because it can cause crashes and is generally not ideal)
- Fixing font regression: Turns out that setting UILabel properties will also override the attributedText property, so checking for nil in `updateTextColor()` and `updateFonts()` was not enough. Adding a boolean to check if we are manually setting the attributedText helped fix the bug.

### Verification

| Simulator Screen Shot - iPhone 12 Pro - 2022-07-19 at 15 06 19 | Simulator Screen Shot - iPhone 12 Pro - 2022-07-19 at 15 09 11 |
| Simulator Screen Shot - iPhone 12 Pro - 2022-07-19 at 15 07 17 | Simulator Screen Shot - iPhone 12 Pro - 2022-07-19 at 15 09 52 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1082)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -187,6 +187,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     ///   - title: The title string
     ///   - subtitle: The subtitle string
     ///   - footer: The footer string
+    ///   - titleFont: The title font; If not set, it will default to the font definition in tokens
+    ///   - subtitleFont: The subtitle font; If not set, it will default to the font definition in tokens
+    ///   - footerFont: The footer font; If not set, it will default to the font definition in tokens   
     ///   - titleLeadingAccessoryView: The accessory view on the leading edge of the title
     ///   - titleTrailingAccessoryView: The accessory view on the trailing edge of the title
     ///   - subtitleLeadingAccessoryView: The accessory view on the leading edge of the subtitle
@@ -309,6 +312,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     ///   - title: The title string
     ///   - subtitle: The subtitle string
     ///   - footer: The footer string
+    ///   - titleFont: The title font; If not set, it will default to the font definition in tokens
+    ///   - subtitleFont: The subtitle font; If not set, it will default to the font definition in tokens
+    ///   - footerFont: The footer font; If not set, it will default to the font definition in tokens
     ///   - titleLeadingAccessoryView: The accessory view on the leading edge of the title
     ///   - titleTrailingAccessoryView: The accessory view on the trailing edge of the title
     ///   - subtitleLeadingAccessoryView: The accessory view on the leading edge of the subtitle
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -126,12 +126,12 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         var customViewSize: MSFTableViewCellCustomViewSize { return self == .oneLine ? .small : .medium }
     }
 
-    struct TextStyles {
-        static let title: TextStyle = .body
-        static let subtitleTwoLines: TextStyle = .footnote
-        static let subtitleThreeLines: TextStyle = .subhead
-        static let footer: TextStyle = .footnote
-    }
+//    struct TextStyles {
+//        static let title: TextStyle = .body
+//        static let subtitleTwoLines: TextStyle = .footnote
+//        static let subtitleThreeLines: TextStyle = .subhead
+//        static let footer: TextStyle = .footnote
+//    }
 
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -146,9 +139,15 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
-    @objc public static var smallHeight: CGFloat { return height(title: "", customViewSize: .small) }
-    @objc public static var mediumHeight: CGFloat { return height(title: "", subtitle: " ") }
-    @objc public static var largeHeight: CGFloat { return height(title: "", subtitle: " ", footer: " ") }
+    @objc public static var smallHeight: CGFloat {
+        return 48
+    }
+    @objc public static var mediumHeight: CGFloat {
+        return 64
+    }
+    @objc public static var largeHeight: CGFloat {
```
laminesm, line 148: Why do we need to define these values here if they are also defined in `TableViewCellTokens`?

sophialee0416: Thanks for bringing this up, I meant to open a thread about this! These are static variables, which means they can be accessed without actually instantiating the TableViewCell. I'm guessing that this is being used by downstream dependencies, and since they're variables, I can't change them to methods. They could also use the tokens to obtain these values from here on out. For now I put in the actual CGFloat value, but maybe I can add deprecation messages/comments like I did for the custom text color methods? @harrieshin @rdeassis

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -211,9 +210,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     ///   - isInSelectionMode: Boolean describing if the cell is in multi-selection mode which shows/hides a checkmark image on the leading edge
     /// - Returns: a value representing the calculated height of the cell
     @objc public class func height(tokens: TableViewCellTokens = .init(),
-                                   title: String,
-                                   subtitle: String = "",
-                                   footer: String = "",
+                                   title: Label,
```
mischreiber, line 213: I see that we're expecting a `Label` view to be passed in here, but I'm not certain that's the best approach. This function (AFAIK) is meant to be a shortcut for helping clients estimate the height of a cell for UITableView layout, so it needs to be a fast lookup, and creating views kinda erases that performance.

Ideally I'd like this API to take in exactly what is needed for measurement, and nothing else. Is there any way we can extract the info we need from the Label to pass in?

sophialee0416: Yes! I ended up removing label, and adding in fonts. Ideally we'd want to get them from the tokens, but labels can now have separate font values from the tokens with the rich text support.

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Resolved offline: removed labels which allowed me to revert to original code

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -321,6 +325,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                                            title: String,
                                            subtitle: String = "",
                                            footer: String = "",
+                                           titleFont: UIFont? = nil,
```
harrieshin, line 328: need to update documentation.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1082,18 +1095,22 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
             self.attributedSubtitle = nil
             subtitleLabel.text = subtitle
         }
+
         if let attributedFooter = attributedFooter {
             self.attributedFooter = attributedFooter
+            isAttributedFooterSet = true
         } else {
             self.attributedFooter = nil
             footerLabel.text = footer
```
harrieshin, line 1116: does that mean in the else case we should set isAttributedFooterSet false? same question for title and subtitle

sophialee0416: The default is always set at false, and setup is called once in the beginning. Should I add it as insurance?

harrieshin: yes! because nothing stops clients to call setup on the cell multiple times

harrieshin (approved):

PR #1087, opened 2022-07-21
[Fluent2 Tokens] Fix usages of arguments in showNotification and hide in MSFNotification

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When we updated the hide method, we simplified the logic for when we do `completionForHide()`, but we unfortunately simplified it too much and removed the case where we need to actually dismiss the notification without animations. We also stopped passing the anchorView and animated down to the call to show the new toast after dismissing the current toast.

### Verification

For testing purposes, I set the dismiss button action to not be animated, and removed the auto dismiss we have in the demo controller.

| Notification_Dismiss_Before | Notification_Dismiss_After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1087)

### DIFF
--- a/ios/FluentUI/Notification/MSFNotificationView.swift
+++ b/ios/FluentUI/Notification/MSFNotificationView.swift
@@ -59,7 +59,7 @@ import UIKit
         let presentationOffset: CGFloat! = notification.tokens.presentationOffset
         if style.isToast, let currentToast = MSFNotification.currentToast, currentToast.window != nil {
             currentToast.hide {
-                self.showNotification(in: view, completion: completion)
+                self.showNotification(in: view, from: anchorView, animated: animated, completion: completion)
             }
             return
         }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Notification/MSFNotificationView.swift
```diff
@@ -59,7 +59,7 @@ import UIKit
         let presentationOffset: CGFloat! = notification.tokens.presentationOffset
         if style.isToast, let currentToast = MSFNotification.currentToast, currentToast.window != nil {
             currentToast.hide {
-                self.showNotification(in: view, completion: completion)
```
harrieshin, line 62: is there still use of self.shownotification(in: completion)?

harrieshin (approved):

sophialee0416 (approved):

ios/FluentUI/Notification/MSFNotificationView.swift
huwilkes: Only one left is in the demo controller.

PR #1089, opened 2022-07-21
[Fluent2 Tokens] Introducing `ControlTokenSet` for overrideTokens

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In response to issue #985, this change brings a new internal system for fetching tokens: the `ControlTokenSet`. This PR includes core-level changes, as well as conversion for `CardNudge` and `Drawer`.

#### `ControlTokenSet`

This new class replaces `ControlTokens` and is responsible for the logic that was previously handled by its `resolvedTokens` property. That old property would resolve between `overrideTokens`, `themeTokens`, and `defaultTokens` on a per-control basis, but we actually need per-*token* granular resolution. This new class provides that!

The per-value lookup is implemented via a custom subscript operator (`[]`) that takes one of the enumerated token keys for the control (of type `TokenSetKey`) and returns the first *non-nil* value in the `overrideTokens -> themeTokens -> defaultTokens` chain. This allows for both per-control and theme-wide token overrides to modify different values. And unlike the defunct `TokenResolver` approach, this checks for *existence* of override, rather than *value*, which is more robust.

**Before:**

```swift
var tokens: CardNudgeTokens { return resolvedTokens } // Special getter that does resolution

var body: some View {
    Rectangle()
        .foregroundColor(Color(dynamicColor: tokens.foregroundColor))
}
```

**After:**

```swift
@ObservedObject var tokenSet: CardNudgeTokenSet = .init()

var body: some View {
    Rectangle()
        .foregroundColor(Color(dynamicColor: tokenSet[.foregroundColor].dynamicColor))
}
```

#### `ControlTokenValue`

The new `ControlTokenValue` enum acts as a union type of all possible token value types used by `ControlTokenSet`, allowing them to be stored alongside one another in a typed lookup table. This is safer than erasing types to `Any`, which would be required without this wrapper.

#### Customization

Previously, all customization was done by providing a subclass of a control's `ControlTokens` class, overriding any relevant properties.

The new approach is much simpler: properties can now be *directly* overridden on a control via its `tokens` property. See below for an example.

**Before:**

```swift
private class OverrideCardNudgeTokens: CardNudgeTokens {
    override var backgroundColor: DynamicColor {
        return DynamicColor(light: globalTokens.sharedColors[.seafoam][.tint50],
                            dark: globalTokens.sharedColors[.seafoam][.shade40])
    }
}

cardNudge.state.overrideTokens = OverrideCardNudgeTokens()

```

**After:**

```swift
cardNudge.tokenSet[.backgroundColor] = .dynamicColor({
    DynamicColor(light: globalTokens.sharedColors[.seafoam][.tint50],
                 dark: globalTokens.sharedColors[.seafoam][.shade40])
})
```

### Verification

In these examples, the `CardNudgeDemoController` is providing overrides at both the control-level and theme-level -- both overrides have a custom background color, but only theme-wide tokens provide custom border width and color values. In the old system, this custom border information is lost, but with this change we now successfully present both sets of tokens.

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1089)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -561,24 +563,13 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         return dismissPresentingViewController(animated: true)
     }
 
-    func updateCurrentTokens(_ tokens: DrawerTokens) {
-        self.tokens = tokens
-    }
+    public typealias TokenSetType = DrawerTokenSet
+    public typealias TokenSetKeyType = DrawerTokenSet.Tokens
+    public var tokenSet: DrawerTokenSet = .init()
```
huwilkes, line 568: I don't quite have the time to dig into this to figure it out myself right now, but does this still have the inheritance issue I was hitting when trying to update the PopupMenu?

mischreiber: It shouldn't, because token lookups will now be indexed by their unique `TokenSetKey` rather than by control type (which can be poisoned by inheritance).

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -561,24 +563,13 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
...
 
-    var defaultTokens: DrawerTokens = .init()
-    var tokens: DrawerTokens = .init()
-    var overrideTokens: DrawerTokens? {
-        didSet {
-            updateDrawerTokens()
-        }
-    }
+    var overrideTokensSink: AnyCancellable?
```
harrieshin, line 570: what does this mean?

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -0,0 +1,147 @@
...
+                              colorTwo: defaultColor,
+                              blurTwo: 10.0,
+                              xTwo: 10.0,
+                              yTwo: 10.0)
+        }
+    }
+}
+
+#if DEBUG
+extension ControlTokenValue: CustomStringConvertible {
```
harrieshin, line 132: maybe a todo is a unit test to figure out if the resolver actually works correctly?

harrieshin (approved):

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -40,35 +40,32 @@ public typealias CardNudgeButtonAction = ((_ state: MSFCardNudgeState) -> Void)
...
-    }
 
     @ViewBuilder
     var icon: some View {
         if let icon = state.mainIcon {
             ZStack {
-                RoundedRectangle(cornerRadius: tokens.circleRadius)
-                    .frame(width: tokens.circleSize, height: tokens.circleSize)
-                    .foregroundColor(Color(dynamicColor: tokens.buttonBackgroundColor))
+                RoundedRectangle(cornerRadius: tokenSet[.circleRadius].float)
```
edjamesmsft, line 60: Mostly for my information, but what is it about this implementation that is making you have to explicitly call out the type of the token like this?

ios/FluentUI/Drawer/DrawerController.swift
mischreiber: Ah, an old name from old PR. I'll update this.

Idea is that it's the Combine-y way of responding to changes to changes in `tokens`, allowing us to update the appearance whenever a token change is observed.

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: Sure! So the short version is that `tokenSet[xyz]` doesn't return a token directly; it returns an enum of type `ControlTokenValue`. Sadly, Swift doesn't support any sort of automatic unboxing, so we need to explicitly extract the raw value of that enum type. To that end I created vars on `ControlTokenValue` for each of the types it potentially represents (`float`, `DynamicColor`, `FontInfo`, and `ShadowInfo`).

edjamesmsft (approved):

PR #1078, opened 2022-07-19
[Fluent2 Tokens] Fix Notification VoiceOver grouping

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Image and Text
  - Put the image and text container in their own stack so they could be read as a single accessible element
  - Added the button trait if there is a message button action, to indicate that selection will do something
  - Added pointer interaction if there is a message button action
- Button 
  - Fixed the default X icon dismiss button reading out "dismiss-20x20" instead of just "Dismiss"
  - Added pointer interaction
  - Added way to set accessibility label for custom trailing image

### Verification

Before:

https://user-images.githubusercontent.com/67026548/179638030-47bf951f-0e36-4585-aa5a-76c970fefe8a.mov

After:

https://user-images.githubusercontent.com/67026548/179870616-e12d9625-a99a-4f96-b6b4-32bccce5d4f7.mov

| Before | After |
|---|---|
| Notification_Pointer_Before | Notification_Pointer_After |

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [x] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [x] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1078)

PR #1029, opened 2022-06-23
Enable Fluent Notification Customization

jeaniehuynh:
Enable flexible width toasts and customizable image color/element spacing

### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change separates the uses of some tokens so that they can be customized without impacting other elements. Image color and element spacing can now be customized. This change also implements isFlexibleWidthToast which allows the notification to be sized either based on the size of the screen (default) or the size of its content.

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 28 14 |  Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 27 00 |
| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 28 24 |  Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 26 05 |
| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 28 24 | Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-06-23 at 02 26 09 |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [x] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [x] Different resolutions (1x, 2x, 3x)
- [x] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1029)

PR #980, opened 2022-05-05
Remove legacy notification view

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change transitions the UIKit demo controller for notification view to use the SwiftUI implementation of the notification instead. Due to sizing issues with GeometryReader in UIKit, @lcapkovic came up with the solution to have the sizing done externally for UIKit within the bridging layer (MSFNotificationView).

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-05-04 at 20 03 22 | Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-05-04 at 20 01 17 |
| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-05-04 at 20 02 59 | Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-05-04 at 20 01 36 |
### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [x] Different resolutions (1x, 2x, 3x)
- [x] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/980)

PR #1090, opened 2022-07-22
Merge tokenized SwiftUI Notification to main

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Merge all the changes for the tokenized Notification to main. These changes include:
- #1087 
- #1083 
- #1086 
- #1078 
- #1076 
- #1070 
- #1055 
- #1042 
- #1029 
- #1023
- #984 
- #980 
- #965 
- #951 
- #896 
- #823 
- #746 
- #715 

### Verification

Built locally, matches current behavior in fluent2-tokens.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1090)


sophialee0416 (approved):

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -0,0 +1,224 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Pre-defined styles of the notification
+@objc public enum MSFNotificationStyle: Int, CaseIterable {
+    case primaryToast
```
harrieshin, line 10: these should be documented too right?

mischreiber: Not seeing any documentation here still.

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -168,23 +174,27 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
...
                     .font(.fluent(tokens.boldTextFont))
+                    .hoverEffect()
                 } else {
                     SwiftUI.Button(action: {
                         isPresented = false
                         buttonAction()
                     }, label: {
                         if let trailingImage = state.trailingImage {
                             Image(uiImage: trailingImage)
+                                .accessibilityLabel(state.trailingImageAccessibilityLabel ?? "")
```
laminesm, line 185: Shouldn't we have a default accessibility label for the trailing image instead of an empty string?

laminesm (approved):

ios/FluentUI/Notification/FluentNotification.swift
huwilkes: By default, the notification will use an X icon and have "Dismiss" as an accessibility label. This is only for if they've set a custom trailing image.

huwilkes: Ah, but, that's not quite what you were suggesting. This matches just the normal default behavior of an image, I think, and lets us add the label if there is one. If `.modifyIf` worked with `if let`, I think that would've been a better choice, but it doesn't. So with this approach we either have the regular default behavior, or the desired label.

PR #1091, opened 2022-07-22
Fix for progress bar accessibility trait bug

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When selecting the progress bar in the UIKit demo, VoiceOver would say "button", but nothing would happen in the SwiftUI demo. It appears this had nothing to do with the progress bar, but it was the `TableViewCell` in the UIKit demo that has accessibility traits set to button by default which is what VoiceOver was reading out. The `updatesFrequently` trait was added to the progress bar as well as the activity indicator accessibility label.

### Verification

| https://user-images.githubusercontent.com/106181067/180839351-93739efb-d23d-4f5a-ad8e-31d583a62af6.mov | https://user-images.githubusercontent.com/106181067/181353958-79c1a0ed-a72a-4848-806e-9152c37fe795.mov |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1091)

### DIFF
--- a/ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBar.swift
+++ b/ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBar.swift
@@ -42,6 +42,7 @@ public struct IndeterminateProgressBar: View, ConfigurableTokenizedControl {
                    maxHeight: height,
                    alignment: .center)
             .background(backgroundColor)
+            .accessibilityAddTraits(.isImage)
             .modifyIf(state.isAnimating, { view in
                 view
                     .onAppear {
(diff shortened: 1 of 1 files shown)


harrieshin:
please attach video for before and after

ios/FluentUI/IndeterminateProgressBar/IndeterminateProgressBar.swift
```diff
@@ -42,6 +42,7 @@ public struct IndeterminateProgressBar: View, ConfigurableTokenizedControl {
                    maxHeight: height,
                    alignment: .center)
             .background(backgroundColor)
+            .accessibilityAddTraits(.isImage)
```
harrieshin, line 45: is image correct? if you use just the native activity indicator or uiprogressview, isn't the trait https://developer.apple.com/documentation/uikit/uiaccessibilitytraits/1620187-updatesfrequently ?

harrieshin:
@laminesm : the after video doesn't seem to have any different behavior?

harrieshin:
what's Apple native activity indicator behavior? and how can we mimic it? as a back up plan, we do have localized string ""Accessibility.ActivityIndicator.Animating.label" ?

harrieshin (approved):

PR #833, opened 2021-12-16
Update usage of MSFSeparator and SwiftUI Divider to Fluent Divider

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated usage of the old separator and SwiftUI divider to the new divider. Also had to make some adjustments to the new divider for it to play nice with the other controls.

After reviewing with design, we also decided to change the color of the divider from stroke.neutral1 to stroke.neutral2 to have better contrast.

### Verification

I've checked all the old usages of the separator or Swift UI divider and made sure the new divider was doing its job in the controls.

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Divider_Old | Divider_New |
| BottomCommanding_Old | BottomCommanding_New |
| Calendar_Old | Calendar_New |
| DateTime_Old | DateTime_New |
| Drawer_Old | Drawer_New |
| ShyHeader_Old | ShyHeader_New |
| PeoplePicker_Old | PeoplePicker_New |
| PeoplePickerResults_Old | PeoplePickerResults_New |
| PopupMenu_Old | PopupMenu_New |
| SideTab_Old | SideTab_New |
| ActivityIndicator_Old | ActivityIndicator_New |
| Indeterminate_Old | Indeterminate_New |
| Notification_Old | Notification_New |
| NotificationOutline_Old | NotificationOutline_New |
| BadgeField_Old | BadgeField_New |
| Avatar_Old | Avatar_New |
| Button_Old | Button_New |
| Color_Old | Color_New |
| TabBar_Old | TabBar_New |
| Shimmer_Old | Shimmer_New |
| List_Old | List_New |
| TableHeaderFooter_Old | TableHeaderFooter_New |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [x] Different resolutions (1x, 2x, 3x)
- [ x Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

[...]

PR #1093, opened 2022-07-23
Fixing separator leading insets.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The calculation for the leading insets of the separator (now replaced with the new MSFDivider control) is incorrect as it takes the safe area left inset into consideration while it's being added to the cell's content view, which already respects the safe area insets.

The previous version of the TableViewCell that used the Separator control added it to the cell view directly instead of the contentView. That was the reason the safeAreaInsets.left need to be considered in the calculation of the separator insets.

### Verification

Ran in both iPhone and iPad running iOS 14 and 15.
Verified that the Separator inset was always rendered correctly as the screenshots below show:

| separatorLeadingInsets_bug_dark | separatorLeadingInsets_fixed_dark |
| separatorLeadingInsets_bug | separatorLeadingInsets_fixed |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1093)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1430,9 +1430,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         guard type == .inset else {
             return 0
         }
-        let baseOffset = safeAreaInsets.left + TableViewCell.selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
-                                                                                    selectionImageMarginTrailing: tokens.selectionImageMarginTrailing,
-                                                                                    selectionImageSize: tokens.selectionImageSize.width)
+        let baseOffset = TableViewCell.selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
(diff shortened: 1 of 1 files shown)


sophialee0416 (approved):

harrieshin:
@sophialee0416  is it just the screenshot but the full divider doesn't actually go through edge to edge on after shot?

harrieshin:
Screen Shot 2022-07-25 at 1 53 04 PM

harrieshin:
@huwilkes https://github.com/microsoft/fluentui-apple/pull/833/files#diff-f8fc51f87d33547de10c5737ad1e05a7a6dab701da49c59880c1b419b49cf18b in actioncell this bug doesn't repro. but the divider is part of the cell'sview. but in tableviewcell it is part of contentview. But if you make the divider part of cell's view, it messes up the cell's height calculation. is there something in the divider that might be messing with the layout?

sophialee0416 (changes requested):
Changing status after observing section border in landscape mode

PR #1095, opened 2022-07-25
[Fluent2 Tokens] Moving `Avatar` calculated colors out of `AvatarTokens`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The calculated colors used by `Avatar` are not really design tokens per se, so this PR moves them out of the `AvatarTokens` class entirely.

This change creates a new struct, `Avatar.CalculatedColors`, that is responsible for all the work of calculating the colors based on a set of our "standard" global colors.

### Verification

Verified that there are no behavioral changes after this change.

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1095)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -138,16 +138,16 @@ public struct Avatar: View, ConfigurableTokenizedControl {
         let presenceIconFrameSideRelativeToOuterRing: CGFloat = presenceIconFrameSideRelativeToInnerRing + outerGapAndRingThicknesCombined
         let overallFrameSide = max(ringOuterGapSize, presenceIconFrameSideRelativeToOuterRing)
 
-        let foregroundColor = state.foregroundColor?.dynamicColor ?? ( !shouldUseCalculatedColors ?
-                                                                       tokens.foregroundDefaultColor :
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -413,6 +377,88 @@ public struct Avatar: View, ConfigurableTokenizedControl {
         }
     }
 
+    /// Handles calculating colors for Avatar foreground and background.
+    private struct CalculatedColors {
```
harrieshin, line 381: can we move it to a separate file so that in case we ever need to expose this to clients we can make the struct public easily?

mischreiber: I know we discussed that a bit this morning, but after reviewing the code I believe that customization will be best served by the existing `backgroundColor` and `foregroundColor` properties on the `AvatarState` object, and this makes sense to keep private. Thoughts?

harrieshin (approved):

laminesm (approved):

edjamesmsft (approved):

PR #1097, opened 2022-07-26
UIBackgroundConfiguration to change the cell color

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Follow the best practice https://developer.apple.com/videos/play/wwdc2020/10027/ to use the UIBackgroundConfiguration which is mutually exclusive to UITableViewCell's backgroundColor property.

- Use the token system to be hooked with UIBackgroundConfiguration
- Dark Elevated colors are not correctly defined for fluent2-token branch (Unknown: why MSFDivider color is different from Fluent1?)
- New api: clients can set cellstyletype to determine whether or not it should be grouped or plain style cell. If clients wants to override the color of the cell they should use the token system
- DemoController: add support for keyboard navigation on
- Cell background color when there is a keyboard focus on the cell is currently the same color as when user taps on the cell.

### Verification

|style| Before                                       | After                                      |
|----|----------------------------------------------|--------------------------------------------|
| dark elevated group style|dark_elevated_grouped | dark_elevated_groupstyle|
| dark elevated plain style| dark_elevated_plain|   dark_elevated_plainstyle|
| dark group style| dark_group | dark_groupstyle|
| dark plain style| dark_plain | dark_plainstyle |
| light group style| light_groupstyle | light_groupstyle
| light plain style|light_plainstyle |light_plainstyle|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1097)

### DIFF
--- a/ios/FluentUI/Table View/TableViewTokens.swift
+++ b/ios/FluentUI/Table View/TableViewTokens.swift
@@ -18,12 +18,18 @@ open class TableViewCellTokens: ControlTokens {
     }
 
     /// The background color of the TableViewCell.
-    open var cellBackgroundColor: DynamicColor { aliasTokens.backgroundColors[.neutral1] }
+    open var cellBackgroundColor: DynamicColor {
+        .init(light: aliasTokens.backgroundColors[.neutral1].light,
+              dark: aliasTokens.backgroundColors[.neutral1].dark,
+              darkElevated: aliasTokens.backgroundColors[.neutral2].darkElevated)
+    }
 
     /// The grouped background color of the TableViewCell.
     open var cellBackgroundGrouped: DynamicColor {
         .init(light: aliasTokens.backgroundColors[.neutral1].light,
-              dark: aliasTokens.backgroundColors[.neutral2].dark)}
+              dark: aliasTokens.backgroundColors[.neutral3].dark,
+              darkElevated: ColorValue(0x212121))
+    }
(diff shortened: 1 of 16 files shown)


ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -18,12 +18,18 @@ open class TableViewCellTokens: ControlTokens {
     }
 
     /// The background color of the TableViewCell.
-    open var cellBackgroundColor: DynamicColor { aliasTokens.backgroundColors[.neutral1] }
+    open var cellBackgroundColor: DynamicColor {
+        .init(light: aliasTokens.backgroundColors[.neutral1].light,
+              dark: aliasTokens.backgroundColors[.neutral1].dark,
+              darkElevated: aliasTokens.backgroundColors[.neutral2].darkElevated)
```
sophialee0416, line 24: Is this the correct color? I think I remember gray950 being 141414?

harrieshin: right, so neutral2,darkelevated is grey8 which is 0x141414.

sophialee0416 (approved):

PR #1101, opened 2022-07-27
[Fluent 2 iOS] PopupMenu colors update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

`PopupMenu` colors were updated to match fluent 2 specs. Instead of using functions that return each of the required colors, I just created lazy variables that get instantiated in the `PopupMenuItem`'s init using the fluentTheme in `accessoryView` or a new `UIView`. 

### Verification

| before_dark | after_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1101)

### DIFF
--- a/ios/FluentUI/Popup Menu/PopupMenuItem.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuItem.swift
@@ -31,15 +31,15 @@ open class PopupMenuItem: NSObject, PopupMenuTemplateItem {
     @objc public lazy var subtitleColor: UIColor = Colors.Table.Cell.subtitle
     /// `image` tint color if it is rendered as template
     @objc public lazy var imageColor: UIColor = Colors.Table.Cell.image
-    /// `title` color when`isSelected` is true. If unset, Colors.primary will be used
+    /// `title` color when`isSelected` is true. If unset, brand foreground 1 will be used
     @objc public var titleSelectedColor: UIColor?
-    /// `subtitle` color when`isSelected` is true.  If unset, Colors.primary will be used
+    /// `subtitle` color when`isSelected` is true.  If unset, brand foreground 1 will be used
     @objc public var subtitleSelectedColor: UIColor?
-    /// tint color if `selectedImage` is rendered as template and `isSelected` is true.  Is unset, Colors.primary will be used
+    /// tint color if `selectedImage` is rendered as template and `isSelected` is true.  If unset, brand foreground 1 will be used
     @objc public var imageSelectedColor: UIColor?
     /// background color of PopupMenuItem corresponding cell
     @objc public var backgroundColor: UIColor = .clear
-    /// checkmark color `isAccessoryCheckmarkVisible` and `isSelected` is true. If unset, Colors.primary will be used
+    /// checkmark color `isAccessoryCheckmarkVisible` and `isSelected` is true. If unset, brand foreground 1 will be used
     @objc public var accessoryCheckmarkColor: UIColor?
 
     @objc public let onSelected: (() -> Void)?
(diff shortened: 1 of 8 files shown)


ios/FluentUI/Popup Menu/PopupMenuItem.swift
```diff
@@ -59,7 +59,35 @@ open class PopupMenuItem: NSObject, PopupMenuTemplateItem {
...
+                                               object: nil)
+    }
+
+    @objc private func themeDidChange(_ notification: Notification) {
+        initRequiredColors()
+    }
+
+    private func initRequiredColors() {
+        guard let accessoryView = self.accessoryView else {
+            let view = UIView()
```
edjamesmsft, line 79: How expensive is it to create and throw away this UIView? Is there any other way to get a view to get the current fluent theme?

laminesm: Another way (besides making the color variables functions) would be to pass in a view in the init. But we would either need to go through all the scripts initializing `PopupMenuItem` and add a view in the init or make it an optional view which would still require us to create a view if nil.

harrieshin: not a view, at this point it should pass in a token. but the property is there for existing clients to override the colors. ultimately everyone should be using override tokens but that more on @huwilkes future work.

harrieshin: basically this nsobject class doesn't really need to know about uiview.

laminesm: So should we just go and pass in directly the color tokens in all the init calls?

laminesm: Could we just make these colors optionals just like the selected colors? Then we could provide the default tokens if these colors aren't set by the client.

ios/FluentUI/Popup Menu/PopupMenuItem.swift
```diff
@@ -26,20 +26,20 @@ open class PopupMenuItem: NSObject, PopupMenuTemplateItem {
     @objc public var isSelected: Bool = false
 
     /// `title` color
-    @objc public var titleColor: UIColor = Colors.Table.Cell.title
+    @objc public var titleColor: UIColor?
```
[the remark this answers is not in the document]
laminesm: I'm looking into it a different branch and it seems the issue might be with `themeDidChange` conflicting with the one in the super class `TableViewCell`.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -198,7 +198,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         return self
     }
 
-    @objc private func themeDidChange(_ notification: Notification) {
+    @objc func themeDidChange(_ notification: Notification) {
```
huwilkes, line 201: @harrieshin wanted this particular change to go from main->fluent2_colors, so I'll spin up a PR for that one, and you can cherry-pick that over.

ios/FluentUI/Popup Menu/PopupMenuItemCell.swift
```diff
@@ -89,6 +89,23 @@ class PopupMenuItemCell: TableViewCell, PopupMenuItemTemplateCell {
...
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        updateColors()
+    }
+
+    @objc override func themeDidChange(_ notification: Notification) {
+        guard let window = window, window.isEqual(notification.object) else {
+            return
+        }
+        updateColors()
```
huwilkes, line 108: I think you'll also want super.themeDidChange.

huwilkes (approved):

PR #1104, opened 2022-07-27
[Fluent2 TokenSet] Merging in more controls

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- `ActivityIndicator`
- `Avatar`
- `AvatarGroup`

In addition to these initial controls, I've also tweaked how overrides are defined in the demo controller. This affects `CardNudgeDemoController` as well, so you'll see a nice before/after there.

### Verification

| [Screenshots incoming] | [Screenshots incoming] |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1104)


ios/FluentUI/ActivityIndicator/ActivityIndicatorTokenSet.swift
```diff
@@ -0,0 +1,70 @@
...
+                case .medium:
+                    return 24
+                case .large:
+                    return 32
+                case .xLarge:
+                    return 36
+                }
+            }
+
+        case .thickness:
```
harrieshin, line 55: we don't have overall global size or thickness tokens? why is it a control token set?

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokenSet.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    /// MSFActivityIndicatorSize enumeration value that will define pre-defined values for side and thickness.
+    public internal(set) var size: MSFActivityIndicatorSize = .large
+
+    override func defaultValue(_ token: Tokens) -> ControlTokenValue {
+        switch token {
+        case .defaultColor:
+            return .dynamicColor { self.aliasTokens.foregroundColors[.neutral4] }
+
+        case .side:
```
harrieshin, line 39: side? should it be size?

ios/FluentUI/Avatar/AvatarTokenSet.swift
```diff
@@ -0,0 +1,249 @@
...
+        case backgroundDefaultColor
+
+        /// The default color of the foreground of the `Avatar`
+        case foregroundDefaultColor
+
+    }
+
+    override func defaultValue(_ token: Tokens) -> ControlTokenValue {
+        switch token {
+        case .avatarSize:
```
harrieshin, line 55: activityindicator it is ".side" but in avatar it is ".avatarSize"?

ios/FluentUI/AvatarGroup/AvatarGroup.swift
```diff
@@ -112,78 +113,90 @@ public struct AvatarGroup: View, ConfigurableTokenizedControl {
...
+                }
+                return avatarSize + interspace - ringGapOffset - ringOuterGap - (currentAvatarHasRing ? ringOuterGap : 0)
+            }()
+
+            let sizeDiff = avatarSize - nextAvatarSize - (currentAvatarHasRing ? 0 : ringGapOffset)
+            let yOrigin = sizeDiff / 2
+
+            // Hand the rendering of the avatar to a helper function to appease Swift's
+            // strict type-checking timeout.
+            VStack {
```
harrieshin, line 180: does "avatarViewInHStack" naming make sense if we have vstack stuff going on inside this block?

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokenSet.swift
laminesm: A comment in the definition says it's the `value for the side of the square frame of an Activity Indicator.`

mischreiber: `side` is correct, or at least it's the historical name of the token.

ios/FluentUI/Avatar/AvatarTokenSet.swift
mischreiber: Yep! Not sure why other than weird legacy/history.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
mischreiber: Fair, not the best name. I'll get a better one.

ios/FluentUI/ActivityIndicator/ActivityIndicatorTokenSet.swift
mischreiber: We do, but (A) they don't map 1:1 to these values, and (B) even if it did, we don't want to directly reference tokens in the control itself. All values should go through a `ControlTokenSet` for code cleanliness and to provide clean override hooks.

mischreiber:
Fixing a regression in AvatarGroup now--will update once that's sorted.

ios/FluentUI/AvatarGroup/AvatarGroup.swift
mischreiber: New signature:
```swift
func avatarView(at index: Int, for avatar: MSFAvatarStateImpl)
```

edjamesmsft (approved):

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift
```diff
@@ -236,45 +236,39 @@ extension ActivityIndicatorDemoController: DemoAppearanceDelegate {
...
     }
 
     func perControlOverrideDidChange(isOverrideEnabled: Bool) {
         defaultColorIndicators.values.forEach { activityIndicator in
-            activityIndicator.state.overrideTokens = (isOverrideEnabled ? PerControlOverrideActivityIndicatorTokens() : nil)
+            perControlOverrideActivityIndicatorTokens.forEach { (key, value) in
+                if isOverrideEnabled {
+                    activityIndicator.tokenSet[key] = value
+                } else {
+                    activityIndicator.tokenSet.removeOverride(key)
```
huwilkes, line 249: Bad idea and probably not worth it, but thought I'd share anyway.
Should this have stack-like behavior? So that if you add an override, then remove it, you don't completely reset any pre-existing overrides?

mischreiber: Each token value (`.backgroundColor`, `.borderWidth`, etc) can only ever have one direct override value. This removal API simply removes the specific override for that token, without affecting any other token value.

I don't think we want to support any sort of FILO-style "queueing" of tokens, if that's what you're suggesting. Our intent is to support exactly two approaches to overrides:

- Direct overrides via modifying the `tokenSet` for a given component.
- Indirect overrides via the `FluentTheme` registration API.

Or am I misunderstanding your question?

huwilkes: No, you got the question. I figured we didn't want to have any sort of token history, but figured I'd ask since I had the thought.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -73,7 +73,10 @@ import SwiftUI
 }
 
 /// View that represents the avatar.
-public struct Avatar: View, ConfigurableTokenizedControl {
+public struct Avatar: View, TokenizedControlView {
+    public typealias TokenSetKeyType = AvatarTokenSet.Tokens
```
huwilkes, line 77: Does this also need `public typealias TokenSetType = AvatarTokensSet`?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -216,14 +219,14 @@ public struct Avatar: View, ConfigurableTokenizedControl {
 
         @ViewBuilder
         var avatarBody: some View {
-            if tokens.style == .group {
+            if state.style == .group {
```
huwilkes, line 222: This should just be `style`, from line 102?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -92,16 +95,16 @@ public struct Avatar: View, ConfigurableTokenizedControl {
         state.primaryText = primaryText
         state.secondaryText = secondaryText
 
-        self.state = state
+        self.init(state)
```
huwilkes, line 98: The documentation comment for this init says it should be used for "internal container views". Should we be using it here, or should we update the comment to just say "internal" and leave out the "container view"?

huwilkes (approved):

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Apparently not--I'll remove that from the other controls as well.

PR #1111, opened 2022-07-28
Adding original width and height methods back for TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding original height and preferredWidth class funcs back into TableViewCell to prevent objc from breaking

### Verification
made sure that there were no visual changes

| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1111)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -212,7 +212,6 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         updateTokens()
     }
 
-    #warning("This method will be deprecated.")
     /// The height of the cell based on the height of its content.
     ///
     /// - Parameters:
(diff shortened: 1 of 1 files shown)


edjamesmsft:
Just for my knowledge, why were they removed in the first place?

sophialee0416:
> Just for my knowledge, why were they removed in the first place?

They weren't removed, but changed; we added attributedText-related parameters and a parameter for tokens (since this is a class method, we can't use the tokens instantiated within the class and had to pass it in). But tokens is a little funky in objc so adding this back in should prevent any breaks caused by these methods

harrieshin:
can we put warnings this will be deprecated?

harrieshin:
> can we put warnings this will be deprecated?

offline discussion. we decided not to warn because there isn't a good alternative for objc clients.

sophialee0416:
> can we put warnings this will be deprecated?

Offline resolution, we'll bring up deprecation at a later time once we figure out objc + tokens compatibility

harrieshin (approved):

PR #1113, opened 2022-07-29
[Fluent 2 iOS] Label colors update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The label colors were updated to fluent 2 colors. The error, warning and disabled styles were removed as they were unused and we don't need to support them in the future.

### Verification

| Screen Shot 2022-07-28 at 1 39 57 PM | Screen Shot 2022-08-08 at 12 53 34 PM |
| Screen Shot 2022-07-28 at 1 40 02 PM | Screen Shot 2022-08-08 at 12 53 40 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1113)

### DIFF
--- a/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
@@ -423,9 +423,6 @@ public final class AliasTokens {
         case brandStroke1
         case brandStroke1Pressed
         case brandStroke1Selected
-        // Shared colors
-        case redForeground1
-        case yellowForeground1
     }
     public lazy var colors: TokenSet<ColorsTokens, DynamicColor> = .init { [weak self] token in
         guard let strongSelf = self else { preconditionFailure() }
@@ -623,12 +620,6 @@ public final class AliasTokens {
         case .brandStroke1Selected:
             return DynamicColor(light: strongSelf.globalTokens.brandColors[.comm60].light,
                                 dark: strongSelf.globalTokens.brandColors[.comm120].light)
-        case .redForeground1:
-            return DynamicColor(light: strongSelf.globalTokens.sharedColors[.red][.shade20],
-                                dark: strongSelf.globalTokens.sharedColors[.red][.tint30])
-        case .yellowForeground1:
-            return DynamicColor(light: strongSelf.globalTokens.sharedColors[.yellow][.shade40],
-                                dark: strongSelf.globalTokens.sharedColors[.yellow][.tint30])
         }
     }
 
--- a/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
@@ -423,6 +423,9 @@ public final class AliasTokens {
         case brandStroke1
         case brandStroke1Pressed
         case brandStroke1Selected
+        // Shared colors
+        case redForeground1
+        case yellowForeground1
     }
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Label/Label.swift
```diff
@@ -17,22 +17,22 @@ public enum TextColorStyle: Int, CaseIterable {
     case warning
     case disabled
 
-    func color(for window: UIWindow) -> UIColor {
+    func color(fluentTheme: FluentTheme) -> UIColor {
```
harrieshin, line 20: interesting... you are actually tokenizing this control in this branch?

ios/FluentUI/Label/Label.swift
```diff
@@ -17,22 +17,22 @@ public enum TextColorStyle: Int, CaseIterable {
...
-            return Colors.textSecondary
+            return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foreground3])
         case .white:
             return .white
         case .primary:
-            return Colors.primary(for: window)
+            return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.brandForeground1])
         case .error:
-            return Colors.error
+            return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.redForeground1])
```
harrieshin, line 31: who uses label error, warning, disable style? do we still need to support them? especially error or warning? does this change impact badgefield or something?

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -423,6 +423,9 @@ public final class AliasTokens {
         case brandStroke1
         case brandStroke1Pressed
         case brandStroke1Selected
+        // Shared colors
+        case redForeground1
```
edjamesmsft, line 427: Are there more than yellow and red foreground colors that should be added as part of this change?

ios/FluentUI/Label/Label.swift
laminesm: Yeah this is the approach we've been mainly using to pull tokens for untokenized controls.

ios/FluentUI/Label/Label.swift
laminesm: I'll dig into it and removed them if they're not used/needed anywhere.

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
laminesm: Not for label but most likely we will need to add a few more for BadgeView and NotificationView.

edjamesmsft: Would have been ideal to add all the new colors in a separate PR, but I'd go ahead and add them here so they're ready to go for future changes.

laminesm: I thought we discussed creating shared colors alias tokens as we encounter them on a per control basis. It would save us some time and prevent us from creating tokens that will be unused.

mischreiber: I don't believe that should be the case: all tokens must be part of our library for clients to consume, so there will never be "unused tokens".

edjamesmsft (approved):

harrieshin:
seems like this branch needs update and updated description of the pr?

PR #1116, opened 2022-07-29
Button, Command Bar, IndeterminateProgressBar

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- `Button`
- `Command Bar` (and related views)
- `IndeterminateProgressBar`

### Verification

NOTE: The difference in color for the other controls in `IndeterminateProgressBar` is known and expected due to the state of the working branch where "after" screenshots are taken. Same with the weird divider widths visible in some demos (especially `Button`).

| before-ipb | after-ipb |
| before-cb1 | after-cb |
| before-b2 | after-b2 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1116)


ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -75,6 +75,7 @@ public enum ControlTokenValue {
     case dynamicColor(() -> DynamicColor)
     case fontInfo(() -> FontInfo)
     case shadowInfo(() -> ShadowInfo)
+    case buttonDynamicColors(() -> ButtonDynamicColors)
```
harrieshin, line 78: hmmm why is buttonynamiccolors different from dynamiccolors? does that mean anything that might have concatinating characteristics needs to be part of this enum?

mischreiber: Effectively yes, any type that can be used as a "token" needs to live in this enum. Buttons use a custom struct to hold several discrete `DynamicColor` instances, so that super-struct needs to be stored.

In other words, a `DynamicColor` is a set of raw color values (light, dark, high contrast, etc), while a `ButtonDynamicColors` struct is a set of `DynamicColors` (standard, disabled, pressed, etc). Different types need different cases in the enum as it's currently designed.

mischreiber: Alternative approach: split up the `ButtonDynamicColors` struct and store each separate color as a separate token value?

harrieshin: yeah i like the split up approach. makes the enum much simplier. what if there is a new thing inside buttondynamiccolors... it is going to be more pain in the future

harrieshin (approved):

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
mischreiber: Discussed offline: we'll leave as-is for now, and then investigate in a future PR.

edjamesmsft (approved):

PR #1119, opened 2022-08-01
[Fluent2 TokenSet] TableViewCell and other cells

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- `ActionsCell`
- `ActivityIndicatorCell`
- `BooleanCell`
- `CenteredLabelCell`
- `TableViewCell`

### Verification

[Screenshots WIP]
|  |  |
### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1119)


ios/FluentUI.Demo/FluentUI.Demo/Demos/OtherCellsDemoController.swift
```diff
@@ -76,28 +54,34 @@ extension OtherCellsDemoController: DemoAppearanceDelegate {
...
-            // "Charcoal"
-            return DynamicColor(light: GlobalTokens().sharedColors[.charcoal][.tint50],
-                                dark: GlobalTokens().sharedColors[.charcoal][.shade40])
-        }
+    private var themeWideOverrideTableViewCellTokenSet: [TableViewCellTokenSet.Tokens: ControlTokenValue] {
+        return [
+            .mainBrandColor: .dynamicColor {
+                // "Charcoal"
+                return DynamicColor(light: GlobalTokens().sharedColors[.charcoal][.tint50],
+                                    dark: GlobalTokens().sharedColors[.charcoal][.shade40])
```
huwilkes, line 62: I guess it was in the original code, but do we really need the `// "Charcoal"` comment when the next two lines both have `GlobalTokens().sharedColors[.charcoal]`?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
```diff
@@ -519,9 +519,8 @@ class TableViewCellFileAccessoryViewDemoController: DemoTableViewController {
```
huwilkes, line 519: For sure outside the scope of this review, but this `extraPadding` line is weird. Doesn't seem like we should really have other, different constants in our demos? Unless they're explicitly token overrides.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -231,34 +216,35 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
-                                   customViewSize: MSFTableViewCellCustomViewSize = .default,
-                                   customAccessoryView: UIView? = nil,
-                                   accessoryType: TableViewCellAccessoryType = .none,
-                                   titleNumberOfLines: Int = 1,
-                                   subtitleNumberOfLines: Int = 1,
-                                   footerNumberOfLines: Int = 1,
-                                   customAccessoryViewExtendsToEdge: Bool = false,
-                                   containerWidth: CGFloat = .greatestFiniteMagnitude,
-                                   isInSelectionMode: Bool = false) -> CGFloat {
+    public class func height(tokenSet: TableViewCellTokenSet? = nil,
```
huwilkes, line 219: Is this branch up to date with all the latest TVC changes that had been going in to main?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -202,7 +187,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     /// The height of the cell based on the height of its content.
     ///
     /// - Parameters:
-    ///   - tokens: The TableViewCell tokens
+    ///   - tokenSet: The TableViewCell tokens
```
huwilkes, line 190: Should it be `///  - tokenSet: The TableViewCell token set`?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1116,9 +1101,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
 
     /// Initializes TableViewCell with the cell style.
     @objc public override init(style: UITableViewCell.CellStyle, reuseIdentifier: String?) {
-        let tokens = TableViewCellTokens()
-        self.tokens = tokens
-
+        self.tokenSet = TableViewCellTokenSet(customViewSize: { .default })
```
huwilkes, line 1104: I don't think this is right. I think this loses any change to `customViewSize`? It should be `slef.tokenSet = TableViewCellTokenSet(cutomViewSize: { self.customViewSize })` where `self.customViewSize` defaults to `.default`, or something, right?

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -4,172 +4,329 @@
 //
 import UIKit
 
-open class TableViewCellTokens: ControlTokens {
-    /// The background color of the TableView.
-    open var backgroundColor: DynamicColor {
-        .init(light: globalTokens.neutralColors[.white],
-              dark: globalTokens.neutralColors[.black])
-    }
+public class TableViewCellTokenSet: ControlTokenSet<TableViewCellTokenSet.Tokens> {
```
huwilkes, line 7: Did you want to also rename this file?

ios/FluentUI/Table View/TableViewTokens.swift
```diff
@@ -4,172 +4,329 @@
...
+        case accessoryDetailButtonColor
+
+        /// The main primary brand color of the theme.
+        case mainBrandColor
+
+        /// The destructive text color in an ActionsCell.
+        case destructiveTextColor
+
+        /// The communication text color in an ActionsCell.
+        case communicationTextColor
```
huwilkes, line 142: I don't think `destructiveTextColor` and `communicationTextColor` were intended to be part of the main TableViewCell token class. Since only `ActionCell` uses them.

ios/FluentUI.Demo/FluentUI.Demo/Demos/OtherCellsDemoController.swift
mischreiber: ⬜

Eh, probably not "needed" but yeah I don't think it's hurting anything enough to warrant removing.

ios/FluentUI.Demo/FluentUI.Demo/Demos/TableViewCellFileAccessoryViewDemoController.swift
mischreiber: Yeah I don't know what's going on here 👎 but agreed, that's a "not this PR, but worth re-evaluating" problem.

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -134,6 +150,19 @@ public enum ControlTokenValue {
                                        disabled: defaultColor)
         }
     }
+
+    public var pillButtonDynamicColors: PillButtonDynamicColors {
+        if case .pillButtonDynamicColors(let pillButtonDynamicColors) = self {
+            return pillButtonDynamicColors()
+        } else {
+            assertionFailure("Cannot convert token to PillButtonDynamicColors: \(self)")
+            let defaultColor = DynamicColor(light: ColorValue(0xE3008C))
```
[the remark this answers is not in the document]
mischreiber: Ah good point. It comes from the global `Hot Pink` primary color, which should hopefully stand out. That said, I think I want to fail a little more quietly here, so I'll put pink behind a debug flag and use black otherwise.

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Almost certainly not; it's likely not even merged with changes in `fluent2-tokens`, since it was spun off before those changes began. I'll spin up a second iteration that at least pulls that all in.

mischreiber: Checked: it is caught up with `fluent2-tokens` but not `main`. I'll wait for that merge to complete before addressing any needed changes.

ios/FluentUI/Table View/TableViewTokens.swift
mischreiber: That's correct, and this is something I would like us to revisit in the future. In the current new system, only the base token set can define the set of token values in the `Tokens` enum. Swift doesn't allow extending enums for pretty good reasons, but that means children can't easily add new values.

There are ways around this, including containerization, but in the name of expediency I'm taking the easiest solution _for now_, with an intent to revisit in the future.

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Good catch. It's a little tricky because you can't reference `self` before calling `super.init()` but I have a solution in mind.

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Sure!

ios/FluentUI/Table View/TableViewTokens.swift
mischreiber: Good catch. This didn't happen automatically because the file name didn't match the class. Changing it manually to match for the future.

ios/FluentUI/Table View/TableViewCell.swift
huwilkes: Clever. Is that better than making the token set lazy?

huwilkes (approved):

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Oh that's even smarter! You can't make it lazy in SwiftUI, but that's totally allowed in UIKit.

sophialee0416 (approved):

PR #917, opened 2022-02-17
Adding `SharedColors` and simple `ColorTokensDemoController`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change adds 392 shared global colors, and adds a new demo controller to show them off!

Three hundred and ninety-two! That's a lot of colors.

They're stored as a single TokenSet of TokenSets, to minimize the impact on our public API footprint.

In the future this can (and will) be expanded to include brand colors and neutrals, but this is a quick standup with the intent of unblocking additional control conversion (specifically `Avatar`, which relies on these values).

### Verification

I exercised the API to fetch these colors with a cool demo controller. Check it out:

https://user-images.githubusercontent.com/4934719/154591003-3919b964-b99f-480d-a708-7f197529ffb2.mp4

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [x] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

PR #1126, opened 2022-08-02
[Fluent 2 iOS] Created new alias tokens for shared colors

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

New alias tokens were added for shared colors. Just like the rest of the color tokens, they were added to the `ColorsTokens` enum.

### Verification

- Ran the demo app to ensure nothing was broken 
- Tested a random selection of tokens on PopupMenu  

### Pull request checklist

This PR has considered:


ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -423,6 +423,172 @@ public final class AliasTokens {
         case brandStroke1
         case brandStroke1Pressed
         case brandStroke1Selected
+        // Dark red
```
huwilkes, line 426: A bit annoying, though it should be fast-ish, but could you put these in the same order as the `SharedColorSets` enum?

huwilkes: Ah, actually, never mind. The actual global tokens aren't even in the same order as the enum.

huwilkes: I think probably reorder to match the enum, but let's check with @mischreiber 

Somewhat relevant comment from the PR they added the enum in:
https://github.com/microsoft/fluentui-apple/pull/917#discussion_r810414366

harrieshin: are we actually planning on using these shared colors? lets not arbirarily add them if we aren't going to use them.

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -423,6 +423,172 @@ public final class AliasTokens {
...
+        case darkRedBackground2
+        case darkRedForeground2
```
mischreiber, line 428: These colors are called out in designs as being a separate "set" of colors: Persona Colors. Shouldn't that mean they live as a separate top-level `TokenSet` within `AliasTokens`?

mischreiber: Additionally, I'm not convinced that these should be alias tokens at all. Let's follow up with design and see if there's sufficient justification for these at all.

mischreiber (changes requested):

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
laminesm: > These colors are called out in designs as being a separate "set" of colors: Persona Colors. Shouldn't that mean they live as a separate top-level `TokenSet` within `AliasTokens`?

So far we've opted to add all new fluent 2 tokens in the same set and wanted to keep it consistent with these alias tokens. But we could move these to a separate set after discussing with design.

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
laminesm: They've been defined for avatar colors.

.swiftlint.yml
```diff
@@ -1,4 +1,4 @@
-whitelist_rules:
+only_rules:
```
edjamesmsft, line 1: It looks like you pulled in some other changes - should these be done as a separate merge into upstream fluent2-colors?

laminesm: This change is part of @harrieshin's PR into main that I cherry picked to fix the recent linting issues. I don't think it's necessary to merge it separately into fluent2-colors.

harrieshin: even though our git branches history is a mess, there is a value of cherry picking one commit at a time to preserve as much history

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -623,6 +623,612 @@ public final class AliasTokens {
         }
     }
 
+    // MARK: Shared Colors
+
+    public enum SharedColorsTokens: CaseIterable {
```
mischreiber, line 628: The latest designs in Figma have moved this out of alias tokens and into the Avatar itself. Feel free to ping me offline if you have questions or need any pointers!

mischreiber (changes requested):

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -645,6 +645,206 @@ public final class AliasTokens {
...
+        case greenBackground2
+        case greenBackground3
+        case greenForeground1
+        case greenForeground2
+        case greenForeground3
+        case greenBorderActive
+        case greenBorder1
+        case greenBorder2
+        // Berry
+        case berryBackground1
```
harrieshin, line 692: talked offline: if this isn't used outside of OOF badge, it should not be an alias token

ios/FluentUI/Core/Theme/Tokens/AliasTokens.swift
```diff
@@ -645,6 +645,206 @@ public final class AliasTokens {
         }
     }
 
+    // MARK: Shared Colors
+
+    public enum SharedColorsTokens: CaseIterable {
+        // Red
+        case redBackground1
```
mischreiber, line 652: Maybe a question for design: why are these called `redBackground` instead of `dangerBackground`? That seems to be the proper semantic intent of these values, and alias tokens are meant to be semantic, right?

laminesm: I'll bring this up to Ian

mischreiber (approved):
Let's make sure we keep up with the work design is doing to rename these tokens and get those applied here before this branch merges to `main`.

PR #1127, opened 2022-08-02
[Fluent2 TokenSet] SegmentedControl, TabBarItem, TabBar, SideTabBar

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- `SegmentedControl`
- `TabBarItem`
- `TabBar`
- `SideTabBar`

Note that we've changed how sub-control token values is slightly changed. The containing controls (`TabBar` and `SideTabBar` still expose the properties for their items, but those properties cannot be optional. Instead, they have default values, along with a debug-only assert. Then, instead of having a full subclass as before, the [Side]TabBar sets the overridden values on  its tab bar items on each update.

### Verification

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1127)


harrieshin (approved):

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
```diff
@@ -26,26 +26,21 @@ class SegmentPillButton: UIButton {
...
-            updateUnreadDot()
-        }
-    }
+    var tokenSet: SegmentedControlTokenSet
 
-    private func updateTokenizedValues() {
-        titleLabel?.font = UIFont.fluent(tokens.font, shouldScale: false)
-        let verticalInset = tokens.verticalInset
-        let horizontalInset = tokens.horizontalInset
+    func updateTokenizedValues() {
```
huwilkes, line 31: Why the switch to internal? Also, is this still getting called? I don't see it anywhere in SegmentedControl.swift, so are the actual buttons of the segmented control getting updates? Unless we call this from the SegmentedControl, I don't think the font or insets will get updates (though, the insets need a completely different update, since they're getting deprecated...)

edjamesmsft (approved):

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
mischreiber: Yep you're right, fixing now

PR #1139, opened 2022-08-05
Introduce isHidden property to CommandBarItem

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The existing CommandBarItem does not allow items to change their hidden state. This change aims to enable that capability and fix a related layout issue.

- Add a public `isHidden` `bool` property to `CommandBarItem` that defaults to `false`. This value is used in the `CommandBarButton updateButtonState` method to update the button's `isHidden` property.
    - A new argument in the `CommandBarItem` initializer was _not added_ so that this will not be a breaking API change for Obj-C consumers. I don't think it's 100% necessary to add it as an argument at this time anyways.
- Add the `MaskedView` private UIView subclass to resolve a layout issue where the mask's frame does update when the size of the `CommandBarView`'s subviews change.
    - A layer mask is applied to the `containerView` inside `CommandBarView` to create a visual texture between the leading and trailing pinned items and the center scrolling items. The mask is set on the `layer` of the `containerView`. Currently, the `CommandBarView.layoutSubview()` attempts to update the mask's frame when the subviews change, but the `layoutSubviews` method isn't always called since the size of the CommandBarView's view generally doesn't change when its subviews change. To fix this, `layoutSubviews` is implemented on actual parent view of the layer mask and the mask frame is updated there.

### Verification

- Verified the behavior using the demo controller
- Verified the behavior with devmain builds

https://user-images.githubusercontent.com/10938746/183263977-e82e90de-2e5d-4cf7-92a3-e5510939b2df.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1139)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
@@ -198,16 +198,21 @@ class CommandBarDemoController: DemoController {
         refreshLeadingItemButton.addTarget(self, action: #selector(refreshDefaultLeadingBarItems), for: .touchUpInside)
         itemCustomizationContainer.addArrangedSubview(refreshLeadingItemButton)
 
-        let customizationStackView = UIStackView()
-        customizationStackView.axis = .horizontal
-        customizationStackView.alignment = .fill
-        customizationStackView.distribution = .fillProportionally
-        customizationStackView.addArrangedSubview(createLabelWithText("'+' Item Enabled"))
+        let itemEnabledStackView = createHorizontalStackView()
+        itemEnabledStackView.addArrangedSubview(createLabelWithText("'+' Item Enabled"))
         let itemEnabledSwitch: UISwitch = UISwitch()
         itemEnabledSwitch.isOn = true
         itemEnabledSwitch.addTarget(self, action: #selector(itemEnabledValueChanged), for: .valueChanged)
-        customizationStackView.addArrangedSubview(itemEnabledSwitch)
-        itemCustomizationContainer.addArrangedSubview(customizationStackView)
+        itemEnabledStackView.addArrangedSubview(itemEnabledSwitch)
+        itemCustomizationContainer.addArrangedSubview(itemEnabledStackView)
+
+        let itemHiddenStackView = createHorizontalStackView()
+        itemHiddenStackView.addArrangedSubview(createLabelWithText("'+' Item Hidden"))
+        let itemHiddenSwitch: UISwitch = UISwitch()
+        itemHiddenSwitch.isOn = false
+        itemHiddenSwitch.addTarget(self, action: #selector(itemHiddenValueChanged), for: .valueChanged)
+        itemHiddenStackView.addArrangedSubview(itemHiddenSwitch)
+        itemCustomizationContainer.addArrangedSubview(itemHiddenStackView)
 
         itemCustomizationContainer.addArrangedSubview(UIView()) //Spacer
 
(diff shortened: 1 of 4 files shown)


harrieshin:
can we add an example of democontroller please?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -279,3 +275,22 @@ extension CommandBar: UIScrollViewDelegate {
         updateShadow()
     }
 }
+
+/// A UIView subclass that udpates its mask frame during layoutSubviews. By default, the layer mask
```
edjamesmsft, line 279: Nit: udpates -> updates

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -279,3 +275,22 @@ extension CommandBar: UIScrollViewDelegate {
...
+private class MaskedView: UIView {
+    init() {
+        super.init(frame: .zero)
+    }
+
+    @available(*, unavailable)
+    required init(coder: NSCoder) {
+        preconditionFailure("init(coder:) has not been implemented")
+    }
+
```
anandrajeswaran, line 291: Don't need any of this init stuff since we don't do any special behavior on init?

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -27,6 +27,7 @@ open class CommandBarItem: NSObject {
         self.isEnabled = isEnabled
         self.isSelected = isSelected
         self.itemTappedHandler = itemTappedHandler
+        self.isHidden = false
```
anandrajeswaran, line 30: Minor: Can initialize in place in the var declaration `    @objc public var isHidden = false {`

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Correct. Removed.

harrieshin:
please test landscape and accessibility to make sure it is properly hidden from VO tree.  also works fine when it is actually shown ontop of keyboard?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -279,3 +275,13 @@ extension CommandBar: UIScrollViewDelegate {
         updateShadow()
     }
 }
+
+/// A UIView subclass that updates its mask frame during layoutSubviews. By default, the layer mask
+/// is not hooked into auto-layout and will not update its frame if its parent frame changes size. This implementation
+/// fixes that.
+private class MaskedView: UIView {
```
harrieshin, line 282: becauz this is so specific for this file for a particular use case, can we be more clear like CommandingContainerView instead?  "maskedView" makes it little confusing to me at a glance be might think this is a uikit thing?

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -334,6 +339,14 @@ class CommandBarDemoController: DemoController {
         }
     }
 
+    func createHorizontalStackView() -> UIStackView {
+        let switchView = UIStackView()
+        switchView.axis = .horizontal
+        switchView.alignment = .fill
+        switchView.distribution = .fillProportionally
```
harrieshin, line 346: make sure it looks fine on larger text mode.

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
nodes11: Updated to improve large text mode support.

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Sure. Updated.

PR #1141, opened 2022-08-08
Update iOS project to have explicit output path

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updating the iOS output files to go somewhere not declared in a script that may or may not have run yet. Moved both the input and output paths to the FluentUILib_common xcconfig. Updated the script to use new xcconfig values and print out what it is copying from where to where.

### Verification
Built and ran the test app locally 
Built the FluentUI-iOS target and checked the build logs. Found `export SCRIPT_OUTPUT_FILE_0\=<Xcode Derived Data>/FluentUI-fhjvlehnpqgcfqhfpfcpsonixlbk/Build/Products/Debug-iphonesimulator/include/FluentUI/FluentUI-Swift.h` instead of `export SCRIPT_OUTPUT_FILE_0\=/FluentUI-Swift.h`
Copy script prints out info on what it's copying from where to where.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1141)


harrieshin:
should clean up the script itself as well right?

huwilkes:
> should clean up the script itself as well right?

I can if you'd like, but I don't think we need to. With it how it is now, we only need to change it in two places if it ever changes, rather than 3.

harrieshin:
out of sync in the future is something i want to make sure we address. which might be why having xcconfig definiton would help us have single source of truth ?

harrieshin (approved):

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -1345,7 +1345,7 @@
 			outputFileListPaths = (
 			);
 			outputPaths = (
-				"${target_dir}/FluentUI-Swift.h",
+				"${BUILT_PRODUCTS_DIR}/include/${PRODUCT_MODULE_NAME}/FluentUI-Swift.h",
```
anandrajeswaran, line 1348: I wonder if this might also be part of the issue - I think it's supposed to be $(<setting name>) instead of ${<setting name>} (parentheses instead of curly braces)?

huwilkes:
> out of sync in the future is something i want to make sure we address. which might be why having xcconfig definiton would help us have single source of truth ?

Moved the paths to the FluentUILib_common xcconfig.

ios/xcode/FluentUILib_common.xcconfig
```diff
@@ -9,3 +9,7 @@ OTHER_LDFLAGS = -ObjC
 PRODUCT_NAME = FluentUI
 SKIP_INSTALL = YES
 SWIFT_VERSION = 5.0
+
+// Objective C header file location paths
+HSP_INPUT = $(DERIVED_SOURCES_DIR)
```
harrieshin, line 14: update the script itself as well? also can we also add FLUENT as prefix of these alias?

harrieshin (approved):

PR #1148, opened 2022-08-09
[Fluent 2 iOS] TableViewCell colors update (2/2)

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

As we are trying to move away from `Colors.swift`, the `Colors` extension was removed from `TableViewCell`.
Some of the colors in the `Table` struct were used by other controls in fluent. In fact, `PersonaListView`, `TableViewHeaderFooterTitleView` & `PopupMenuItemCell` had dependencies to some of these colors and they were all updated to use their appropriate color tokens instead.
The Objective-C colors were changed into functions that take in a `FluentTheme` and output the right token. 

### Verification

| before_persona_light | after_persona_light |
| before_persona_dark | after_persona_dark |
| before_title_light | after_title_light |
| before_title_dark | after_title_dark |
| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1148)

### DIFF
--- a/ios/FluentUI/People Picker/PersonaListView.swift
+++ b/ios/FluentUI/People Picker/PersonaListView.swift
@@ -78,7 +78,7 @@ open class PersonaListView: UITableView {
     @objc override init(frame: CGRect, style: UITableView.Style) {
         super.init(frame: frame, style: style)
 
-        backgroundColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
+        updateBackgroundColor()
         separatorStyle = .none
         tableFooterView = UIView(frame: .zero)
 
@@ -92,6 +92,22 @@ open class PersonaListView: UITableView {
 
         dataSource = self
         delegate = self
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+    }
+
+    @objc private func themeDidChange(_ notification: Notification) {
+        guard let window = window, window.isEqual(notification.object) else {
+            return
+        }
+        updateBackgroundColor()
+    }
+
+    private func updateBackgroundColor() {
+        backgroundColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
     }
 
     @objc public required init(coder aDecoder: NSCoder) {
--- a/ios/FluentUI/People Picker/PersonaListView.swift
+++ b/ios/FluentUI/People Picker/PersonaListView.swift
@@ -78,7 +78,7 @@ open class PersonaListView: UITableView {
     @objc override init(frame: CGRect, style: UITableView.Style) {
         super.init(frame: frame, style: style)
 
-        backgroundColor = Colors.Table.background
+        backgroundColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
         separatorStyle = .none
         tableFooterView = UIView(frame: .zero)
 
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -84,18 +84,6 @@ public enum TableViewCellBackgroundStyleType: Int {
     }
 }
 
-// MARK: - Table Colors
-
-public extension Colors {
-    // Objective-C support
-    @objc static var tableBackground: UIColor { return surfacePrimary }
-    @objc static var tableBackgroundGrouped: UIColor { return UIColor(light: surfaceSecondary, dark: surfacePrimary) }
-    @objc static var tableCellBackground: UIColor { return surfacePrimary }
-    @objc static var tableCellBackgroundGrouped: UIColor { return UIColor(light: surfacePrimary, dark: surfaceSecondary) }
-    @objc static var tableCellBackgroundSelected: UIColor { return surfaceTertiary }
-    @objc static var tableCellImage: UIColor { return iconSecondary }
-}
-
 // MARK: - TableViewCell
 
 /**
@@ -148,6 +136,26 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     @objc public static var smallHeight: CGFloat { return height(title: "", customViewSize: .small) }
     @objc public static var mediumHeight: CGFloat { return height(title: "", subtitle: " ") }
     @objc public static var largeHeight: CGFloat { return height(title: "", subtitle: " ", footer: " ") }
+
+    /// TableViewCell colors with obj-c support
+    @objc public static func tableBackground(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
+    }
+    @objc public static func tableBackgroundGrouped(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.canvasBackground])
+    }
+    @objc public static func tableCellBackground(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
+    }
+    @objc public static func tableCellBackgroundGrouped(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background3])
+    }
+    @objc public static func tableCellBackgroundSelected(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1Pressed])
+    }
+    @objc public static func tableCellImage(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foreground3])
+    }
 
     /// Identifier string for TableViewCell
     @objc public static var identifier: String { return String(describing: self) }
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -87,26 +87,6 @@ public enum TableViewCellBackgroundStyleType: Int {
 // MARK: - Table Colors
 
 public extension Colors {
-    internal struct Table {
-        struct Cell {
-            static var image: UIColor = iconSecondary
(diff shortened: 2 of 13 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -80,26 +80,6 @@ public enum TableViewCellBackgroundStyleType: Int {
...
-            static var textDivider: UIColor = textSecondary
-            static var textLink = UIColor(light: Palette.communicationBlueShade10.color, dark: communicationBlue)
-        }
-
-        static var background: UIColor = surfacePrimary
-        static var backgroundGrouped = UIColor(light: surfaceSecondary, dark: surfacePrimary)
-    }
-
     // Objective-C support
     @objc static var tableBackground: UIColor { return surfacePrimary }
```
harrieshin, line 84: all the objc colors need to be updated then?

ios/FluentUI/People Picker/PersonaListView.swift
```diff
@@ -92,6 +92,22 @@ open class PersonaListView: UITableView {
...
+
+    @objc private func themeDidChange(_ notification: Notification) {
+        guard let window = window, window.isEqual(notification.object) else {
+            return
+        }
+        updateBackgroundColor()
+    }
+
+    private func updateBackgroundColor() {
+        backgroundColor = UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
```
harrieshin, line 110: now that we have elevated colors we should test those

ios/FluentUI/Table View/TableViewCell.swift
laminesm: Yes they'd need to. I just spoke with @huwilkes regarding this and since we'd need to remove Colors at some point maybe I should go ahead and remove the objc colors here as well. Then we can update the controls to pull in the right tokens in devmain. Would that be a good idea?

harrieshin: i think we still need to have some sort of objc support for few colors. but we will need to look at client codes before making that decision.

laminesm: Sounds good. I'll work on finding a way to pull them out of `Colors`.

ios/FluentUI/People Picker/PersonaListView.swift
laminesm: The PR description was updated with screenshots of before and after elevated colors

ios/FluentUI/People Picker/PersonaListView.swift
```diff
@@ -92,6 +92,22 @@ open class PersonaListView: UITableView {
...
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+    }
+
+    @objc private func themeDidChange(_ notification: Notification) {
+        guard let window = window, window.isEqual(notification.object) else {
+            return
+        }
+        updateBackgroundColor()
```
edjamesmsft, line 106: Should this not also have updateColors?

laminesm: This is the only color update in `PersonaListView`

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -84,38 +84,6 @@ public enum TableViewCellBackgroundStyleType: Int {
...
-            static var textDivider: UIColor = textSecondary
-            static var textLink = UIColor(light: Palette.communicationBlueShade10.color, dark: communicationBlue)
-        }
-
-        static var background: UIColor = surfacePrimary
-        static var backgroundGrouped = UIColor(light: surfaceSecondary, dark: surfacePrimary)
-    }
-
-    // Objective-C support
-    @objc static var tableBackground: UIColor { return surfacePrimary }
```
sophialee0416, line 111: Have we double checked that this isn't being used in any downstream dependencies?

laminesm: These colors were used in a bunch of devmain files. The plan is to update those usages when porting fluent 2 colors into devmain.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -169,6 +137,26 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     @objc public static var mediumHeight: CGFloat { return height(title: "", subtitle: " ") }
     @objc public static var largeHeight: CGFloat { return height(title: "", subtitle: " ", footer: " ") }
 
+    /// TableViewCell colors with obj-c support
+    @objc public static func tableBackground(fluentTheme: FluentTheme) -> UIColor {
```
mischreiber, line 141: Nit: can these functions have the word `Color` in them?

```swift
    @objc public static func tableBackgroundColor(fluentTheme: FluentTheme) -> UIColor {
```

harrieshin:
i kind of want all of this change except for the tableviewcell.swift and some pieces you are refering to fluent2 color to be in mainbranch first. Because most of the stuff is demoapp fixes only which has no impact on the framework. changes in exsiting fluent controls like personalist and tableheaderfooterview that just listens to the theme without any public api changes. these are all good things to have in main and not specific to fluent2-colors branch.

edjamesmsft:
I disagree, Harrie. These changes are all directly related to the work to remove Colors.swift and are not entirely independent of the work being done in the fluent2-colors branch. Additionally, I don't see this change as crucial to have in main and it would introduce yet more delay into the schedule for the fluent 2 work. Unless you feel VERY strongly that this should be broken into two PRs (one in main, then merged to fluent 2 colors, then the remaining change pushed on top), I would prefer this to go in now, and eventually it will make it's way back to main.

harrieshin:
because overall fluent2-color branch merge to main is being pushed back i think there are changes that should also live in Fluent2-token branch which is why i think some of the fluent-theme listening changes can live in main.

laminesm:
> because overall fluent2-color branch merge to main is being pushed back i think there are changes that should also live in Fluent2-token branch which is why i think some of the fluent-theme listening changes can live in main

Essentially all non tokenized controls we have updated so far have FluentTheme listening changes. Is there any particular reason we would want to have just a handful of non tokenized controls into fluent2-tokens?

harrieshin:
> > because overall fluent2-color branch merge to main is being pushed back i think there are changes that should also live in Fluent2-token branch which is why i think some of the fluent-theme listening changes can live in main
> 
> Essentially all non tokenized controls we have updated so far have FluentTheme listening changes. Is there any particular reason we would want to have just a handful of non tokenized controls in fluent2-tokens?

yup. which at the end, speeds up fluent2-token branch to be complete faster?

laminesm:
I've just started working on it and we would also need to pull in the updated `AliasTokens` and `GlobalTokens` from fluent2-colors. It might create some conflicts in main as many tokens have been changed/moved around/renamed. Should I still go ahead with it or just use the closest matching token from the main branch?

harrieshin:
its not the end of the world if we have to copy and past in fluent-token side as well. @huwilkes @sophialee0416 yall fine with that?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -169,6 +137,26 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
 
+    /// TableViewCell colors with obj-c support
+    @objc public static func tableBackgroundColor(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
+    }
+    @objc public static func tableBackgroundGroupedColor(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.canvasBackground])
+    }
+    @objc public static func tableCellBackgroundColor(fluentTheme: FluentTheme) -> UIColor {
+        UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background1])
```
harrieshin, line 148: i would have thought you will be referencing to tableviewtokens.swift itself not referencing fluenttheme.aliastoken.

edjamesmsft:
> its not the end of the world if we have to copy and past in fluent-token side as well. @huwilkes @sophialee0416 yall fine with that? the merges are going to be harder that's why i suggested on a separate change which makes commit conflict much smaller.

I'm not certain I understand what you're proposing here. But, if it makes merges harder, I would strongly recommend avoiding it. Our merges are already going to be a monster, I would hate to make it any worse!

huwilkes:
> > its not the end of the world if we have to copy and past in fluent-token side as well. @huwilkes @sophialee0416 yall fine with that? the merges are going to be harder that's why i suggested on a separate change which makes commit conflict much smaller.
> 
> I'm not certain I understand what you're proposing here. But, if it makes merges harder, I would strongly recommend avoiding it. Our merges are already going to be a monster, I would hate to make it any worse!

I think @laminesm was asking if we should bring some of the update tokens from fluent2-colors -> main, which is a hard no. We don't want to add any of the new tokens to main until we're ready to remove the old ones. We could add the notification for theme updates on main if we wanted, but the change from `Colors.tableCellBackgroundSelected` to `TableViewCell.tableCellBackgroundSelectedColor(fluentTheme: cell.fluentTheme)` is a breaking change I really don't think we want to move to main right now.

harrieshin:
> > > its not the end of the world if we have to copy and past in fluent-token side as well. @huwilkes @sophialee0416 yall fine with that? the merges are going to be harder that's why i suggested on a separate change which makes commit conflict much smaller.
> > 
> > 
> > I'm not certain I understand what you're proposing here. But, if it makes merges harder, I would strongly recommend avoiding it. Our merges are already going to be a monster, I would hate to make it any worse!
> 
> I think @laminesm was asking if we should bring some of the update tokens from fluent2-colors -> main, which is a hard no. We don't want to add any of the new tokens to main until we're ready to remove the old ones. We could add the notification for theme updates on main if we wanted, but the change from `Colors.tableCellBackgroundSelected` to `TableViewCell.tableCellBackgroundSelectedColor(fluentTheme: cell.fluentTheme)` is a breaking change I really don't think we want to move to main right now.

right my first comment was not including tableviewcell.swift file and only about the theme listening

harrieshin (approved):

PR #1151, opened 2022-08-11
Update Notification to make default action optional

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The UIKit notification had a default action by default, so I added that back in to the SwiftUI one, but now we want that to be optional. Also cleaned up additional bools left in the style.

### Verification

| Notification_DefaultDismiss_Before | Notification_DefaultDismiss_After2 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1151)


ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -42,6 +42,9 @@ import SwiftUI
     /// To show an action button, provide values for both `actionButtonTitle` and  `actionButtonAction`.
     var actionButtonAction: (() -> Void)? { get set }
 
+    /// Bool to control if the Notification has a dismiss action by default.
+    var hasDefaultDismissActionButton: Bool { get set }
```
harrieshin, line 46: nitpick? "show" sounds better than "has"?

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -63,9 +66,11 @@ public struct FluentNotification: View, ConfigurableTokenizedControl {
     ///   - attributedTitle: Optional attributed text to draw above the message area. If set, it will override the title parameter.
     ///   - image: Optional icon to draw at the leading edge of the control.
     ///   - trailingImage: Optional icon to show in the action button if no button title is provided.
+    ///   - trailingImageAccessibilityLabel: Optional localized accessibility label for the trailing image.
     ///   - actionButtonTitle:Title to display in the action button on the trailing edge of the control.
     ///   - actionButtonAction: Action to be dispatched by the action button on the trailing edge of the control.
-    ///   - messageButtonAction: Action to be dispatched by tapping on the toast/bar notification.
+    ///   - hasDefaultDismissButton: Bool to control if the Notification has a dismiss action by default.
+    ///   - messageButtonAction: Action to be dispatched by tapping on the toast/bar notification.   
     public init(style: MSFNotificationStyle,
```
harrieshin, line 74: are we creating new convenence method to not disturb exsiting objc clients?

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -46,12 +46,6 @@ import UIKit
     var animationDurationForHide: TimeInterval { return Constants.animationDurationForHide }
     var animationDampingRatio: CGFloat { return isToast ? Constants.animationDampingRatioForToast : 1 }
 
-    var needsFullWidth: Bool { return !isToast }
-    var needsSeparator: Bool { return  self == .primaryOutlineBar }
-    var supportsTitle: Bool { return isToast }
-    var supportsImage: Bool { return isToast }
```
harrieshin, line 52: why are these other 4 things deleted?

edjamesmsft: I suspect because they're not actually consumed anywhere (just checked in code).

ios/FluentUI/Notification/FluentNotification.swift
huwilkes: No, this is just the SwiftUI one, it doesn't have @objc

ios/FluentUI/Notification/NotificationTokens.swift
huwilkes: Right, they were used in the old UIKit one, but aren't being used anywhere, so I figured the clean up just got missed in the transition.

harrieshin: should be part of the description of the pr

ios/FluentUI/Notification/FluentNotification.swift
harrieshin: ah i see. but then i am confused why we adding this? can't swift client just override state themselves after creating a new instance?

huwilkes: I think we need to add it here in order for the .presentNotification modifier to really be able to customize the state. So I think it's mostly for SwiftUI usage. Objective C is also able to just modify the state after init. So if the concern were just Objective C exposure of the state, we could drastically simplify a lot of our inits.

mischreiber: Unlike other SwiftUI controls, `Notification` passes all of its state in `init` instead of via modifiers. I don't know why that is; these should probably all be broken out into separate functions in `NotificationModifiers`.

ios/FluentUI/Notification/NotificationTokens.swift
mischreiber: Fortunately these were never public APIs, so there's nothing breaking about it.

ios/FluentUI/Notification/FluentNotification.swift
edjamesmsft: I feel like that should be done in a separate PR though.

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -437,5 +448,11 @@ class MSFNotificationStateImpl: NSObject, ControlConfiguration, MSFNotificationS
         self.actionButtonTitle = actionButtonTitle
         self.actionButtonAction = actionButtonAction
         self.messageButtonAction = messageButtonAction
+        self.showFromBottom = showFromBottom
+        if let showDefaultDismissActionButton = showDefaultDismissActionButton {
```
harrieshin, line 452: i don't have swift compiler in my head but self.showDefaultDismisseActionButton = showDefaultDismissActionButton ?? style.isToast dosn't work?

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -410,6 +418,7 @@ class MSFNotificationStateImpl: NSObject, ControlConfiguration, MSFNotificationS
 
     @objc init(style: MSFNotificationStyle) {
         self.style = style
+        self.showDefaultDismissActionButton = style.isToast
```
harrieshin, line 421: is this needed if you are double checking in line 455?

huwilkes: This one doesn't call the convenience init, so it isn't double checking on 455.

ios/FluentUI/Notification/FluentNotification.swift
huwilkes: Ah, yeah, that should also work.

ios/FluentUI/Notification/FluentNotification.swift
harrieshin: hmmm everything should evenually funnel to one init? so we don't have duplicated stuff?

harrieshin (approved):

PR #1152, opened 2022-08-11
Make `GlobalTokens` immutable

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In order to align with the broader goals of Fluent UI design customization, this change makes all global tokens fully immutable. I've done so via the following:

* `GlobalTokens` is now a struct composed of only pure static functions that return prebaked token values. It cannot be initialized (`init` is marked as `unavailable`).
* `brandColors`, which _can_ be customized, have been moved to `AliasTokens`.
* `FluentTheme`, `AliasTokens`, and `ControlTokens` all lose their references to an instance of `GlobalTokens` and instead will refer to the static global values.
* All components that pulled values from an instance of `GlobalTokens` have been updated to use the new static lookup APIs.

### Verification

Sanity pass of all tokenized controls demonstrates no changes in appearance or behavior. Existing customization continues as expected.

Additionally, we've got a big binary size savings!

| 30,446,587 bytes (29,733 kB) | 30,304,923 bytes (29,595 kB) |

Improvement: *138 kB* 🥳

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1152)


ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -50,10 +43,10 @@ import SwiftUI
         }
     }
 
-    static var shared: FluentTheme = .init()
+    /// The aassociated `AliasTokens` for this theme.
```
edjamesmsft, line 46: Nit: "associated"

ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -7,17 +7,10 @@ import SwiftUI
 
 /// Base class that allows for customization of global, alias, and control tokens.
 @objc public class FluentTheme: NSObject, ObservableObject {
-    /// Initializes and returns a new `FluentTheme`, with optional custom global and alias tokens.
+    /// Initializes and returns a new `FluentTheme`.
```
edjamesmsft, line 10: Nit: still worth it to call out that alias tokens can be customized?

mischreiber: Good idea. I'll add it to the discussion field.

harrieshin (approved):

huwilkes (approved):

PR #1156, opened 2022-08-12
Implement Shimmer in SwiftUI

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Implement ShimmerView and ShimmerLinesView in SwiftUI

Known Issues:
1. The existing logic is off with the shimmer angle (even in the UIKit implementation, the angles vary between the views). The shimmerAngle token's comment says "Angle of the direction of the gradient, in radian. 0 means horizontal, Pi/2 means vertical," but the formula for radians is degrees * (pi/180), not -(pi/degrees) as it currently is. I will address this in a separate change. For now, the SwiftUI implementation uses the .topLeading and .bottomTrailing UnitPoints for the gradient.

Simulator Screen Recording - iPad Pro (12 9-inch) (5th generation) - 2022-08-12 at 02 25 51

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1156)


huwilkes:
Can you grab a before/after binary size comparison?

ios/FluentUI/Shimmer/ShimmerViewSwiftUI.swift
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Determines whether the view itself is shimmered or an added cover on top is shimmered
+    var shouldAddShimmeringCover: Bool { get set }
+
+    /// Whether to use the height of the view (if the view is a label), else default to token value.
+    var usesTextHeightForLabels: Bool { get set }
+}
+
+/// View Modifier that adds a "shimmering" effect to any view
+public struct ShimmerViewSwiftUI: ViewModifier, TokenizedControlView {
```
huwilkes, line 22: I don't think this should be named ShimmerViewSwiftUI. I think the UIKit one needs to be renamed MSFShimmerView. Unless it's a naming collision with an existing SwiftUI control, then it should be FluentShimmerView. But I think it's the first one. And the UIKit one should probably also be demoted to just a wrapper. Though reducing that in another PR would make sense. For now, we should probably just change the name of the file and class.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ShimmerLinesViewDemoController_SwiftUI.swift
```diff
@@ -0,0 +1,179 @@
...
+                VStack {
+                    Text("This is a single label being shimmered.")
+                        .padding(.leading)
+                }
+                .frame(maxWidth: .infinity, alignment: .leading)
+                .shimmering(style: style,
+                            usesTextHeightForLabels: usesTextHeightForLabels,
+                            animationId: namespace,
+                            isLabel: true)
+                .padding()
```
sophialee0416, line 38: Did we want to use a specific padding value here (and throughout the rest of the file)?

ios/FluentUI/Shimmer/ShimmerLinesViewSwiftUI.swift
```diff
@@ -0,0 +1,99 @@
...
+    var lineCount: Int { get set }
+
+    /// The percent the first line (if 2+ lines) should fill the available horizontal space.
+    var firstLineFillPercent: CGFloat? { get set }
+
+    /// The percent the last line should fill the available horizontal space.
+    var lastLineFillPercent: CGFloat? { get set }
+}
+
+/// View that represents the ShimmerLines view
```
sophialee0416, line 24: nit: period

ios/FluentUI/Shimmer/ShimmerModifiers.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+public extension View {
+    /// Adds an animated shimmering effect to any view
```
sophialee0416, line 9: period here too

sophialee0416: (and throughout the changes)

ios/FluentUI/Shimmer/ShimmerTokenSet.swift
```diff
@@ -82,6 +85,14 @@ public class ShimmerTokenSet: ControlTokenSet<ShimmerTokenSet.Tokens> {
                     }
                 }
 
+            case .viewTint:
+                return .dynamicColor {
+                    return DynamicColor(light: ColorValue(0xF1F1F1) /* gray50 */,
+                                            lightHighContrast: ColorValue(0x919191) /* gray400 */,
```
sophialee0416, line 91: are these comments meant to be here?

jeaniehuynh: that's how i saw it done elsewhere in the repo (e.g. ResizingHandleTokenSet and TabBarItemTokenSet)

jeaniehuynh:
> Can you grab a before/after binary size comparison?

Before: 34,391,068 bytes (34.8 MB on disk)
After: 35,032,396 bytes (35.4 MB on disk)

huwilkes:
> > Can you grab a before/after binary size comparison?
> 
> Before: 34,391,068 bytes (34.8 MB on disk) After: 35,032,396 bytes (35.4 MB on disk)

That's quite the increase. Do you think you could get the after size without the new demo, so we can see how much just the new control is adding?

jeaniehuynh:
> > > Can you grab a before/after binary size comparison?
> > 
> > 
> > Before: 34,391,068 bytes (34.8 MB on disk) After: 35,032,396 bytes (35.4 MB on disk)
> 
> That's quite the increase. Do you think you could get the after size without the new demo, so we can see how much just the new control is adding?

Without the demo: 34,674,828 bytes (35.1 MB on disk)

ios/FluentUI/Shimmer/MSFShimmerLinesView.swift
```diff
@@ -0,0 +1,121 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/**
+ Specialized ShimmerView that shows 1 or more shimmering lines.
+ */
+@objc(MSFShimmerLinesView)
+open class MSFShimmerLinesView: MSFShimmerView {
```
huwilkes, line 12: You don't need the @objc on a separate line now, so you should be able to do just `@objc open class MSFShimmerLinesView:MSFShimmerView {` all in one go. Same in a few other place.

mischreiber:
> > > > Can you grab a before/after binary size comparison?
> > > 
> > > 
> > > Before: 34,391,068 bytes (34.8 MB on disk) After: 35,032,396 bytes (35.4 MB on disk)
> > 
> > 
> > That's quite the increase. Do you think you could get the after size without the new demo, so we can see how much just the new control is adding?
> 
> Without the demo: 34,674,828 bytes (35.1 MB on disk)

I think we're still gonna have problems pitching a 280kB size increase for this one component...

rdeassis (approved):
Don't forget to resolve the merge conflicts on the project file.
Thanks for using the modifyIf and addressing all the other comments and feedback. :)

PR #1161, opened 2022-08-12
Fixing Avatar and AvatarGroup build warnings caused by the deprecated animation modifier.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The [animation(_:) modifier](https://developer.apple.com/documentation/swiftui/view/animation(_:)-7mq1i) is now deprecated for iOS 15. This means that when the target deployment version is bumped to iOS 15, deprecation warnings will be issued by the compiler.

This PR addresses the use of the animation modifier in the Avatar and AvatarGroup SwiftUI controls.

The fix for the AvatarGroup is just a matter of removing the animation modifier as the transition modifier is doing the job of animating the position and visibility of avatars.

The fix for the Avatar control consists of calling individual animation modifier based on all the values of the state object that need to be monitored to trigger those animations.

### Verification

Basic sanity validation of Avatar scenarios.
Verified all animations remain working properly.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1161)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -267,10 +267,29 @@ public struct Avatar: View, ConfigurableTokenizedControl {
             }
         }
 
+        let standardAnimation = Animation.linear(duration: animationDuration)
+
         return avatarBody
             .pointerInteraction(state.hasPointerInteraction)
             .modifyIf(state.isAnimated, { thisView in
-                thisView.animation(.linear(duration: animationDuration))
+                thisView
+                    .animation(standardAnimation,
+                               value: [state.hasRingInnerGap,
+                                       state.isRingVisible,
+                                       state.isTransparent,
+                                       state.isOutOfOffice])
+                    .animation(standardAnimation,
+                               value: [state.backgroundColor,
+                                       state.foregroundColor,
+                                       state.ringColor])
+                    .animation(standardAnimation,
+                               value: state.size)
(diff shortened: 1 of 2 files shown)


laminesm (approved):

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -267,10 +267,29 @@ public struct Avatar: View, ConfigurableTokenizedControl {
...
         }
 
+        let standardAnimation = Animation.linear(duration: animationDuration)
+
         return avatarBody
             .pointerInteraction(state.hasPointerInteraction)
             .modifyIf(state.isAnimated, { thisView in
-                thisView.animation(.linear(duration: animationDuration))
+                thisView
+                    .animation(standardAnimation,
```
edjamesmsft, line 276: Is there a cost impact to having six calls to animation rather than just one? Is there any way to condense this?

rdeassis: This call is not the one responsible for performing the animation itself. It's just setting it up based on the properties passed.

Previously the way to condense all of them was to just have a call into the animation modifier with no value specified. It'd apply to any changes in the SwiftUI view. Apple just deprecated this call and now we need to have explicit calls to all properties that will cause the view to animate.

PR #1164, opened 2022-08-16
Add GitHub action to compare binary size of the Demo app

huwilkes:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

See https://github.com/huwilkes/fluentui-apple/pull/2 for a history of the actions on each(ish) commit.
Adding a new action that will compare the size of the release, device builds of the demo app from the base and head commits of a PR. I couldn't figure out what trickery Apple uses to get Finder to be able to show the size down the the byte, so we'll have to settle for the size in kilobytes derived from the size in blocks.

### Verification
Added some bloat in the PR from my fork, saw the script reporting a difference in sizes.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1164)


.github/workflows/sizeComparison.yml
```diff
@@ -0,0 +1,66 @@
...
+    strategy:
+      fail-fast: false
+    steps: 
+    - uses: actions/checkout@v3
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Switch to current version of Xcode
+      run: scripts/xcode_select_current_version.sh
+    - name: scripts/xcodebuild_wrapper.sh 'ios_device_build Demo.Development Release build -archivePath $GITHUB_WORKSPACE/headBuild archive'
+      run: scripts/xcodebuild_wrapper.sh ios_device_build Demo.Development Release build -archivePath $GITHUB_WORKSPACE/${{ github.event.pull_request.head.sha }} archive
```
harrieshin, line 25: does this trigger another build? or can we leverage the matrix build triggered by ci.yml?

huwilkes: This currently triggers another build. We might be able to add a dependency on ci.yml, and move the uploading of the artifact there, but I'm not sure.

huwilkes: Though, if we don't do the build here, we'd probably lose the guarantee that we'd actually have a build from the base hash. I'm not sure how much control we have over a the lifetime of an artifact.

harrieshin: what does the build artifact name looks like? how do we guarentee its lifecycle and its uniqueness per PR trigger (and same pr having multiple commit push/iteration)

huwilkes: The artifact name is set in the upload-artifact action, and I'm just using the commit hash as the name. So each trigger will produce two, one for the base hash and one for the head hash. We could probably streamline it by making it check for if the base one already exists, but I hadn't looked in to how to actually do that.

harrieshin: can you also add comments on what the exact output artifect would look like per steps?

.github/workflows/sizeComparison.yml
```diff
@@ -0,0 +1,66 @@
...
+    steps: 
+    - uses: actions/checkout@v3 # Checkout repo at head sha
+      with:
+        ref: ${{ github.event.pull_request.head.sha }}
+    - name: Switch to current version of Xcode
+      run: scripts/xcode_select_current_version.sh
+    - name: scripts/xcodebuild_wrapper.sh 'ios_device_build Demo.Development Release build -archivePath $GITHUB_WORKSPACE/headBuild archive' # Build and archive release device build
+      run: scripts/xcodebuild_wrapper.sh ios_device_build Demo.Development Release build -archivePath $GITHUB_WORKSPACE/${{ github.event.pull_request.head.sha }} archive
+    - name: Create size file # Get size of demo app in "kilobytes" (2*512B blocks) and store in a file with the hash as the name
+      run: du -sk $GITHUB_WORKSPACE/${{ github.event.pull_request.head.sha }}.xcarchive/Products/Applications/FluentUI.Demo.app | cut -f1 >> $GITHUB_WORKSPACE/${{ github.event.pull_request.head.sha }}
```
dannyvv, line 27: is the `pull-request` object populated on 'push' jobs as well?

dannyvv: GitHub's docs don't really have a 'reference guide'  section. it all just examples and general terms....

.github/workflows/sizeComparison.yml
```diff
@@ -0,0 +1,66 @@
...
+      uses: actions/upload-artifact@v3
+      with:
+        name: ${{ github.event.pull_request.head.sha }}
+        path: ${{ github.event.pull_request.head.sha }}
+  archiveBaseBuild:
+    runs-on: macos-12
+    strategy:
+      fail-fast: false
+    steps: 
+    - uses: actions/checkout@v3
```
dannyvv, line 38: If we can't create a reusable job or composite action, I'd love to see more symmetry and use `with: ref: ${{ github.event.pull_request.base.sha }}` So that I don't have to think that BASE contains the 'merged commit' and HEAD is what the PR is based on... Where locally HEAD before submitting it contains your changes... 

Maybe even name then different than base and head... i.e. `withchanges` and `withoutchanges` or `before` and `after`.

.github/workflows/sizeComparison.yml
huwilkes: I believe so. It seemed to be properly updating when I was testing it.

huwilkes: Ah. I misunderstood the question. I'm not actually sure if it gets populated on just a push to a branch, and not during a PR. I can check.

.github/actions/action.yml
```diff
@@ -0,0 +1,24 @@
...
+    type: string
+runs:
+  using: "composite"
+  steps:
+    - uses: actions/checkout@v3 # Checkout repo at sha
+      with:
+        ref: ${{ inputs.sha }}
+    - name: Switch to current version of Xcode
+      run: scripts/xcode_select_current_version.sh
+    - name: scripts/xcodebuild_wrapper.sh 'ios_device_build Demo.Development Release build -archivePath $GITHUB_WORKSPACE/sha archive' # Build and archive release device build
```
harrieshin, line 16: if this is in a different yml, can we just utilize the existing build ci and push a new artifact? rather than doing another build?

harrieshin: only the head build will be needed

huwilkes: I'm not sure there's a way to just add the action on a single build from the matrix?

mischreiber (approved):
As discussed offline, I'd prefer that we be able to fetch old build sizes by pulling old build artifacts. But if that doesn't look feasible, then I'd rather take this than not.

huwilkes:
> As discussed offline, I'd prefer that we be able to fetch old build sizes by pulling old build artifacts. But if that doesn't look feasible, then I'd rather take this than not.

While I think it's possible, it isn't possible with just upload/download artifacts, and seems to require adding some extra steps I wasn't able to get to this week: https://docs.github.com/en/rest/actions/artifacts

I'll check in for now, since I think this adds plenty of value as is, and we can revisit in the future to try and optimize.

PR #1165, opened 2022-08-16
PillButtonBar: Fixing warnings for iOS 15 target deployment version.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We're soon going to be bumping our target deployment version to iOS 15.
This will trigger compile time warnings regarding deprecated APIs in iOS 15.

This PR addresses the PillButtonBar that now needs to use the UIButton Configuration to set title, background color and content insets.

### Verification

Exercised scenarios in the demo app to ensure no regressions were caused:

https://user-images.githubusercontent.com/68076145/184774897-021dea42-b3ef-410c-8f3e-5d38e069d7ff.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1165)

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButtonBar.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButtonBar.swift
@@ -125,6 +125,7 @@ open class PillButtonBar: UIScrollView {
 
     private var stackView: UIStackView = {
         let view = UIStackView()
+        view.distribution = .fillProportionally
         view.alignment = .center
         view.spacing = Constants.minButtonsSpacing
         return view
@@ -411,8 +412,8 @@ open class PillButtonBar: UIScrollView {
         buttonExtraSidePadding = ceil(totalPadding / CGFloat(buttonEdges))
         for button in buttons {
             button.layoutIfNeeded()
-            button.contentEdgeInsets.right += buttonExtraSidePadding
-            button.contentEdgeInsets.left += buttonExtraSidePadding
+            button.configuration?.contentInsets.leading += buttonExtraSidePadding
+            button.configuration?.contentInsets.trailing += buttonExtraSidePadding
             button.layoutIfNeeded()
         }
     }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Pill Button Bar/PillButtonBar.swift
```diff
@@ -125,6 +125,7 @@ open class PillButtonBar: UIScrollView {
 
     private var stackView: UIStackView = {
         let view = UIStackView()
+        view.distribution = .fillProportionally
```
harrieshin, line 128: why this change?

rdeassis: The buttons using the configuration were now the same width by default. The stack view gets the widest button and apply that same width for all the buttons unless the distribution .fillProportionally is set.

harrieshin: i see. then shouldn't it be only set for ios15 and newer?

rdeassis: I can make it specific for that, but it seems like it'll be just redundant in the worst case scenario for iOS 14. It didn't cause any regressions for iOS 14.
Let me know your preference. I'm fine with either option.

harrieshin: i see as long as it didn't cause regression in iOS 14 and the scrolling calculation

harrieshin (approved):

rdeassis:
> test RTL and larger text?

Yes!

Large text:
Screen Shot 2022-08-17 at 9 02 45 AM

RTL:
Screen Shot 2022-08-17 at 9 03 59 AM

ios/FluentUI/Pill Button Bar/PillButtonBar.swift
rdeassis: Yes. I validated on both iOS 14 and 15 and no regressions were introduced by this extra call.

PR #1167, opened 2022-08-16
Adjusting separatorLeadingInset calculation in TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

An update to #1093 

In the previous PR, it was discovered that since the Divider was added to the cell's contentView, we needed to accommodate for the additional safeAreaInset. This means that while it is no longer needed for the `.inset` Separator Type, we need to include it for the `.full` type instead.

### Verification

| image | image |
| image | image |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1694,13 +1694,17 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     }
 
     func separatorLeadingInset(for type: SeparatorType) -> CGFloat {
-        guard type == .inset else {
+        switch type {
+        case .none:
             return 0
+        case .inset:
+            let baseOffset = TableViewCell.selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
+                                                              selectionImageMarginTrailing: tokenSet[.selectionImageMarginTrailing].float,
+                                                              selectionImageSize: tokenSet[.selectionImageSize].float)
+            return baseOffset + paddingLeading + tokenSet[.customViewDimensions].float + tokenSet[.customViewTrailingMargin].float
+        case .full:
+            return -safeAreaInsets.left
         }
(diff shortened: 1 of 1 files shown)


edjamesmsft:
Is it expected that the color changed in the after shot?

edjamesmsft:
Have you checked dark mode?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1694,13 +1694,17 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
...
+        switch type {
+        case .none:
             return 0
+        case .inset:
+            let baseOffset = TableViewCell.selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
+                                                              selectionImageMarginTrailing: tokenSet[.selectionImageMarginTrailing].float,
+                                                              selectionImageSize: tokenSet[.selectionImageSize].float)
+            return baseOffset + paddingLeading + tokenSet[.customViewDimensions].float + tokenSet[.customViewTrailingMargin].float
+        case .full:
+            return -safeAreaInsets.left
```
harrieshin, line 1706: in RTL do you need to check safeareainsets.right? easier to test if the iphone x bezel is in the other side?

sophialee0416: Sanity check:

Left bezel:
image

Right bezel:
image

sophialee0416:
> Is it expected that the color changed in the after shot?

I don't really notice a color change, could you point it out? I didn't touch the colors so shouldn't have been impacted in this change

ios/FluentUI/Table View/TableViewCell.swift
harrieshin: i don't see a code update?

sophialee0416: Ok new sanity check with new changes:

image
image

harrieshin (approved):

PR #1169, opened 2022-08-17
Improve font scaling for Fluent text

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Previously, we've been scaling all text using the ramp for Apple's `body` text style. This works pretty well for _most_ text sizes, but results in too much growth for large text and too little for smaller text. This is because each text style has a fully custom, hand-built scaling ramp, which we were not respecting.

A complication: our Fluent typography system technically supports any arbitrary text size for a component, which won't neatly map to Apple's system.

Solution: let's inspect font sizes and intelligently map them to specific text styles for the purposes of size scaling.

### Verification

I added a new Typography demo controller and used it to compare before/after shots of the different typefaces:

| Size     | Before            | After               | Apple (as FYI) |
|----------|-------------------|---------------------|----------------|
| Standard ("Large") size | before-l |  after-l | apple-l |
| Largest accessibility size | before-a11y | after-a11y | apple-a11y |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1169)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TypographyTokensDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TypographyTokensDemoController.swift
@@ -1,0 +1,84 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import FluentUI
+import UIKit
+
+class TypographyTokensDemoController: DemoTableViewController {
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        tableView.register(UITableViewCell.self, forCellReuseIdentifier: Constants.cellID)
+    }
+
+    override func numberOfSections(in tableView: UITableView) -> Int {
+        return 1
+    }
+
+    override func tableView(_ tableView: UITableView, numberOfRowsInSection section: Int) -> Int {
+        return AliasTokens.TypographyTokens.allCases.count
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        let cell = tableView.dequeueReusableCell(withIdentifier: Constants.cellID, for: indexPath)
+        let typographyToken = AliasTokens.TypographyTokens.allCases[indexPath.row]
+
+        cell.selectionStyle = .none
+
+        var contentConfiguration = cell.defaultContentConfiguration()
+        let text = "\(typographyToken.text)"
+        contentConfiguration.attributedText = NSAttributedString(string: text,
+                                                                 attributes: [
+                                                                    .font: UIFont.fluent(aliasTokens.typography[typographyToken])
+                                                                 ])
+        contentConfiguration.textProperties.alignment = .center
+        cell.contentConfiguration = contentConfiguration
+
+        return cell
+    }
+
+    private var aliasTokens: AliasTokens {
+        guard let fluentTheme = self.view.window?.fluentTheme else {
+            return AliasTokens()
+        }
+        return fluentTheme.aliasTokens
+    }
+
+    private struct Constants {
+        static let cellID: String = "cellID"
+    }
+}
+
+// MARK: - Private extensions
+
+private extension AliasTokens.TypographyTokens {
+    var text: String {
+        switch self {
+        case .display:
+            return "Display"
+        case .largeTitle:
+            return "Large Title"
+        case .title1:
+            return "Title 1"
+        case .title2:
+            return "Title 2"
+        case .title3:
+            return "Title 3"
+        case .body1Strong:
+            return "Body 1 Strong"
+        case .body1:
+            return "Body 1"
+        case .body2Strong:
+            return "Body 2 Strong"
+        case .body2:
+            return "Body 2"
+        case .caption1Strong:
+            return "Caption 1 Strong"
+        case .caption1:
+            return "Caption 1"
+        case .caption2:
+            return "Caption 2"
+        }
+    }
+}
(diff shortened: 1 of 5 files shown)


huwilkes:
Would it be possible to get shots of the Apple fonts at these sizes? Sort of hard to tell how much closer we are with just "wrong" and "right" pictures.

ios/FluentUI/Core/Theme/Tokens/FontInfo.swift
```diff
@@ -17,47 +17,75 @@ public struct FontInfo {
...
+        (15.0, .subheadline),
+        (13.0, .footnote),
+        (12.0, .caption),
+        (11.0, .caption2)
+    ]
+
+    public var textStyle: Font.TextStyle {
+        // Defaults to smallest supported text style for mapping, before checking if we're bigger.
+        var textStyle = Font.TextStyle.caption2
+        for tuple in Self.sizeTuples {
```
huwilkes, line 46: Should we really be going through a for loop every time someone wants to get the textStyle from a FontInfo? Even if it's not an especially long loop.

ios/FluentUI/Core/Theme/Tokens/FontInfo.swift
```diff
@@ -17,47 +17,75 @@ public struct FontInfo {
...
+            // Named font
+            unscaledFont = font.withWeight(uiWeight(fontInfo.weight))
         } else {
-            return .systemFont(ofSize: size, weight: uiWeight(fontInfo.weight))
+            // System font
+            unscaledFont = .systemFont(ofSize: fontInfo.size, weight: uiWeight(fontInfo.weight))
+        }
+
+        if shouldScale {
+            let fontMetrics = UIFontMetrics(forTextStyle: uiTextStyle(fontInfo.textStyle))
```
huwilkes, line 85: I can't seem to find it in the documentation, and it probably does / seems to, but does this maintain the weight that we pass in? The documentation just sort of mentions the base font and doesn't explicitly say that it'll preserve the weight of that font.

mischreiber: Yes it does -- the scaled typography in the attached screenshots maintain their weights. Without the posted weights, those fonts would look _much_ less bold.

ios/FluentUI/Core/Theme/Tokens/FontInfo.swift
mischreiber: I agree that it's a little unusual. Would you prefer an additional layer of cache? I'm not sure that there's any measurable CPU hit for this constrained loop, but I can double-check.

mischreiber: Took a trace, and it looks like we spend ~1ms of time in that method, for a workload with 1.18 seconds of CPU time (just scrolling a table view up and down, constantly regenerating cells). 

I'm leaning towards living with that, but I'm open to looking for further optimization if it feels appropriate.

huwilkes: That seems lightweight enough that we're probably okay as is. Thanks for checking!

huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/TypographyTokensDemoController.swift
```diff
@@ -0,0 +1,84 @@
...
+
+    private var aliasTokens: AliasTokens {
+        guard let fluentTheme = self.view.window?.fluentTheme else {
+            return AliasTokens()
+        }
+        return fluentTheme.aliasTokens
+    }
+
+    private struct Constants {
+        static let cellID: String = "cellID"
```
laminesm, line 49: You should be able to use `TableViewCell.identifier` instead instead of storing this in this struct.

mischreiber: Ah I technically could, but I chose not to because I'm not using the Fluent `TableViewCell` here, and I'd rather not mix and match the two.

(FWIW this class is an almost perfect copy of the `ColorTokensDemoController`, which uses the same paradigm.)

harrieshin (approved):

ios/FluentUI/Core/Theme/Tokens/FontInfo.swift
mischreiber: Upon further thinking, I agree with @huwilkes's first point. Gonna test a simple `Dictionary` lookup cache and report back on perf impact.

mischreiber: Okay, confirmed that there's no real benefit (the dictionary lookups end up taking roughly the same amount of time as the tiny constrained loop).

ios/FluentUI.Tests/FontTests.swift
```diff
@@ -63,16 +29,16 @@ class FontTests: XCTestCase {
         let font = UIFont.fluent(fontInfo)
         let otherFontDescriptor = UIFontDescriptor(name: name.appending("-Semibold"), size: size)
         let otherFont = UIFont(descriptor: otherFontDescriptor, size: size)
-        XCTAssertEqual(font, otherFont)
+        XCTAssertEqual(font.fontDescriptor, otherFont.fontDescriptor)
```
edjamesmsft, line 32: Why did you change this check?

mischreiber: Unfortunately the new way we construct fonts fails Apple's `isEqual` check, despite the fact that every observable external attribute of the font (including its visual appearance!) matches. Since their implementation of equality is opaque to us, the only reasonable workaround I could find was to check the font descriptors, which are simple enumerations of the properties of the font (e.g. weight, size, typeface name).

PR #1172, opened 2022-08-17
CCB: Fixing warnings for iOS 15 target deployment version.

rdeassis:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We're soon going to be bumping our target deployment version to iOS 15.
This will trigger compile time warnings regarding deprecated APIs in iOS 15.

This PR addresses the CommandBar that now needs to use the UIButton Configuration to set its content insets, title and background colors.

### Verification

Demo app scenario validations:

| iOS 14                                       | iOS 15                                      |
| ccb_ios14_light | ccb_ios15_light |
| ccb_ios14_dark | ccb_ios15_dark |

### Pull request checklist

This PR has considered:


ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -38,12 +38,21 @@ class CommandBarButton: UIButton {
         super.init(frame: .zero)
 
         translatesAutoresizingMaskIntoConstraints = false
-        setImage(item.iconImage, for: .normal)
+
+        if #available(iOS 15.0, *) {
+            var buttonConfiguration = UIButton.Configuration.plain()
+            buttonConfiguration.image = item.iconImage
+            buttonConfiguration.contentInsets = LayoutConstants.contentInsets
+            buttonConfiguration.background.cornerRadius = 0
```
harrieshin, line 46: is this necessary? i thought default was zero https://developer.apple.com/documentation/uikit/uibackgroundconfiguration/3600321-cornerradius

harrieshin (approved):

PR #1175, opened 2022-08-18
Expose CommandBarDelegate to notify implementer of CommandBar events

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There is a need for consumers of the `CommandBar` to be notified when scroll events occur.

Created the `CommandBarDelegate` that exposes a method, `commandBarDidScroll(_ commandBar: CommandBar)` that notifies the implementer when a scroll event occurs.

Some alternative approaches were explored including:
1. Utilizing `NSNotification` to publish event globally
2. Have the consumer use KVM to monitor for scroll events. 

Both of these seemed less than ideal so I went with the `CommandBarDelegate` approach.

### Verification

- Verified static library compiles
- Updated the `CommandBarDemoController` with a switch to turn on animations when the `CommandBarDelegate` events occur. The animation simply changes the `CommandBar` background color to blue and then back to the original background color. See video below.

https://user-images.githubusercontent.com/10938746/185459748-9aff1bca-4fc3-440b-af7d-e8a10e8b8f8c.mov

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
@@ -142,6 +142,7 @@ class CommandBarDemoController: DemoController {
     }
 
     var defaultCommandBar: CommandBar?
+    var animateCommandBarDelegateEvents: Bool = false
 
     let textField: UITextField = {
         let textField = UITextField()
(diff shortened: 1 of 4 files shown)


anandrajeswaran:
Usually for the delegate pattern we put the protocol in the same file as the object using the delegate (for example: UITableViewDelegate lives in UITableView.h). Any reason for this one to split out to a separate project?

anandrajeswaran (approved):

PR #1177, opened 2022-08-19
Update ActivityIndicator demo to use UITableViewCell

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The ActivityIndicator demo was using our TableViewCell, which was causing each cell to be read as a button through VoiceOver. Switching to UITableViewCell to match our other demos fixes this, and lets VoiceOver actually navigate to the ActivityIndicator.

### Verification
Before:

https://user-images.githubusercontent.com/67026548/185714074-253ad542-1f2e-42f1-a7d3-1e384adf5d6f.mov

After:

https://user-images.githubusercontent.com/67026548/185714085-25cb6134-71a2-4969-9bc7-904b3b15ca90.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1177)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift
@@ -65,16 +65,36 @@ class ActivityIndicatorDemoController: DemoTableViewController {
 
             return cell
         case.demoOfSize:
-            let cell = TableViewCell()
+            let cell = UITableViewCell()
 
             let activityIndicatorSize = MSFActivityIndicatorSize.allCases.reversed()[indexPath.row]
             let activityIndicatorDictionaries = [defaultColorIndicators, customColorIndicators]
             let activityIndicatorPath = indexPath.section - 2
-            let activityIndicator = activityIndicatorDictionaries[activityIndicatorPath][activityIndicatorSize]
+            guard let activityIndicator = activityIndicatorDictionaries[activityIndicatorPath][activityIndicatorSize] else {
+                return cell
+            }
 
-            cell.setup(title: activityIndicatorSize.description,
-                       customView: activityIndicator)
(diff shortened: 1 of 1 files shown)


mischreiber (approved):
Looks fine for now, but I also want to know why `TableViewCell` is always a button anyway. Is that a property we should expose? Something feels off here.

huwilkes:
> Looks fine for now, but I also want to know why `TableViewCell` is always a button anyway. Is that a property we should expose? Something feels off here.

We set it in `updateAccessibility` of TableViewCell.swift. I also agree that it's a bit broken, but that seemed like a much more in depth change that need more careful consideration than just matching the rest of our demos for now.
``` 
    private func updateAccessibility() {
        accessibilityTraits.insert(.button)

        if isEnabled {
            accessibilityTraits.remove(.notEnabled)
        } else {
            accessibilityTraits.insert(.notEnabled)
        }
    }
```

harrieshin:
> > Looks fine for now, but I also want to know why `TableViewCell` is always a button anyway. Is that a property we should expose? Something feels off here.
> 
> We set it in `updateAccessibility` of TableViewCell.swift. I also agree that it's a bit broken, but that seemed like a much more in depth change that need more careful consideration than just matching the rest of our demos for now.
> 
> ```
>     private func updateAccessibility() {
>         accessibilityTraits.insert(.button)
> 
>         if isEnabled {
>             accessibilityTraits.remove(.notEnabled)
>         } else {
>             accessibilityTraits.insert(.notEnabled)
>         }
>     }
> ```

this is by design. most of our cell are used like a button in menu or file picker so it was a design request to make cell as button as default. but that shouldn't stop client team to remove .button trait on their own right? do we have to use a brand new cells?

harrieshin:
how does our cell look in darkmode then?

PR #1179, opened 2022-08-19
Fix CardNudge truncation in large text sizes

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

CardNudge was truncating text in large font scenarios for accessibility. This change adds Large Content Viewer support and prevents scaling of font to ensure all text is visible and accessible from LCV.

### Verification

Built and ran simulator, setting font size to large accessibility size to test LCV.

| Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-19 at 15 55 13 | Simulator Screen Recording - iPhone 13 Pro Max - 2022-08-19 at 15 54 15 |
| Simulator Screen Shot - iPhone 13 Pro Max - 2022-08-19 at 15 55 19 | Simulator Screen Recording - iPhone 13 Pro Max - 2022-08-19 at 15 54 48 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1179)

### DIFF
--- a/ios/FluentUI/Card Nudge/CardNudge.swift
+++ b/ios/FluentUI/Card Nudge/CardNudge.swift
@@ -84,6 +84,7 @@ public struct CardNudge: View, ConfigurableTokenizedControl {
                 .fontWeight(.medium)
                 .lineLimit(1)
                 .foregroundColor(Color(dynamicColor: tokens.textColor))
+                .showsLargeContentViewer(text: state.title)
 
             if hasSecondTextRow {
                 HStack(spacing: tokens.accentPadding) {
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -80,10 +81,10 @@ public struct CardNudge: View, ConfigurableTokenizedControl {
     var textContainer: some View {
         VStack(alignment: .leading, spacing: tokens.interTextVerticalPadding) {
             Text(state.title)
-                .font(.subheadline)
-                .fontWeight(.medium)
+                .font(.fluent(tokens.aliasTokens.typography[.body2Strong], shouldScale: false))
```
mischreiber, line 84: Can we move these tokens to a control-level token in `CardNudgeTokens`? I'd rather we avoid directly referencing global or alias tokens directly inside components.

edjamesmsft: Good call! I was following the previous implementation, but I agree, design tokens should be inside the tokens class. :)

huwilkes (approved):

PR #1181, opened 2022-08-22
Expose resetScrollPosition API on CommandBar

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In some use cases a consumer may want to reset the scroll position of the scrollable items. This can be achieved by calling `UIScrollView.scrollRectToVisible(x: Int, y: Int, width: Int, height: Int)`.  

A method, `resetScrollPosition`, is exposed on the `CommandBar` which calls `UIScrollView.scrollRectToVisible(...)` on the internal `UIScrollView`. The `UIScrollView` supports RTL by default so the `x` position does not need to be calculated. `0` works in all cases.

This new method is not called as part of the `itemGroups` `didSet` functionality because we do not want this to be the default. Consumers can call this as needed to suit their needs.

### Verification

- Build static library
- Updated and verified in `CommandBarDemoController`
- Verified behavior with devmain applications

https://user-images.githubusercontent.com/10938746/186038274-534df813-8db9-4424-8dd9-70879fb8e6f8.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1181)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -96,8 +96,9 @@ open class CommandBar: UIView {
     }
 
     /// Sets the scoll position  to the start of the scroll view
-    @objc public func resetScrollPosition() {
-        scrollView.scrollRectToVisible(CGRect(x: 0, y: 0, width: 1, height: 1), animated: false)
+    @objc public func resetScrollPosition(_ animated: Bool = false) {
+        /// A `CGRect` with a `width` and `height` both greater than `0` is required for the scrolling to occur
+        scrollView.scrollRectToVisible(CGRect(x: 0, y: 0, width: 1, height: 1), animated: animated)
     }
 
     // MARK: Overrides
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -95,6 +95,11 @@ open class CommandBar: UIView {
         trailingCommandGroupsView.updateButtonsState()
     }
 
+    /// Sets the scoll position  to the start of the scroll view
+    @objc public func resetScrollPosition() {
+        scrollView.scrollRectToVisible(CGRect(x: 0, y: 0, width: 1, height: 1), animated: false)
```
harrieshin, line 100: should we pass in the animated boolean?

anandrajeswaran, line 100: Does CGRectZero (.zero?) not work? If not, should at least add a comment why we need the magic 1x1 sized rect.

nodes11: `CGRectZero` does not work. The `width` and `height` need to greater than 0 in order the scrolling to occur.

nodes11: Updated.

harrieshin (approved):

PR #1185, opened 2022-08-24
[fluent2-tokens] Tokenize PopupMenu

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Update the PopupMenu to now use tokens. Since it's already a Drawer controller, this mostly involved just finding a way to expose a separate set of tokens.
Note: There seems to be a bug with the Drawer/PopupMenu where if you call `.replaceAllOverrides`, it will cause the theme of the controller to change, and lose the theme overrides. I'll create a work item to track this and solve it separately.

### Verification
Trying to set a `PopupMenuController`'s token set through `.tokenSet` will crash in debug but not release.
| PopupMenu_Tokens_Before | PopupMenu_Tokens_After |
| PopupMenu_Tokens_CustomColors_Before | PopupMenu_Tokens_CustomColors_After |
| PopupMenu_Tokens_Dark_Before | PopupMenu_Tokens_Dark_After |
| N/A | PopupMenu_Tokens_Overrides_After_Light |
| N/A | PopupMenu_Tokens_Overrides_After_Dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1185)


ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
```diff
@@ -206,4 +221,38 @@ class PopupMenuDemoController: DemoController {
...
+
+    private var themeWideOverridePopupMenuTokens: [PopupMenuTokenSet.Tokens: ControlTokenValue] {
+        return [
+            .drawerContentBackground: .dynamicColor { DynamicColor(light: GlobalTokens.init().sharedColors[.plum][.primary]) }
+        ]
+    }
+
+    private var perControlOverridePopupMenuTokens: [PopupMenuTokenSet.Tokens: ControlTokenValue] {
+        return [
+            .drawerContentBackground: .dynamicColor { DynamicColor(light: GlobalTokens.init().sharedColors[.red][.primary]) }
```
edjamesmsft, line 255: Just to anticipate a bug, did you double check that these two color overrides satisfy accessibility ratios in light and dark mode?

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -31,6 +31,18 @@ open class PopupMenuController: DrawerController {
...
     }
 
+    public typealias TokenSetKeyType = PopupMenuTokenSet.Tokens
+    public typealias TokenSetType = PopupMenuTokenSet
+    public override var tokenSet: DrawerTokenSet {
+        get {
+            return popupTokenSet
+        }
+        set {
+            preconditionFailure("PopupMenuController tokens must be set through popupTokenSet")
```
edjamesmsft, line 41: Do we need to stop execution in release builds here? I am leery of hard crashing unless we absolutely have to.

edjamesmsft (approved):

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -31,6 +31,18 @@ open class PopupMenuController: DrawerController {
         static let descriptionVerticalMargin: CGFloat = 12
     }
```
huwilkes, line 32: There were some talks about moving this sort of info out of token sets, so I just didn't touch these for now. I can update if needed, though.

harrieshin: what do you mean? we can't use the spacing token?

huwilkes: I think @mischreiber was saying we don't want to expose them anymore? I can probably update the code to just reference the values, and not add them to the token set.

mischreiber: This is good for now, but funnily enough I'll probably move these values back here once I merge https://github.com/microsoft/fluentui-apple/pull/1152 and global tokens become immutable static values again.

ios/FluentUI/Popup Menu/PopupMenuController.swift
mischreiber: Agreed -- this seems fine for a debug-only assert, but should probably resolve to a no-op in shipping code.

mischreiber (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
```diff
@@ -206,4 +221,38 @@ class PopupMenuDemoController: DemoController {
...
+
+    func isThemeWideOverrideApplied() -> Bool {
+        return self.view.window?.fluentTheme.tokens(for: PopupMenuTokenSet.self)?.isEmpty == false
+    }
+
+    // MARK: - Custom tokens
+
+    private var themeWideOverridePopupMenuTokens: [PopupMenuTokenSet.Tokens: ControlTokenValue] {
+        return [
+            .drawerContentBackground: .dynamicColor { DynamicColor(light: GlobalTokens.init().sharedColors[.plum][.primary]) }
```
laminesm, line 249: nit: can `GlobalTokens.init()` just be `GlobalTokens()`?

PR #1188, opened 2022-08-24
Creating `GradientInfo` and exposing as optional override on `FluentNotification`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change introduces a new struct, `GradientInfo`, to represent the building blocks of a gradient in Fluent. It comes with three properties: 

* `colors`: an array of `DynamicColor`  values to represent each color stop in the gradient.
* `startPoint`: a point representing the starting point of the gradient.
* `endPoint`: a point representing the ending point of the gradient.

As per Apple docs, `startPoint` and `endpoint` should both range from `(0.0, 0.0)` to `(1.0, 1.0)`, representing a percentage along both the `x` and `y` axes.

Additionally, this change adds an optional background gradient to `FluentNotification`, and maps its setting across SwiftUI and UIKit. Some notes here:

* This gradient is _not_ a Fluent design token--that system does not currently support gradients. Instead, this value is stored as a settable state property in UIKit and a view modifier in SwiftUI.
* I had to do some minor refactoring of notification presentation in SwiftUI to allow for arbitrary view modifiers to be expressed. Eventually I'd like most of the values currently passed into `FluentNotification.init()` to move to view modifiers, but for now I've left them as-is. If reviewers would like me to move more then I will gladly do so in a follow-up PR.

### Verification

New gradient works in light and dark modes, in both UIKit and SwiftUI.

Package size is _reduced_ by 2 kB, from 30,450 kB to 30,448 kB 🥳

| UIKit                                        | SwiftUI                                      |
| UIKit | SwiftUI |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1188)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -200,8 +200,10 @@ class NotificationViewDemoController: DemoController {
             notification.state.message = "The background of this notification has been customized with a gradient."
             notification.state.image = UIImage(named: "play-in-circle-24x24")
             // It's a lovely blue-to-pink gradient
-            let colors: [DynamicColor] = [DynamicColor(light: ColorValue(0xFFF1FA), dark: ColorValue(0x601044)),
-                                          DynamicColor(light: ColorValue(0xE0F6FF), dark: ColorValue(0x005A7F))]
+            let colors: [DynamicColor] = [DynamicColor(light: GlobalTokens().sharedColors[.pink][.tint50],
+                                                       dark: GlobalTokens().sharedColors[.pink][.shade40]),
+                                          DynamicColor(light: GlobalTokens().sharedColors[.cyan][.tint50],
+                                                       dark: GlobalTokens().sharedColors[.cyan][.shade40])]
             notification.state.backgroundGradient = GradientInfo(colors: colors,
                                                                  startPoint: .init(x: 0.0, y: 1.0),
                                                                  endPoint: .init(x: 1.0, y: 0.0))
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -200,10 +200,10 @@ class NotificationViewDemoController: DemoController {
             notification.state.message = "The background of this notification has been customized with a gradient."
             notification.state.image = UIImage(named: "play-in-circle-24x24")
             // It's a lovely blue-to-pink gradient
-            let colors: [DynamicColor] = [DynamicColor(light: GlobalTokens().sharedColors[.pink][.tint50],
-                                                       dark: GlobalTokens().sharedColors[.pink][.shade40]),
-                                          DynamicColor(light: GlobalTokens().sharedColors[.cyan][.tint50],
-                                                       dark: GlobalTokens().sharedColors[.cyan][.shade40])]
+            let colors: [DynamicColor] = [DynamicColor(light: GlobalTokens.sharedColors(.pink, .tint50),
+                                                       dark: GlobalTokens.sharedColors(.pink, .shade40)),
+                                          DynamicColor(light: GlobalTokens.sharedColors(.cyan, .tint50),
+                                                       dark: GlobalTokens.sharedColors(.cyan, .shade40))]
             notification.state.backgroundGradient = GradientInfo(colors: colors,
                                                                  startPoint: .init(x: 0.0, y: 1.0),
                                                                  endPoint: .init(x: 1.0, y: 0.0))
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -21,6 +21,7 @@ class NotificationViewDemoController: DemoController {
         case primaryToastWithStrikethroughAttribute
         case neutralBarWithFontAttribute
         case neutralToastWithOverriddenTokens
+        case neutralToastWithGradientBackground
         case warningToastWithFlexibleWidth
 
         var displayText: String {
@@ -51,6 +52,8 @@ class NotificationViewDemoController: DemoController {
                 return "Neutral Bar with Font Attribute"
             case .neutralToastWithOverriddenTokens:
                 return "Neutral Toast With Overridden Tokens"
+            case .neutralToastWithGradientBackground:
+                return "Neutral Toast With Gradient Background"
             case .warningToastWithFlexibleWidth:
                 return "Warning Toast With Flexible Width"
             }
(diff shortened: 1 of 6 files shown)


mischreiber:
Note: the `API break` tag only affects the SwiftUI API. Our UIKit API is unaffected.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -192,6 +195,21 @@ class NotificationViewDemoController: DemoController {
                 notification.hide()
             }
             return notification
+        case .neutralToastWithGradientBackground:
+            let notification = MSFNotification(style: .neutralToast)
+            notification.state.message = "The background of this notification has been customized with a gradient."
+            notification.state.image = UIImage(named: "play-in-circle-24x24")
+            // It's a lovely blue-to-pink gradient
+            let colors: [DynamicColor] = [DynamicColor(light: ColorValue(0xFFF1FA), dark: ColorValue(0x601044)),
+                                          DynamicColor(light: ColorValue(0xE0F6FF), dark: ColorValue(0x005A7F))]
```
huwilkes, line 204: Any chance we could get more readable colors?

huwilkes: Your comment _does_ say it's a lovely blue-to-pink, but if we're getting the colors from somewhere, I'm curious what those are.

ios/FluentUI/Core/Theme/Tokens/GradientInfo.swift
```diff
@@ -0,0 +1,45 @@
...
+import SwiftUI
+
+/// Represents a gradient as used by FluentUI.
+@objc public class GradientInfo: NSObject {
+    /// Initializes a gradient to be used in Fluent.
+    ///
+    /// - Parameters:
+    ///   - colors: The color of the shadow for shadow 1.
+    ///   - startPoint: The blur of the shadow for shadow 1.
+    ///   - endPoint: The horizontal offset of the shadow for shadow 1.
```
huwilkes, line 17: This comment looks suspiciously `ShadowInfo`y

ios/FluentUI/Notification/NotificationModifiers.swift
```diff
@@ -8,65 +8,28 @@ import SwiftUI
...
+    /// Specifies a custom background 
+    func backgroundGradient(_ gradientInfo: GradientInfo?) -> FluentNotification {
+        state.backgroundGradient = gradientInfo
+        return self
+    }
+
+    /// An optional gradient to use as the background of the notification.
+    ///
+    /// If this property is nil, then this notification will use the background color defined by its design tokens.
     func overrideTokens(_ tokens: NotificationTokens?) -> FluentNotification {
```
huwilkes, line 33: The comments on both modifiers are not quite right.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
```diff
@@ -175,8 +179,19 @@ struct NotificationDemoView: View {
...
-                             overrideTokens: $overrideTokens.wrappedValue ? NotificationOverrideTokens() : nil)
+                             showFromBottom: showFromBottom)
+            .backgroundGradient(showBackgroundGradient ? backgroundGradient : nil)
+            .overrideTokens($overrideTokens.wrappedValue ? NotificationOverrideTokens() : nil)
+        }
+    }
+
+    private var backgroundGradient: GradientInfo {
+        // It's a lovely blue-to-pink gradient
+        let colors: [DynamicColor] = [DynamicColor(light: ColorValue(0xFFF1FA), dark: ColorValue(0x601044)),
```
huwilkes, line 190: It's harder to change the text color for just if the gradient is on, so any chance we could switch to orange-to-pink or something for SwiftUI, for better contrast on the blue text?

huwilkes: Oh, or maybe just switch to grey-scale if the style uses brand colors, and the lovely blue-to-pink otherwise? Also, would it be possible to get an example of one of those with more than 2 colors?

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
mischreiber: I didn't do a full accessibility test, but to my eye the notification still looks pretty readable in both dark and light mode with these colors? What did you have in mind?

ios/FluentUI/Core/Theme/Tokens/GradientInfo.swift
mischreiber: Oh no, you've discovered my secret copy-and-paste approach to comment blocks! (will fix)

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
huwilkes: Ah, by "readable", I meant not so much hex. I think the contrast is generally fine. Though maybe the dark mode blue could be a bit darker to match the dark mode blue of the primary style notifications

mischreiber: Ahhh gotcha, sure!

ios/FluentUI/Notification/NotificationModifiers.swift
mischreiber: Copy-paste failure.

ios/FluentUI/Core/Theme/Tokens/GradientInfo.swift
```diff
@@ -0,0 +1,45 @@
...
+    public init(colors: [DynamicColor],
+                startPoint: CGPoint,
+                endPoint: CGPoint) {
+        self.colors = colors
+        self.startPoint = startPoint
+        self.endPoint = endPoint
+    }
+
+    /// The array of colors to apply to this linear gradient.
+    public let colors: [DynamicColor]
```
harrieshin, line 27: looking at CAGardientlayer documentation should we also have locations as optional thing?

harrieshin (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
mischreiber: I'd actually rather not try to do any "smarts" on our end about text color -- like all customization, this is something that consuming apps are responsible for ensuring is accessible.

huwilkes: That's fair. I also just meant in the demo, not as something the control would try to enforce.

ios/FluentUI/Core/Theme/Tokens/GradientInfo.swift
mischreiber: SwiftUI's `LinearGradient` _does_ require locations, so I'm inclined to side with the pickier of the two.

mischreiber: Oh I see what you mean -- adding it as an additional property. Yep okay I can rework that to be added -- the API for SwiftUI is a little different, but I can unify the two.

harrieshin: because this gradient struct needs to work for uikit stuff too so....

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
mischreiber: Using global colors now, with better dark mode contrast.

huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
mischreiber: While I agree with your point, I also think it's important to not "automatically" switch toggles here. I want people testing the app to see exactly what happens if they put a gradient background with the standard "primary" text -- and why they probably shouldn't do so, since we won't automatically change our style in the component itself once they're consuming it.

PR #1194, opened 2022-08-26
Update Avatar size ramp

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A new Avatar size 20 has been added and size 52 has been changed to size 56. All the values of the new size 20 Avatar were  validated by design. 
Changes were made to all Avatar demo controllers to accommodate this update.

### Verification

**Avatar demo**
| Size 56                                       | Size 20                                      |
| size56_swiftui_light | size20_swiftui_light |
| size56_swiftui_dark | size20_swiftui_dark |
| size56_light | size20_light |
| size56_dark | size20_dark |

**AvatarGroup demo**
| avatargroup_light | avatargroup_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1194)


ios/FluentUI/Avatar/AvatarTokens.swift
```diff
@@ -13,17 +13,19 @@ open class AvatarTokens: ControlTokens {
...
+            return 20
+        case .size24:
             return 24
-        case .medium:
+        case .size32:
             return 32
-        case .large:
+        case .size40:
             return 40
-        case .xlarge:
```
edjamesmsft, line 24: Do we know what impact this change is going to have for customers? It's different to adding just a new value. Will we need to add a size52 in the future? Do we know who's using this value now and how crucial it is to their layouts?

ios/FluentUI/PersonaButton/PersonaButtonTokens.swift
```diff
@@ -13,9 +13,9 @@ import UIKit
     var avatarSize: MSFAvatarSize {
         switch self {
         case .large:
-            return .xxlarge
+            return .size72
         case .small:
-            return .xlarge
+            return .size56
```
mischreiber, line 18: These `Avatar` instances still appear as 52pt in Figma. Has design confirmed that this component should be changing as well?

ios/FluentUI/Avatar/AvatarTokens.swift
laminesm: Good point - I didn't dig into this. All I know is design requested this change to align our avatar sizes and naming with web avatars. Maybe @harrieshin has more info about this?

ios/FluentUI/PersonaButton/PersonaButtonTokens.swift
laminesm: Right, the figmas don't seem to have been updated. I assume since design requested we change size 52 they should be aware any other control using 52pt avatar will change to 56pt avatar. I can check with design on Monday (since they're both oof) but this will probably not make it this fork then.

laminesm: Confirmed the changes with design.

mischreiber (approved):

harrieshin (approved):

PR #1195, opened 2022-08-26
Create function in ShadowInfo to apply chosen shadow to a specified view 

laminesm:
`###` Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

I've added the applyShadow function which applies one of the shadows in ShadowInfo to a specified view. 
- This change makes it easier for us and customers to apply shadows to views. I will make sure to use this function for all UIKit controls using shadows as part of the fluent 2 colors effort. 
- The blur values of our shadow tokens need to be halved in order to display shadows properly. This wasn't public knowledge and adding this function will reduce the risk of misuse of these values for us and our customers.

### Verification

Used the new function to apply shadow 28 to large cards in the demo app.

| Shadow 1                                       | Shadow 2                                      |
| shadow1 | shadow2 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1195)

### DIFF
--- a/ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
@@ -4,6 +4,7 @@
 //
 
 import CoreGraphics
+import UIKit
 
 /// Represents a two-part shadow as used by FluentUI.
 public struct ShadowInfo {
@@ -59,4 +60,13 @@ public struct ShadowInfo {
 
     /// The vertical offset of the shadow for shadow 2.
     public let yTwo: CGFloat
+
+    /// Applies the chosen shadow to a specified view
+    public func applyShadow(for view: UIView, isShadowOne: Bool) {
+        view.layer.shadowColor = isShadowOne ? UIColor(dynamicColor: colorOne).cgColor : UIColor(dynamicColor: colorTwo).cgColor
+        view.layer.shadowRadius = isShadowOne ? (blurOne / 2) : (blurTwo / 2)
+        view.layer.shadowOpacity = 1
+        view.layer.shadowOffset = CGSize(width: isShadowOne ? xOne : xTwo,
+                                         height: isShadowOne ? yOne : yTwo)
+    }
 }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,13 @@ public struct ShadowInfo {
 
     /// The vertical offset of the shadow for shadow 2.
     public let yTwo: CGFloat
+
+    /// Applies the chosen shadow to a specified view
+    public func applyShadow(for view: UIView, isShadowOne: Bool) {
+        view.layer.shadowColor = isShadowOne ? UIColor(dynamicColor: colorOne).cgColor : UIColor(dynamicColor: colorTwo).cgColor
+        view.layer.shadowRadius = isShadowOne ? (blurOne / 2) : (blurTwo / 2)
```
harrieshin, line 67: we should add comments on this?

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,13 @@ public struct ShadowInfo {
 
     /// The vertical offset of the shadow for shadow 2.
     public let yTwo: CGFloat
+
+    /// Applies the chosen shadow to a specified view
+    public func applyShadow(for view: UIView, isShadowOne: Bool) {
```
huwilkes, line 65: Why only apply one shadow? If our designs have two shadows, shouldn't we just apply both here, rather than needing to call this twice?

laminesm: I don't think we can apply both shadows like that. Essentially the layer will just use the latest set values. Also I believe for most of our shadows we don't use both. It's usually either the elevation shadow or the border one.

huwilkes: If most of our shadows don't use both shadows, then `ShadowInfo` shouldn't include both?

laminesm: Talked offline and it seems both shadows need to be applied for the controls. I will work on it.

Saadnajmi: +1 on applying both shadows. AFAIK, it's very intentional in design to _most of the time_ want both

PR #1197, opened 2022-08-27
Fix Voice Over accessibility for title in TableViewHeaderFooterView w…

isatoderici:
…hen presented in popover

### Platforms Impacted

- [x] iOS

### Description of changes

ISSUE

- When Voice Over accessibility is enabled, the TableViewHeaderFooterView title
has a wrong selection when presented in a popover on iPad.

- The title in this header/footer is a UITextView and in the existing implementation we override the default accessibility behavior by making the UITextView an accessibility element.
This implementation leads to a wrong text selection for title in popover on iPad.

FIX

- The fix is to remove the accessibility element assignment from UITextView and keep the accessibility trait insertion as .header when TableViewHeaderFooterView is used as header and remove the accessibility trait insertion when the view is used as footer. This implementation already exists but it was followed by a 'Bug in iOS - ...' implementation where  accessibility element for UITextView was set up to 'false' and then back to 'true'. That implementation broke the VO selection for multi-line of text title for both header and footer scenarios. Removing that logic fixed the issue properly.

A NICE TO HAVE
- Replace the UITextView with a UILabel for the title. That would imply custom implementation for hyper-texts and would impact the TableViewHeaderFooterView interface signature because we would have to mention the clickable portion for the label text in order to specify it in the attributed string range and then would be assigned to the label. (Maybe that can be done later.)

### Verification

- Verified the change for Voice Over accessibility in table header and footer in popovers or in presented controllers.
- Verified that the Voice Over selection works well for header and footer when there is a single line title and multiple lines title only; multiple lines and single line title with accessory view.
- Verified that Voice Over accessibility works well in portrait/landscape and split view modes for iPad and iPhone.

| Before                                                                  | After                                                                  |
| https://user-images.githubusercontent.com/65185855/187558496-8ed11cb6-871b-4941-af84-bd151e7f30a4.mov | https://user-images.githubusercontent.com/65185855/187559563-a43665bd-f8ff-4c52-8997-13f26ccd278b.mov |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1197)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -365,15 +365,6 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
     ///   - leadingView: An optional custom view that appears near the leading edge of the view.
     private func setup(style: Style, accessoryButtonTitle: String, leadingView: UIView? = nil) {
         updateTitleViewFont()
-        switch style {
-        case .header, .divider, .dividerHighlighted, .headerPrimary:
-            titleView.accessibilityTraits.insert(.header)
-        case .footer:
-            titleView.accessibilityTraits.remove(.header)
-            // Bug in iOS - need to manually refresh VoiceOver text for accessibilityTraits
-            titleView.isAccessibilityElement = false
-            titleView.isAccessibilityElement = true
-        }
 
         accessoryButton = !accessoryButtonTitle.isEmpty ? createAccessoryButton(withTitle: accessoryButtonTitle) : nil
         self.leadingView = leadingView
@@ -543,7 +534,6 @@ private class TableViewHeaderFooterTitleView: UITextView {
         isEditable = false
         isScrollEnabled = false
         clipsToBounds = false    // to avoid clipping of "deep-touch" UI for links
-        isAccessibilityElement = true
         self.textContainer.lineBreakMode = .byTruncatingTail
         self.textContainer.lineFragmentPadding = 0
         textContainerInset = .zero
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -365,15 +365,6 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView {
     ///   - leadingView: An optional custom view that appears near the leading edge of the view.
     private func setup(style: Style, accessoryButtonTitle: String, leadingView: UIView? = nil) {
         updateTitleViewFont()
-        switch style {
-        case .header, .divider, .dividerHighlighted, .headerPrimary:
```
harrieshin, line 369: this is quite a bit of behavior regression that we can't use voiceover rotor to navigate around header traits.

anandrajeswaran: +1 - can we keep the header traits but get rid of the isAccessibilityElement? Worst case that is a clear improvement and makes the VoiceOver bug much more rare. Best case the new bug is not a real scenario in any clients consuming this control

anandrajeswaran:
Fix looks good to me, but probably should make sure @harrieshin is explicitly good with it and probably update the PR notes to make clear which of the proposed fixes we went with

harrieshin:
would you please update your description with video include with the voiceover captions specially around the footer with link to make sure it isn't regressed?

sophialee0416:
Might also want to update the table markdown in your description so that the side-to-side comparison is easier to look at

harrieshin (approved):

PR #1203, opened 2022-08-30
Align Notification Border with Figma

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Changed subtle bar to have outline only on top/bottom instead of all 4 sides to match Figma
- Changed border color to match Figma
- Fixed issue with showFromBottom toggle not working

| Simulator Screen Recording - iPad Pro (12 9-inch) (5th generation) - 2022-08-30 at 13 04 38 | Simulator Screen Recording - iPad Pro (12 9-inch) (5th generation) - 2022-09-01 at 13 17 41 |
| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1203)


ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -57,22 +57,22 @@ import UIKit
 /// Design token set for the `Notification` control.
 open class NotificationTokens: ControlTokens {
     /// Defines the style of the notification.
-    public internal(set) var style: MSFNotificationStyle = .primaryToast
+    public internal(set) var style: MSFNotificationStyle = .accentToast
```
huwilkes, line 60: We may want to keep the name how it was, and then change it with the Fluent 2 Colors branch merge, rather than breaking the Notification style api.

ios/FluentUI/Notification/NotificationTokens.swift
```diff
@@ -175,15 +175,15 @@ open class NotificationTokens: ControlTokens {
...
-        case .primaryOutlineBar:
-            return aliasTokens.strokeColors[.neutral1]
+        case .subtleBar:
+            return aliasTokens.strokeColors[.neutral2]
         }
     }
 
     /// The width of the outline around the frame of a notification
-    open var outlineWidth: CGFloat { GlobalTokens.borderSize(.thin) }
+    open var outlineWidth: CGFloat { 0.5 }
```
harrieshin, line 186: why we not using border tokens?

harrieshin: do we not have half the point?

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -467,7 +469,7 @@ class MSFNotificationStateImpl: NSObject, ControlConfiguration, MSFNotificationS
          actionButtonAction: (() -> Void)? = nil,
          showDefaultDismissActionButton: Bool? = nil,
          messageButtonAction: (() -> Void)? = nil,
-         showFromBottom: Bool = true) {
+         showFromBottom: Bool? = nil) {
```
harrieshin, line 472: why we cahnging this?

ios/FluentUI/Notification/NotificationTokens.swift
huwilkes: Looks like `.thin` is supposed to be 1 pixel on iOS, so it _should_ be right, but we have 1 set as the value, so we're actually returning 1 point. We should probably update our borderSize tokens to return something * screen pixel

ios/FluentUI/Notification/FluentNotification.swift
jeaniehuynh: was fixing an issue where SwiftUI's showFromBottom was hardcoded as "true" and went ahead and aligned the parameter with the other parameters in the inits

ios/FluentUI/Notification/NotificationTokens.swift
jeaniehuynh: 0.5 is not in our border tokens and the figma shows the border width as being 0.5:
image

harrieshin: seems like something we need to follow up with design?

huwilkes (approved):

ios/FluentUI/Notification/FluentNotification.swift
mischreiber: I kinda want to push back on this: I don't think optional Bool values are generally a good design pattern, as it's ambiguous what the default behavior will be. Keeping it non-optional and just setting the default value to `false` seems easier to understand, unless there's a convention/pattern out there that I'm not familiar with.

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -470,7 +472,7 @@ class MSFNotificationStateImpl: ControlState, MSFNotificationState {
         self.actionButtonTitle = actionButtonTitle
         self.actionButtonAction = actionButtonAction
         self.messageButtonAction = messageButtonAction
-        self.showFromBottom = showFromBottom
+        self.showFromBottom = showFromBottom ?? true
```
mischreiber, line 475: Isn't this logic the same as defaulting to `true`?

PR #1204, opened 2022-08-31
introduced a Medium Fluent Button Size

amoggo:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
We are introducing a new Button Look that has specific height, spacing and corners. The current Button size heights available are large (28pt) and small (20pt). We want a 24pt height so a "medium" size is being added in this PR. Additionally, it was discovered that the `verticalPadding` constraint in `ButtonSizeParameters` didn't inform the absolute height of the button ie. the Button height grew/shrunk to accommodate the image size. To address the spec which required a 24pt height when a 20x20 image is used, a `minButtonHeight` param is introduced.

Button.swift - Introduce the new "Medium" size
TestButtonViewController.swift - Add the new size variant to the test sheet showing all combinations of buttons available

### Verification
- Standard Build + Test app running
- Switching the Button Pane in the test app to select available options under "Image Position", Button State, Width and Height, to ensure proper functioning. 

| Screen Shot 2022-08-30 at 18 48 29 | Screen Shot 2022-09-08 at 16 33 08|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1204)


macos/FluentUITestViewControllers/TestButtonViewController.swift
```diff
@@ -195,6 +203,10 @@ class TestButtonViewController: NSViewController, NSMenuDelegate {
 			NSTextField(labelWithString: "Large Secondary"),
 			NSTextField(labelWithString: "Large Acrylic"),
 			NSTextField(labelWithString: "Large Borderless"),
+			NSTextField(labelWithString: "Medium (tighter-spacing) Primary"),
```
anandrajeswaran, line 206: what does `tighter-spacing` mean to convey? Since all Medium's have it can we just omit it for brevity?

macos/FluentUI/Button/Button.swift
```diff
@@ -597,6 +598,16 @@ private struct ButtonSizeParameters {
 		titleToImageSpacing: 10,
 		titleToImageVerticalSpacingAdjustment: 7
 	)
+	
+	static let medium = ButtonSizeParameters(
+		fontSize: 13,  // line height: 17
+		cornerRadius: 6,
+		verticalPadding: 2.0,  // overall height: 24
```
brentpe, line 605: Somethings not right. These measured perfectly on my local build in the OfficeSpaceTestApp, but in the FluentUITest app the size is actually 20pts high (same as the small size). Need to investigate why that might be the case.

macos/FluentUITestViewControllers/TestButtonViewController.swift
brentpe: Yeah. It was thought that the tighter spacing was making the buttons look no larger than the "small" size, but that's not the case as the test app is showing them smaller for some reason (see above comment). So once that's resolved we can remove the "tighter-spacing" label.

macos/FluentUI/Button/Button.swift
brentpe: Looks like it's the images being used that are pushing the size from 20 to 24pt.

brentpe: Looks like we just need to adjust the "verticalPadding" from 2.0 to 4.0.

macos/FluentUI/Button/Button.swift
```diff
@@ -524,6 +524,7 @@ class ButtonCell: NSButtonCell {
 @objc(MSFButtonSize)
 public enum ButtonSize: Int, CaseIterable {
 	case large
+	case medium
```
edjamesmsft, line 527: If we have large, medium, small, would it be worth it to consider changing to an objective size name for the enums, like we did for Avatar sizes? e.g., instead of medium, it would be size24?

edjamesmsft: (I recognize this would be a breaking change, but may be better to do it sooner than later)

macos/FluentUI/Button/Button.swift
brentpe: Hmm, if you make the above change, then the buttons in the OfficeSpaceTestApp are 28pts tall.

macos/FluentUI/Button/Button.swift
harrieshin: i think we should hold off on public enum change until we have a solid fluent 2 mac design toolkit.

macos/FluentUI/Button/Button.swift
brentpe: Button height grows to fit the images but also decreases when smaller images are used, so we may want to add a minimum height constraint.

macos/FluentUI/Button/Button.swift
```diff
@@ -524,6 +524,7 @@ class ButtonCell: NSButtonCell {
 @objc(MSFButtonSize)
```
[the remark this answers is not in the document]
amoggo: Based on the current ButtonSizeParams logic, the Padding portions that account for height and width of the button are not absolute in that the dimensions of the button are auto-scaled to accommodate the size of the Image with it. Additionally documenting a height and width would only be duplicating the numbers already noted in the static variable assignments in ButtonSizeParamaters - so it would require a 2-place maintenance. So I'm questioning if it might be worth it as its straightforward to simply look closer at the sizes to gauge the dimensions?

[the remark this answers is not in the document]
harrieshin: we document all of our public variables, enums, function. think of all Apple appkit documentation is standard.

macos/FluentUI/Button/Button.swift
```diff
@@ -587,6 +602,7 @@ private struct ButtonSizeParameters {
 	fileprivate let titleVerticalPositionAdjustment: CGFloat
 	fileprivate let titleToImageSpacing: CGFloat
 	fileprivate let titleToImageVerticalSpacingAdjustment: CGFloat
+	fileprivate var minButtonHeight: CGFloat?
```
harrieshin, line 605: might as well not make it options. based on the comment of the code, large min height is 28 and small is 20?

macos/FluentUI/Button/Button.swift
```diff
@@ -368,6 +371,17 @@ open class Button: NSButton {
 			needsDisplay = true
 		}
 	}
+
+	open override var intrinsicContentSize: CGSize {
+		let superSize = super.intrinsicContentSize
+		let height = { () -> CGFloat in
+			if let minButtonHeight = self.minButtonHeight {
+				return superSize.height <= minButtonHeight ? minButtonHeight : superSize.height
```
brentpe, line 379: Nit: Any reason you don't just use "<" vs. "<=" here?

macos/FluentUI/Button/Button.swift
```diff
@@ -598,15 +613,16 @@ private struct ButtonSizeParameters {
...
 		fontSize: 13,  // line height: 17
 		cornerRadius: 6,
-		verticalPadding: 2.0,  // overall height: 24
+		verticalPadding: 2.0, // overall height: 24
 		horizontalPadding: 5,
 		titleVerticalPositionAdjustment: 0,
 		titleToImageSpacing: 3,
-		titleToImageVerticalSpacingAdjustment: 7
+		titleToImageVerticalSpacingAdjustment: 7,
+		minButtonHeight: 24
```
brentpe, line 625: Yeah, per the other comment also adding minButtonHeight values for small and large seems worth it as the risk of those buttons changing a few pixels taller would be small and arguably by design since the sizes are already mentioned in the structs.

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
brentpe: Maybe just add a quick comment for each size indicating the "minimum height is X, but the button will grow in height if needed to fit its image".

macos/FluentUI/Button/Button.swift
amoggo: It's the same thing?

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
amoggo: Done. I've documented the button heights because they now have a minimum constant.

macos/FluentUI/Button/Button.swift
amoggo: One thing to call out. I've adjusted the min button heights as 28 and 20 for large and small respectively. However I noticed that for retina screen, the small button is 38 pts high (so amounting to 19 instead of 20) while for non-retina it would be as expected viz.20

Now with the minButtonHeight added as 20, it will be 20 in both cases.

macos/FluentUI/Button/Button.swift
brentpe: > It's the same thing?

Yeah, it does produce the same result. Seems minor, but to me I end up having to do the extra "greater than **or equal to**" step whereas just using > seems more exacting.

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
brentpe: Seems fine to me as-is, but perhaps it would be more accurate to say "28 pts minimum" or would that cause confusion?

brentpe (approved):

macos/FluentUI/Button/Button.swift
```diff
@@ -523,7 +537,14 @@ class ButtonCell: NSButtonCell {
 /// Indicates the size of the button
 @objc(MSFButtonSize)
 public enum ButtonSize: Int, CaseIterable {
+	
+	/// Button Height - 28 pts
```
brentpe, line 541: Nit: "/// Button Height - minimum 28 pts" etc., to be more accurate?

macos/FluentUI/Button/Button.swift
```diff
@@ -342,10 +342,13 @@ open class Button: NSButton {
 
 	private static let borderWidth: CGFloat = 1
 
+	private var minButtonHeight: CGFloat?
```
harrieshin, line 345: why is this variable needed?

harrieshin: it also doesn't need to be optional any more?

harrieshin (approved):

PR #1207, opened 2022-08-31
Bottom sheet collapsed height resolver API

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

It's a common scenario for clients to dynamically adjust the bottom sheet collapsed height. Typically these decisions depend on:
- the current trait collection (vertical size class)
- available height

Doing this well can have caveats, like ensuring the height change is properly timed and animated. We can make this easier for clients by providing a height resolver API. The clients just provide a closure that gets the necessary context as input (max height and current trait collection) and returns the desired height given the context. The bottom sheet ensures the closure is called at the right times.
This way clients can ensure the sheet is sized properly without having to hook up to lifecycle events. This pattern is similar to Apple's detent height resolver - https://developer.apple.com/documentation/uikit/uisheetpresentationcontroller/detent/3976719-custom

#### Implementation details
The collapsed sheet height is in a hot code path that gets called a lot during layout and animations, so we cache the resolved height values to ensure we only call the client when necessary.

I split out the `ContentHeightResolutionContext` away from the bottom sheet to make it a bit more generic. There's no need to tie this pattern to the bottom sheet specifically.

### Verification

Test app sanity on trait collection changes.

Here's the demo app behavior on iPhone. Trait collection changes when screen is rotated, so a new height gets resolved from the client (demo controller is the client in this case):

https://user-images.githubusercontent.com/3610850/190026787-594f5549-8569-469b-8885-0e562b728a5c.mov

On iPad the sheet configured with the same resolver maintains the same height in all orientations, because the trait collection stays the same:

https://user-images.githubusercontent.com/3610850/190026862-ebec7940-ef9a-4484-9258-f24e06bf2a9c.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1207)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -843,6 +875,24 @@ public class BottomSheetController: UIViewController {
...
     }
 
+    // Output of `collapsedHeightResolver` wrapped in a cache.
+    private var resolvedCollapsedSheetHeight: CGFloat {
+        let oldContext = lastCollapsedSheetHeightResolutionContext
+        let newContext = ContentHeightResolutionContext(maximumHeight: maxSheetHeight - view.safeAreaInsets.bottom , containerTraitCollection: view.traitCollection)
+
+        if oldContext?.maximumHeight != newContext.maximumHeight || !(oldContext?.containerTraitCollection.containsTraits(in: newContext.containerTraitCollection) ?? false) {
+            lastResolvedCollapsedSheetHeight = collapsedHeightResolver?(newContext) ?? 0
+            lastCollapsedSheetHeightResolutionContext = newContext
```
nodes11, line 885: I think this is just my misunderstanding, but when when will `lastCollapsedSheetHeightResolutionContext ` be initialized for the first time? It looks like it would happen here, but in order for it to get here its value would need to be non-nil so that it passes the optional check on line 883.

lcapkovic: When it's `nil` initially, the equality heck with newContext will fail, so we will go in this branch and set it to newContext. So in other words a nil context is "dirty" and will cause a call to the resolver.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -843,6 +875,24 @@ public class BottomSheetController: UIViewController {
         return maxHeight
     }
 
+    // Output of `collapsedHeightResolver` wrapped in a cache.
```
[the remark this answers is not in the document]
lcapkovic: Typically I've been using // for private stuff and /// for public comments. I'm not sure if that's any good, your suggestion definitely makes sense. But I'll probably keep it this way as I'd have to change all the comments and that would make the PR huge.

nodes11 (approved):

ios/FluentUI/Utilities/ContentHeightResolutionContext.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFContentHeightResolutionContext)
+public class ContentHeightResolutionContext: NSObject {
```
harrieshin, line 9: bit unclear to me why this won't be a simple struct and it can't live within bottomsheet folder because i don't know what the usage would be outside of this context

lcapkovic: Why not a simple struct - I wish I could use that, but unfortunately those don't work with @objc.

Why not bottom sheet specific:
One reason is that none of the height resolver pattern is bottom sheet specific, it could be used in other classes. 

Another, more valid reason is that I want to use this internally for presentation configuration, decoupled from the actual presentation. This way app clients that want to present something can provide a height behavior to Silhouette without necessarily knowing about (or committing to) a bottom sheet controller being used behind the scenes to do the presentation.

lcapkovic: If you think it would be cleaner to have this under the bottom sheet folder, I'd be happy to do that.

lcapkovic: Wait, I think I could make this swift-only, which would let me keep it as a struct. That's probably a greater benefit than the above reusability stuff...

lcapkovic: Moved back under bottom sheet folder. Kept an objc compatible class after a suggestion from Mike.

ios/FluentUI/Bottom Sheet/ContentHeightResolutionContext.swift
```diff
@@ -0,0 +1,20 @@
...
+
+@objc(MSFContentHeightResolutionContext)
+public class ContentHeightResolutionContext: NSObject {
+    init(maximumHeight: CGFloat, containerTraitCollection: UITraitCollection) {
+        self.maximumHeight = maximumHeight
+        self.containerTraitCollection = containerTraitCollection
+    }
+
+    /// Maximum height of the content.
+    @objc public let maximumHeight: CGFloat
```
mischreiber, line 16: Do these values need to be `@objc` and `public`? Do we expect them to be read outside Fluent, and by Objective-C clients?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -843,6 +875,24 @@ public class BottomSheetController: UIViewController {
...
+    private var resolvedCollapsedSheetHeight: CGFloat {
+        let oldContext = lastCollapsedSheetHeightResolutionContext
```
mischreiber, line 880: Probably could do with a 

```swift
guard let oldContext = lastCollapsedSheetHeightResolutionContext else {
    return lastResolvedCollapsedSheetHeight
}
```

Then the logic below would simplify a fair bit.

lcapkovic: That seems like it's different logic? If there's no previously used context, then last resolved height will have no meaning, so we need to go to the resolver to get a fresh value.

ios/FluentUI/Bottom Sheet/ContentHeightResolutionContext.swift
lcapkovic: Yes to both. Clients will be doing the height resolving, and many of them are objc (including the use I'm working on).

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
```diff
@@ -24,8 +24,10 @@ class BottomSheetDemoController: UIViewController {
         let bottomSheetViewController = BottomSheetController(headerContentView: headerView, expandedContentView: expandedContentView)
         bottomSheetViewController.hostedScrollView = personaListView
         bottomSheetViewController.headerContentHeight = BottomSheetDemoController.headerHeight
-        bottomSheetViewController.collapsedContentHeight = 70
         bottomSheetViewController.delegate = self
+        bottomSheetViewController.collapsedHeightResolver = { context in
+            return context.containerTraitCollection.verticalSizeClass == .regular ? 100 : 70
```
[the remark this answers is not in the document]
lcapkovic: Will take on this in a separate PR.

ios/FluentUI/Bottom Sheet/ContentHeightResolutionContext.swift
```diff
@@ -0,0 +1,20 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc(MSFContentHeightResolutionContext)
+public class ContentHeightResolutionContext: NSObject {
```
harrieshin, line 9: add documentation on this public class please?

harrieshin:
test on iPad as well please? attach video of the usuage?

harrieshin (approved):

lcapkovic:
> test on iPad as well please? attach video of the usuage?

Tested on iPad + attached vids of the demo on both iPhone and iPad.

ios/FluentUI/Bottom Sheet/ContentHeightResolutionContext.swift
lcapkovic: Done

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -843,6 +875,24 @@ public class BottomSheetController: UIViewController {
...
+        let newContext = ContentHeightResolutionContext(maximumHeight: maxSheetHeight - view.safeAreaInsets.bottom, containerTraitCollection: view.traitCollection)
+
+        if oldContext?.maximumHeight != newContext.maximumHeight || !(oldContext?.containerTraitCollection.containsTraits(in: newContext.containerTraitCollection) ?? false) {
```
anandrajeswaran, line 883: Maybe I'm just weak, but optional chaining combined with a boolean operator (`||`) really confuses me. This seems like a good case for a
```
if let oldContext = oldContext,
oldContext.maximumHeight != newContext.maximumHeight || !oldContext.containerTraitCollection.containsTraits(in: newContext.containerTraitCollection) {
}
```

PR #1209, opened 2022-09-01
[fluent2 tokens] Fix Drawer background color logic interaction with tokens and add Appearance menu to demo

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Removed our logic for tracking if the user has set a custom background color, and will now just return the correct token on get and set all three tokens on set.

Also, added the appearance editor to the Drawer demo.

### Verification

Drawer:
DrawerColorLogic_Drawer

PopupMenu:
DrawerColorLogic_PopupMenu

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1209)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -765,6 +765,7 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
             } else {
                 backgroundColor = UIColor(dynamicColor: tokenSet[.drawerContentBackground].dynamicColor)
             }
+            useCustomBackgroundColor = false
```
mischreiber, line 768: I know why, but it's very odd to be setting `useCustomBackgroundColor = false` in a block that is already `if !useBackgroundColor`. This tells me that there are unintended side effects to setting the background color, which strikes me as fragile. Is there any other way we can factor this code to be less fragile?

If not, I think this at least warrants a comment explaining why it needs to be set here.

huwilkes: Talked a bit offline, came up with a good solution. Just removed useCustomBackgroundColor since it's private and moved all the logic to get the different background colors to the getter of backgroundColor.

mischreiber (approved):

jeaniehuynh (approved):

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -111,10 +111,25 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
...
+            }
+            return UIColor(dynamicColor: color)
+        }
+        set {
+            guard let newColor = newValue.dynamicColor else {
+                return
+            }
+            tokenSet[.popoverContentBackground] = .dynamicColor { newColor }
+            tokenSet[.navigationBarBackground] = .dynamicColor { newColor }
+            tokenSet[.drawerContentBackground] = .dynamicColor { newColor }
```
edjamesmsft, line 132: At this point, what's the purpose of getting these values independently? If they're all getting set to the same color, I mean

huwilkes: They have different default values, and also if the client goes through tokens to override them they could just override a single one. But, if the client does `drawer.background = customColor`, we'll just set all the different backgrounds, to match the old behavior

PR #1210, opened 2022-09-01
Implement functionality to show a custom view within a CommandBarItem

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In some cases, users may want to display a custom view for a control inside of the `CommandBar`. This change aims to address that. 

Two new properties are exposed on the public `CommandBarItem` interface to support this:
1. `@objc public var customControlView: (() -> UIView)?` -- This closure allows consumers to provide a block which returns a `UIView` when called. Internally, this closure is called when creating the buttons for the `CommandBar` view and the view returned is constrained to the edge of the buttons.

When a `customControlView` is provided by the consumer, the `isAccessibilityElement` property is set to `false` for the `CommandBarButton` object so that the custom view can provide itself or its subviews as the accessibility elements. This is important for voice over mode otherwise users cannot navigate through the contents of the provided view.

### Verification
Screen Shot 2022-08-31 at 8 17 24 PM

- Verified in devmain apps with a real-world use case
- Added usage to the demo controller and verified in the demo controller

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1210)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBarButton.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButton.swift
@@ -39,23 +39,29 @@ class CommandBarButton: UIButton {
 
         translatesAutoresizingMaskIntoConstraints = false
 
-        if #available(iOS 15.0, *) {
-            var buttonConfiguration = UIButton.Configuration.plain()
-            buttonConfiguration.image = item.iconImage
-            buttonConfiguration.contentInsets = LayoutConstants.contentInsets
-            buttonConfiguration.background.cornerRadius = 0
-            configuration = buttonConfiguration
+        if let makeCustomButtonView = item.customControlView {
+            addCustomView(makeCustomButtonView())
+            /// Disable accessiblity for the button so that the custom view can provide itself or its subviews as the accessilbity element(s)
+            isAccessibilityElement = false
         } else {
-            setImage(item.iconImage, for: .normal)
-            contentEdgeInsets = LayoutConstants.contentEdgeInsets
+            if #available(iOS 15.0, *) {
+                var buttonConfiguration = UIButton.Configuration.plain()
+                buttonConfiguration.image = item.iconImage
+                buttonConfiguration.contentInsets = LayoutConstants.contentInsets
+                buttonConfiguration.background.cornerRadius = 0
+                configuration = buttonConfiguration
+            } else {
+                setImage(item.iconImage, for: .normal)
+                contentEdgeInsets = LayoutConstants.contentEdgeInsets
+            }
+
+            let accessibilityDescription = item.accessibilityLabel
+            accessibilityLabel = (accessibilityDescription != nil) ? accessibilityDescription : item.title
+            accessibilityHint = item.accessibilityHint
+
+            menu = item.menu
+            showsMenuAsPrimaryAction = item.showsMenuAsPrimaryAction
         }
-
-        let accessibilityDescription = item.accessibilityLabel
-        accessibilityLabel = (accessibilityDescription != nil) ? accessibilityDescription : item.title
-        accessibilityHint = item.accessibilityHint
-
-        menu = item.menu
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -70,28 +76,31 @@ class CommandBarButton: UIButton {
         isSelected = isPersistSelection && item.isSelected
         isHidden = item.isHidden
 
-        // always update icon and title as we only display one; we may alterenate between them, and the icon may also change
-        let iconImage = item.iconImage
-        let title = item.title
-        let accessibilityDescription = item.accessibilityLabel
+        if item.customControlView == nil {
```
harrieshin, line 79: this type of nil check feels not Swift

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -108,6 +108,12 @@ open class CommandBarItem: NSObject {
     /// Set `isSelected` to desired value in this handler. Default implementation is toggling `isSelected` property.
     @objc public var itemTappedHandler: ItemTappedHandler
 
+    /// If set, the `UIView` returned from the closure will be shown inside `CommandBarButton` instead of the `title` and `image`
+    @objc public var customControlView: (() -> UIView)?
```
harrieshin, line 112: how do we handle ipad pointer interaction on this ?

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -103,8 +103,10 @@ class CommandBarCommandGroupsView: UIView {
     }
 
     @objc private func handleCommandButtonTapped(_ sender: CommandBarButton) {
-        sender.item.handleTapped(sender)
-        sender.updateState()
+        if !sender.item.ignoresItemTappedHandler {
```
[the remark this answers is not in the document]
nodes11: My initial thought was to provide additional flexibility to the client if was providing a view that didn't by default have a gesture recognizer. However, after some more consideration, I think I agree that no putting the selector on the custom view and forcing the client to handle the gestures would be more straightforward.

ios/FluentUI/Command Bar/CommandBarItem.swift
nodes11: Currently the `CommandBarButton` does not support any additional pointer interaction other than the default on-press highlighting. Since this is a custom view, I don't think there should be any default behavior provided by our the core component. Any pointer interactions should be defined by the view provider.

harrieshin (approved):

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -141,6 +153,17 @@ class CommandBarButton: UIButton {
...
 
+    private func addCustomView(_ view: UIView) {
+        view.translatesAutoresizingMaskIntoConstraints = false
+        addSubview(view)
+
+        /// Constrain view to edges of the button
+        view.leadingAnchor.constraint(equalTo: leadingAnchor).isActive = true
+        view.trailingAnchor.constraint(equalTo: trailingAnchor).isActive = true
+        view.topAnchor.constraint(equalTo: topAnchor).isActive = true
+        view.bottomAnchor.constraint(equalTo: bottomAnchor).isActive = true
```
anandrajeswaran, line 164: Every time any constraint is activated, Apple's layout engine does some computation. This is a small amount of activations, but in practice we've seen bugs before where we've activated constraints in a loop and it was extremely slow. We were able to make non-trivial perf gains by creating the constraints in the loop and then making one call to activateConstraints after.

Again, it's only 4 so the practical impact is probably not game changing, but as best practice we should activate these all at once using NSLayoutConstraint.activateConstraints([...])

PR #1206, opened 2022-08-31
Merging `ControlTokenSet` to main branch

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change represents a large merging of code from `fluent2-tokens` branch to support the new `ControlTokenSet` system of reading and customizing token properties.

No functional changes with this PR; this simply brings the `ControlTokenSet` format that has already been in use in `fluent2-tokens` over to `main`. For more information, check out #1146 and its attached list of PRs.

Note: this is not a full merge of `fluent2-tokens` into `main`; it's a very selective cherry-picking of changes. Thanks to that, there isn't a "real" merge commit associated with this PR.

### Verification

Existing Fluent 2 components continue to behave as expected, with no impact to customization.

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1206)

PR #1212, opened 2022-09-01
Restore lost code in separator merge

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Merge #1206 broke separator layout in `TableViewCell`. This restores the missing code.

### Verification

| Simulator Screen Shot - iPhone 13 - 2022-09-01 at 15 05 40 | Simulator Screen Shot - iPhone 13 - 2022-09-01 at 15 17 37 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1212)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -1513,6 +1513,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
 
         layoutContentSubviews()
         contentView.flipSubviewsForRTL()
+
+        layoutSeparator(topSeparator, with: topSeparatorType, at: 0)
+        layoutSeparator(bottomSeparator, with: bottomSeparatorType, at: frame.height - bottomSeparator.frame.height)
     }
 
     open func layoutContentSubviews() {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1688,20 +1691,17 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
             width: frame.width - separatorLeadingInset(for: type),
             height: separator.frame.height
         )
+        separator.flipForRTL()
     }
 
     func separatorLeadingInset(for type: SeparatorType) -> CGFloat {
-        switch type {
-        case .none:
+        guard type == .inset else {
```
mischreiber, line 1698: Question: this logic differs between `main` and `fluent2-tokens`. Which is more correct? Which should we go with in the future?

mischreiber: @sophialee0416

sophialee0416: In the future, we should go with the switch statement. A little context: The SwiftUI Divider control takes safe insets into consideration, which was causing a regression with the divider appearance (because MSFSeparator did not). So to summarize, switch statement + MSFDivider, guard + MSFSeparator
https://github.com/microsoft/fluentui-apple/pull/1167

sophialee0416 (approved):

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: Great, thanks!

PR #1216, opened 2022-09-06
Remove all usage of UIScreen.main

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

UIScreen.main is deprecated in iOS 16 (https://developer.apple.com/documentation/uikit/uiscreen/1617815-main).
- Most of the usage of UIScreen main is legacy code that tries to figure out device pixel. But, in reality, in order to avoid anti-aliasing we should be just making sure we draw on integer points. 
- If we are trying to draw border, let's just draw on half point for now. 

Size decrease of 118,784 bytes in Fluent static library device 
|----------------------------------------------|-----------------------------|
|27,**251,632** bytes | 27,**132,848** bytes|

### Verification

| component                                       | Before                                       | After                                      |
|---------------------------------|----------------------------------------------|--------------------------------------------|
|badgeField|image| image |
|BadgeView |image| image|
|Card| image| image| 
|DateTimePicker|image | image|
|DrawerController| image | image|
|people picker | image| image|
|Segmented Control |image | image|
|Shimmer| image| image|
|TableViewCell | image| image|
|TableViewHeaderFooterView|image |image|
|ToolTip|image| image|
|other cells|image | image|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1216)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -603,7 +603,7 @@ open class DrawerController: UIViewController {
     private func updatePreferredContentSize(isExpanded: Bool) {
         isPreferredContentSizeBeingChangedInternally = true
         if isExpanded {
-            let screenHeight: CGFloat = UIScreen.main.bounds.height
+            let screenHeight: CGFloat = view.window?.screen.bounds.height ?? 0
```
huwilkes, line 606: Should there be a `guard let window = view.window else {` at the top of this function instead? Not sure we really want to go through all this if there isn't a window.

ios/FluentUI/Extensions/UIScreen+Extension.swift
```diff
@@ -1,29 +0,0 @@
...
-extension UIScreen {
-    var devicePixel: CGFloat { return 1 / scale }
-
-    internal func roundToDevicePixels(_ value: CGFloat) -> CGFloat {
-        /*
-         Round to 3 digits after floating point to better match the device rounding on 3x devices.
-         Avoids a situation where calculated size is smaller than screen size by a difference in 4th+ position after floaing point.
-         For example:
-         calculated = 52.666666666666671
-         actual = 52.666667938232422
```
huwilkes, line 17: Is this still an issue? Do we have an actual bug of this we fixed we can check against?

ios/FluentUI/Extensions/UIView+Extensions.swift
```diff
@@ -50,11 +50,11 @@ extension UIView {
         }
 
         if horizontally {
-            frame.origin.x = UIScreen.main.roundDownToDevicePixels(0.5 * (superview.frame.width - frame.width))
+            frame.origin.x = ceil(0.5 * (superview.frame.width - frame.width))
```
huwilkes, line 53: Should we just use centerX/Y constraints instead?

ios/FluentUI/People Picker/PersonaListView.swift
```diff
@@ -194,14 +194,12 @@ extension PersonaListView: UITableViewDataSource {
                 guard let cell = dequeueReusableCell(withIdentifier: ActivityIndicatorCell.identifier, for: indexPath) as? ActivityIndicatorCell else {
                     return UITableViewCell()
                 }
-                cell.hideSystemSeparator()
```
edjamesmsft, line 197: Do you not need to replace this with:

// hide system separator so we can draw our own...
separatorInset = UIEdgeInsets(top: 0, left: CGFloat.greatestFiniteMagnitude, bottom: 0, right: 0)

Like you do in ActionsCell?

edjamesmsft: And same with the other two instances

ios/FluentUI/People Picker/PersonaListView.swift
```diff
@@ -194,14 +194,12 @@ extension PersonaListView: UITableViewDataSource {
...
                 return cell
             case .displayingSearchResults:
                 guard let cell = dequeueReusableCell(withIdentifier: CenteredLabelCell.identifier, for: indexPath) as? CenteredLabelCell else {
                     return UITableViewCell()
                 }
                 cell.setup(text: searchResultText)
-                cell.hideSystemSeparator()
```
huwilkes, line 204: The `ActivityInidicatorCell` takes care of it in init, and the `ActionsCell` does it in layoutSubviews, but do we not need to push the system separator off screen anymore with the `CenteredLabelCell`?

huwilkes (approved):

ios/FluentUI/People Picker/PersonaListView.swift
harrieshin: this is on purpose.  because the personlistview the tableview itself has hideseparator style.

ios/FluentUI/Drawer/DrawerController.swift
harrieshin: i thought about this. but then, we shouldn't be really updating any contentsize if we aren't part of the window hierarchy. there isn't any preferredcontentsize to update.

ios/FluentUI/Extensions/UIView+Extensions.swift
harrieshin: eventaully yes, we should get rid of all cetnerInsuperview function all together but that should be in a separate PR

ios/FluentUI/People Picker/PersonaListView.swift
harrieshin: this is on purpose. because the personlistview the tableview itself hides the separator. so no behavioral changes.

ios/FluentUI/Extensions/UIScreen+Extension.swift
harrieshin: i assume this bug still exists. however, i think we should avoid aliasing as much as we can in our frame logic

ios/FluentUI/Drawer/DrawerController.swift
harrieshin: addressed in next commit.

PR #1220, opened 2022-09-07
Removing spacing and padding values from token sets

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Future versions of Fluent will no longer support full customization of all padding/spacing in a component, due to the inherent complexity of such a broad mandate. These spacings will now be more selectively supported, but will default to private internal implementations.

This means lots of deleted tokens! The following components have been updated:

* `ActivityIndicator`
* `Avatar`
* `AvatarGroup`
* `CardNudge`
* `HeadsUpDisplay`
* `IndeterminateProgressBar`
* `FluentNotification`
* `ActionsCell`
* `CenteredLabelCell`
* `PersonaButton`
* `TableViewCell`

### Verification

Binary size is a big win here, with 78.6 kB of overall savings.

| 31,857,875 bytes | 31,777,347 bytes |

| b-act | a-act |
| b-ava | a-ava |
| b-avg | a-avg |
| b-car | a-car |
| b-hud | a-hud |
| b-ipg | a-ipb |
| b-not | a-not |
| b-oth | a-oth |
| b-pbc | a-pbc |
| b-tvc | a-tvc |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1220)


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -322,7 +323,7 @@ public struct Avatar: View, TokenizedControlView {
     }
 
     var contentSize: CGFloat {
-        return tokenSet[.avatarSize].float
+        return Self.avatarSize(state.size)
     }
```
huwilkes, line 327: Do we still need contentSize?

ios/FluentUI/Notification/FluentNotification.swift
```diff
@@ -401,6 +401,21 @@ public struct FluentNotification: View, TokenizedControlView {
...
+    // MARK: Constants
+
+    /// The value for the horizontal padding between the elements within a notification and its frame
+    private static let horizontalPadding: CGFloat = GlobalTokens.spacing(.medium)
+
+    /// The value for the vertical padding between the elements within a multi-line notification and its frame
+    private static let verticalPadding: CGFloat = GlobalTokens.spacing(.small)
+
+    /// The value for the horizontal spacing between the elements within a notification
+    private static let horizontalSpacing: CGFloat = GlobalTokens.spacing(.medium)
```
huwilkes, line 414: We had a specific ask for the horizontal spacing to be customizable, so we may want to keep this one.

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Yep, it's used by `AvatarGroup`, outside of `Avatar`.

ios/FluentUI/Notification/FluentNotification.swift
mischreiber: Looks like we need `minimumHeight` too :(

ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -178,6 +178,20 @@ public struct CardNudge: View, TokenizedControlView {
         self.state = state
         self.tokenSet = CardNudgeTokenSet(style: { state.style })
     }
+
+    // MARK: Constants
+    private static let iconSize: CGFloat = GlobalTokens.iconSize(.xSmall)
```
harrieshin, line 183: long term thinking. eventhough these arent overridable shouldn't these definition live in token.swift file still because it should be all magically staticly defined?

harrieshin (approved):

ios/FluentUI/Card Nudge/CardNudge.swift
mischreiber: Y'know, I went back and forth there a lot. We technically take a veeeery minor size hit by putting them in TokenSet because the values then need to be `internal` instead of `private`. That said, I definitely do prefer that approach from a code cleanliness perspective. What do you think?

mischreiber: Moving to the proposed solution, which actually _reduces_ package size!

PR #1223, opened 2022-09-07
[Fluent 2 iOS] Contextual Command Bar colors update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The CCB was updated to use Fluent 2 tokens.

### Verification

| before_normal_light | after_normal_light |
| before_normal_dark | after_normal_dark |
| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1223)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBarButton.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButton.swift
@@ -58,6 +58,18 @@ class CommandBarButton: UIButton {
         showsMenuAsPrimaryAction = item.showsMenuAsPrimaryAction
 
         updateState()
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+    }
+
+    @objc private func themeDidChange(_ notification: Notification) {
+        guard let window = window, window.isEqual(notification.object) else {
+            return
+        }
+        updateStyle()
     }
 
     @available(*, unavailable)
--- a/ios/FluentUI/Command Bar/CommandBarButton.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButton.swift
@@ -152,12 +152,4 @@ class CommandBarButton: UIButton {
                                                     bottom: 8.0,
                                                     right: 10.0)
     }
-
-    private struct ColorConstants {
-        static let normalTintColor: UIColor = Colors.textPrimary
-        static let normalBackgroundColor = UIColor(light: Colors.gray50,
-                                                   dark: Colors.gray600)
-        static let highlightedBackgroundColor = UIColor(light: Colors.gray100,
-                                                        dark: Colors.gray900)
-    }
 }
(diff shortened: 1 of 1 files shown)


harrieshin:
that's pretty significant change! we should make sure we communicate to our customers.

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -96,48 +108,49 @@ class CommandBarButton: UIButton {
...
-        }
+    private var normalTintColor: UIColor {
+        return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.foreground1])
+    }
 
-        return UIColor(light: Colors.primary(for: window),
-                       dark: .black)
+    private var selectedTintColor: UIColor {
+        return UIColor(light: UIColor(colorValue: fluentTheme.aliasTokens.colors[.brandForeground4].light),
+                       dark: UIColor(colorValue: fluentTheme.aliasTokens.colors[.foreground1].dark!))
```
harrieshin, line 117: why do we need "!" for dark colors?

laminesm: The dark ColorValue parameter is optional so I need to unwrap it

laminesm: > The dark ColorValue parameter is optional so it needs to be unwrapped

mischreiber: We should never force unwrap, as it will crash if something goes wrong. You'll need to do a safe unwrapping elsewhere and feed the value into the UIColor constructor.

laminesm: It'll now directly pull the right global token if the dark alias color hasn't been set. Please let me know if this approach works

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -96,48 +108,59 @@ class CommandBarButton: UIButton {
...
-                       dark: .black)
+        return UIColor(light: UIColor(colorValue: fluentTheme.aliasTokens.colors[.brandForeground4].light),
+                       dark: UIColor(colorValue: darkColorValue))
     }
 
     private var selectedBackgroundColor: UIColor {
-        guard let window = window else {
-            return UIColor(light: Colors.Palette.communicationBlueTint30.color,
-                           dark: Colors.Palette.communicationBlue.color)
+        guard let darkColorValue = fluentTheme.aliasTokens.colors[.background5Selected].dark else {
```
mischreiber, line 126: Idea for simplifying this code a bit, let me know which you prefer:

```swift
let darkColorValue = fluentTheme.aliasTokens.colors[.background5Selected].dark ?? GlobalTokens.neutralColors(.grey38)
return UIColor(light: UIColor(colorValue: fluentTheme.aliasTokens.colors[.brandBackground4].light),
               dark: UIColor(colorValue: darkColorValue))
```

laminesm: I can actually bypass unwrapping the value with the `DynamicColor` initializer so I'll use that instead.

harrieshin (approved):

PR #1227, opened 2022-09-08
Clean up Button usage of APIs being deprecated in iOS 15

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This change updates the button to use configuration instead of contentEdgeInsets and titleEdgeInsets, and removes usages of imageRect (mostly).

### Verification

| ButtonInsets_Before | ButtonInsets_After2 |
| ButtonInsets_Big_Before | ButtonInsets_Big_After2 |
| ButtonInsets_RTL_Before | ButtonInsets_RTL_After2 |
| ButtonInsets_14_Before | ButtonInsets_14_After2 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1227)


ios/FluentUI/Button/Button.swift
```diff
@@ -221,6 +237,12 @@ open class Button: UIButton {
 
         titleLabel?.font = style.titleFont
         titleLabel?.adjustsFontForContentSizeCategory = true
+
+        if #available(iOS 15, *) {
+            var configuration = UIButton.Configuration.plain()
```
harrieshin, line 242: why is this only time we set the self.configuration? any client updtes to edgeinset should trigger configuration update directly? it seems like only indirect call vias adjustCustomContentEdgeInsetsForImage() is handled

harrieshin: before line 324 contentEdgeInset in update() alwasy did the right thing?

ios/FluentUI/Button/Button.swift
```diff
@@ -321,7 +343,7 @@ open class Button: UIButton {
         layer.borderWidth = style.hasBorders ? Constants.borderWidth : 0
 
         if !isUsingCustomContentEdgeInsets {
-            contentEdgeInsets = style.contentEdgeInsets
+            edgeInsets = style.contentEdgeInsets
```
sophialee0416, line 346: @harrieshin Just trying to understand your previous comment, does that mean we should be triggering config here as well?

huwilkes: Setting edgeInsets will trigger the didSet of edgeInsets, so I think just adding it there should be fine.

ios/FluentUI/Button/Button.swift
```diff
@@ -221,6 +243,12 @@ open class Button: UIButton {
 
         titleLabel?.font = style.titleFont
```
harrieshin, line 244: similar to our pillbutton fix we probably need to update font for ios 15?

huwilkes: Do you mean the bug with the segmented control? If we had the FAB on this branch we would, but since these buttons are meant to scale this should be okay.

ios/FluentUI/Button/Button.swift
```diff
@@ -181,21 +201,23 @@ open class Button: UIButton {
     }
 
     open override func imageRect(forContentRect contentRect: CGRect) -> CGRect {
```
harrieshin, line 203: hmmm we need more docuemntatiopon and assert handling for ios 15? they shouldn't be calling this anymore right?

huwilkes: Right, they shouldn't be calling it, but UIButton already has that. Are the warnings not inherited as well?

harrieshin: we can makr this has @deprecated here if it is a override?

harrieshin (approved):

ios/FluentUI/Button/Button.swift
huwilkes: Turns out trying to deprecate something the system calls makes it crash. Added an assert instead.

PR #1236, opened 2022-09-12
Fix RTL behavior for CommandBar leading and trailing Items

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

It was discovered that the RTL behavior for the leading/trailing items in the `CommandBar` was off. The leading and trailing groups are switching sides as expected, but the additional transform being applied in `configureHierarchy` is causing the views inside the groups to be shown in LTR order when the device is using a RTL language.

**Fix:**
Remove the transform for the leading and trailing groups inside `configureHierarchy` and let the default behavior from the `UIStackView` handle the transition.

**Regression Info:**
This regression was originally introduced with commit [1e8d6806062c12aee072caff2b10cf8254e9a51b](https://github.com/microsoft/fluentui-apple/commit/1e8d6806062c12aee072caff2b10cf8254e9a51b). The 0.6, 0.7 and 0.8 versions will be updated with this fix.

Only consumers who have updated use cases to show more than 1 item in the leading or trailing positions of the `CommandBar` will see this regression, otherwise it is unobservable in an application.

### Verification

Verified in the `CommandBarDemoController` with both LTR and RTL languages. Confirmed that when there is more than 1 item in the leading or trailing item group, that the items are shown in reversed order.

Before LTR Screenshot:
before-ltr

Before RTL Screenshot (note: the keyboard icon should be in the rightmost position):
before-rtl

After LTR Screenshot:
after-ltr

After RTL Screenshot (note: the keyboard icon is in the rightmost position):
after-rtl

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1236)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -232,9 +232,7 @@ open class CommandBar: UIView {
             // Flip the scroll view to invert scrolling direction. Flip its content back because it's already in RTL.
             let flipTransform = CGAffineTransform(scaleX: -1, y: 1)
             scrollView.transform = flipTransform
-            leadingCommandGroupsView.transform = flipTransform
             mainCommandGroupsView.transform = flipTransform
-            trailingCommandGroupsView.transform = flipTransform
             containerMaskLayer.setAffineTransform(flipTransform)
         }
 


harrieshin:
which commit regressed this behavior? so we understand which stable branches this may need to be applied?

nodes11:
> which commit regressed this behavior? so we understand which stable branches this may need to be applied?

It was [this commit](https://github.com/microsoft/fluentui-apple/commit/1e8d6806062c12aee072caff2b10cf8254e9a51b). It's worth noting that the `CommandBar` could only show 1 item in the leading and trailing positions before the commit causing the regression was made. So unless consumers of the `CommandBar` have updated use cases to show more than 1 item in the leading or trailing positions, the regression is unobservable in an application.

harrieshin:
Please change the PR description that referes to the commit hash that regress accessibility. We would need to update this is 0.6 and 0.7 and 0.8 branches as well

harrieshin (approved):

PR #1241, opened 2022-09-13
Move width constraint for MSFNotification with flexible width from show() to a var

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding our own width constraint was causing the Notification to stay squished if the user made the available width smaller, then increased it again. Allowing the user to set the only width constraint allows the Notification to resize properly.

### Verification

| NotificationWidth_Before | NotificationWidth_After |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -217,6 +217,8 @@ class NotificationViewDemoController: DemoController {
                                                isFlexibleWidthToast: true)
             notification.state.message = "This toast has a flexible width which means the width is based on the content rather than the screen size."
             notification.tokenSet.replaceAllOverrides(with: notificationOverrideTokens)
+            let widthConstraint = notification.widthAnchor.constraint(equalTo: view.widthAnchor, constant: -2 * notification.tokenSet[.presentationOffset].float)
+            notification.flexibleWidthConstraints = [widthConstraint]
             notification.state.actionButtonAction = { [weak self] in
                 self?.showMessage("`Dismiss` tapped")
                 notification.hide()
--- a/ios/FluentUI/Notification/MSFNotification.swift
+++ b/ios/FluentUI/Notification/MSFNotification.swift
@@ -100,12 +100,12 @@ import UIKit
         constraints.append(animated ? constraintWhenHidden : constraintWhenShown)
         constraints.append(self.centerXAnchor.constraint(equalTo: view.centerXAnchor))
 
-        let horizontalPadding = -2 * notification.tokenSet[.presentationOffset].float
-        let widthAnchor = self.widthAnchor
-        let viewWidthAnchor = view.widthAnchor
-        if isFlexibleWidthToast {
-            constraints.append(widthAnchor.constraint(lessThanOrEqualTo: viewWidthAnchor, constant: horizontalPadding))
+        if isFlexibleWidthToast, let widthConstraints = flexibleWidthConstraints {
+            constraints.append(contentsOf: widthConstraints)
         } else {
+            let horizontalPadding = -2 * notification.tokenSet[.presentationOffset].float
+            let widthAnchor = self.widthAnchor
+            let viewWidthAnchor = view.widthAnchor
             let isHalfLength = state.style.isToast && traitCollection.horizontalSizeClass == .regular
             if isHalfLength {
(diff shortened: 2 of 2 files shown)


ios/FluentUI/Notification/MSFNotification.swift
```diff
@@ -196,6 +196,12 @@ import UIKit
     }
 
     @objc public static var allowsMultipleToasts: Bool = false
+
+    /// Only to be used when `isFlexibleWidthToast` is `true`
+    /// A set of constraints to be applied to the `MSFNotification`
+    /// in `show` after it has been added to the view hierarchy.
+    @objc public var flexibleWidthConstraints: [NSLayoutConstraint]?
+
     var isFlexibleWidthToast: Bool
```
huwilkes, line 205: Should `isFlexibleWidthToast` be public instead of internal so it can be set/seen outside of the init?

ios/FluentUI/Notification/MSFNotification.swift
```diff
@@ -196,6 +196,12 @@ import UIKit
     }
 
     @objc public static var allowsMultipleToasts: Bool = false
+
+    /// Only to be used when `isFlexibleWidthToast` is `true`
+    /// A set of constraints to be applied to the `MSFNotification`
+    /// in `show` after it has been added to the view hierarchy.
+    @objc public var flexibleWidthConstraints: [NSLayoutConstraint]?
```
harrieshin, line 203: why do we need this at all? shouldn't clients add the constrainst themselves by grabbing the view?

huwilkes: This is mostly so it's easier for them to use `.show` still. Otherwise I had to write a custom animation, or have the width constraint be a layout cycle late. I also ran into a crash when trying to show a second notification in a different window due to show being delayed while the first notification was dismissed.

harrieshin: ok. so what was the issues to change line 107 to just equalTo rather than lessThanOrEqualTo ?

huwilkes: Then it would be too long if the text was short. We want the width of the notification to hug the text.

harrieshin: then your demo app changes will still have the smae issue?

huwilkes: It would if the text didn't take up the whole space, yes. But, it's behaving how we wanted. And now we aren't setting any constraints internally to mess with that.

harrieshin: having the nsconstraint var public in case another property is set to true by the client doesn't seem like the right design as a ux framework. either we do all the calculation correctly (figure out the intrinsic content size and update. the width anchor) or client code actually sets the width themselves is the much cleaner approach.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -248,11 +247,66 @@ class NotificationViewDemoController: DemoController {
...
+            notification.show(in: view) { $0.hide(after: 3.0) }
+        }
     }
 
     @objc private func showSwiftUIDemo() {
         navigationController?.pushViewController(NotificationViewDemoControllerSwiftUI(),
                                                  animated: true)
     }
+
+    private func showNotification(notification: MSFNotification) {
```
harrieshin, line 263: so bascially we need to update the show an documentation that client shouldn't use it if they want flexiblewidth? what is a good way to calculate the best estimate of swiftuinotfication instrinsic size to uikit callers?

ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -248,11 +247,66 @@ class NotificationViewDemoController: DemoController {
...
+        NSLayoutConstraint.activate([
+            constraintWhenHidden,
+            notification.centerXAnchor.constraint(equalTo: view.centerXAnchor),
+            notification.widthAnchor.constraint(equalTo: view.widthAnchor, constant: -2 * presentationOffset )
+        ])
+        notification.alpha = 0
+        view.layoutIfNeeded()
+        flexibleWidthNotificationHidden = constraintWhenHidden
+        flexibleWidthNotificationShown = constraintWhenShown
+        UIView.animate(withDuration: 0.6,
```
sophialee0416, line 281: Should these values be a constant somewhere? and with the other values around the PR

PR #1254, opened 2022-09-14
Hide CommandBarButtonGroupView if all CommandBarButtons are hidden

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

**Issue:** 
If all the items in a `CommandBarButtonGroupView` are hidden, the system spacers are still shown in the stack view and create visible empty space. See video below for example.

This is a bug with the `CommandBar`. Not technically a regression since the ability to hide a `CommandBarButton` was introduced with [this commit](https://github.com/microsoft/fluentui-apple/commit/061da18cf8b15b59ea6b3affd1f59e3414f9b43b) and this change is fixing a case that was missed with the original commit.

**Root Cause:** 
The `CommandBarItem.propertyChangedUpdateBlock` was calling `CommandBarButton.updateState()` directly to update the `isHidden` state of a `CommandBarButton`. This works to hide each of the individual buttons, but the superview, the `CommadBarButtonGroupView.stackView`, isn't aware of when its subviews change their hidden state. If all of the subviews are hidden, the superview needs to also mark itself as hidden so that the system spacers inside the stackView do not remain visible.

**Fix:**
- The `CommandBarItem.propertyChangedUpdateBlock` block that calls `CommandBarButton.updateState()` now takes an additional `Bool` parameter to indicate if the  `CommadBarButtonGroupView` should also update its state. If the parameter is `true` the block calls `hideGroupIfNeeded()` on the group.
- Add a call to `hideGroupIfNeeded()` from `CommadBarButtonGroupView. init(buttons:)` to ensure the group view is hidden if it is initialized with buttons that are set as hidden.
- Modified the loop used to create a `CommadBarButtonGroupView` so that an update call can be made on the group if needed.
- Updated `CommandBarDemoController` to hide the `"Delete"` item instead of the `"+"` item. Since the `"Delete"` item is the item in its own group, the `CommadBarButtonGroupView` will have to hide itself when the `"Delete"` item is hidden, testing the functionality added in this PR.

### Verification

**Before (Observe space before the Delete item):**

https://user-images.githubusercontent.com/10938746/190278112-bc926e97-86a2-4161-94f7-5da78b853f0c.mov

**After (Observe space before the Delete item):**

https://user-images.githubusercontent.com/10938746/190278101-a59f7b84-1b1e-4b8d-9c73-317e2dd0a42c.mov

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1254)

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
@@ -28,15 +28,6 @@ class CommandBarButtonGroupView: UIView {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
-    /// Update the passed button and hide the group view if requested
-    func update(_ button: CommandBarButton, updateGroupView: Bool) {
-        button.updateState()
-
-        if updateGroupView {
-            hideGroupIfNeeded()
-        }
-    }
-
     /// Hides the group view if all the views inside the `stackView` are hidden
      func hideGroupIfNeeded() {
         var allViewsHidden = true
--- a/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
@@ -20,11 +20,21 @@ class CommandBarButtonGroupView: UIView {
 
         configureHierarchy()
         applyInsets()
+        hideGroupIfNeeded()
     }
 
     @available(*, unavailable)
     required init?(coder: NSCoder) {
         preconditionFailure("init(coder:) has not been implemented")
+    }
+
+    /// Update the passed button and hide the group view if requested
+    func update(_ button: CommandBarButton, updateGroupView: Bool) {
+        button.updateState()
+
+        if updateGroupView {
+            hideGroupIfNeeded()
+        }
     }
 
     private lazy var stackView: UIStackView = {
@@ -54,6 +64,20 @@ class CommandBarButtonGroupView: UIView {
             buttons.first?.contentEdgeInsets.left += LayoutConstants.leftRightBuffer
             buttons.last?.contentEdgeInsets.right += LayoutConstants.leftRightBuffer
         }
+    }
+
+    /// If all views inside the `stackView` are hidden, the group itself should also be hidden. Otherwise the system spacers
+    /// will remain unhidden and cause additional visible space in the layout.
+    private func hideGroupIfNeeded() {
+        var allViewsHidden = true
+        for view in stackView.arrangedSubviews {
+            if !view.isHidden {
+                allViewsHidden = false
+                break
+            }
+        }
+
+        isHidden = allViewsHidden
     }
 
     private struct LayoutConstants {
--- a/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
@@ -29,7 +29,7 @@ class CommandBarButtonGroupView: UIView {
     }
 
     /// Hides the group view if all the views inside the `stackView` are hidden
-     func hideGroupIfNeeded() {
+    func hideGroupIfNeeded() {
         var allViewsHidden = true
         for view in stackView.arrangedSubviews {
             if !view.isHidden {
--- a/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
+++ b/ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
@@ -37,6 +37,19 @@ class CommandBarButtonGroupView: UIView {
         }
     }
 
+    /// Hides the group view if all the views inside the `stackView` are hidden
+     func hideGroupIfNeeded() {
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -125,7 +125,7 @@ open class CommandBarItem: NSObject {
     }
 
     /// Called after a property is changed to trigger the update of a corresponding button
-    var propertyChangedUpdateBlock: ((CommandBarItem) -> Void)?
+    var propertyChangedUpdateBlock: ((_: CommandBarItem, _: Bool) -> Void)?
```
harrieshin, line 128: doccument each param what it means

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -20,13 +20,23 @@ class CommandBarButtonGroupView: UIView {
...
+        hideGroupIfNeeded()
     }
 
     @available(*, unavailable)
     required init?(coder: NSCoder) {
         preconditionFailure("init(coder:) has not been implemented")
     }
 
+    /// Update the passed button and hide the group view if requested
+    func update(_ button: CommandBarButton, updateGroupView: Bool) {
```
harrieshin, line 32: something about this make me feel messy. this is meant to be each button need to update. not all property update tirggers group stackview to be hidden. rather than pipling all of propertyChangedUpdateBlock to trigger this, can we just call the hideGroupIfneeded() when the hidden property changes?

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -125,7 +125,9 @@ open class CommandBarItem: NSObject {
     }
 
     /// Called after a property is changed to trigger the update of a corresponding button
-    var propertyChangedUpdateBlock: ((CommandBarItem) -> Void)?
+    /// - Parameter CommandBarItem: Instance of `CommandBarItem` the closure is being invoked from
+    /// - Parameter Bool: Indicates if the item's group state should also be updated
+    var propertyChangedUpdateBlock: ((_: CommandBarItem, _: Bool) -> Void)?
```
harrieshin, line 130: i think because it isn't really apparent what this bool is for. lets actually have a variable? rather than _ ?

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -77,15 +77,28 @@ class CommandBarCommandGroupsView: UIView {
...
+                    preconditionFailure("Button is not initialized in map")
+                }
+                return button
+            }
+
+            let group = CommandBarButtonGroupView(buttons: buttons)
+
+            for item in items {
+                if let button = itemsToButtonsMap[item] {
+                    item.propertyChangedUpdateBlock = { _, updateGroupViewState in
```
harrieshin, line 91: "updateGroupViewState" : i am also bad at naming things but can we think how this can be more clear swift boolean naming?

ios/FluentUI/Command Bar/CommandBarItem.swift
nodes11: I've added parameter names, however, Swift's type system requires that the parameters of function types are unnamed so the `_` before each parameter name is left in place. [This document](https://github.com/apple/swift-evolution/blob/master/proposals/0111-remove-arg-label-type-significance.md) provides some detail on why it works this way.

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
nodes11: `shouldUpdateGroupState` seems a bit better to me. The `should` prefix hints at the fact that it is a boolean.

ios/FluentUI/Command Bar/CommandBarItem.swift
harrieshin: thanks for the doc reference. can we do old school and in the places you call the propertychangedupateblock we do /* shouldUpdateGroupState*/ next do the boolean? i really think it isn't super apparant for the readers.

harrieshin (approved):

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -20,13 +20,27 @@ class CommandBarButtonGroupView: UIView {
...
 
+    /// Hides the group view if all the views inside the `stackView` are hidden
+    func hideGroupIfNeeded() {
+        var allViewsHidden = true
+        for view in stackView.arrangedSubviews {
+            if !view.isHidden {
+                allViewsHidden = false
+                break
+            }
+        }
```
anandrajeswaran, line 39: swifty way to do this:
```
isHidden = !stackView.arrangedSubviews.contains(where: { !$0.isHidden })

PR #1262, opened 2022-09-16
Adding back `.paddingLeading` token to `TableViewCellTokenSet`

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The `paddingLeading` token in TableViewCell needs to be added back to accommodate for a downstream dependency. It may be reconsidered for removal again in the future, but for now it will be a part of the `TableViewCellTokenSet` again.

Note:
- Original PR for padding removal: https://github.com/microsoft/fluentui-apple/pull/1220

### Verification

TableViewCell works as expected.

| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1262)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -170,7 +170,10 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     @objc public static let defaultNumberOfLinesForLargerDynamicType: Int = -1
 
     /// The default leading padding in the cell.
-    @objc public static var defaultPaddingLeading: CGFloat { TableViewCellTokenSet.paddingLeading }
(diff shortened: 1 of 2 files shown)


laminesm (commented):
Out of curiosity why would we choose to add the padding back to the token set instead of making the dependency file use the padding from the extension?

sophialee0416:
> Out of curiosity why would we choose to add the padding back to the token set instead of making the dependency file use the padding from the extension?

I don't think we'd be able to override the padding value from the extension because those values are constants.

huwilkes (approved):

PR #1258, opened 2022-09-15
[Fluent 2 iOS] Merge Main

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Merging main into fluent2-colors. This includes the following changes:

- https://github.com/microsoft/fluentui-apple/pull/1196
- https://github.com/microsoft/fluentui-apple/pull/1197
- https://github.com/microsoft/fluentui-apple/pull/1198
- https://github.com/microsoft/fluentui-apple/pull/1206
- https://github.com/microsoft/fluentui-apple/pull/1212
- https://github.com/microsoft/fluentui-apple/pull/1213
- https://github.com/microsoft/fluentui-apple/pull/1203
- https://github.com/microsoft/fluentui-apple/pull/1210
- https://github.com/microsoft/fluentui-apple/pull/1217
- https://github.com/microsoft/fluentui-apple/pull/1215
- https://github.com/microsoft/fluentui-apple/pull/1216
- https://github.com/microsoft/fluentui-apple/pull/1218
- https://github.com/microsoft/fluentui-apple/pull/1221
- https://github.com/microsoft/fluentui-apple/pull/1219
- https://github.com/microsoft/fluentui-apple/pull/1231
- https://github.com/microsoft/fluentui-apple/pull/1220
- https://github.com/microsoft/fluentui-apple/pull/1229
- https://github.com/microsoft/fluentui-apple/pull/1233
- https://github.com/microsoft/fluentui-apple/pull/1234
- https://github.com/microsoft/fluentui-apple/pull/1232
- https://github.com/microsoft/fluentui-apple/pull/1230
- https://github.com/microsoft/fluentui-apple/pull/1236
- https://github.com/microsoft/fluentui-apple/pull/1194
- https://github.com/microsoft/fluentui-apple/pull/1207
- https://github.com/microsoft/fluentui-apple/pull/1244
- https://github.com/microsoft/fluentui-apple/pull/1242

### Verification

Built and ran simulator app, running through each major conflict's demo app (ActivityIndicator, Avatar, HUD, IndeterminateProgressBar, Notification, PersonaButtonCarousel, TableViewTokens), ensuring functionality did not regress.

(screenshots are incoming - there are many)

| Main                                       | fluent2-colors before                                      | fluent2-colors after      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
[...]

PR #1271, opened 2022-09-27
[Fluent 2 iOS] Fix HUD color contrast issues

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The hud now uses new tokens that fix the existing color contrast issues. 

### Verification

Changes were tested on the demo app.

| before_activity_light | after_activity_light |
| before_activity_dark |  after_activity_dark |
| before_failure_light | after_failure_light |
| before_failure_dark | after_failure_dark |
| before_nolabel_light | after_nolabel_light |
| before_nolabel_dark | after_nolabel_dark |
| before_swiftui_light | after_swiftui_nolabel_light |
| before_swiftui_dark | after_swiftui_nolabel_dark |
| before_swiftui_label_light | after_swiftui_light |
| before_swiftui_label_dark | after_swiftui_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1271)


ios/FluentUI/HUD/HUDTokens.swift
```diff
@@ -10,13 +10,13 @@ import UIKit
 open class HeadsUpDisplayTokens: ControlTokens {
 
     /// The color of the squared background of the Heads-up display.
-    open var backgroundColor: DynamicColor { aliasTokens.colors[.backgroundInverted] }
+    open var backgroundColor: DynamicColor { aliasTokens.colors[.backgroundDarkStatic] }
 
     /// The color of the activity indicator presented by the Heads-up display.
-    open var activityIndicatorColor: DynamicColor { aliasTokens.colors[.foreground2] }
+    open var activityIndicatorColor: UIColor { .init(light: UIColor(dynamicColor: aliasTokens.colors[.foregroundLightStatic]).withAlphaComponent(0.4), dark: UIColor(dynamicColor: aliasTokens.colors[.foregroundLightStatic]).withAlphaComponent(0.6)) }
```
huwilkes, line 16: With #1258 merged, I think adding a token of type `UIColor` is a bit of a problem. I think we may be better off keeping this returning `DynamicColor`, and putting the different opacities somewhere else. Though, I also don't really get why we need the different opacities.

laminesm: The different opacities were set by design to get the desired contrast ratios. Can you clarify why we shouldn't return a `UIColor`?

huwilkes: With `ControlTokenValue`, each token is actually a collection of every possible type, so adding a new type would actually lead to a lot of bloat that we don't want.

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
```diff
@@ -47,13 +36,22 @@ public class HeadsUpDisplayTokenSet: ControlTokenSet<HeadsUpDisplayTokenSet.Toke
...
             case .activityIndicatorColor:
+                let colorToken = theme.aliasTokens.colors[.foregroundLightStatic]
+                let lightColorValue = ColorValue(r: colorToken.light.r,
+                                                 g: colorToken.light.g,
+                                                 b: colorToken.light.b,
+                                                 a: 0.4)
+                let darkColorValue = ColorValue(r: colorToken.dark?.r ?? colorToken.light.r,
+                                                g: colorToken.dark?.g ?? colorToken.light.g,
+                                                b: colorToken.dark?.b ?? colorToken.light.b,
+                                                a: 0.6)
```
huwilkes, line 47: Looks like the design has light and dark mode both using `FG Light Static` @ different alphas.

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
```diff
@@ -47,13 +36,22 @@ public class HeadsUpDisplayTokenSet: ControlTokenSet<HeadsUpDisplayTokenSet.Toke
                 }
 
             case .activityIndicatorColor:
+                let colorToken = theme.aliasTokens.colors[.foregroundLightStatic]
+                let lightColorValue = ColorValue(r: colorToken.light.r,
+                                                 g: colorToken.light.g,
+                                                 b: colorToken.light.b,
+                                                 a: 0.4)
```
huwilkes, line 43: We talked about this some offline, but I think we need to follow up with design on this one. I don't think we should be making new colors by changing the alpha value of other tokens. I don't think that should block this change, but I think we should try to resist this becoming a new trend.

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
```diff
@@ -47,13 +36,22 @@ public class HeadsUpDisplayTokenSet: ControlTokenSet<HeadsUpDisplayTokenSet.Toke
                 }
 
             case .activityIndicatorColor:
+                let colorToken = theme.aliasTokens.colors[.foregroundLightStatic]
```
huwilkes, line 39: Not sure it's quite in the scope of this one, but the design also seems to have a "Progress indicator background" value defined and the code doesn't.

laminesm: Some other controls like cards have different designs as well, and for the scope of the colors work the focus has been on updating the colors, fonts, sizes, etc.

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
edjamesmsft: If we rely on alpha/transparency, couldn't this be a problem if we have an unexpected color behind the control? If we don't explicitly set a fully opaque color, then we have no way to ensure we are actually fulfilling design goals (including accessibility ratios).

laminesm: Design confirmed that this is not to create a one-off usage token for the HUD component. This white token/opacity combination will likely be used to deal with future contrast issues. I would take this method over creating single use tokens especially if we're likely to run into similar contrast issues in the future.

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
laminesm: That's how it's implemented as well but it's pulling its dark `ColorValue` for dark mode. 

Screen Shot 2022-09-28 at 2 57 39 PM

huwilkes (approved):

ios/FluentUI/HUD/HeadsUpDisplayTokenSet.swift
mischreiber: Personally I don't see this as any different than having a one-off token: the hardcoded alpha value is every bit as custom to this control as a hardcoded color would be. Let's follow up more with design, because I don't think this is setting a good precedent for dealing with creating grays from white.

mischreiber: Separately, we shouldn't be hard-coding values like `0.4` directly in code like this. Magic numbers are bad!

mischreiber (changes requested):

harrieshin (approved):

mischreiber (approved):

PR #1287, opened 2022-10-08
[Fluent 2 iOS] Updating Button to Fluent 2

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated button to the new designs. Note, danger button does not have pressed/highlighted state in the current designs. Additionally, there is no focus ring. When we move to the new SwiftUI button and update the colors for that control, we will be in closer alignment with designs. Note also that the pressed state now goes dark instead of light in light mode (and the inverse in dark mode).

### Verification

| buttonOldLight | buttonNewLight |
| buttonOldDark | buttonNewDark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1287)


ios/FluentUI/Button/Button.swift
```diff
@@ -396,33 +391,41 @@ open class Button: UIButton {
...
         }
 
+        if isFocused {
+            return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.brandBackground1Selected])
+        }
+
+        return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.brandBackground1])
+    }
+
+    private func primaryDangerFilledBackgroundColor() -> UIColor {
```
mischreiber, line 406: Does this need to be a function, or could it be a `var` with a custom getter?

```swift
private var primaryDangerFilledBackgroundColor: UIColor {
    // TODO: in the future, we want to have highlighted state defined for danger buttons.
    // For now, highlighted/isPressed are not differentiated for danger buttons.
    return UIColor(dynamicColor: fluentTheme.aliasTokens.sharedColors[.dangerBackground2])
}

Comment (mischreiber) [reply ios/FluentUI/Button/Button.swift:406]: I think the main reason it'd be a function is if it takes any arguments, otherwise we prefer to use vars.

**Review by laminesm — COMMENTED** [review]
Would also need to add an observer for theme change notifications

File: ios/FluentUI/Button/Button.swift [hunk 2]
```diff
@@ -325,24 +313,31 @@ open class Button: UIButton {
...
     }
 
     private func highlightedTitleAndImageColor(for window: UIWindow) -> UIColor {
         if style.isFilledStyle {
-            return Colors.Button.titleWithFilledBackground
+            return style.isDangerStyle ? dangerFilledTitleAndImageColor : titleWithFilledBackground
         }
 
-        return style.isDangerStyle ? Colors.Palette.dangerTint20.color : Colors.primaryTint20(for: window)
+        return style.isDangerStyle ? dangerTitleAndImageColor : UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.brandStroke1Pressed])
```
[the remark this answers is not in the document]
edjamesmsft: The inlined colors were the ones already inlined - private lazy var were the consts. Additionally, the ones that were consts are generally ones that were used more than once, so private lazy vars means less redundancy/more reuse/lower risk of making a simple mistake.

ios/FluentUI/Button/Button.swift
```diff
@@ -96,18 +96,6 @@ public enum ButtonStyle: Int, CaseIterable {
...
-    struct Button {
-        public static var background: UIColor = .clear
-        public static var backgroundFilledDisabled: UIColor = surfaceQuaternary
-        public static var borderDisabled: UIColor = surfaceQuaternary
-        public static var titleDisabled: UIColor = textDisabled
-        public static var titleWithFilledBackground: UIColor = textOnAccent
-    }
-}
-
 // MARK: - Button
```
mischreiber, line 99: Leaving an open comment to stop the automerge to give @laminesm a chance to look too! I'll resolve this and let automerge do its thing if there are no objections later today.

mischreiber (approved):

ios/FluentUI/Button/Button.swift
```diff
@@ -280,6 +273,13 @@ open class Button: UIButton {
         updateProposedTitleLabelWidth()
     }
 
+    @objc private func themeDidChange(_ notification: Notification) {
```
laminesm, line 276: Can you add this check:
```
      return
}
```

PR #1290, opened 2022-10-12
Button - Added support for Trailing Image

amoggo:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
- Changes were made to the Fluent Button to support a Trailing Image in addition to the main (Primary) Button Image. This trailing image is optional, and is designed such that it is independent of the imagePosition of the primary image and will always be anchored to the Trailing side of the button. 
- Layout Alignment Logic: Because this trailing image is added as a subview to the Fluent Button, auto layout constraints were added such that it remains anchored to the Trailing end of the button, accounting for the margin padding as well as adjusting the alignment of the other button contents like title and primary image to account for the new image. This alignment logic follows suit with the pre-existing logic for aligning the titleRect and imageRect where initially the area in question is centered, then it is offset based on the other button content dimensions and finally the Button borders are resized to account for its contents. 
- RTL support was also added to the Trailing Image such that in RTL languages the Image view will still get anchored to the correct side of the button whilst maintaining the right spacing. 
- ContentTintColor was accounted for in that the trailing image tint will be identical to the tint of the button content - primary image and title
- Updating the trailing image (including the removal of it) was also accounted for

### Verification in the Test App
- RTL was tested by changing system language to Arabic
- tested for all Image Position settings
- Tested for all the Button State settings
- Tested for most of the background pane colors and materials. 
- Tested the modification of trailing image and removing it after previously setting it.

The Button with the Trailing Image set:
Screen Shot 2022-10-11 at 22 55 20

| Before       | After                                      |
| Screen Shot 2022-10-11 at 21 50 46 | Screen Shot 2022-10-11 at 22 07 25 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1290)


macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
...
+					NSLayoutConstraint.activate([
+						secondaryImageView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -cell.horizontalPadding),
+						secondaryImageView.centerYAnchor.constraint(equalTo: centerYAnchor)
+					])
+					updateContentTintColor()
+				}
+			}
+		}
+
+	private var secondaryImageView: NSImageView = NSImageView()
```
anandrajeswaran, line 188: Can we not construct an NSImageView unless secondaryImage is non null (make this optional). That helps with a bit of perf and helps keep the view hierarchy clean when this property isn't used

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
 		}
 	}
 
+	/// When set to a non `nil` value ,the Secondary Image will always be placed on the trailing edge of the button in both LTR and
+	/// RTL language settings. It is designed to remain independent of  the `imagePosition` value of the NSButton which deals
+	/// with the main Button `image`.
+	public var secondaryImage: NSImage? {
+			didSet {
```
anandrajeswaran, line 161: minor: Unless GitHub UI is deceiving me (and it certainly could be) - looks like multiple tabs here?

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
...
+				guard oldValue != secondaryImage else {
+					return
+				}
+
+				guard let cell = cell as? ButtonCell else {
+					return
```
anandrajeswaran, line 167: This one is probably a preconditionFailure rather than a silent failure as it is probably illegal to use a Fluent button with a different cell?

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
...
+				}
+				
+				// Handle replacement/removal of a previously set image
+				if secondaryImageView.isDescendant(of: self) {
```
anandrajeswaran, line 171: will secondaryImageView ever exist and not be a descendant? Per other comment about making it optional, could then just always remove from superview if it exists?

anandrajeswaran: Or actually in the case where it already exists we could just avoid re-adding to the hierarchy and instead just change the image of the NSImageView?

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
...
+
+				if let secondaryImage = secondaryImage {
+					secondaryImageView = NSImageView(image: secondaryImage)
+					secondaryImageView.translatesAutoresizingMaskIntoConstraints = false
+					addSubview(secondaryImageView)
+					NSLayoutConstraint.activate([
+						secondaryImageView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -cell.horizontalPadding),
+						secondaryImageView.centerYAnchor.constraint(equalTo: centerYAnchor)
+					])
+					updateContentTintColor()
```
anandrajeswaran, line 183: To avoid needing to update contentTintColor on the rest of the button while it's not changing, can update `updateContentTintColor` to call a new `updateSecondaryImageTintColor` which sets secondaryImageView's contentTintColor to the button's contentTintColor. And then directly call into `updateSecondaryImageTintColor` here instead

macos/FluentUI/Button/Button.swift
```diff
@@ -374,7 +410,13 @@ open class Button: NSButton {
 
 	open override var intrinsicContentSize: CGSize {
 		let superSize = super.intrinsicContentSize
-		return CGSize(width: superSize.width,
+		var trailingImageAdjustment: CGFloat = 0
+
+		// Account for extra space needed by `secondaryImage`
+		if secondaryImage != nil, let cell = cell as? ButtonCell {
```
anandrajeswaran, line 416: I think usually style is to separate if clauses into multiple lines

macos/FluentUI/Button/Button.swift
```diff
@@ -374,7 +410,13 @@ open class Button: NSButton {
 
 	open override var intrinsicContentSize: CGSize {
 		let superSize = super.intrinsicContentSize
-		return CGSize(width: superSize.width,
+		var trailingImageAdjustment: CGFloat = 0
```
anandrajeswaran, line 413: can make this a let (prefer immutability) and avoid assigning it an initial value by assigning the value to 0 in an else block

macos/FluentUI/Button/Button.swift
```diff
@@ -431,11 +473,19 @@ class ButtonCell: NSButtonCell {
...
+		// First, center the Primary Image
 		var x = (rect.width - imageSize.width) / 2
 		var y = (rect.height - imageSize.height) / 2
 
 		if xOffsetSign != 0 {
+			// Second, offset the Primary Image from the Title
 			x += CGFloat(xOffsetSign) * (titleSize.width + titleToImageSpacing) / 2
+
+			// Third, offset the Title from the Secondary Image
+			if let controlView = self.controlView as? Button,
```
anandrajeswaran, line 485: minor: can omit self. here - I think type inference easily gets the `if let property = someProperty as? SomeType` pattern

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
 		}
 	}
 
+	/// When set to a non `nil` value ,the Secondary Image will always be placed on the trailing edge of the button in both LTR and
```
anandrajeswaran, line 157: minor: I think "in both RTL language settings" is implicit because you used the word "trailing" rather than "right"

anandrajeswaran (changes requested):

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
...
+	/// RTL language settings. It is designed to remain independent of  the `imagePosition` value of the NSButton which deals
+	/// with the main Button `image`.
```
brentpe, line 159: Minor: Instead of "deals with the main Button 'image', you could say "applies to the built-in 'image' property for NSButtons." Or just "applies to the built-in 'image' property."

macos/FluentUI/Button/Button.swift
```diff
@@ -491,11 +541,19 @@ class ButtonCell: NSButtonCell {
...
 
+		// First, center the Title
 		var x = (rect.width - titleSize.width) / 2
 		var y = (rect.height - titleSize.height) / 2 + titleVerticalPositionAdjustment
 
 		if xOffsetSign != 0 {
+			// Second, offset the Title from the Primary Image
 			x += CGFloat(xOffsetSign) * (imageSize.width + titleToImageSpacing) / 2
+
+			// Third, offset the Title from the Secondary Image
```
laminesm, line 552: nit: you can omit the `First`, `Second` and `Third` parts of the comments as the control flow already dictates the order.

macos/FluentUI/Button/Button.swift
```diff
@@ -154,6 +154,39 @@ open class Button: NSButton {
 		}
 	}
 
+	/// When set to a non `nil` value ,the Secondary Image will always be placed on the trailing edge of the button in both LTR and
+	/// RTL language settings. It is designed to remain independent of  the `imagePosition` value of the NSButton which deals
+	/// with the main Button `image`.
+	public var secondaryImage: NSImage? {
```
brentpe, line 160: Minor: Maybe put this right below the "image" property to keep them together, since these aren't really sorted alphabetically anyhow.

brentpe, line 160: Needs @objc prefix so we can access it in Objective-C code.

macos/FluentUI/Button/Button.swift
```diff
@@ -131,6 +131,42 @@ open class Button: NSButton {
...
+			guard oldValue != trailingImage else {
+				return
+			}
+
+			guard let cell = cell as? ButtonCell else {
+				preconditionFailure("The FluentUI `Button` should only be used with the `ButtonCell` cell class.")
+			}
+
+			// Handle replacement/removal of a previously set image
+			trailingImageView?.image = nil
```
anandrajeswaran, line 147: Can do this in a corresponding `else` block for the `if` condition on line 149

anandrajeswaran: And then we should also remove it from our view hierarchy and nil it out in that case

macos/FluentUI/Button/Button.swift
```diff
@@ -264,6 +300,17 @@ open class Button: NSButton {
 		} else {
 			contentTintColor = contentTintColorRest
 		}
+		updateTrailingImageContentTintColor()
+	}
+
+	private func updateTrailingImageContentTintColor() {
+		if !isEnabled {
```
anandrajeswaran, line 307: can enclose this in `if let trailingImageView`? Then we don't need any optional unwraps in each of the cases and we save some trivial work if we don't have a trailing image

macos/FluentUITestApp/Assets.xcassets/ic_fluent_chevron_down_12_regular.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
...
+    }
+  ],
+  "info": {
+    "author": "xcode",
+    "version": 1
+  },
+  "properties": {
+    "template-rendering-intent": "template"
+  }
+}
```
brentpe, line 15: Nit: This shows there's no blank line at the end of this file and there should be.

amoggo: This is copied over from FluentIcon library where there's no blank line at the end.

brentpe (approved):

anandrajeswaran (approved):

PR #1293, opened 2022-10-13
Update iPad CommandBar group spacing, propagate accessibilityValue from CommandBarItem to CommandBarButton, and add accentImage to CommandBarItem

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

- `CommandBarCommandGroupsView` 
  - Change group spacing for iPad back to `16.0`. Verified by Fluent design board through email.
- `CommandBarItem` 
  - Override `accessibilityValue` property. Providers of `CommandBarItem` can set this value to provide additional accessibility info describing the button state if desired.
  - Expose `accentImage` property. This property enables apps not using SFSymbols to "fill" images with color by inserting a second image behind the default image.
- `CommandBarButton` 
  - Set the `UIButton.accessibilityValue` equal to the `item.accessibilityValue` inside the `init` and `updateState` methods.
  -  Create a private `accentImageView` `UIImageView?` property
  -  Implement `layoutSubviews`. If the item has a `accentImage` set, initialize the `accemtImageView` with the `accentImage`, insert the `accentImageView` as a subview behind the `UIButton.imageView`, and set the frame of the `accentImageView` to match the `UIButton.imageView`.
- CommandBarDemoController
  - Add `delete24Filled` asset from microsoft/fluentui-system-icons to the Demo assets catalog
  - Configure the delete item to use the `accentImage` property to test the new api

### Verification

**Spacing Changes**

Before:
Screen Shot 2022-10-13 at 9 47 48 AM

After:
Screen Shot 2022-10-13 at 9 45 31 AM

**Accent Image API Changes**

Before Light Mode:
Screen Shot 2022-10-14 at 4 16 57 PM

Before Dark Mode:
Screen Shot 2022-10-14 at 4 17 10 PM

After Light Mode:
Screen Shot 2022-10-14 at 4 00 39 PM

After Dark Mode:
Screen Shot 2022-10-14 at 4 00 22 PM

Accent Image Toggle Demo:

https://user-images.githubusercontent.com/10938746/196231609-a7c3d757-89f4-4b4b-8deb-3e7dad90dc4e.mov

- Verified spacing updates only show on iPad
- Verified the accessibilityValue propagates as expected from the `CommandBarItem` to the `CommandBarButton`
- Verified an accessibilityValue use case in devmain applications
- Verified accentImage in demo controller and added new example
- Verified accentImage use case in devmain applications

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1293)


ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -18,7 +18,7 @@ class CommandBarCommandGroupsView: UIView {
 
         buttonGroupsStackView.translatesAutoresizingMaskIntoConstraints = false
         buttonGroupsStackView.axis = .horizontal
-        buttonGroupsStackView.spacing = LayoutConstants.buttonGroupSpacing
+        buttonGroupsStackView.spacing = (UIDevice.current.userInterfaceIdiom == .phone) ? LayoutConstants.buttonGroupSpacingPhone : LayoutConstants.buttonGroupSpacingPad
```
mischreiber, line 21: Seems like we should be using size classes rather than device interface idiom here.

mischreiber (changes requested):

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
nodes11: I made the change to use size classes locally but forgot to amend it to the commit 🤦‍♂️. Will update.

nodes11:
Apologies for the scope creep in this PR. One additional API was added to allow the CommandBarItem to have an `accentImage` that sits behind the default `iconImage`. An example was added in the demo controller.

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -73,6 +74,22 @@ class CommandBarButton: UIButton {
...
+
+        guard let accentImage = item.accentImage, let imageView = imageView else {
+            return
+        }
+        
+        let iconImageFrame = imageView.frame
+
+        accentImageView = UIImageView(image: accentImage)
+        if let accentImageView = accentImageView {
+            insertSubview(accentImageView, belowSubview: imageView)
```
anandrajeswaran, line 88: we shouldn't be modifying the view hierarchy in layoutSubviews. We should fix up the hierarchy when accentImage is set and then either can use layoutSubviews or just add constraints to keep them positioned in the same place. But layout happens often and manipulating the view hierarchy should only happen when needed

anandrajeswaran (changes requested):

ios/FluentUI/Command Bar/CommandBarButton.swift
nodes11: Pulled the view hierarchy manipulation out of layoutSubviews.

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/CommandBar/delete24Filled.imageset/Contents.json
```diff
@@ -0,0 +1,15 @@
+{
+  "images": [
+    {
+      "filename": "ic_fluent_delete_24_filled.svg",
```
harrieshin, line 4: why we using svg? most of the fluent uses pdf

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -78,6 +79,16 @@ class CommandBarButton: UIButton {
...
     }
 
+    override func layoutSubviews() {
+        super.layoutSubviews()
+
+        guard let accentImageView = accentImageView, let imageView = imageView else {
+            return
+        }
+
+        accentImageView.frame = imageView.frame
```
harrieshin, line 89: why we doing frame calculation instead of autolayout?

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/CommandBar/delete24Filled.imageset/Contents.json
nodes11: Overlooked the filetype. Updated to the pdf asset.

ios/FluentUI/Command Bar/CommandBarButton.swift
nodes11: No specific reason other than the internal example I was looking at used frame calculation. Updated the `accentImageView` to constraint itself to the `UIButton.imageView` `centerXAnchor` and `centerYAnchor`. Confirmed that the image is positioned correctly on size class changes, rotations, and if a title shown in the button.

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -135,6 +139,27 @@ class CommandBarButton: UIButton {
                         dark: Colors.primary(for: window))
     }
 
+    private var accentImageView: UIImageView?
+
+    private func updateAccentImage(accentImage: UIImage?) {
```
anandrajeswaran, line 144: minor: to be more swift-y, would recommend either `updateAccentImage(_ accentImage: UIImage?)` or `update(accentImage: UIImage?)`. So callers don't have to duplicate `accentImage`

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -135,6 +139,27 @@ class CommandBarButton: UIButton {
...
+        if accentImage != accentImageView?.image {
+            if let accentImage = accentImage {
+                accentImageView = UIImageView(image: accentImage)
+                accentImageView?.translatesAutoresizingMaskIntoConstraints = false
+                if let accentImageView = accentImageView, let imageView = imageView {
```
anandrajeswaran, line 149: minor: for clarity and debuggability, usually good to put separate `if let` clauses on separate lines

anandrajeswaran (approved):

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -135,6 +139,28 @@ class CommandBarButton: UIButton {
...
+    private func updateAccentImage(_ accentImage: UIImage?) {
+        if accentImage != accentImageView?.image {
+            if let accentImage = accentImage {
+                accentImageView = UIImageView(image: accentImage)
+                accentImageView?.translatesAutoresizingMaskIntoConstraints = false
+                if let accentImageView = accentImageView,
+                    let imageView = imageView {
+                    insertSubview(accentImageView, belowSubview: imageView)
+                    NSLayoutConstraint.activate([
+                        accentImageView.centerXAnchor.constraint(equalTo: imageView.centerXAnchor),
```
harrieshin, line 153: nothing guarantee that accentimageview is the same size as imageview. right? is center X anchor good enough?

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -69,6 +69,15 @@ open class CommandBarItem: NSObject {
         }
     }
 
+    /// Image displayed behind the `iconImage` to provide fill or accent if desired
+    @objc public var accentImage: UIImage? {
```
harrieshin, line 73: curious, should we be more conservative that even if user passes in default rendering image, we will treat it like templated image?

nodes11: In this case, I think it's best to not override the renderingMode and let the provider of the `CommandBarItem`s determine the rendering mode. If the icon is tinted with a specific color and with `original` set as template before being set to the `CommandBarItem.accentImage`, changing the renderingMode to `template` would cause the color disappear and the icon would respect the system theme color which is not what the user intended.

nodes11: If we wanted to treat all images as templates, I think we would need to expose an additional `accentImageTintColor` property which could be used to tint the images on `UIUserInterfaceStyle` changes.

nodes11: After some more thinking, I think exposing the accentImageTintColor property to tint the image and forcing the accentImage to use `alwaysTemplate` does remove some ambiguity from how the component works. I've made the necessary changes in the latest revision.

ios/FluentUI/Command Bar/CommandBarButton.swift
nodes11: I think constraining the center X and center Y anchors of the two image views is the best way can handle this without a lot of complicated resizing logic. Trying to handle cases where images have different scale ratios could get messy very fast. I think the responsibility should be the application to provide images of the same size.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -75,6 +75,24 @@ class CommandBarDemoController: DemoController {
...
+                return UIImage(named: "delete24Filled")
+            default:
+                return nil
+            }
+        }
+
+        var accentImageTintColor: UIColor? {
+            switch self {
+            case .delete:
+                return Colors.communicationBlue
```
harrieshin, line 90: can you make it primary color please? in fluent demo app we can easily change from blue to green branding color to accomodate various client sceanrios.

ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -75,6 +75,24 @@ class CommandBarDemoController: DemoController {
...
+            default:
```
harrieshin, line 91: we try to avoid default and spell out all the enum because default tends to make compiler happy and we can easily make mistakes while adding new enums.

nodes11: Update here as well as other places where default is used in this file.

harrieshin (approved):

PR #1295, opened 2022-10-15
Make PillButtonBarItem title mutable and set its font earlier

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Despite the naming, PillButtons and their corresponding PillButtonBarItems get used outside of the PillButtonBar.
We have a specific use where we need the title to be mutable (persistent pill button that represents a current tab switcher selection). This PR adds title mutability to the PillButtonBarItem.

This PR also includes a slight adjustment to how iOS 15+ attribute title creation is done. We now set the font as soon as possible, and only leave the color changing to the updateAppearance function. This resolves a bug where intrinsic size was incorrect before updateAppearance runs when dynamic size was enabled.

### Verification

Made a new demo VC that is tailored to Pill buttons outside of the pill button bar.

Screenshot 2022-10-14 at 5 46 32 PM

| Screenshot 2022-10-14 at 5 55 21 PM | Screenshot 2022-10-14 at 5 56 06 PM |

(no visual change to pill button bar intended)

Tested iOS 14 as well to verify the old non-configuration paths still work.

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -118,6 +118,9 @@ open class PillButton: UIButton {
                                                                   bottom: Constants.bottomInset,
                                                                   trailing: Constants.horizontalInset)
             self.configuration = configuration
+
+            // This updates the attributed title stored in self.configuration,
+            // so it needs to be called after we set it.
             updateAttributedTitle()
 
             configurationUpdateHandler = { [weak self] _ in
@@ -185,7 +188,6 @@ open class PillButton: UIButton {
         } else {
             setTitle(pillBarItem.title, for: .normal)
         }
-
     }
 
     @available(iOS 15, *)
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -62,6 +62,11 @@ open class PillButton: UIButton {
                                                selector: #selector(isUnreadValueDidChange),
                                                name: PillButtonBarItem.isUnreadValueDidChangeNotification,
                                                object: pillBarItem)
+
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(titleValueDidChange),
+                                               name: PillButtonBarItem.titleValueDidChangeNotification,
+                                               object: pillBarItem)
     }
 
     var unreadDotColor: UIColor = Colors.gray100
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -174,6 +179,25 @@ open class PillButton: UIButton {
...
+            setTitle(pillBarItem.title, for: .normal)
+        }
+
+    }
+
+    @available(iOS 15, *)
+    private func updateAttributedTitle() {
+        var attributedTitle = AttributedString(pillBarItem.title)
+        attributedTitle.font = Constants.font
+        configuration?.attributedTitle = attributedTitle
```
nodes11, line 195: I believe there is also an iOS 15+ issue with VoiceOver and the PillButton where the selected item wouldn't read the button title as the accessibility description/label. It's as if the UIButton.Configuration.attributedTitle isn't propagated to the accessibility property as it is should be. Could the accessibility value be set here to resolve this issue?

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -174,6 +179,25 @@ open class PillButton: UIButton {
         setNeedsLayout()
     }
 
+    @objc private func titleValueDidChange() {
+        if #available(iOS 15.0, *) {
+            updateAttributedTitle()
+        } else {
+            setTitle(pillBarItem.title, for: .normal)
+        }
+
```
nodes11, line 188: nit: extra line

nodes11 (approved):

ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -108,12 +113,12 @@ open class PillButton: UIButton {
     private func setupView() {
         if #available(iOS 15.0, *) {
             var configuration = UIButton.Configuration.plain()
-            configuration.attributedTitle = AttributedString(pillBarItem.title)
             configuration.contentInsets = NSDirectionalEdgeInsets(top: Constants.topInset,
                                                                   leading: Constants.horizontalInset,
                                                                   bottom: Constants.bottomInset,
                                                                   trailing: Constants.horizontalInset)
             self.configuration = configuration
+            updateAttributedTitle()
```
anandrajeswaran, line 121: From the change notes, is the order in which you call this important? Relative to the line before? If so, good to add a comment explaining for which calls the order matters and what bug one should look for if they try to change the order in the future

ios/FluentUI/Pill Button Bar/PillButton.swift
lcapkovic: will check this

harrieshin: i have a separate bug for this. can you update your branch first and merge? i can take a look

lcapkovic: Rebased on main to include your fix. Relocated it to the update function because it needs to rerun on every title change.

harrieshin (approved):

PR #1296, opened 2022-10-17
Create functionality to apply multiple shadows on a view

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

**UIKit**
The `applyShadow` function creates `CALayer`s for the two shadows. Both shadow layers are then added as sublayers to the view.

In addition, I created a `Shadowable` protocol to keep track of each view's shadow sublayers. Any view that needs a shadow (or its superview for controls like `Tooltip` that apply shadows to a subview) needs to conform to that protocol. Keeping track of the shadow sublayers ensures we don't have duplicated shadows when resizing or moving views. 

`CardView` was updated to use shadow 2 token as well.

**SwiftUI**
A new `ViewModifier` was created to take in a `ShadowInfo` and applies both shadows to a `View`.

### Verification

The tests were run on the demo app.

| shadow1_light | shadow1_dark |
| shadow2_light | shadow2_dark |

**CardView**
| before_dark | after_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1296)


ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,47 @@ public struct ShadowInfo: Equatable {
...
+        guard let superview = view.superview else {
+            return
+        }
+
+        let shadowView = UIView()
+
+        shadowView.frame = view.frame
+        shadowView.layer.cornerRadius = view.layer.cornerRadius
+
+        superview.insertSubview(shadowView, belowSubview: view)
```
harrieshin, line 80: that's interesting if the shadow is applied for superview of the passined view, if the view resizes or moves around(animation), how does the frame get updated accordingly?

laminesm: I missed that, I can add functionality to update the shadow frame if the view's frame changes.

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,63 @@ public struct ShadowInfo: Equatable {
 
     /// The vertical offset of the shadow for shadow 2.
     public let yTwo: CGFloat
+
+    /// The number that the figma blur needs to be divided by to properly display shadows
```
lyzhan7, line 64: Can you add a link to this documentation: https://github.com/microsoft/apple-ux-guide/blob/gh-pages/Shadows.md

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,63 @@ public struct ShadowInfo: Equatable {
...
+    private let blurDivisor: CGFloat = 2.0
```
lyzhan7, line 65: Optional - in FURN we called this appleShadowBlurAdjustment and did * 0.5 instead of /2.0. Up to you if we want to be more aligned or not, the 'apple' part doesn't make as much sense here but maybe it could be just 'shadowBlurAdjustment'

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,63 @@ public struct ShadowInfo: Equatable {
...
+    }
+
+    private static func initializeShadowLayer(layer: CALayer,
+                                              shadow: ShadowInfo,
+                                              view: UIView,
+                                              isShadowOne: Bool = false) {
+        layer.frame = view.bounds
+        layer.shadowColor = UIColor(dynamicColor: isShadowOne ? shadow.colorOne : shadow.colorTwo).cgColor
+        layer.shadowRadius = isShadowOne ? shadow.blurOne : shadow.blurTwo
+        layer.shadowOpacity = 1
```
lyzhan7, line 111: Hmm in FURN we didn't do shadowOpacity = 1, we had to extract the opacity value from the alpha component of the color. The screenshots you have look fine though

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -4,6 +4,7 @@
 //
 
 import CoreGraphics
+import UIKit
 
 /// Represents a two-part shadow as used by FluentUI.
 public struct ShadowInfo: Equatable {
```
lyzhan7, line 10: Optional: just an observation reading some of your code, it seems like some of it could be simplified by splitting the ShadowInfo struct into two structs, one that's just 'color, blur, x, y' and then another that's just 'shadow1, shadow2' (or in FURN we call the two shadows 'key' and 'ambient'). Then you might not need as many isShadowOne? checks.

Ex.
```
initializeShadowLayer(layer: shadow1, shadow: shadowInfo, view: view, isShadowOne: true)
initializeShadowLayer(layer: shadow2, shadow: shadowInfo, view: view)
```
could be replaced with
```
initializeShadowLayer(layer: shadow1, shadow: shadowInfo.shadow1, view: view)
initializeShadowLayer(layer: shadow2, shadow: shadowInfo.shadow2, view: view)
```

And the function itself would be a little simpler

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -27,11 +28,11 @@ public struct ShadowInfo: Equatable {
                 xTwo: CGFloat,
                 yTwo: CGFloat) {
         self.colorOne = colorOne
-        self.blurOne = blurOne
+        self.blurOne = blurOne / blurDivisor
         self.xOne = xOne
         self.yOne = yOne
         self.colorTwo = colorTwo
-        self.blurTwo = blurTwo
+        self.blurTwo = blurTwo / blurDivisor
```
harrieshin, line 35: can we just make sure our global token is already have the right apple value rather than doing a divison?  @lyzhan7 what did you do for furn?

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
laminesm: In UIKit we need to set the shadowOpacity to 1 since the alpha is already set in shadowColor.

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
laminesm: We had a team discussion about this some time ago and agreed to use the figma values and just adjust them.

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -50,6 +50,14 @@ extension View {
     func showsLargeContentViewer(text: String? = nil, image: UIImage? = nil) -> some View {
         modifier(LargeContentViewerModifier(text: text, image: image))
     }
+
+    /// Applies multiple shadows on a View
+    /// - Parameters
+    ///  - shadowInfo: The values of the two shadows to be applied
+    /// - Returns: The modified view.
+    func applyShadow(shadowInfo: ShadowInfo) -> some View {
+            modifier(ShadowModifier(shadowInfo: shadowInfo))
```
mischreiber, line 59: Nit: spacing

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
mischreiber: Might be worth leaving a comment explaining that, for future engineers who see this code.

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,54 @@ public struct ShadowInfo: Equatable {
...
     /// The vertical offset of the shadow for shadow 2.
     public let yTwo: CGFloat
+
+    /// The number that the figma blur needs to be divided by to properly display shadows
+    private let shadowBlurAdjustment: CGFloat = 0.5
+}
+
+public class ShadowUtil {
+
+    public static func applyShadow(_ shadowInfo: ShadowInfo, for view: UIView) {
```
mischreiber, line 70: Question: could this just be an extension on `ShadowInfo` to eliminate the need for a new class? So instead of calling 

```swift
ShadowUtil.applyShadow(shadowInfo, for: view)
```

you could just call

```swift
shadowInfo.apply(to: view)
```

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,54 @@ public struct ShadowInfo: Equatable {
...
+        guard let superview = view.superview, var shadowable = (view as? Shadowable) ?? (view.superview as? Shadowable)  else {
+            return
```
mischreiber, line 72: Maybe add an `assertionFailure()` here with a note that someone attempted to apply a shadow to a non-`Shadowable` view?

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
lyzhan7: Also might want to update the comment now not to say "divide". Maybe just "adjusted by"

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
lyzhan7: We make the adjustment inside FURN as well, rather than in the token pipeline. I think we considered updating the tokens values but decided against it, @mischreiber @Saadnajmi might remember more about this

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
laminesm: Thanks for catching that!

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -59,4 +60,54 @@ public struct ShadowInfo: Equatable {
...
+        layer.shadowOpacity = 1
+        layer.shadowOffset = CGSize(width: isShadowOne ? shadow.xOne : shadow.xTwo,
+                                    height: isShadowOne ? shadow.yOne : shadow.yTwo)
+        layer.needsDisplayOnBoundsChange = true
+        layer.cornerRadius = view.layer.cornerRadius
+        layer.backgroundColor = view.backgroundColor?.cgColor
+    }
+}
+
+public protocol Shadowable {
```
harrieshin, line 110: please add documentation for the shadowable and its pbulic variables

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
```diff
@@ -27,11 +28,11 @@ public struct ShadowInfo: Equatable {
                 xTwo: CGFloat,
                 yTwo: CGFloat) {
         self.colorOne = colorOne
-        self.blurOne = blurOne
+        self.blurOne = blurOne * shadowBlurAdjustment
```
harrieshin, line 31: @lyzhan7 : so for FURN global token is iOS shawdow already halved from the token definition? or is furn programmatically dividing it by 2?

Saadnajmi: FURN programmatically divides by 2. https://github.com/microsoft/fluentui-react-native/blob/3f16965e77fc0e4aedd093b2206a2e523c6bcdca/packages/experimental/Shadow/src/shadowStyle.ts#L11

Sambit650 (approved):

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
lyzhan7: Sorry my FUA emails are going to a different folder, still trying to figure out the right email filters. Yes we decided that we should programmatically divide by 2

ios/FluentUI/Core/Theme/Tokens/ShadowInfo.swift
lyzhan7: @laminesm were you planning to make this change?

harrieshin (approved):

PR #1297, opened 2022-10-19
ios15+ VoiceOver doesn't seem to read selected PillButton

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There seems to be an Apple bug that selected Button doesn't get read out by VoiceOver when using UIButton.Configuration to set the button title rather using setTitle().  For now the workaround is to manually set the accessibility label.

### Verification
Test both on iOS 16 simulator and on device

| image | Screen Shot 2022-10-18 at 8 51 44 PM |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -109,6 +109,10 @@ open class PillButton: UIButton {
         if #available(iOS 15.0, *) {
             var configuration = UIButton.Configuration.plain()
             configuration.attributedTitle = AttributedString(pillBarItem.title)
+
+            // Workaround for Apple bug: when UIButton.Configuration is used with UIControl's isSelected = true, accessibilityLabel doesn't get set automatically
+            accessibilityLabel = pillBarItem.title
+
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -109,6 +109,10 @@ open class PillButton: UIButton {
         if #available(iOS 15.0, *) {
             var configuration = UIButton.Configuration.plain()
             configuration.attributedTitle = AttributedString(pillBarItem.title)
+
+            // Workaround for Apple bug: when UIButton.Configuration is used with UIControl's isSelected = true, accessibilityLabel doesn't get set automatically
+            accessibilityLabel = pillBarItem.title
```
anandrajeswaran, line 114: This is only called in init, right? So if we change the title later this won't be updated? Unfortunately that is an actual scenario in our usage of PillButton.

harrieshin: well.. this version of the PR doesn't have lukas change that actually makes the pillbutton title mutable.

lcapkovic (approved):

PR #1219, opened 2022-09-06
Move `fluentTheme` from window to view

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

With this change, any arbitrary `UIView` can now host its own `FluentTheme`. Any view's `fluentTheme` property will now look up the nearest parent with a custom theme and use that, instead of relying just on the window.

Side effect: Components should now check whether their view is a descendent of the changing view before redrawing, as the sender is not guaranteed to be a `UIWindow`.

### Verification

Binary size comparison: this saves 19.1 kB. Neat!

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| 31,882,819 bytes | 31,863,203 bytes |

And a demo that theme changing still works as expected:

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before | after |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [x] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1219)

PR #1299, opened 2022-10-20
Tokenize Tooltip and add title functionality

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Tokenize Tooltip: Due to the static-nature and dependencies of Tooltip, Tooltip passes the tokens to TooltipView and TooltipPositionController via their inits
- Add functionality that enables titles for Tooltips

| Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-20 at 09 07 12 |Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 14 05 04|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 11 40 09|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 14 05 12|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 11 42 07|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 14 00 32|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 11 41 38|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-31 at 00 24 58|
|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 11 41 17|Simulator Screen Shot - iPad Pro (12 9-inch) (5th generation) - 2022-10-28 at 14 01 03|

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1299)


ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -4,12 +4,26 @@
 //
 
 import UIKit
+import Combine
```
harrieshin, line 7: just curious. why do we need combine?

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -63,29 +77,77 @@ open class Tooltip: NSObject {
...
+    ///   - title: The optional bolded text to be displayed above the message on the new tooltip view.
     ///   - anchorView: The view to point to with the new tooltip's arrow.
     ///   - preferredArrowDirection: The preferrred direction for the tooltip's arrow. Only the arrow's axis is guaranteed; the direction may be changed based on available space between the anchorView and the screen's margins. Defaults to down.
     ///   - offset: An offset from the tooltip's default position.
     ///   - screenMargins: The margins from the window's safe area insets used for laying out the tooltip. Defaults to 16.0 pts on all sides.
     ///   - dismissMode: The mode of tooltip dismissal. Defaults to tapping anywhere.
     ///   - onTap: An optional closure used to do work after the user taps
-    @objc public func show(with message: String, for anchorView: UIView, preferredArrowDirection: ArrowDirection = .down, offset: CGPoint = CGPoint(x: 0, y: 0), screenMargins: UIEdgeInsets = defaultScreenMargins, dismissOn dismissMode: DismissMode = .tapAnywhere, onTap: (() -> Void)? = nil) {
+    @objc public func show(with message: String,
```
harrieshin, line 105: to avoid bunch of objc clients to be re-written we should have a separate convenience metod that has title param and leave this as is.

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: For `AnyCancellable` (type of tokenSetSink)

ios/FluentUI/Tooltip/TooltipPositionController.swift
```diff
@@ -76,12 +76,30 @@ class TooltipPositionController {
...
+    private let titleLabelFont: UIFont
     private let boundingRect: CGRect
     private let offset: CGPoint
 
     private var tooltipSize: CGSize = .zero
-
-    init(anchorView: UIView, message: String, boundingRect: CGRect, preferredArrowDirection: Tooltip.ArrowDirection, offset: CGPoint, arrowMargin: CGFloat, arrowWidth: CGFloat) {
+    init(anchorView: UIView,
+         message: String,
+         title: String? = nil,
```
harrieshin, line 91: because tooltippositioncontroller is a private clss anyway, can't we just pass in the tooltiptokenset rather than adding tons of new param? any time we update a token we have to update multiple different functions?

harrieshin: then we can get rid of bunch of private let too?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: but if the title param is given a default value of nil, the existing callers of the show method don't need to be changed, right?

harrieshin: swift works that way. not Objc

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -4,12 +4,26 @@
...
 /// A styled tooltip that is presented anchored to a view.
 @objc(MSFTooltip)
-open class Tooltip: NSObject {
+open class Tooltip: UIView, TokenizedControlInternal {
+    // MARK: - TokenizedControl
+    public typealias TokenSetKeyType = TooltipTokenSet.Tokens
+    public var tokenSet: TooltipTokenSet = .init()
+    var tokenSetSink: AnyCancellable?
+
+    @objc private func themeDidChange(_ notification: Notification) {
```
laminesm, line 19: With #1219 change you might need to add this check
```
            return
}
```

laminesm: Also, did you add a notification observer for theme change in the init?

ios/FluentUI/Tooltip/TooltipTokenSet.swift
```diff
@@ -0,0 +1,71 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Design token set for the `Tooltip` control.
+public class TooltipTokenSet: ControlTokenSet<TooltipTokenSet.Tokens> {
+    public enum Tokens: TokenSetKey {
+        /// The color of the body of the tooltip.
+        case tooltipColor
```
laminesm, line 12: nit: maybe you could add a new line between cases. It'd improve readability and be consistent with how other controls' TokenSets are commented.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -136,12 +153,127 @@ class TooltipView: UIView {
...
+        layer.insertSublayer(ambientShadow, at: 0)
+
+        let perimeterShadow = CALayer()
+        perimeterShadow.frame = bounds
+        perimeterShadow.shadowPath = UIBezierPath(roundedRect: bounds, cornerRadius: backgroundCornerRadius).cgPath
+        perimeterShadow.shadowColor = UIColor(dynamicColor: shadowInfo.colorTwo).cgColor
+        perimeterShadow.shadowOpacity = 1
+        perimeterShadow.shadowOffset = CGSize(width: shadowInfo.xTwo, height: shadowInfo.yTwo)
+        perimeterShadow.shadowRadius = shadowInfo.blurTwo
+        layer.insertSublayer(perimeterShadow, at: 0)
```
laminesm, line 220: Seems like you are adding both shadow layers on the same index.

jeaniehuynh: it just inserts it at that place! so the first one gets moved to index 1

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: I tried to do this, but am crashing when not passing in a title because it can't tell which method I'm trying to use. Should I just give the function a different name?

harrieshin:
can you please add increased contrast mode, dark mode, and the regular tooltip(without title) in the list of screenshot please?

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -4,12 +4,33 @@
...
 import UIKit
+import Combine
 
 // MARK: Tooltip
 
 /// A styled tooltip that is presented anchored to a view.
 @objc(MSFTooltip)
-open class Tooltip: NSObject {
+open class Tooltip: NSObject, TokenizedControlInternal {
+    var fluentTheme: FluentTheme {
```
harrieshin, line 14: we still want the public vars on top and private var below grouped together? https://microsoft.github.io/swift-guide/FileOrganization.html

harrieshin:
can we also show case all the different arrow direction? since there was some code changes around that logic?

harrieshin:
larger text?

ios/FluentUI/Tooltip/TooltipPositionController.swift
```diff
@@ -11,8 +11,42 @@ class TooltipPositionController {
...
+                                                     arrowDirection: preferredArrowDirection,
+                                                     tokenSet: tokenSet)
+        let backupSize = TooltipView.sizeThatFits(backupBoundingRect.size,
+                                                  message: message,
+                                                  title: title,
+                                                  arrowDirection: preferredArrowDirection.opposite,
+                                                  tokenSet: tokenSet)
+
+        var usePreferred = true
+        if preferredArrowDirection.isVertical {
```
harrieshin, line 29: these 2 if statements can't be combined?

ios/FluentUI/Tooltip/TooltipPositionController.swift
```diff
@@ -11,8 +11,42 @@ class TooltipPositionController {
...
+            if preferredBoundingRect.height < preferredSize.height && backupBoundingRect.height >= backupSize.height {
+                usePreferred = false
+            }
+        } else {
```
harrieshin, line 33: can't this be one else if statement?

ios/FluentUI/Tooltip/TooltipTokenSet.swift
```diff
@@ -0,0 +1,93 @@
...
+        case messageLabelTextStyle
+
+        /// The TextStyle of the title label.
+        case titleLabelTextStyle
+
+        /// The horizontal padding between the text and edges of the tooltip.
+        case paddingHorizontal
+
+        /// The vertical padding between the text and edges of the tooltip with both a title and message.
+        case paddingVerticalWithTitle
```
harrieshin, line 33: similar to notfication tokens, i thought the spacings are not tokenizable. we should only expose public tokens that we are going to repsect. text color, style seems reasonable. maybe get clarification from design team?

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -136,12 +147,115 @@ class TooltipView: UIView {
...
+                                                                 tokenSet: tokenSet,
+                                                                 isMessage: true)
+        messageLabel.frame.size = preferredMessageSize
+        if let titleLabel = titleLabel, let title = title {
+            let preferredTitleSize = TooltipView.labelSizeThatFits(textContainer.frame.size,
+                                                                   text: title,
+                                                                   tokenSet: tokenSet,
+                                                                   isMessage: false)
+            titleLabel.frame.size = preferredTitleSize
+            messageLabel.frame.origin = CGPoint(x: 0, y: titleLabel.frame.height + tokenSet[.spacingVertical].float)
```
harrieshin, line 181: please double check RTL layout as well.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -98,12 +86,35 @@ class TooltipView: UIView {
...
+
+        if let titleLabel = titleLabel, let title = title {
+            titleLabel.text = title
+            titleLabel.textAlignment = textAlignment
+            titleLabel.isAccessibilityElement = false
+        }
+
+        addSubview(textContainer)
 
         // Shadow
```
harrieshin, line 98: can adding shadow be its own private function. i think this function is getting way to big

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -98,12 +86,35 @@ class TooltipView: UIView {
         messageLabel.text = message
         messageLabel.textAlignment = textAlignment
         messageLabel.isAccessibilityElement = false
```
harrieshin, line 88: nit: should isaccessibilityelement move to the uilabel's lazy var

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -98,12 +86,35 @@ class TooltipView: UIView {
...
-        addSubview(messageLabel)
+
+        if let titleLabel = titleLabel, let title = title {
```
harrieshin, line 90: curious why we need to check both existence of title AND titlelabel. shouldn't title be good enough?

ios/FluentUI/Tooltip/TooltipView.swift
harrieshin: same thing for titlelable

ios/FluentUI/Tooltip/TooltipView.swift
jeaniehuynh: already checked!

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -63,30 +84,65 @@ open class Tooltip: NSObject {
     private var gestureView: UIView?
     private var dismissMode: DismissMode = .tapAnywhere
 
-    private override init() { }
+    private override init() {
```
harrieshin, line 87: `    @objc public static let defaultScreenMargins = UIEdgeInsets(top: Constants.defaultMargin, left: Constants.defaultMargin, bottom: Constants.defaultMargin, right: Constants.defaultMargin)`

this is interesting. is this being deprecated for tokens?

harrieshin, line 87: `    /// The alignment of the text in the tooltip. Defaults to natural alignment (left for LTR languages, right for RTL languages).
    @objc public var textAlignment: NSTextAlignment = .natural`

@mischreiber : to text alignment make sense as a token or as a property?

ios/FluentUI/Tooltip/TooltipView.swift
jeaniehuynh: we'd have to force unwrap titleLabel otherwise

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: not sure what you mean, can you elaborate?

harrieshin: before in fluent1, we had exposed bunch of property to customize tooltip. if we are moving to token system, and we think this is something we continue to support as customization, we should move over?

jeaniehuynh: FYI: This is API breaking because we'll need to remove screenMargins parameter from show methods. Is that okay, @harrieshin?

harrieshin: i think we should have deprecated declare so we can slowly migrate people over. do you happen to know who uses this btw?

jeaniehuynh: sounds good and not extensively, I only know of its usage in CocoaUI for teaching callouts

harrieshin:
the dark mode colors shouldn't have change. we should still follow fluent1 colors until fluent2-colors token branch is merged to main branch.

ios/FluentUI/Tooltip/TooltipTokenSet.swift
jeaniehuynh: talked with Mike since he made the initial change, we're going to make them token constants!

apple/Resources/FluentUI-apple.xcassets/FluentColors/gray900.colorset/Contents.json
```diff
@@ -1,20 +1,20 @@
...
     {
-      "idiom" : "universal",
       "color" : {
         "color-space" : "srgb",
         "components" : {
-          "red" : "0x21",
           "alpha" : "1.000",
           "blue" : "0x21",
-          "green" : "0x21"
+          "green" : "0x21",
```
harrieshin, line 9: why did this file change?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
```diff
@@ -20,20 +20,23 @@ class TooltipDemoController: DemoController {
         navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Show on title", style: .plain, target: self, action: #selector(showTitleTooltip))
 
         container.addArrangedSubview(createButton(title: "Show single-line tooltip below", action: { sender in
-            Tooltip.shared.show(with: "This is pointing up.", for: sender, preferredArrowDirection: .up)
+            Tooltip.shared.show(with: "This is pointing up.", title: nil, for: sender, preferredArrowDirection: .up)
```
harrieshin, line 23: why doe did you need to update the democontroller calls if it is just nil?

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -133,6 +101,34 @@ open class Tooltip: NSObject {
...
+    ///
+    /// - Parameters:
+    ///   - message: The text to be displayed on the new tooltip view.
+    ///   - anchorView: The view to point to with the new tooltip's arrow.
+    ///   - preferredArrowDirection: The preferrred direction for the tooltip's arrow. Only the arrow's axis is guaranteed; the direction may be changed based on available space between the anchorView and the screen's margins. Defaults to down.
+    ///   - offset: An offset from the tooltip's default position.
+    ///   - screenMargins: The margins from the window's safe area insets used for laying out the tooltip. Defaults to 16.0 pts on all sides.
+    ///   - dismissMode: The mode of tooltip dismissal. Defaults to tapping anywhere.
+    ///   - onTap: An optional closure used to do work after the user taps
+    @available(*, deprecated, renamed: "show(with:title:for:preferredArrowDirection:offset:dismissOn:onTap:)")
```
harrieshin, line 115: only the screenmargin is really deprecated. the rest we can still support as is.

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -159,6 +155,87 @@ open class Tooltip: NSObject {
...
+    }
+
+    @objc(MSFTooltipDismissMode)
+    public enum DismissMode: Int {
+        case tapAnywhere
+        case tapOnTooltip
+        case tapOnTooltipOrAnchor
+    }
+
+    @available(*, deprecated, message: "Screen margins value has been tokenized. Setting this var no longer has any effect and it will be removed in a future update.")
```
harrieshin, line 196: give info what api they should be using.

apple/Resources/FluentUI-apple.xcassets/FluentColors/gray900.colorset/Contents.json
jeaniehuynh: oh oops, i must've clicked something on accident when checking the hex. i'll revert

ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
jeaniehuynh: just to get rid of the deprecation warnings, should I leave it?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: there's no other api to be using. the screen margins is built into the token system now

ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
harrieshin: should there be 3 different show() functions now?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: removing the deprecation for the method turns the usage of the deprecated screenmargins into a compiler error (not just a warning)

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -97,13 +34,34 @@ class TooltipView: UIView {
...
+
+        addSubview(textContainer)
 
         // Shadow
-        layer.shadowOpacity = Constants.shadowOpacity
-        layer.shadowRadius = Constants.shadowRadius
-        layer.shadowOffset = CGSize(width: 0.0, height: Constants.shadowOffset)
+        let backgroundCornerRadius = tokenSet[.backgroundCornerRadius].float
+        let shadowInfo = tokenSet[.shadowInfo].shadowInfo
+        let ambientShadow = CALayer()
```
harrieshin, line 48: this this code different from updateShadows() function you just created?

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -136,12 +94,141 @@ class TooltipView: UIView {
...
+        self.frame = positionController.tooltipRect
+
+        // Update shadows
+        updateShadows(tokenSet: tokenSet)
+    }
+
+    private func updateShadows(tokenSet: TooltipTokenSet) {
+        let backgroundCornerRadius = tokenSet[.backgroundCornerRadius].float
+        let shadowInfo = tokenSet[.shadowInfo].shadowInfo
+        let ambientShadow: CALayer = (layer.sublayers?[1])!
```
harrieshin, line 209: we should not rely on force unwrap which give higher chance of crash during runtime.

ios/FluentUI/Tooltip/TooltipView.swift
jeaniehuynh: this adds the shadows (done once in the init) and the new method updates the existing shadows (readding the shadows makes the shadow opacity get darker and darker with the added layers)

harrieshin: with bit of the refactor, it seems like the code can be reused ?

ios/FluentUI/Tooltip/TooltipTokenSet.swift
```diff
@@ -0,0 +1,98 @@
...
+
+        /// The width of the arrow of the tooltip.
+        case arrowWidth
+    }
+
+    init() {
+        super.init { token, theme in
+            switch token {
+            case .tooltipColor:
+                return .dynamicColor { DynamicColor(light: ColorValue(r: 33.0 / 255.0,
```
harrieshin, line 43: nit: if you do these floating math it is little bit inaccurate compare to hex itself. we can't use DynamicColor(light: ColorValue(0xFFFFFF) ? like you did in shimmer?

jeaniehuynh: what about the alpha?

harrieshin (approved):
Assuming there might be update in the max width logic in next PR and one more comment on the color, signing off. thanks.

ios/FluentUI/Tooltip/TooltipTokenSet.swift
harrieshin: argh. great point!

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -213,7 +213,7 @@ class TooltipView: UIView {
                                           text: String,
                                           tokenSet: TooltipTokenSet,
                                           isMessage: Bool) -> CGSize {
-        let boundingWidth = min(tokenSet[.maximumWidth].float - (2 * TooltipTokenSet.paddingHorizontal), size.width)
+        let boundingWidth = UIApplication.shared.preferredContentSizeCategory.isAccessibilityCategory ? size.width : tokenSet[.maximumWidth].float - (2 * TooltipTokenSet.paddingHorizontal)
```
harrieshin, line 216: we should be avoiding using UIApplication.Shared usuage @jeaniehuynh

jeaniehuynh: ah okay, what's the alternative? also curious as to why

PR #1300, opened 2022-10-20
Fix `TabBarItemView` losing tint color when backgrounded

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There's a known issue with iOS 16 instances of `UIImageView` dropping the `tintColor` of their `highlightedImage`, especially when the app is backgrounded and restored. While we wait for an official fix, we can work around the issue by manually swapping out the main `image` property.

### Verification

Verified that the old issue does not repro, and that images continue to highlight as expected.

#### Before:
https://user-images.githubusercontent.com/4934719/197026053-e3083e3d-130c-481a-9148-0be1f9d64517.mp4

#### After:
https://user-images.githubusercontent.com/4934719/197026049-2b4f3c63-e99f-40ff-ae74-818511905800.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1300)

### DIFF
--- a/ios/FluentUI/Tab Bar/TabBarItemView.swift
+++ b/ios/FluentUI/Tab Bar/TabBarItemView.swift
@@ -20,6 +20,7 @@ class TabBarItemView: UIControl {
         didSet {
             titleLabel.isHighlighted = isSelected
             imageView.isHighlighted = isSelected
+            updateImage()
             updateColors()
             if isSelected {
                 accessibilityTraits.insert(.selected)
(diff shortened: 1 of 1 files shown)


laminesm (approved):

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -242,10 +243,16 @@ class TabBarItemView: UIControl {
...
     }
 
-    private func updateLayout() {
-        imageView.image = item.unselectedImage(isInPortraitMode: isInPortraitMode, labelIsHidden: titleLabel.isHidden)
-        imageView.highlightedImage = item.selectedImage(isInPortraitMode: isInPortraitMode, labelIsHidden: titleLabel.isHidden)
+    private func updateImage() {
+        // Normally we'd set imageView.image and imageView.highlightedImage separately. However, there's a known issue with
+        // UIImageView in iOS 16 where highlighted images lose their tint color in certain scenarios. While we wait for a fix,
+        // this is a straightforward workaround that gets us the same effect without triggering the bug.
+        imageView.image = imageView.isHighlighted ?
```
harrieshin, line 250: should we do available iOS 16?

harrieshin (approved):

ios/FluentUI/Tab Bar/TabBarItemView.swift
mischreiber: I thought about leaving the old code for iOS 14/15, but at the end of the day I'd rather we have one supported solution, since there's no real performance impact or other negative side-effect.

PR #1147, opened 2022-08-09
[Fluent 2 iOS] TableViewCell colors update (1/2)

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

`TableViewCell` colors were updated to match fluent 2 specs. 

**NOTE**: These changes do not include:
- Color changes to `ActionsCell` -> This will be updated in a separate PR when the work on shared colors alias tokens is done.
- Removal of `Colors` extension -> A new PR will be up shortly dealing with this and addressing the dependencies.

### Verification

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before_1_light | after_1_light |
| before_1_dark | after_1_dark |
| before_2_light | after_2_light |
| before_2_dark | after_2_dark |
| before_3_light | after_3_light |
| before_3_dark | after_3_dark |
| before_4_light | after_4_light |
| before_4_dark | after_4_dark |
| before_grouped_light | after_grouped_light |
| before_grouped_dark | after_grouped_dark |
| before_selected_light | after_selected_light |
| before_selected_dark | after_selected_dark |
| before_selected_grouped_light | after_selected_grouped_light |
| before_selected_grouped_dark | after_selected_grouped_dark |
| before_dark_elevated | after_dark_elevated |
| before_dark_elevated_grouped | after_dark_elevated_grouped |
| before_light | after_light |
| before_grouped_light | after_grouped_light |
| before_dark | after_dark |
| before_grouped_dark | after_grouped_dark |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1147)

PR #1320, opened 2022-10-25
Fix TableViewCell grouped style header color

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A bug was filed pointing out that the `TableViewCell` grouped style looked incorrect as the background was all white instead of being light grey in the header/footer areas. To fix the issue, `TableViewHeaderFooterView` background has been reverted to clear color as it used to be in fluent 1. 

### Verification

| before_dark | after_dark |
| before_keyboard_light | after_keyboard_light |
| before_keyboard_dark | after_keyboard_dark |
| before_grouped_tvc_light | tvc_light |
| before_grouped_tvc_dark | tvc_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1320)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCellTokenSet.swift
+++ b/ios/FluentUI/Table View/TableViewCellTokenSet.swift
@@ -81,10 +81,16 @@ public class TableViewCellTokenSet: ControlTokenSet<TableViewCellTokenSet.Tokens
         super.init { token, theme in
             switch token {
             case .backgroundColor:
-                return .dynamicColor { theme.aliasTokens.colors[.background1] }
+                return .dynamicColor {
+                    DynamicColor(light: theme.aliasTokens.colors[.background1].light,
+                                 dark: theme.aliasTokens.colors[.background1].dark)
+                }
 
             case .backgroundGroupedColor:
-                return .dynamicColor { theme.aliasTokens.colors[.canvasBackground] }
+                return .dynamicColor {
+                    DynamicColor(light: theme.aliasTokens.colors[.canvasBackground].light,
+                                 dark: theme.aliasTokens.colors[.canvasBackground].dark)
+                }
 
             case .cellBackgroundColor:
                 return .dynamicColor { theme.aliasTokens.colors[.background1] }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Table View/TableViewCellTokenSet.swift
```diff
@@ -81,10 +81,16 @@ public class TableViewCellTokenSet: ControlTokenSet<TableViewCellTokenSet.Tokens
         super.init { token, theme in
             switch token {
             case .backgroundColor:
-                return .dynamicColor { theme.aliasTokens.colors[.background1] }
+                return .dynamicColor {
+                    DynamicColor(light: theme.aliasTokens.colors[.background1].light,
```
harrieshin, line 85: why do we need this change? seems like both light and dark are coming from .backgoroudn1?

harrieshin, line 85: is there an intended usage of background1 to still use the elevated colors?

harrieshin:
"As part of this PR, I made changes to TableViewCell's background colors to not use elevated colors so it stays consistent with its behavior in fluent 1." i am confused by this. Fluent 1 tableviewcell colors to change when elevated. please test iPad. @sophialee0416 you would have more context?

laminesm:
> "As part of this PR, I made changes to TableViewCell's background colors to not use elevated colors so it stays consistent with its behavior in fluent 1." i am confused by this. Fluent 1 tableviewcell colors to change when elevated. please test iPad. @sophialee0416 you would have more context?

Initially I looked at #1147 but looking at #1148 it seems like only the tableview background had elevated colors.

ios/FluentUI/Table View/TableViewCellTokenSet.swift
laminesm: The following controls use BG1:

- `BadgeField`
- `CardNudge`
- `Notification` (primary outline)
- `PersonaButtonCarousel`
- `PersonaListView`
- `SegmentedControl`

So far only `SegmentedControl` explicitly ensures it doesn't use elevated colors. I don't know how to determine if any of these controls 'need' elevated colors.

harrieshin:
Please add screenshot of ipad split mode with group styled tableviewcell. In main branch the cell and the background color changes

harrieshin:
Please check the ipad keyboard selection color of the cell too

harrieshin:
whatever you do. make sure actioncells and booleancells etc are also consistent behavior.

harrieshin (approved):

PR #1321, opened 2022-10-25
Fix bottom sheet VO order when it's hosting a navigation controller

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Bottom sheet VO order is incorrect when a navigation controller is hosted in the sheet. It seems like there is some default system behavior that gives nested navigation controllers a higher VO priority.

#### In regular use, VO order is:
Parent nav controller
main content (or dimming view when visible)
sheet resizing handle
sheet header view (if present)
sheet expanded content.

#### But when the sheet hosts a navigation controller, the order changes to this:
Parent nav controller
(skips main content, resizing handle, and sheet header)
Sheet-hosted nav controller
the rest

To fix this, let's explicitly specify the order of a11y elements in the bottom sheet view.

### Verification

Modified the demo app to include nav controller case.

#### Broken - before

https://user-images.githubusercontent.com/3610850/197902663-55a95c16-1c6d-4dc4-a7ac-825f1da96f66.mov

#### Fixed - after

https://user-images.githubusercontent.com/3610850/197902697-8b86ba31-3c75-46c8-b87d-c1197a9fb457.mov

Tested on device and simulator with a11y inspector.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1321)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -417,6 +417,8 @@ public class BottomSheetController: UIViewController {
         stackView.axis = .vertical
         stackView.translatesAutoresizingMaskIntoConstraints = false
 
+        // Some types of content (like navigation controllers) can mess up the VO order.
+        // Explicitly specifying a11y elements helps prevents this.
         bottomSheetContentView.accessibilityElements = [resizingHandleView]
 
         if let headerView = headerContentView {
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -417,11 +417,15 @@ public class BottomSheetController: UIViewController {
         stackView.axis = .vertical
         stackView.translatesAutoresizingMaskIntoConstraints = false
 
+        bottomSheetContentView.accessibilityElements = [resizingHandleView]
+
         if let headerView = headerContentView {
             stackView.addArrangedSubview(headerView)
+            bottomSheetContentView.accessibilityElements?.append(headerView)
         }
 
         stackView.addArrangedSubview(expandedContentView)
+        bottomSheetContentView.accessibilityElements?.append(expandedContentView)
         bottomSheetContentView.addSubview(stackView)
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -417,11 +417,15 @@ public class BottomSheetController: UIViewController {
         stackView.axis = .vertical
         stackView.translatesAutoresizingMaskIntoConstraints = false
 
+        bottomSheetContentView.accessibilityElements = [resizingHandleView]
```
harrieshin, line 420: not clear why we need to manually add it ourselves?

harrieshin: if the dimmingView is shown then do we need to mannually inject it too?

lcapkovic: Added a comment which explains why we're doing this.

lcapkovic: >if the dimmingView is shown then do we need to mannually inject it too?
No, this only enforces ordering within the actual sheet view. Dimming view is outside of this and gains focus before any of these.

harrieshin: i am sure you already tried it but https://developer.apple.com/documentation/objectivec/nsobject/1615143-shouldgroupaccessibilitychildren on the bottomsheetcontentview doesn't help? VO doing its own thing?

lcapkovic: oh I haven't tried that! Let me see...

lcapkovic: Tried `shouldGroupAccessibilityChildren`, doesn't fix this issue unfortunately.

harrieshin (approved):

PR #1330, opened 2022-10-28
[Fluent 2 iOS] Drawer colors update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

`Drawer` was updated to match its fluent 2 tokens. Color variables `drawerBackground` and `popoverBackground` were removed as they're not used outside of `DrawerController`, hence no longer need objc exposure.

### Verification

| before_normal_light | after_normal_light |
| before_normal_dark | after_normal_dark |
| before_popover_light | after_popover_light |
| before_popover_dark | after_popover_dark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1330)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -123,13 +110,23 @@ open class DrawerController: UIViewController {
...
     /// Set `backgroundColor` to customize background color of the drawer
-    @objc open var backgroundColor: UIColor = Colors.Drawer.background {
+    @objc open lazy var backgroundColor: UIColor = DrawerController.drawerBackgroundColor(fluentTheme: view.fluentTheme) {
         didSet {
             useCustomBackgroundColor = true
             view.backgroundColor = backgroundColor
         }
     }
 
+    @objc static func drawerBackgroundColor(fluentTheme: FluentTheme) -> UIColor {
```
harrieshin, line 120: this this static function actually objc friendly? is fluenttheme actually be created in objc ? @mischreiber

mischreiber: Yep, `fluentTheme` is fully Objective-C friendly!

mischreiber, line 120: Can these two static functions be `private`? Doesn't look like they're referenced anywhere else.

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -123,13 +110,23 @@ open class DrawerController: UIViewController {
     }
 
     /// Set `backgroundColor` to customize background color of the drawer
-    @objc open var backgroundColor: UIColor = Colors.Drawer.background {
+    @objc open lazy var backgroundColor: UIColor = drawerBackgroundColor(fluentTheme: view.fluentTheme) {
         didSet {
             useCustomBackgroundColor = true
             view.backgroundColor = backgroundColor
```
slzublid, line 116: A check for `viewIsLoaded` should be added before accessing the view, similar to other didSets in this file.

```
if viewIsLoaded {
     view.backgroundColor = backgroundColor
}
```

Accessing the `view` before is loaded will force a `loadView()`, which will call `viewDidLoad()` and a custom `presentationStyle` will be ignored if the other of setting the properties are not right.

Issue filed here: https://github.com/microsoft/fluentui-apple/issues/1338

slzublid, line 116: Hi Lamine, could you fix this issue in this same PR?

a check for `isViewLoaded` should be added before accessing the `view`, or it will cause the `loadView()` and `viewDidLoad()` to be called before other properties are set and some undesired behaviour could happen, like the presentationStyle be ignored.

```
if isViewLoaded {
    view.backgroundColor = backgroundColor
}
```

I filed an issue here: #1338 

I have a workaround, which is to set backgroundColor last, but it would be good to have this fixed for future versions.

Thanks!

laminesm: As discussed offline, I'll be making the fix in a different PR targeting the main branch to get it out in the next fluent release. Thank you for filing the issue!

harrieshin (approved):

PR #1333, opened 2022-10-31
Remove UIApplication.shared usage

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We avoid using UIApplication.shared, so let's get traitCollection from view/window instead.

### Text Size Change:
https://user-images.githubusercontent.com/31874971/200698442-7ab0862c-59f4-4eea-a1c6-be34c1e4d867.MP4

### Window Size Change:
https://user-images.githubusercontent.com/31874971/200698520-9954fe37-546d-4c87-b973-576a075b1a48.MP4

### Pull request checklist

This PR has considered:


harrieshin:
i would expect you still need to listen to the traitcollection changes and update the size accordingly? https://developer.apple.com/documentation/uikit/uitraitenvironment/1623516-traitcollectiondidchange?language=objc easy test would be use the debug override to change the sizes while the tooltip demo is shown?

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -81,10 +81,15 @@ class TooltipView: UIView {
         updateAppearance(tokenSet: tokenSet)
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        layoutSubviews()
```
harrieshin, line 85: must call super

harrieshin, line 85: only should do the layout if there is a change that is relavant? can you test as you change ipad's compact to regular mode, rotation, size change? video capture on the description?

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -50,55 +50,56 @@ open class Tooltip: NSObject, TokenizedControlInternal {
...
-                gestureView.passthroughView = anchorView
-                gestureView.onPassthroughViewTouches = { _ in
+        self.tooltipViewController = TooltipViewController(message: message,
+                                                           title: title,
+                                                           textAlignment: textAlignment,
+                                                           positionController: positionController,
+                                                           tokenSet: tokenSet)
+        if let tooltipViewController = tooltipViewController, let tooltipView = tooltipViewController.view {
+            let rootViewController = window.rootViewController
+            rootViewController?.addChild(tooltipViewController)
```
harrieshin, line 60: if we are keeping this addchild, willmove() logic, that means in dismissal of tooltip we need to clean up properly from the rootviewcontroller.

jeaniehuynh: I called tooltipViewController.willMove(toParent: nil) and tooltipViewController.removeFromParent(), what else needs to be done to clean up?

ios/FluentUI/Tooltip/TooltipPositionController.swift
```diff
@@ -33,14 +33,17 @@ class TooltipPositionController {
     func updateArrowDirectionAndTooltipSize(for message: String, title: String? = nil, tokenSet: TooltipTokenSet) {
         let preferredBoundingRect = boundingRect.inset(by: anchorViewInset(for: preferredArrowDirection))
         let backupBoundingRect = boundingRect.inset(by: anchorViewInset(for: preferredArrowDirection.opposite))
-        let preferredSize = TooltipView.sizeThatFits(preferredBoundingRect.size,
+        let isAccessibilityContentSize = window.traitCollection.preferredContentSizeCategory.isAccessibilityCategory
+        let preferredSize = TooltipViewController.sizeThatFits(preferredBoundingRect.size,
```
harrieshin, line 37: fundamentally sizeThatFits on viewcontroller doesn't make sense. why do we still need toptippostioncontroller separate from your newly tooltipviewcontroller.swift?

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -40,51 +40,45 @@ class TooltipView: UIView {
...
+        super.viewWillLayoutSubviews()
 
         updateAppearance(tokenSet: tokenSet)
     }
 
+    override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
+        super.traitCollectionDidChange(previousTraitCollection)
+
+        if previousTraitCollection?.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+            viewWillLayoutSubviews()
```
harrieshin, line 73: i am bit surprise by this line of code. i would expect viewwilllayoutsubivew should get automatically called?

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -40,51 +40,45 @@ class TooltipView: UIView {
...
-        backgroundView.frame = bounds
-
-        if positionController.arrowDirection.isVertical {
-            arrowImageView.frame.origin.x = positionController.arrowPosition
-            backgroundView.frame.size.height -= arrowImageView.frame.height
-        } else {
-            arrowImageView.frame.origin.y = positionController.arrowPosition
-            backgroundView.frame.size.width -= arrowImageView.frame.width
+        coordinator.animate(alongsideTransition: nil) { _ in
+            self.updateAppearance(tokenSet: self.tokenSet)
```
harrieshin, line 59: how noisy is this call? anything we should optimize?

ios/FluentUI/Tooltip/TooltipPositionController.swift
jeaniehuynh: discussed offline, this will be handled in a separate PR!

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: I made sure to follow the documentation and don't see anything missing from clean up, so I'm going to resolve this comment. Let me know if you see something else I should have!

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -50,55 +50,59 @@ open class Tooltip: NSObject, TokenizedControlInternal {
...
-            tooltipView.addGestureRecognizer(UITapGestureRecognizer(target: self, action: #selector(handleTapGesture)))
-            if self.dismissMode == .tapOnTooltipOrAnchor {
-                gestureView.passthroughView = anchorView
-                gestureView.onPassthroughViewTouches = { _ in
+        self.tooltipViewController = TooltipViewController(message: message,
+                                                           title: title,
+                                                           textAlignment: textAlignment,
+                                                           positionController: positionController,
+                                                           tokenSet: tokenSet)
+        if let tooltipViewController = tooltipViewController,
```
harrieshin, line 58: `
       if let tooltipViewController = tooltipViewController,
           let tooltipView = tooltipViewController.view,
           let rootViewController = window.rootViewController {
`

all these let. does it require If statement? curious if it is creating unnecessary indentation

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -50,55 +50,59 @@ open class Tooltip: NSObject, TokenizedControlInternal {
...
+            }, completion: { _ in
+                UIAccessibility.post(notification: .screenChanged, argument: tooltipView)
+            })
 
-        isShowing = true
+            isShowing = true
 
-        UIDevice.current.beginGeneratingDeviceOrientationNotifications()
-        NotificationCenter.default.addObserver(self, selector: #selector(orientationChanged), name: UIDevice.orientationDidChangeNotification, object: nil)
+            UIDevice.current.beginGeneratingDeviceOrientationNotifications()
```
harrieshin, line 103: none of this code should be needed any more with your awsome new viewwillresize?

ios/FluentUI/Tooltip/TooltipPositionController.swift
```diff
@@ -84,7 +87,7 @@ class TooltipPositionController {
         return CGRect(origin: tooltipOrigin, size: tooltipSize)
     }
 
-    private lazy var tooltipOrigin: CGPoint = {
+    private var tooltipOrigin: CGPoint {
```
harrieshin, line 90: why can't it be lazy?

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -7,30 +7,30 @@ import UIKit
 
 // MARK: TooltipView
```
harrieshin, line 8: update comments

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -147,7 +130,36 @@ class TooltipView: UIView {
         set { }
     }
 
-    private func updateAppearance(tokenSet: TooltipTokenSet) {
+    func updateAppearance(tokenSet: TooltipTokenSet) {
+        // Update tokenSet
+        self.tokenSet = tokenSet
+
+        // Update tooltip size
+        positionController.updateArrowDirectionAndTooltipSize(for: message, title: titleMessage, tokenSet: tokenSet)
```
harrieshin, line 138: having the position of arrow direction to the framesize itself to be in a different smaller private function makes sense to me. it feels werid why updateAppearance(token) which is meant to be more about token update dictates the frame and size.

ios/FluentUI/Tooltip/TooltipPositionController.swift
jeaniehuynh: lazy can't be used on a computed property. we need tooltipOrigin to be recalculated each time it is called to adjust to new anchor position

harrieshin: i guess that the part i am confused. either this is source of truth and computation should be done somewhere else as a function?

jeaniehuynh: can you explain more? isn't the point of a computed property for it to be computed? why should that be separated into two parts (the var and a func to compute)?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: oh I thought you didn't want to change existing behavior. I'll update it to reposition instead of hide

harrieshin: you cdan still make it dismissed but you don't need this observer anymore?

harrieshin: before they needed this hack because it was just a uiview. now that you have converted to viewcontroller there is much better way to tell the orientation has changed

jeaniehuynh: gotcha. that's a bit tricky because viewWillTransitionTo gets called for screen rotation _and_ window resizing. there's not really a clean way to determine whether the method was called due to rotating or multi-task resizing. when viewWillTransitionTo gets called, UIDevice.current.orientation has already updated to the new orientation. otherwise i would compare the orientation outside the coordinator completion with the orientation within the coordinator completion. we could store a variable that keeps track of the orientation to compare, but i think the existing implementation with observers is better

harrieshin: the UIDevice.current is as wrong practice as uiapplicationshared. iPad now can share their screen with monitor so automatically this usuage is incorrect. the traitcollectiondidchange can be another place to figure out the rotation happened to check before and after state?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: requires force unwrapping otherwise, but i'll do guard let instead as you suggested on Teams!

ios/FluentUI/Tooltip/TooltipPositionController.swift
harrieshin: we have 3 things tooltiprect/tooltiporigin/idealtooltiporigin. My assumption is at the end tooltipRect.origin is the source of truth for tooltipviewcontroller. now that tooltiporigin is meant to be computational i don't know why this is yet another variable? to me it should be more a function?

ios/FluentUI/Tooltip/Tooltip.swift
jeaniehuynh: yup! traitcollectiondidchange works (though that still worked when it was just uiview too)

ios/FluentUI/Tooltip/TooltipViewController.swift
jeaniehuynh: if message or title textstyle token changes, size also changes and if size changes too much, arrow direction may need to change.

ios/FluentUI/Tooltip/TooltipViewController.swift
harrieshin: does it need to call the whole update appearance or just part that frame needs to update?

ios/FluentUI/Tooltip/Tooltip.swift
harrieshin: if we just check the currently view bounds height to the size that will be transition height, if the height changes, does that mean it is rotated? in window size changes the height stays the same and usual the width changes.

jeaniehuynh: seems hacky to me, there are edge cases to that. for example, changing the app from slide over to split view and vice versa changes the window height too

harrieshin: thanks! i see. then i think to remove all the hackiness together, lets just reposition the tooltip all the time rather than guess the rotation.

jeaniehuynh: sounds good! :)

harrieshin (approved):

PR #1334, opened 2022-10-31
Aligning Avatar color hash color algorithm with web

sophialee0416:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

Aligning our color hash algorithm to match with FluentUI Web, and fixing macOS AvatarView bug where email and name order was switched.

References:
- https://github.com/microsoft/fluentui-react-native/pull/2201
- https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-avatar/src/components/Avatar/useAvatar.tsx#L200

### Verification

| image |image |
| image | image |
| image | image |
| image | image |

### Pull request checklist

This PR has considered:


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -435,10 +435,14 @@ public struct Avatar: View, TokenizedControlView {
         /// - Returns: hash code of string
         private static func javaHashCode(_ text: NSString) -> Int32 {
             var hash: Int32 = 0
-            for i in 0..<text.length {
-                // Allow overflows, mimicking Java behavior
-                hash = 31 &* hash &+ Int32(text.character(at: i))
-            }
+            var len = text.length - 1
```
harrieshin, line 438: might be worth cross referencing where this code comes from?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -435,10 +435,14 @@ public struct Avatar: View, TokenizedControlView {
...
+            while len >= 0 {
```
harrieshin, line 439: any particular reason for while vs for?

harrieshin:
separate pr for macOS or are you doing to do it here?

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: How do I do that?

sophialee0416:
> separate pr for macOS or are you doing to do it here?

Yep coming in a hot second!

ios/FluentUI/Avatar/Avatar.swift
harrieshin: comment to the web(Fluent react) file?

sophialee0416: I think I already did that in the description, should I do it in the code as well?

harrieshin: i think so. also make sure the link is the actual file not the link reference to a commit by a particular user

macos/FluentUITestViewControllers/TestAvatarViewController.swift
```diff
@@ -15,8 +15,8 @@ private struct TestIdentity {
 
 /// Some sample identities
 private struct TestData {
-	static let annie = TestIdentity(name: "Annie Lindqvist", email: "Annie.Lindqvist@example.com", image: NSImage(named: TestAvatarViewController.personaFemale))
```
harrieshin, line 18: do we need this change?

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: Not really, just thought it'd be concise. I guess I can do for loop if you have a preference?

macos/FluentUITestViewControllers/TestAvatarViewController.swift
sophialee0416: Oops sorry, I'll remove that.

ios/FluentUI/Avatar/Avatar.swift
harrieshin: more to be as similar as web was my thought for ease of reference

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -431,16 +431,15 @@ public struct Avatar: View, TokenizedControlView {
         }
 
         /// Hash algorithm to determine Avatar color.
+        /// Referenced from: https://github.com/microsoft/fluentui/blob/9a6f5b7a938f95518b20624f37031112ea267914/packages/react-components/react-avatar/src/components/Avatar/useAvatar.tsx#L200
```
harrieshin, line 434: this link doesn't look right. shouldn't it link to the master branch version?

harrieshin: update macoS file too and the description.

harrieshin:
please update your branch to latest of main

Saadnajmi:
For double-verification, could we locally copy the changes into FURN's copy of FluentUI Apple in it's Pods folder, and compare the NativeAvatar with the JS Avatar ?

sophialee0416:
Currently updating the Test files, will update once they are done

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -430,15 +430,18 @@ public struct Avatar: View, TokenizedControlView {
             return Int(abs(javaHashCode(combinedHashable)))
         }
 
-        /// To ensure iOS and Android achieve the same result when generating string hash codes (e.g. to determine avatar colors) we've copied Java's String implementation of `hashCode`.
+        /// Hash algorithm to determine Avatar color.
+        /// Referenced from: https://github.com/microsoft/fluentui/blob/master/packages/react-components/react-avatar/src/components/Avatar/useAvatar.tsx#L200
         /// Must use Int32 as JVM specification is 32-bits for ints
         /// - Returns: hash code of string
         private static func javaHashCode(_ text: NSString) -> Int32 {
```
lyzhan7, line 437: I think "java hash code" refers to the old algorithm, might want to rename this to "webHashCode"

sophialee0416: Thinking of changing it to just `hashCode` actually, would that sound ok?

harrieshin (approved):

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Sounds good! You should probably also remove the reference to `JVM`, since that was the Java Virtual Machine from the old algorithm too.

mischreiber (approved):

macos/FluentUIUnitTest/AvatarViewTests.swift
```diff
@@ -140,20 +140,20 @@ class AvatarViewTests: XCTestCase {
 
 	func testHashAlgorithm () {
 		var colorSet = AvatarView.getInitialsColorSet(fromPrimaryText: "", secondaryText: "Annie.Lindqvist@example.com")
```
lyzhan7, line 142: Optional - could add some of the automated tests that FURN has just to double check FURN/FUA are aligned (esp since it looks like these tests are all mostly based off email instead of name)

Screenshot 2022-11-01 at 10 18 46 AM

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -427,18 +427,21 @@ public struct Avatar: View, TokenizedControlView {
             }
```
lyzhan7, line 427: Hmm I realize this is unrelated to your PR but I think the logic here might be different from how FURN handles primary text/secondary text (I think the equivalent we have in FURN is idForColor and name - if both are provided we don't combine them, we just use idForColor). Not sure what the desired behavior is here (or if they even need to be aligned?), but I can update FURN if needed

Saadnajmi: They need to be aligned, or else the hash wouldn't be the same

lyzhan7: The issue might be that the API for avatar for FURN vs. FUA are different - I'm not sure if we can make the comparison that idForColor = email, idForColor seems a lot more general

Screenshot 2022-11-01 at 10 50 59 AM

macos/FluentUIUnitTest/AvatarViewTests.swift
sophialee0416: Is the Id supposed to be idForColor?

lyzhan7: Yes

lyzhan7: I think it's probably fine to omit those test cases since FUA doesn't really have an equivalent of "idForColor", unless we decide the FUA Avatar API should change to match FURN (or vice versa). I can add the FUA test cases of email to FURN, I think that would make sense. If anyone else has thoughts on what test cases should be on FUA/FURN feel free to chime in

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: Resolving this for now as this conversation is out of scope for this PR, but not disregarding this topic!

sophialee0416:
> For double-verification, could we locally copy the changes into FURN's copy of FluentUI Apple in it's Pods folder, and compare the NativeAvatar with the JS Avatar ?

Will be revisiting this once we figure out the hash input alignment regarding idForColor/name :)

PR #1349, opened 2022-11-07
Implement the Unread Dot for TabBar and SideTabBar

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding an Unread Dot to indicate whether a TabBarItem has been interacted with or not. The unread dot is mutually exclusive with badge numbers. 

### Verification

Built and ran simulator, tested with RTL and LTR languages, various screen sizes and orientations. Verified that SideTabBar and TabBar unread dot behavior is as expected, and the dependent Bottom Commanding Controller does not have an unread dot.

| tabBarLightBefore | tabBarLightAfter |
| tabBarDarkBefore | tabBarDarkAfter |
| sideTabBarLightBefore | sideTabBarLightAfter |
| sideTabBarDarkBefore | sideTabBarDarkAfter |
| Bottom Commanding Controller: | Simulator Screen Shot - iPhone 14 Pro - 2022-11-07 at 15 24 44 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1349)


ios/FluentUI/Tab Bar/TabBarItem.swift
```diff
@@ -62,6 +62,30 @@ open class TabBarItem: NSObject {
...
     }
 
+    /// Initializes `TabBarItem`
+    /// - Parameter title: Used for tabbar item view's label and for its accessibilityLabel.
+    /// - Parameter image: Used for tabbar item view's imageView and for its accessibility largeContentImage unless `largeContentImage` is specified.
+    /// - Parameter selectedImage: Used for imageView when tabbar item view is selected.  If it is nil, it will use `image`.
+    /// - Parameter landscapeImage: Used for imageView when tabbar item view in landscape. If it is nil, it will use `image`. The image will be used in portrait mode if the tab bar item shows a label.
+    /// - Parameter landscapeSelectedImage: Used for imageView when tabbar item view is selected in landscape. If it is nil, it will use `selectedImage`. The image will be used in portrait mode if the tab bar item shows a label.
+    /// - Parameter isUnread: Used for when the tabbar item view has not yet been selected. If never selected, an unread dot will be displayed whent the control is initialized. Default value is false.
+    @objc public convenience init(title: String,
```
harrieshin, line 72: adding as init of tabbaritem doesn't seem right. unread state is something that we can modify after item is already created. we don't have special init for badgevalue either which matches how eng would setup UITabbaritem too https://developer.apple.com/documentation/uikit/uitabbaritem?language=objc

ios/FluentUI/Tab Bar/TabBarItem.swift
```diff
@@ -90,6 +114,19 @@ open class TabBarItem: NSObject {
     /// Notification sent when the tab bar item's badge value changes.
     static let badgeValueDidChangeNotification = NSNotification.Name(rawValue: "TabBarItemBadgeValueDidChangeNotification")
 
+    /// Notification sent when item's `isUnread` value changes.
+    static let isUnreadValueDidChangeNotification = NSNotification.Name(rawValue: "TabBarItemisUnreadValueDidChangeNotification")
+
+    /// This value will determine whether or not to show the mark that represents the "unread" state.
+    /// The default value of this property is false.
+    @objc public var isUnread: Bool = false {
```
harrieshin, line 122: should be with the other public var on top of the file. https://microsoft.github.io/swift-guide/FileOrganization.html

harrieshin, line 122: would the better naming be isUnreadDotVisible ? because you can set badge value which also means something is unread yet.

ios/FluentUI/Tab Bar/TabBarItem.swift
```diff
@@ -90,6 +114,19 @@ open class TabBarItem: NSObject {
     /// Notification sent when the tab bar item's badge value changes.
     static let badgeValueDidChangeNotification = NSNotification.Name(rawValue: "TabBarItemBadgeValueDidChangeNotification")
 
+    /// Notification sent when item's `isUnread` value changes.
+    static let isUnreadValueDidChangeNotification = NSNotification.Name(rawValue: "TabBarItemisUnreadValueDidChangeNotification")
+
+    /// This value will determine whether or not to show the mark that represents the "unread" state.
```
harrieshin, line 120: the documentation and the code should be clear which will win. BadgeValue? or isUnread?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -173,6 +186,33 @@ class TabBarItemView: UIControl {
         }
     }
 
+    @objc private func isUnreadValueDidChange() {
+        isUnreadDotVisible = item.isUnread
+        updateUnreadDot()
+        setNeedsLayout()
```
harrieshin, line 192: do we need to update accessibilityLabel what we are going to say if there is an unread dot?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -173,6 +186,33 @@ class TabBarItemView: UIControl {
...
+    }
+
+    private let unreadDotLayer: CALayer = {
+        let unreadDotLayer = CALayer()
+        unreadDotLayer.bounds.size = CGSize(width: Constants.unreadDotSize, height: Constants.unreadDotSize)
+        unreadDotLayer.cornerRadius = Constants.unreadDotSize / 2
+        return unreadDotLayer
+    }()
+
+    private lazy var unreadDotColor: UIColor = UIColor(colorValue: GlobalTokens.sharedColors(.red, .primary))
```
harrieshin, line 214: while fluent2-token branch hasn't merged, shouldn't it be declared with rest of private struct Constants { ?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TabBarViewDemoController.swift
```diff
@@ -61,7 +61,9 @@ class TabBarViewDemoController: DemoController {
 
     private func setupTabBarView() {
         // remove the old tab bar View
+        var openIsUnread = true
```
harrieshin, line 64: boolean names should always start with clear verb  https://microsoft.github.io/swift-guide/Naming.html#variables-and-properties

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: None of the other controls with unread dots currently support accessibilityLabel functionality. Because of this, we do not have a localized string we can use for this label right now. 

Do you want me to add an "unread" label in the unread dot case for now and get a loc pass done? Happy to chat offline about how to do this.

Separately, we should update SegmentedControl and PillButtonBar to have accessibilityLabels for their unreadDots.

edjamesmsft: I've gone ahead and updated the loc file with an unread dot string. Tested with accessibility inspector.

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -98,16 +102,21 @@ class TabBarItemView: UIControl {
...
 
         NSLayoutConstraint.activate([
-			container.centerXAnchor.constraint(equalTo: centerXAnchor),
-			container.centerYAnchor.constraint(equalTo: centerYAnchor),
-			container.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor)
-		])
+            container.centerXAnchor.constraint(equalTo: centerXAnchor),
+            container.centerYAnchor.constraint(equalTo: centerYAnchor),
+            container.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor)
+    ])
```
harrieshin, line 108: seems like wrong tabbing?

ios/FluentUI/Tab Bar/TabBarItemView.swift
harrieshin: we should follow whatever apple does for app badges on home screen. would need to figure out if it is part of acessibilitylabel or accessibilityhint

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -173,6 +187,32 @@ class TabBarItemView: UIControl {
...
+    @objc private func isUnreadValueDidChange() {
+        isUnreadDotVisible = item.isUnreadDotVisible
+        updateUnreadDot()
+        updateAccessibilityLabel()
+        setNeedsLayout()
+    }
+
+    private var isUnreadDotVisible: Bool = false {
+        didSet {
+            if oldValue != isUnreadDotVisible {
```
harrieshin, line 199: according to your other comment, if there is already a badgevalue that means we should completely igore the new value of isUnreadDotvisible?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -98,16 +102,21 @@ class TabBarItemView: UIControl {
...
-			container.centerYAnchor.constraint(equalTo: centerYAnchor),
-			container.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor)
-		])
+            container.centerXAnchor.constraint(equalTo: centerXAnchor),
+            container.centerYAnchor.constraint(equalTo: centerYAnchor),
+            container.widthAnchor.constraint(lessThanOrEqualTo: widthAnchor)
+    ])
 
         NotificationCenter.default.addObserver(self,
                                                selector: #selector(badgeValueDidChange),
```
harrieshin, line 111: if badgeValuedidchange getcalled, when there was already isunreaddot visible, then we need to have the logic to hide the dot, and redrawe the badgevalue.

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: No, this is fixing incorrect tabbing. Switched from tabs to spaces!

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: This currently works as intended and as you have described.

harrieshin: where is that code?

ios/FluentUI/Tab Bar/TabBarItemView.swift
laminesm: Specifically line 108 seems to need an extra tab

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -343,6 +397,12 @@ class TabBarItemView: UIControl {
         badgeValue = item.badgeValue
     }
 
+    // The priority logic for accessibility label is:
+    //      If the badge is visibile:
```
jeaniehuynh, line 401: nit: visible is misspelled here

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: Refactored a bunch of code. I've confirmed that the subview is removed when the unread dot is not visible.

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: Eh, this feels like a safety release valve in case of weird behavior from the consumer. I'd rather keep this logic the same just in case they decide to remove a badge but not change the read/unread state.

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: This whitespace is correct in the larger context of the code!

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: The current badge logic updates the accessibilityLabel. I will follow that pattern and if we need to update it to use accessibilityHint instead, I think we should address that in another change for both badge and unread dot, to keep this change constrained.

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -211,6 +251,8 @@ class TabBarItemView: UIControl {
 
     let badgeView: UILabel = BadgeLabel(frame: .zero)
 
+    let unreadDotView: UILabel = BadgeLabel(frame: .zero)
```
harrieshin, line 254: why does it  have to be a label? if it does, then the badgeView can be used for unreaddot scenario as well?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -162,6 +172,11 @@ class TabBarItemView: UIControl {
         static let badgeBorderWidth: CGFloat = 2
         static let badgeHorizontalPadding: CGFloat = 10
         static let badgeCorderRadii: CGFloat = 10
+        static let unreadDotPortraitOffsetX: CGFloat = 6.0
+        static let unreadDotOffsetX: CGFloat = 4.0
+        static let unreadDotOffsetY: CGFloat = 20.0
+        static let unreadDotSize: CGFloat = 8.0
+        static let unreadDotColor: UIColor = UIColor(colorValue: GlobalTokens.sharedColors(.red, .primary))
```
harrieshin, line 179: curious is the badge color different from unreaddot color?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -22,6 +22,11 @@ class TabBarItemView: UIControl {
             imageView.isHighlighted = isSelected
             updateColors()
             if isSelected {
+                if item.isUnreadDotVisible {
+                    item.isUnreadDotVisible = false
+                    // updateUnreadDot()
```
harrieshin, line 27: remove the comment?

ios/FluentUI/Tab Bar/TabBarItemView.swift
```diff
@@ -173,6 +188,23 @@ class TabBarItemView: UIControl {
...
+    }
+
+    private var isUnreadDotVisible: Bool = false {
+        didSet {
+            if oldValue != isUnreadDotVisible {
+                if isUnreadDotVisible {
+                    updateBadgeView()
+                }
+            }
+        }
```
harrieshin, line 205: is all your logic of didset{} needed? now that you have unified updateBadgeView() it should just be handle by isUnreadValueDidChange and badgevaluedidchange because once the isunreaddot is set to false, you already triggered mask to be nil.

ios/FluentUI/Tab Bar/TabBarItemView.swift
edjamesmsft: Until we merge this change to fluent 2 colors, the badge color is correct enough. I will remove the unreadDotColor since we're not using it any more.

harrieshin (approved):

PR #1250, opened 2022-09-14
Add transformer for SegmentedControl's button title

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Added a transformer to set the font size of the Segmented Control's buttons so they don't scale, since we use the Large Content Viewer.

### Verification
| iOS Version | Before | After |
|---|---|---|
| 15.5 | SegmentedControlLargeText_Before | SegmentedControlLargeText_After |
| 14.5 | SegmentedControlLargeText_14_Before | SegmentedControlLargeText_14_After |

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [x] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1250)

PR #1354, opened 2022-11-10
[fluent2 tokens] Add scrolling and faded edges to the SegmentedControl

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Updated the `SegmentedControl` to:
 - have all the buttons in a stackView, which simplifies the layout logic
 - have all the content in a scrollView, which allows the `SegmentedControl` to scroll
 - apply a `CAGradientLayer` backed mask to the entire view, which allows us to fade out an edge and communicate to the user that there is more content to scroll to.
 - fixed the large text bug that #1250 fixed, but was unsuccessfully merged 

### Verification

| Before | After |
|---|---|
| n/a | SegmentedControl_GradientMask |
| n/a | SegmentedControl_Scrolling_After2 |
| SegmentedControl_iPhone_Before | SegmentedControl_iPhone_After2 |
| SegmentedControl_iPhone_LargeText_Before | SegmentedControl_iPhone_LargeText_After2 |
| SegmentedControl_iPhone_RTL_Before | SegmentedControl_iPhone_RTL_After2 |
| SegmentedControl_iPhone_Dark_Before | SegmentedControl_iPhone_Dark_After2 |
| SegmentedControl_iPhone_ThemeChange_Before | SegmentedControl_iPhone_ThemeChange_After |
| SegmentedControl_iPhone_Rotation_Before | SegmentedControl_iPhone_Rotation_After |
| SegmentedControl_iPad_MultiScene_Before | SegmentedControl_iPad_MultiScene_After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1354)


ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -278,32 +333,19 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
+        // flipSubviewsForRTL only works on direct children subviews
+        pillContainerView.flipSubviewsForRTL()
 
         flipSubviewsForRTL()
         layoutSelectionView()
+
+        updateGradientMaskColors()
+        CATransaction.begin()
+        CATransaction.setDisableActions(true)
+        gradientMaskLayer.frame = layer.bounds
```
laminesm, line 347: Why do we need to wrap these in a CATransaction?

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -60,19 +70,23 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
+    // Hierarchy:
     //
-    // pillContainerView (used to create 16pt inset on either side)
-    // |--backgroundView (fill container view, uses restTabColor)
-    // |--buttons (uses restLabelColor)
-    // |--pillMaskedLabelsContainerView (fill container view, uses selectedTabColor)
-    // |  |.mask -> selectionView
-    // |  |--pillMaskedLabels (uses selectedLabelColor)
-    // |  |--pillMaskedImages (uses selectedLabelColor)
+    // .mask -> gradientMask
```
laminesm, line 75: What does the . mean?

huwilkes: `.mask` is the mask of the `SegmentedControl` itself, so same as `self.mask`. Currently being set on 187.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -474,41 +508,53 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
+        }
+        pillContainerViewConstraints = [
+            pillContainerView.topAnchor.constraint(equalTo: scrollView.topAnchor,
+                                                   constant: contentInset.top),
+            pillContainerView.leadingAnchor.constraint(equalTo: leadingAnchor, constant: contentInset.leading),
+            pillContainerView.trailingAnchor.constraint(equalTo: trailingAnchor, constant: -contentInset.trailing),
+            pillContainerView.bottomAnchor.constraint(equalTo: scrollView.bottomAnchor,
+                                                   constant: -contentInset.bottom)
+        ]
+        NSLayoutConstraint.activate(pillContainerViewConstraints)
```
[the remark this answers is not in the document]
huwilkes: We need to disable those constraints at the start of the function, so we need to keep them in a var.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -27,18 +27,28 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
 
     @objc public var isAnimated: Bool = true
     @objc public var numberOfSegments: Int { return items.count }
-    @objc public var shouldSetEqualWidthForSegments: Bool = true
+    /// Defines if the segments of the `SegmentedControl` are equal in width.
+    @objc public var shouldSetEqualWidthForSegments: Bool = true {
```
harrieshin, line 31: it will be less intrusive if this was optional datasource

huwilkes: I can see an argument for making `isFixedWidth` optional, and returning the value of `shouldSetEqualWidthForSegments` if it's nil, but could you help me understand how making `shouldSetEqualWidthForSegments` optional would be better? It wasn't optional before, and the default behavior hasn't changed, so I think making it optional would just add a third state.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
huwilkes: Ah, looks like we don't. I think I needed them when I was just drawing the gradients, but that doesn't seem to be the case anymore. Thanks!

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -60,19 +70,23 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
-    // |--backgroundView (fill container view, uses restTabColor)
-    // |--buttons (uses restLabelColor)
-    // |--pillMaskedLabelsContainerView (fill container view, uses selectedTabColor)
-    // |  |.mask -> selectionView
-    // |  |--pillMaskedLabels (uses selectedLabelColor)
-    // |  |--pillMaskedImages (uses selectedLabelColor)
+    // .mask -> gradientMask
+    // scrollView
+    // |--pillContainerView (used to create 16pt inset on either side)
+    // |  |--backgroundView (fill container view, uses restTabColor)
```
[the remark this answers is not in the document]
huwilkes: Looks like we don't need it. Thanks!

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -117,21 +165,20 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
         pillContainerView.addSubview(pillMaskedContentContainerView)
         addButtons(items: items)
-        // We need to add pillMaskedContentContainerView to the container view
-        // before the buttons in order to activate the label constraints, but
-        // we want pillMaskedContentContainerView to show above the buttons.
-        pillContainerView.bringSubviewToFront(pillMaskedContentContainerView)
         pillContainerView.addInteraction(UILargeContentViewerInteraction())
-        addSubview(pillContainerView)
+        scrollView.addSubview(pillContainerView)
+        addSubview(gradientMask)
```
harrieshin, line 177: the mask and scrollview shouldn't be part of the UIView hierarchy until those public apis are on? the facdt you made it variable also makes me think i can switch and it will rerender even after initialization?

huwilkes: Which public APIs are you referring to? And they're vars instead of lets so that they could  be lazy vars and reference self during their setup. Are you saying we should dynamically remove and add the scroll view and mask based on `isFixedWidth`?

huwilkes: Though, looks like the scrollView could still be a let. But the gradient mask needs to get the colors from self, so that's how it wound up as a lazy var.

ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
```diff
@@ -49,15 +56,18 @@ class SegmentedControlDemoController: DemoController {
 
         addTitle(text: "On Brand Pill")
```
harrieshin, line 57: can we give more details in this section label "equal width " just to be extra clear with contrast?

ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
```diff
@@ -35,9 +41,10 @@ class SegmentedControlDemoController: DemoController {
 
         addTitle(text: "Primary Pill with unequal buttons")
 
-        addPillControl(items: Array(segmentItems.prefix(2)),
+        addPillControl(items: Array(segmentItems.prefix(10)),
```
harrieshin, line 44: can we keep the prefix 2 as well

harrieshin:
mostly talked offline but here is the summary
1. make sure the demo app has all the different combination of segmented controls and "switch style" control
2. VoiceOver: what does it say? does 3-finger swipe scrolls the view? disable/enable state?
3. VO: now that stackview is there, does indexing happen automatically? selected/unslected state properly annouced by VO?
4. masklayer or any UIView should only exist when it is needed
5. when client sets selectedSegmentedIndex do we scroll to the right position?

harrieshin:
label should probably change from api break to new api?

huwilkes:
> mostly talked offline but here is the summary
> 
> 1. make sure the demo app has all the different combination of segmented controls and "switch style" control
> 2. VoiceOver: what does it say? does 3-finger swipe scrolls the view? disable/enable state?
> 3. VO: now that stackview is there, does indexing happen automatically? selected/unslected state properly annouced by VO?
> 4. masklayer or any UIView should only exist when it is needed
> 5. when client sets selectedSegmentedIndex do we scroll to the right position?

1. It has everything now except for fixed with + unequal buttons, which isn't a combo we want to promote, anyway.
2+3. 3 finger swipe will scroll, reads dimmed for disabled, has selection and index.
4. view hierarchy now changes dynamically
5. the selected index is now scrolled to, as long as it's selected after the control's been drawn.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -545,4 +585,39 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
+            NSLayoutConstraint.activate(scrollViewConstraints)
+        }
+    }
+
+    private var maximumContentOffset: CGFloat {
+        return stackView.frame.size.width - scrollView.frame.size.width
+    }
+
+    // MARK: UIScrollViewDelegate
+    public func scrollViewDidScroll(_ scrollView: UIScrollView) {
```
harrieshin, line 620: nit: it should be closer to other public functions

harrieshin (approved):

ios/FluentUI/SegmentedControl/SegmentedControl.swift
huwilkes: Moved to an extension, which seems to be a common pattern in our code. Will follow up with the team in general to see which approach should be preferred in the future.

PR #1356, opened 2022-11-10
Bringing fluent2 CCB to main.

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Bringing fluent2 CCB to main.

| File | Main | CCB | Delta |
|-----|------|------|------|
| CommandBar.o | 153,656 bytes | 221,896 bytes | +68,240 bytes |
| CommandBarButton.o | 103,760 bytes | 120,624 bytes | +16,864 bytes |
| CommandBarButtonGroupView.o | 59,104 bytes | 64,248 bytes | +5,144 bytes |
| CommandBarCommandGroupsView.o | 211,256 bytes | 210,896 bytes | -360 bytes |
| CommandBarItem.o | 113,072 bytes | 113,072 bytes | 0 bytes |
| CommandBarTokenSet.o | N/A | 67,696 bytes | +67,696 bytes |
| libFluent.a | 28,762,616 bytes | 28,947,552 bytes | +184,936 bytes |

### Verification

| Main before                                       | Main after                                      | fluent2-tokens |
|----------------------------------------------|--------------------------------------------| -----|
| Simulator Screen Recording - iPhone 14 Pro - 2022-11-11 at 13 22 23 | Simulator Screen Recording - iPhone 14 Pro - 2022-11-11 at 13 19 28 | Simulator Screen Recording - iPhone 14 Pro - 2022-11-11 at 13 40 33 |
| image | image | image |

Theming verification:
Simulator Screen Recording - iPhone 14 Pro - 2022-12-05 at 14 25 43

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1356)


ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -18,7 +19,7 @@ public protocol CommandBarDelegate: AnyObject {
  Provide `itemGroups` in `init` to set the buttons in the scrollable area. Optional `leadingItemGroups` and `trailingItemGroups` add buttons in leading and trailing positions. Each `CommandBarItem` will be represented as a button.
  */
 @objc(MSFCommandBar)
-open class CommandBar: UIView {
+public class CommandBar: UIView, TokenizedControlInternal {
```
harrieshin, line 22: hmmm do we need to keep it open?

ios/FluentUI/Command Bar/CommandBarItem.swift
```diff
@@ -124,15 +124,6 @@ open class CommandBarItem: NSObject {
         }
     }
 
-    /// A string that uniquely identifies the element, typically for automation purposes
-    @objc public var accessibilityIdentifier: String? {
```
harrieshin, line 128: this can't be right. is yours or fluent2 branch up to date?

sophialee0416: Seems that the fluent2-tokens branch hasn't merged with main with this change, I'll revert it

ios/FluentUI/Command Bar/CommandBar.swift
sophialee0416: https://github.com/microsoft/fluentui-apple/pull/832 Looks like we changed it in the binary size reduction effort, should we keep this change @mischreiber ?

ios/FluentUI/Command Bar/CommandBarButton.swift
```diff
@@ -218,12 +201,19 @@ class CommandBarButton: UIButton {
...
+// MARK: CommandBarButton UIPointerInteractionDelegate
 
-    private struct ColorConstants {
-        static let normalTintColor: UIColor = Colors.textPrimary
-        static let normalBackgroundColor = UIColor(light: Colors.gray50,
-                                                   dark: Colors.gray600)
-        static let highlightedBackgroundColor = UIColor(light: Colors.gray100,
-                                                        dark: Colors.gray900)
+extension CommandBarButton: UIPointerInteractionDelegate {
+    @available(iOS 13.4, *)
```
sophialee0416, line 209: @harrieshin Will we still need this line if our minimum is now 15?

ios/FluentUI/Command Bar/CommandBar.swift
harrieshin: i would do a quick double check if any of our clients are subclassing commandbar...

sophialee0416: I don't think we have any subclassing business downstream!

harrieshin:
lets make sure we test voiceover and the sampel "delete" button which has its own tinting capability.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -180,6 +219,7 @@ open class CommandBar: UIView {
 
     private lazy var scrollView: UIScrollView = {
         let scrollView = UIScrollView()
+        let itemInterspace: CGFloat = tokenSet[.itemInterspace].float
```
nodes11, line 222: Is this being used?

ios/FluentUI/Command Bar/CommandBarButton.swift
harrieshin: yeah this can be removed for sure. the main is already min io14 anyway.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -264,6 +306,19 @@ open class CommandBar: UIView {
         containerMaskLayer.locations = locations.map { NSNumber(value: Float($0)) }
     }
 
+    private func updateButtonTokens() {
+        leadingCommandGroupsView.updateButtonGroupViews()
+        mainCommandGroupsView.updateButtonGroupViews()
+        trailingCommandGroupsView.updateButtonGroupViews()
+    }
+
+    @objc private func handleCommandButtonTapped(_ sender: CommandBarButton) {
```
nodes11, line 315: Is this method needed? I believe this is also in `CommandBarCommandGroupsView`.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -62,17 +63,14 @@ class CommandBarButtonGroupView: UIView {
...
         }
     }
 
-    private struct LayoutConstants {
-        static let cornerRadius: CGFloat = 8.0
-        static let buttonPadding: CGFloat = 2.0
-        static let leftRightBuffer: CGFloat = 2.0
-    }
+    private var tokenSet: CommandBarTokenSet
+    private static let leftRightBuffer: CGFloat = 2.0
```
nodes11, line 75: Maybe keep this within a Constants struct to follow the Microsoft Swift Style guidelines?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -64,13 +65,18 @@ open class CommandBar: UIView {
     @objc public init(itemGroups: [CommandBarItemGroup],
                       leadingItemGroups: [CommandBarItemGroup]? = nil,
                       trailingItemGroups: [CommandBarItemGroup]? = nil) {
+        self.tokenSet = CommandBarTokenSet()
```
nodes11, line 68: I think `self.` can be omitted here

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -113,6 +138,20 @@ open class CommandBar: UIView {
...
     }
 
+    // MARK: - TokenizedControl
+
+    public typealias TokenSetKeyType = CommandBarTokenSet.Tokens
+    public var tokenSet: CommandBarTokenSet
+
+    var tokenSetSink: AnyCancellable?
+
+    @objc private func themeDidChange(_ notification: Notification) {
```
nodes11, line 148: nit: This method should be moved down with the other private methods

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -113,6 +138,20 @@ open class CommandBar: UIView {
         updateShadow()
     }
 
+    // MARK: - TokenizedControl
+
+    public typealias TokenSetKeyType = CommandBarTokenSet.Tokens
+    public var tokenSet: CommandBarTokenSet
+
+    var tokenSetSink: AnyCancellable?
```
nodes11, line 146: nit: Internal properties should placed be in between the public and private members

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -264,6 +306,19 @@ open class CommandBar: UIView {
         containerMaskLayer.locations = locations.map { NSNumber(value: Float($0)) }
     }
 
+    private func updateButtonTokens() {
+        leadingCommandGroupsView.updateButtonGroupViews()
```
nodes11, line 310: I am not sure that exposing `updateButtonGroupVIews` as public and calling it here is safe. The implementation of `updateButtonGroupVIews` calls `updateItemsToButtonsMap` to refresh the items-to-buttons map and then creates new buttons for each item, but the newly created buttons are never added to the stack view. This could cause an issue where the state of the `CommandBarItem` is not consistent the corresponding `CommandBarButton`. If this sort of update behavior is needed, I would suggest trying to expose `updateButtonsShown ` instead as it will update the stack view after calling `updateButtonGroupVIews`.

ios/FluentUI/Command Bar/CommandBar.swift
sophialee0416: Looks like it's being used in the groups view but not here, as it looks like it was deleted in #987. I can go ahead and remove this method

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
harrieshin: the buffer should technically be part of the tokenset's private static const like other fluent2 controls

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -75,7 +76,7 @@ class CommandBarCommandGroupsView: UIView {
     }
 
     /// Refreshes the `buttonGroupViews` array of `CommandBarButtonGroupView`s that are displayed in the view
-    private func updateButtonGroupViews() {
+    func updateButtonGroupViews() {
```
jeaniehuynh, line 79: this can be moved up now that it's no longer private

harrieshin:
size comparison please?

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -113,6 +138,11 @@ open class CommandBar: UIView {
         updateShadow()
     }
 
+    // MARK: - TokenizedControl
+
+    public typealias TokenSetKeyType = CommandBarTokenSet.Tokens
+    public var tokenSet: CommandBarTokenSet
```
harrieshin, line 144: curious what's the archtecture difference here which some control tokenSet is a "lazy" public var and this one isn't? @sophialee0416 & @mischreiber

huwilkes: I think it's probably be cause the other controls needed to pass in a style or something to their token set, so they needed to be lazy. Since the `CommandBarTokenSet` doesn't rely on anything from the control, it can just be initialized.

ios/FluentUI/Command Bar/CommandBarTokenSet.swift
```diff
@@ -0,0 +1,67 @@
...
+public class CommandBarTokenSet: ControlTokenSet<CommandBarTokenSet.Tokens> {
+    public enum Tokens: TokenSetKey {
+        case backgroundColor
+        case groupBorderRadius
+        case groupInterspace
+        case itemBackgroundColor
+        case itemFixedIconColor
+        case itemIconColor
+        case itemInterspace
+        case leftRightBuffer
```
sophialee0416, line 18: Note: Make this into a token constant

ios/FluentUI/Command Bar/CommandBarTokenSet.swift
```diff
@@ -0,0 +1,67 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+/// Design token set for the `CommandBar` control.
+public class CommandBarTokenSet: ControlTokenSet<CommandBarTokenSet.Tokens> {
+    public enum Tokens: TokenSetKey {
+        case backgroundColor
```
harrieshin, line 11: these should all have comments

ios/FluentUI/Command Bar/CommandBarTokenSet.swift
```diff
@@ -0,0 +1,67 @@
...
+        case groupBorderRadius
+        case groupInterspace
+        case itemBackgroundColor
+        case itemFixedIconColor
+        case itemIconColor
+        case itemInterspace
```
harrieshin, line 17: do these have to be public? we don't expect client to be able to change things

ios/FluentUI/Command Bar/CommandBarTokenSet.swift
```diff
@@ -0,0 +1,67 @@
...
+import UIKit
+
+/// Design token set for the `CommandBar` control.
+public class CommandBarTokenSet: ControlTokenSet<CommandBarTokenSet.Tokens> {
+    public enum Tokens: TokenSetKey {
+        case backgroundColor
+        case groupBorderRadius
+        case groupInterspace
+        case itemBackgroundColor
+        case itemFixedIconColor
```
harrieshin, line 15: is this needed?

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -116,6 +116,7 @@ public enum ControlTokenValue {
     case dynamicColor(() -> DynamicColor)
     case fontInfo(() -> FontInfo)
     case shadowInfo(() -> ShadowInfo)
+    case buttonDynamicColors(() -> ButtonDynamicColors)
```
huwilkes, line 120: Ah, @mischreiber just commented on my button PR for this one, but turns out we're not supposed to add this. Not sure why it's still around in the fluent2-tokens branch if it isn't supposed to make it to main, though.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -79,7 +85,26 @@ open class CommandBar: UIView {
...
+                self?.updateButtonTokens()
+            }
+        }
+    }
+
+    public override func didMoveToWindow() {
+        super.didMoveToWindow()
+
+        tokenSet.update(fluentTheme)
+        updateButtonTokens()
```
huwilkes, line 107: Updating the theme will call the sink from line 98, which will also call updateButtonTokens. So we probably should remove it from didMoveToWindow, to avoid calling it more often than we need to.

huwilkes (commented):
It doesn't necessarily need to happen in this PR, but we should make sure that overrides work.

ios/FluentUI/Command Bar/CommandBar.swift
sophialee0416: I think we need this for the theme-wide override, removing this doesn't update the tokens when I exit the window.

sophialee0416:
> It doesn't necessarily need to happen in this PR, but we should make sure that overrides work.

Added an additional gif for validation!

huwilkes (approved):

PR #1362, opened 2022-11-15
Make ShimmerLinesView inherit ShimmerView init

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

According to [Apple docs](https://docs.swift.org/swift-book/LanguageGuide/Initialization.html), "superclass initializers are automatically inherited if certain conditions are met." A couple things were wrong with the implementation done during tokenization that didn't meet those conditions and thus would break API upon going into devmain:
1. lineCount, firstLineFillPercent, and lastLineFillPercent need to be properties so that they can have a default value
2. ShimmerLinesView cannot have its own init

### Verification

Tested init of ShimmerLinesView

| Before | After |
|  image | simulator_screenshot_0C7244D1-C7D7-495D-92DC-F9FE5A37262C |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1362)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
@@ -30,6 +30,7 @@ struct Demos {
         DemoDescriptor("NotificationView", NotificationViewDemoController.self),
         DemoDescriptor("Other cells", OtherCellsDemoController.self),
         DemoDescriptor("PersonaButtonCarousel", PersonaButtonCarouselDemoController.self),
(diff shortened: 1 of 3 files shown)


harrieshin:
can we also add test cases in the demo app calling these apis and properties, please?

harrieshin (approved):
thanks for the quick turn around

Issue #1338, opened 2022-11-02
Setting backgroundColor on DrawerController caused the view to load and ignore other properties


slzublid:
### Environment Information

- **Platform**:
  - [X] iOS
  - [ ] macOS
- **Package version(s)**: 0.6.2

### Please provide a reproduction of the bug:

When customizing the DrawerController, if the backgroundColor is set, it will load the view before other properties are set, ignoring, for example, presentationStyle, if that's set after backgroundColor.

#### Actual behavior:

view.background is set before view is loaded, causing a loadView() call before other properties are set

#### Expected behavior:

isViewLoaded should be checked before setting view.background

### Priorities and help requested:

Workaround is to set backgroundColor last, right before presenting the controller

imWildCat:
@sophialee0416 fyi

The version is 0.6.2. `DrawerController` also has the patch from https://github.com/microsoft/fluentui-apple/commit/5fbff443da8d921bccdffd6d897849c7414a7d63

sophialee0416:
Closing issue seeing that #1379 has merged.

PR #1379, opened 2022-11-17
Check isViewLoaded before setting Drawer background color

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

Setting the background color in `DrawerController` forces a call to `loadView` if the root view isn't initialized yet. This PR adds a `isViewLoaded` check before setting the view's background color. 
* #1338 

Binary change:

| Before | After |
|--------|-------|
| 32,126,133 bytes | 32,126,133 bytes |

(a summary of the changes made, often organized by file)

### Verification

| before | after |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1379)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -126,7 +126,9 @@ open class DrawerController: UIViewController {
     @objc open var backgroundColor: UIColor = Colors.Drawer.background {
         didSet {
             useCustomBackgroundColor = true
-            view.backgroundColor = backgroundColor
+            if isViewLoaded {
+                view.backgroundColor = backgroundColor
+            }
         }
     }
 


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -126,7 +126,9 @@ open class DrawerController: UIViewController {
     @objc open var backgroundColor: UIColor = Colors.Drawer.background {
         didSet {
             useCustomBackgroundColor = true
-            view.backgroundColor = backgroundColor
+            if isViewLoaded {
+                view.backgroundColor = backgroundColor
+            }
```
edjamesmsft, line 131: What happens in the case of the view NOT being loaded? Will there be unexpected behavior? Should you have some fallback behavior for setting the background color when the view does end up getting loaded, like a lazy set? No idea if that's possible though.

laminesm: If the view isn't loaded, setting the background color will force the view to load which has led to unwanted behavior from a partner as described in the issue linked in the description. We don't need any fallback behavior since the background color is set in `viewDidLoad` anyway.

harrieshin (approved):

PR #1385, opened 2022-11-19
Defensive fixes for a UIViewPropertyAnimator crash

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| Before | After |
|--------|-------|
|31,642,171 bytes |31,639,099 bytes |

### Problem 
Clients can crash if they make certain bottom sheet modifications during `UIViewPropertyAnimator` completion callbacks.

### Cause
`UIViewPropertyAnimator` has a bug / undocumented behavior - when it's calling out to its completion handlers, the animator reports `isRunning==true` and it's in a private `UIViewAnimatingState` (value 5). When it's in this state, calling `finishAnimation` will cause it to throw an exception.

We have a `completeAnimationsIfNeeded` helper function that frequently gets called on bottom sheet environment changes, like the host view size / safe area. This helper checks the `isRunning` prop to make sure it's only trying to force complete animators that are running. But if an animator is in the process of completing, and the client reacts to that completion in a way that triggers `completeAnimationsIfNeeded` synchronously, we can end up triggering the exception.

### Fix
Two fixes to be safe, but either would work to mitigate the problem:
1. Remove our reference to the current animator as soon as possible. This eliminates most / all possibilities for this bug, because we'll bail early in our `completeAnimationsIfNeeded` helper as it no longer has an animator to act on.
2. Add an additional check to `completeAnimationsIfNeeded` - in addition to checking `isRunning`, let's also make sure `state == .active`. Documentation says this is always true, but it's clearly not.

### Verification

Force crash condition by modifying safe area in animator completion. Verify crash doesn't occur anymore with the fixes.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1385)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -300,8 +300,7 @@ public class BottomSheetController: UIViewController {
             animator = stateChangeAnimator(to: targetState)
 
             currentStateChangeAnimator = animator
-            animator?.addCompletion { [weak self] finalPosition in
-                self?.currentStateChangeAnimator = nil
+            animator?.addCompletion { finalPosition in
                 completion?(finalPosition)
             }
         }
@@ -733,9 +732,6 @@ public class BottomSheetController: UIViewController {
 
             if animated {
                 currentStateChangeAnimator = animator
-                animator.addCompletion { [weak self] _ in
-                    self?.currentStateChangeAnimator = nil
-                }
                 animator.startAnimation()
             } else {
                 animator.startAnimation() // moves the animator into active state so it can be stopped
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -790,6 +786,13 @@ public class BottomSheetController: UIViewController {
             guard let strongSelf = self else {
                 return
             }
+
+            // It's important we drop the reference to the animator as early as possible.
+            // Otherwise we could accidentally try modifying the animator while it's calling out to its completion handler, which can lead to a crash.
+            if let animator = strongSelf.currentStateChangeAnimator,animator == translationAnimator {
```
anandrajeswaran, line 792: This looks like the CI linter error
```suggestion
            if let animator = strongSelf.currentStateChangeAnimator, animator == translationAnimator {
```

anandrajeswaran (approved):

harrieshin:
do you mind cherry picking this to main_0.9 branch too please?

lcapkovic:
> "Documentation says this is always true" - good place for a link in the PR notes

Only in the code docs unfortunately:
// Running indicates that the animation is running either in the forward or the reversed direction.
// The state of a running animation is always active.
var isRunning: Bool { get }

nodes11 (approved):

PR #1386, opened 2022-11-19
expand our archs support

harrieshin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
some of our clients relies on explicit copies of FluentUI.swiftmodule folder with various arch. To make that step little bit easier, this change force any local build of fluent iOS library to generate arm64 arm64e and x86_64 for appropriate simulator and device builds.

### Verification
Looking at the output of FluentUI.swiftmodule and see it generateds .swiftdoc, swiftinterface and .swiftmodule.

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1386)


ios/xcode/FluentUI_common.xcconfig
```diff
@@ -52,3 +52,8 @@ SWIFT_TREAT_WARNINGS_AS_ERRORS = YES
 TARGETED_DEVICE_FAMILY = 1,2
 VERSIONING_SYSTEM = apple-generic
 VERSION_INFO_PREFIX =
+
+ARCHS[sdk=iphonesimulator*] = $(ARCHS_STANDARD) arm64 x86_64
+ARCHS[sdk=iphoneos*] = $(ARCHS_STANDARD) arm64e
+ONLY_ACTIVE_ARCH = NO
```
anandrajeswaran, line 58: Are people really copying debug bits? This seems like a rather large build time tax for everyone to pay for that?

mischreiber: Yep, debug bits are used in several of our consumer's test apps for their own debug builds.

mischreiber (commented):
I think I'd rather a solution that only adds `arm64e` for our pipeline builds. This impacts every local dev build in Xcode as well, which'll mean double build time for every device build for every developer.

harrieshin:
> I think I'd rather a solution that only adds `arm64e` for our pipeline builds. This impacts every local dev build in Xcode as well, which'll mean double build time for every device build for every developer.

dumbquestion: if we do it just for our pipeline, what's the easiest way to grab those .swiftmodules files from github virtual pool?

mischreiber:
> > I think I'd rather a solution that only adds `arm64e` for our pipeline builds. This impacts every local dev build in Xcode as well, which'll mean double build time for every device build for every developer.
> 
> dumbquestion: if we do it just for our pipeline, what's the easiest way to grab those .swiftmodules files from github virtual pool?

Will follow up offline - builds should be accessible from the ADO pipeline that actually performs the building.

Alternative idea: should we just create a shell script in the `scripts/` directory that can produce an arm64e build on demand?

PR #1392, opened 2022-11-23
Refactor Tooltip

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

  - Add TooltipView.swift
  - Remove TooltipPositionController.swift
- ios/FluentUI/Tooltip/Helpers/TooltipPositionController.swift
  - Remove TooltipPositionController
- ios/FluentUI/Tooltip/Tooltip.swift
  - Update API calls
- ios/FluentUI/Tooltip/TooltipView.swift
  - Separate view-building logic from view controller so methods like sizeThatFits make more sense 
- ios/FluentUI/Tooltip/TooltipViewController.swift
  - Move view-building logic into TooltipView
  - Move TooltipPositionController into TooltipViewController.swft
- ios/docs/Controls/Tooltip.md
  - Update naming in documentation

Binary change:
| Before | After |
|--------|-------|
|    34,356,786 bytes (34.7 MB on disk)  |   34,347,380 bytes (34.7 MB on disk)    |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1392)


jeaniehuynh:
Note: In terms of MVC architecture, I think Tooltip should be a UIViewController and everything done in TooltipViewController could be done in Tooltip considering how much of the view-controller logic Tooltip.swift does. However, that would be API-breaking and I figure we don't want to do that. Lmk what you think @harrieshin

harrieshin:
moving the view related stuff to tooltipview.swift makes alot of sense. I still don't understand the need of tooltippositioner class. most of the function can be just helper (even static function) in tooltipviewcontroller it self. tooltipview doesn't need an instance of positioner at all. it just needs to know the arrow direction which tooltipviewcontroller can just feed it in tooltipview that can be easily adjust

ios/docs/Controls/Tooltip.md
```diff
@@ -39,7 +39,7 @@ Tooltip.shared.show(with: "This is the message of the tooltip.",
 `Tooltip` in Swift, `MSFTooltip` in Objective-C/UIKit
 ### Source Code
 [Tooltip.swift](https://github.com/microsoft/fluentui-apple/blob/fluent2-tokens/ios/FluentUI/Tooltip/Tooltip.swift)  
-[TooltipPositionController.swift](https://github.com/microsoft/fluentui-apple/blob/fluent2-tokens/ios/FluentUI/Tooltip/TooltipPositionController.swift)  
+[TooltipPositioner.swift](https://github.com/microsoft/fluentui-apple/blob/fluent2-tokens/ios/FluentUI/Tooltip/TooltipPositioner.swift)
```
harrieshin, line 42: update?

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -9,285 +9,191 @@ import UIKit
...
+        return window.bounds.inset(by: window.safeAreaInsets).inset(by: UIEdgeInsets(top: screenMargin,
+                                                                                     left: screenMargin,
+                                                                                     bottom: screenMargin,
+                                                                                     right: screenMargin))
     }
 
-    private func transformForArrowImageView() -> CGAffineTransform {
-        switch positionController.arrowDirection {
+    private let tooltipView: TooltipView
+    private static var anchorView: UIView = UIView()
```
harrieshin, line 174: these don't need to be static variables?

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -9,285 +9,191 @@ import UIKit
...
+    private(set) static var arrowDirection: Tooltip.ArrowDirection = .down
+    private static var preferredArrowDirection: Tooltip.ArrowDirection = .down
+    private static var window: UIView = UIWindow()
```
harrieshin, line 177: we don't need static windows? the viewcontroller knows its window

jeaniehuynh: if it's not static, then it can't be used in the static methods (Error when trying to use `self.view.window`: Instance member 'view' cannot be used on type 'TooltipViewController')

ios/FluentUI/Tooltip/TooltipViewController.swift
jeaniehuynh: by "these," do you mean all of the variables from lines 174-182? how can they be used in static methods if the they aren't static as well?

ios/FluentUI/Tooltip/TooltipViewController.swift
harrieshin: right but if the caller is actually the same class you don't need to refer to it as TooltipViewController.arrowPosition() anymore you can just do arrowPosition because it is just helper method

ios/FluentUI/Tooltip/TooltipViewController.swift
harrieshin: yes, only determine things that has to be static variables. windows or Views seeems unlikely if tooltipviewwcontroller already has them as properties. arrow direction aslo seems to be not static? because we should only have 1 direction that can change based on the position.

ios/FluentUI/Tooltip/TooltipViewController.swift
jeaniehuynh: Sorry I'm a bit confused...arrowPosition()'s caller isn't the same class. It's only called twice, both in TooltipView.swift, not TooltipViewController.swift. But also, how does that relate to original window comment?

ios/FluentUI/Tooltip/TooltipViewController.swift
jeaniehuynh: I'm not sure how to get around these errors without making the variables static: Instance member 'arrowDirection' cannot be used on type 'TooltipViewController'

Should I make everything non-static and then created a shared instance of TooltipViewController instead like we do in Tooltip.cs?

ios/FluentUI/Tooltip/TooltipViewController.swift
harrieshin: any UIView that is present has access to the window. so tooltipview should have a window if needed. the flow has to be changed. rather than tooltipview asking static tooltipviewcontroller what the size should be to then update its own size, once the tooltipviewcontroller is initialized, it knows where the position should be and tell the tooltipview for with the necessary information that the view needs.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -51,51 +59,41 @@ class TooltipView: UIView {
...
-                                                          isAccessibilityContentSize: isAccessibilityContentSize,
-                                                          tokenSet: tokenSet,
-                                                          isMessage: false)
-            width = max(width, titleLabelFittingSize.width)
-            height += titleLabelFittingSize.height
-        }
-
+    func updateTooltipSizeAndOrigin() {
+        updateArrowDirectionAndTooltipRect(for: message, title: titleMessage, tokenSet: tokenSet)
+        self.frame.size = tooltipRect.size
```
harrieshin, line 64: nit: is it currently ambiguous that you need to call self.frame and self.bounds instead of just frame and bound?

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -168,7 +159,7 @@ open class Tooltip: NSObject, TokenizedControlInternal {
             tooltipView.alpha = 0.0
         }, completion: { _ in
             tooltipView.removeFromSuperview()
-            UIAccessibility.post(notification: .screenChanged, argument: self.tooltipViewController?.positionController.anchorView)
+            UIAccessibility.post(notification: .screenChanged, argument: self.anchorView)
```
harrieshin, line 162: to be defensive, we should always capture [weak self] in the completion blocks

` completion: { [weak self] in
                    return
                }
strongSelf.foo
`

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -0,0 +1,425 @@
...
+        messageLabel.textAlignment = textAlignment
+
+        if let titleLabel = titleLabel, let title = title {
+            titleLabel.text = title
+            titleLabel.textAlignment = textAlignment
+        }
+
+        self.addSubview(textContainer)
+
+        // Shadow
```
harrieshin, line 50: can we add a todo not to integrate  with https://github.com/microsoft/fluentui-apple/pull/1296 ? @laminesm we should track this separately as well

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -0,0 +1,425 @@
...
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+// MARK: TooltipView
+
+class TooltipView: UIView {
+
+    init(anchorView: UIView,
```
harrieshin, line 12: it might be worth adding a comment that visually represent the hierarachy how we coded up laying out all the views within the tooltipview. and what offset/size var means what.  ex) https://github.com/microsoft/fluentui-apple/blob/037b7f6254bf83f168cbe97d9a9656f7df58d0e4/ios/FluentUI/SegmentedControl/SegmentedControl.swift#L74

harrieshin (approved):

PR #1395, opened 2022-11-23
Removing nil assignment to textColor in TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:
| Before | After |
|--------|-------|
|    28,475,600    |   28,475,600    |

There was an EXC_BAD_ACCESS error when trying to set the textColor of a TableViewCell label. It could be that the error was caused because the textColor was set to `nil` in multiple places. This change removes that since textColor already has a default assigned to it if not set.

### Verification

Opened TableViewCellDemoController, tested text color changes. No crashes observed.

| image | image |
| image | image |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Label/Label.swift
+++ b/ios/FluentUI/Label/Label.swift
@@ -44,7 +44,6 @@ public enum TextColorStyle: Int, CaseIterable {
 open class Label: UILabel {
     @objc open var colorStyle: TextColorStyle = .regular {
         didSet {
-            _textColor = nil
             updateTextColor()
         }
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Label/Label.swift
```diff
@@ -98,9 +97,6 @@ open class Label: UILabel {
     }
 
     private func initialize() {
-        // textColor is assigned in super.init to a default value and so we need to reset our cache afterwards
-        _textColor = nil
-
```
huwilkes, line 103: Seems like we added this for some reason. Were you able to track down why and make sure we aren't regressing whatever behavior this was trying to fix?

mischreiber (commented):
It seems most likely that the crash would have been caused by performing this nil assignment after the cell was already deallocated--do you think either of these removed lines are likely candidates for that?

sophialee0416:
Found a potential different cause for the bug, closing this PR. But I think the nil assignment is still suspicious...

PR #1396, opened 2022-11-23
Fix bottom sheet host height change detection

lcapkovic:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| Before | After |
|--------|-------|
|31,681,160 bytes|31,681,352 bytes|

#### Bug
In some cases when sheet was animating while the host view was resizing, we could end up with an outdated sheet position until the next layout pass.

#### Cause
In the bottom sheet, we detect trait collection, safe area, and root view height changes to auto-complete ongoing animations. This guarantees the animation won't target a frame that is now out of date.

We were missing a class of size changes not captured by `viewWillTransitionTo`. This becomes an issue on pre-iPhone X devices in some cases where the host height changes while we're hiding. On iPhone X and later in most cases this is luckily covered by the safe area change handler - that's also the reason why we didn't catch it for so long.

#### Fix
Switch to a catch-all method of height detection - manually doing it in `viewDidLayoutSubviews`. The logic of how it's handled stays the same - complete animations and cancel gestures.

### Verification

No visible change.

Validation:
- repro bug in test app, validate repro is gone after fix
- validate bug is fixed in client

Additional sanity:
- quickly toggle between isHidden true and false
- change screen orientation during animation
- change screen orientation during gesture
- repeat all above for bottom commanding

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1396)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -469,6 +469,18 @@ public class BottomSheetController: UIViewController {
     }
 
     public override func viewDidLayoutSubviews() {
+        let newHeight = view.bounds.height
+        if currentRootViewHeight != newHeight && currentExpansionState == .transitioning {
+            // The view height has changed and we can't guarantee the animation target frame is valid anymore.
+            // Let's complete animations and cancel ongoing gestures to guarantee we end up in a good state.
+            completeAnimationsIfNeeded(skipToEnd: true)
+
+            if panGestureRecognizer.state != .possible {
+                panGestureRecognizer.state = .cancelled
+            }
+        }
+        currentRootViewHeight = newHeight
+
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -469,6 +469,18 @@ public class BottomSheetController: UIViewController {
     }
 
     public override func viewDidLayoutSubviews() {
```
anandrajeswaran, line 471: Seems like an existing bug, but shouldn't we call super.viewDidLayoutSubviews at some point in our implementation?

anandrajeswaran (approved):

nodes11 (approved):

PR #1405, opened 2022-11-28
Update GlobalTokens to include new ramps

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

`IconSizeToken`, `BorderSizeToken`, and `BorderRadiusToken` enums have been updated to their new defined names to align fluent sizing tokens across platforms. 
Our existing sizing tokens were simply renamed, and a few new tokens have been added.

Binary change:

| Before | After |
|--------|-------|
| 32,543,810 bytes | 32,560,546 bytes |

### Verification

The controls affected by this change were tested on the demo app.

| before_avatar | after_avatar |
| before_button | after_button |
| before_card_nudge | after_card_nudge |
| before_hud | after_hud |
| before_notification | after_notification |
| before_shimmer | after_shimmer |
| before_tvc | after_tvc |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1405)

### DIFF
--- a/ios/FluentUI/Avatar/AvatarTokenSet.swift
+++ b/ios/FluentUI/Avatar/AvatarTokenSet.swift
@@ -53,17 +53,17 @@ public class AvatarTokenSet: ControlTokenSet<AvatarTokenSet.Tokens> {
                 return .float({
                     switch style() {
                     case .default, .accent, .outlined, .outlinedPrimary, .overflow:
-                        return GlobalTokens.corner(.cornerRadiusNone)
+                        return GlobalTokens.corner(.radiusNone)
                     case .group:
                         switch size() {
                         case .size16, .size20:
-                            return GlobalTokens.corner(.cornerRadius20)
+                            return GlobalTokens.corner(.radius20)
                         case .size24, .size32:
-                            return GlobalTokens.corner(.cornerRadius40)
+                            return GlobalTokens.corner(.radius40)
                         case .size40, .size56:
-                            return GlobalTokens.corner(.cornerRadius80)
+                            return GlobalTokens.corner(.radius80)
                         case .size72:
-                            return GlobalTokens.corner(.cornerRadius120)
+                            return GlobalTokens.corner(.radius120)
                         }
                     }
                 })
@@ -105,11 +105,11 @@ public class AvatarTokenSet: ControlTokenSet<AvatarTokenSet.Tokens> {
                 return .float({
                     switch size() {
                     case .size16, .size20, .size24:
-                        return GlobalTokens.stroke(.strokeWidth10)
+                        return GlobalTokens.stroke(.width10)
                     case .size32, .size40, .size56:
-                        return GlobalTokens.stroke(.strokeWidth20)
+                        return GlobalTokens.stroke(.width20)
                     case .size72:
-                        return GlobalTokens.stroke(.strokeWidth40)
+                        return GlobalTokens.stroke(.width40)
                     }
                 })
 
@@ -117,9 +117,9 @@ public class AvatarTokenSet: ControlTokenSet<AvatarTokenSet.Tokens> {
                 return .float({
                     switch size() {
                     case .size16, .size20, .size24, .size32, .size40, .size56:
-                        return GlobalTokens.stroke(.strokeWidth20)
+                        return GlobalTokens.stroke(.width20)
                     case .size72:
-                        return GlobalTokens.stroke(.strokeWidth40)
+                        return GlobalTokens.stroke(.width40)
                     }
                 })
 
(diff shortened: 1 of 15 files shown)


ios/FluentUI/Notification/NotificationTokenSet.swift
```diff
@@ -166,9 +166,9 @@ public class NotificationTokenSet: ControlTokenSet<NotificationTokenSet.Tokens>
                 return .float {
                     switch style().isToast {
                     case true:
-                        return GlobalTokens.borderRadius(.xLarge)
+                        return GlobalTokens.borderRadius(.cornerRadius120)
                     case false:
-                        return GlobalTokens.borderSize(.none)
+                        return GlobalTokens.borderSize(.strokeWidthNone)
```
huwilkes, line 171: Could you change the false case to pull the zero value from the same set of tokens as the non-zero value?

edjamesmsft (approved):

ios/FluentUI/Notification/NotificationTokenSet.swift
edjamesmsft: To clarify: they should both be borderRadius if possible

ios/FluentUI/Avatar/AvatarTokenSet.swift
```diff
@@ -53,17 +53,17 @@ public class AvatarTokenSet: ControlTokenSet<AvatarTokenSet.Tokens> {
                 return .float({
                     switch style() {
                     case .default, .accent, .outlined, .outlinedPrimary, .overflow:
-                        return GlobalTokens.borderRadius(.none)
+                        return GlobalTokens.borderRadius(.cornerRadiusNone)
                     case .group:
                         switch size() {
                         case .size16, .size20:
-                            return GlobalTokens.borderRadius(.small)
+                            return GlobalTokens.borderRadius(.cornerRadius20)
```
harrieshin, line 60: shouldn't the globaltokens.corneradious?

ios/FluentUI/Core/Theme/Tokens/GlobalTokens.swift
```diff
@@ -1711,56 +1735,62 @@ public class GlobalTokens: NSObject {
...
             return 12
-        case .circle:
+        case .cornerRadiusCircular:
             return 9999
         }
     }
 
     // MARK: - BorderSize
 
     public enum BorderSizeToken: TokenSetKey {
```
harrieshin, line 1767: it shouldn't be bordersize anymore?

ios/FluentUI/Core/Theme/Tokens/GlobalTokens.swift
```diff
@@ -1633,36 +1633,60 @@ public class GlobalTokens: NSObject {
...
+        case size160
+        case size200
+        case size240
+        case size280
+        case size320
+        case size360
+        case size400
+        case size480
+        case size520
+        case size560
```
lyzhan7, line 1652: So my understanding of the new size ramp that was recently created is that it was primarily meant to replace the existing spacing ramp. On FURN, we're currently replacing all the existing references to globalTokens.spacing with globalTokens.size. There are currently no global tokens specific to icon size. Is the new size ramp also meant to be used for icon size, or should there be an entirely different ramp for icon size?

laminesm: In FUA we're using both spacing and iconSize. Spacing is mostly used for paddings though. I agree it'd be a good idea to just use size since all but one of the spacing tokens are included in the new size tokens, and that will probably save us some binary as well. Out of curiosity, did design explicitly they wanted us to use size tokens for everything?

lyzhan7: Icon size tokens haven't been talked about at all in the FURN token discussions - if there haven't been conversations with design about this on the FUA side I think this would be good to bring up in the Fluent Apple x Design Discussion channel - i.e. should the existing size ramp also be used for icon size? Or are there plans for a new ramp specific to icon size

lyzhan7:
@laminesm based on the conversation in Teams, it seems like the updated size ramp you used for the IconSize ramp should actually be used to update the SpacingToken ramp (starting around line 1695 of this file)- the IconSize ramp is not yet finalized on the design side and that needs to be done first. I think it's likely that these will be two separate ramps since the use cases are pretty different (ex. the new size ramp is only used for padding/margins/etc. whereas an iconSize ramp would be just for icons). 

Thinking we should scope this PR to just the global token ramps that have already been finalized (corner radius/stroke width/size) so it isn't being blocked by design. Once design finalizes the icon size ramp there can be follow work to update FUA with it.

lyzhan7 (changes requested):
@laminesm based on the conversation in Teams, it seems like the updated size ramp you used for the IconSize ramp should actually be used to update the SpacingToken ramp (starting around line 1695 of this file)- the IconSize ramp is not yet finalized on the design side and that needs to be done first. I think it's likely that these will be two separate ramps since the use cases are pretty different (ex. the new size ramp is only used for padding/margins/etc. whereas an iconSize ramp would be just for icons).

Thinking we should scope this PR to just the global token ramps that have already been finalized (corner radius/stroke width/size) so it isn't being blocked by design. Once design finalizes the icon size ramp there can be follow work to update FUA with it.

lyzhan7 (approved):
Looks good to me

PR #1423, opened 2022-12-02
[main_0.9] Fix scope of window check in Segmented Control 

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| Before | After |
|--------|-------|
| 31,263,518 bytes | 31,263,518 bytes |

Moved the check for if there is a window to be just around the logic for calculating the size of the Segmented Control, so that if UIKit decides to try and get the size of the SegmentedControl before adding it to a window we don't just return zero. Also updated the SegmentedControl to invalidate intrinsic content size if it moves to a new window and the size is based on the size of the window.

### Verification

| SegmentedControl_WindowCheck_Before | SegmentedControl_WindowCheck_After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1423)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -483,6 +483,9 @@ open class SegmentedControl: UIControl {
     open override func didMoveToWindow() {
         super.didMoveToWindow()
         updateWindowSpecificColors()
+        if shouldSetEqualWidthForSegments {
+            invalidateIntrinsicContentSize()
+        }
     }
 
     /// Used to retrieve the view from the segment at the specified index
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -442,9 +442,6 @@ open class SegmentedControl: UIControl {
     }
 
     open override func sizeThatFits(_ size: CGSize) -> CGSize {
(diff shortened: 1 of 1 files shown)


laminesm (approved):

mischreiber (changes requested):
Can you merge to `main` first and then cherry-pick, instead of making two parallel changes? It'll make tracking the change history easier.

huwilkes:
> Can you merge to `main` first and then cherry-pick, instead of making two parallel changes? It'll make tracking the change history easier.

I was going to just not pick the main one to 0.9, and then cherry-pick the 0.9 one to the other branches that needed the update, since the control changed enough underneath that it doesn't really make sense as a cherry-pick.

mischreiber (approved):

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -484,6 +483,9 @@ open class SegmentedControl: UIControl {
     open override func didMoveToWindow() {
         super.didMoveToWindow()
         updateWindowSpecificColors()
+        if shouldSetEqualWidthForSegments {
+            invalidateIntrinsicContentSize()
```
imWildCat, line 487: TIL! Thanks for the fix!

PR #1424, opened 2022-12-02
Fix Avatar crash on complicated unicode names

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

Fixing Avatar hashing in iOS and macOS by preemptively upscaling our unicode point value from a `unichar` (aka an `unsigned short`) to an `Int32`. This ensures that, when bit-shifted by up to 8 bits, all possible UTF-16 values will still fit within the given space without overflowing.

In JavaScript (where this algorithm was first written) this isn't a problem, since all values are automatically upsized to have enough bits. Swift is pickier and will crash if an arithmetic operation results in overflow. No overflow means no crash!

While here, I discovered a couple of other minor issues with how we derive hash codes on both iOS and macOS:
* Both platforms ignore a possible scenario where a `secondaryText` value is provided with no `primaryText` value. Instead of falling back to nothing, let's use whatever we have (aligns with web). Logic is thus greatly simplified.
* On macOS there's a bug where primary and secondary flip during certain layout passes. This resolves that, again to match both iOS and web.

### Verification

Verified with a known set of crashing unicode emoji (though it is also possible to crash with many other sets of UTF-16 characters). Added unit tests to ensure this code stays working and matches what FluentUI web does (note the gray `E` avatar).

| Before | After | FluentUI web |
|-------------------|--------------------|--------------------|
| Screenshot 2022-12-02 at 11 27 56 AM | Screenshot 2022-12-02 at 11 27 17 AM | Screenshot 2022-12-02 at 11 34 24 AM |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -433,7 +433,7 @@ public struct Avatar: View, TokenizedControlView {
             var hash: Int32 = 0
             for len in (0..<text.length).reversed() {
                 let ch = text.character(at: len)
-                let shift = len % 8
+                let shift = Int32(len % 8)
                 hash ^= Int32((ch << shift) + (ch >> (8 - shift)))
               }
 
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -413,17 +413,7 @@ public struct Avatar: View, TokenizedControlView {
         }
 
         static func initialsHashCode(fromPrimaryText primaryText: String?, secondaryText: String?) -> Int {
-            var combined: String
-            if let secondaryText = secondaryText, let primaryText = primaryText, secondaryText.count > 0 {
-                combined = primaryText + secondaryText
-            } else if let primaryText = primaryText {
-                combined = primaryText
-            } else if let secondaryText = secondaryText {
-                combined = secondaryText
-            } else {
-                combined = ""
-            }
-
+            let combined = (primaryText ?? "") + (secondaryText ?? "")
             let combinedHashable = combined as NSString
             return Int(abs(hashCode(combinedHashable)))
         }
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -418,6 +418,8 @@ public struct Avatar: View, TokenizedControlView {
                 combined = primaryText + secondaryText
             } else if let primaryText = primaryText {
                 combined = primaryText
+            } else if let secondaryText = secondaryText {
+                combined = secondaryText
             } else {
                 combined = ""
             }
@@ -432,8 +434,8 @@ public struct Avatar: View, TokenizedControlView {
         private static func hashCode(_ text: NSString) -> Int32 {
             var hash: Int32 = 0
             for len in (0..<text.length).reversed() {
-                let ch = text.character(at: len)
-                let shift = Int32(len % 8)
+                let ch = Int32(text.character(at: len))
+                let shift = len % 8
                 hash ^= Int32((ch << shift) + (ch >> (8 - shift)))
               }
 
--- a/macos/FluentUI/AvatarView/AvatarView.swift
+++ b/macos/FluentUI/AvatarView/AvatarView.swift
@@ -360,7 +360,7 @@ open class AvatarView: NSView {
 		var updatedBackgroundColor = avatarBackgroundColor
 		var updatedInitialsColor = initialsFontColor
 		if !isCustomAvatarBackgroundColorConfigured || !isCustomAvatarInitialColorConfigured {
-			let defaultColorSet = AvatarView.getInitialsColorSet(fromPrimaryText: contactEmail, secondaryText: contactName)
+			let defaultColorSet = AvatarView.getInitialsColorSet(fromPrimaryText: contactName, secondaryText: contactEmail)
 
 			if !isCustomAvatarBackgroundColorConfigured {
 				updatedBackgroundColor = defaultColorSet.background.resolvedColor(appearance)
@@ -499,9 +499,13 @@ open class AvatarView: NSView {
 		if initials == nil,
 			let email = email?.trimmingCharacters(in: whitespaceNewlineAndZeroWidthSpace),
 			!email.isEmpty {
-			let initialCharacter = email[email.startIndex]
-			if initialCharacter.isValidInitialsCharacter {
-				initials = String(initialCharacter)
+			let emailComponentsWithUnicodeLetterFirstCharacters = email.filter {
+				$0.isValidInitialsCharacter
+			}
+
+			// Find the first valid unicode character in the email address and use that
+			if let initial = emailComponentsWithUnicodeLetterFirstCharacters.first {
+				initials = String(initial)
 			}
 		}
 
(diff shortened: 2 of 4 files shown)


macos/FluentUITestViewControllers/TestAvatarViewController.swift
```diff
@@ -19,6 +19,7 @@ private struct TestData {
 	static let maor = TestIdentity(name: "Maor Sharett", email: "Maor.Sharett@example.com", image: nil)
 	static let annieBoyl = TestIdentity(name: "Annie Boyl Lind", email: "annie.boyl@example.com", image: nil)
 	static let kat = TestIdentity(name: nil, email: "Kat.Larrson@example.com", image: nil)
+	static let turkey = TestIdentity(name: "🦃", email: "🦃🦃🦃🦃🦃🦃🦃🦃@example.com", image: nil)
```
huwilkes, line 22: Can we add this nice bird to our iOS demo?

huwilkes (approved):

macos/FluentUITestViewControllers/TestAvatarViewController.swift
huwilkes: Also, seems a bit odd that the first letter we find is the e from @example.com. Are we sure this algorithm isn't biased against certain languages?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -433,7 +433,7 @@ public struct Avatar: View, TokenizedControlView {
             var hash: Int32 = 0
             for len in (0..<text.length).reversed() {
                 let ch = text.character(at: len)
-                let shift = len % 8
+                let shift = Int32(len % 8)
```
lyzhan7: Would it be useful to add a comment on why converting to Int32 is necessary?

lyzhan7 (approved):

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -413,15 +413,7 @@ public struct Avatar: View, TokenizedControlView {
...
-            var combined: String
-            if let secondaryText = secondaryText, let primaryText = primaryText, secondaryText.count > 0 {
-                combined = primaryText + secondaryText
-            } else if let primaryText = primaryText {
-                combined = primaryText
-            } else {
-                combined = ""
-            }
-
+            let combined = (primaryText ?? "") + (secondaryText ?? "")
```
edjamesmsft, line 416: Is concatenating two empty strings different than just one empty string in Swift? Just double checking to make sure this is semantically equivalent to what it's replacing!

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -360,7 +360,7 @@ open class AvatarView: NSView {
 		var updatedBackgroundColor = avatarBackgroundColor
 		var updatedInitialsColor = initialsFontColor
 		if !isCustomAvatarBackgroundColorConfigured || !isCustomAvatarInitialColorConfigured {
-			let defaultColorSet = AvatarView.getInitialsColorSet(fromPrimaryText: contactEmail, secondaryText: contactName)
+			let defaultColorSet = AvatarView.getInitialsColorSet(fromPrimaryText: contactName, secondaryText: contactEmail)
```
edjamesmsft, line 363: If order matters, is it worth it to update getInitialColorSet's param names to reflect that? If we can't due to breaking public change, should we ensure it's at least documented well? Is there any other input verification we can do on secondaryText to ensure we have these ordered correctly?

macos/FluentUI/AvatarView/AvatarView.swift
```diff
@@ -499,9 +499,13 @@ open class AvatarView: NSView {
...
 			!email.isEmpty {
-			let initialCharacter = email[email.startIndex]
-			if initialCharacter.isValidInitialsCharacter {
-				initials = String(initialCharacter)
+			let emailComponentsWithUnicodeLetterFirstCharacters = email.filter {
+				$0.isValidInitialsCharacter
+			}
+
+			// Find the first valid unicode character in the email address and use that
+			if let initial = emailComponentsWithUnicodeLetterFirstCharacters.first {
```
edjamesmsft, line 507: How do we ensure this behavior is used consistently across all of our platforms (since this is outside of the hashing code)?

macos/FluentUITestViewControllers/TestAvatarViewController.swift
edjamesmsft: I'd suggest adding a few more test cases - different languages, different names, different combinations of emojis and characters.

mischreiber: Two thoughts:
* Our iOS demo currently doesn't have much coverage at all for name variety -- it's architected to mainly pivot on appearance style. Adding various names as a test case would potentially require a substantial rethinking of the control, which I'm not currently budgeted to undertake. But I'll make sure it gets on our backlog, with this particular test case called out.
* I agree that it's odd that we skip all the way to "e", but the Avatar is currently behaving as specced -- to only use Latin alphabet letters as initials. This aligns with all Fluent platforms, and reconsideration needs to be a cross-platform feature request.

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Sure! This function is a little opaque in general.

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Good question! Yep, output is identical.

macos/FluentUI/AvatarView/AvatarView.swift
mischreiber: I don't think we can at this point, at least not without modifying the public API to align with iOS. I agree that it's not good, but unfortunately we'd have to prioritize that against our existing iOS work.

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Comment added. Also this whole PR just reminds me how badly I want to refactor this codebase to allow for code sharing between iOS and macOS.

macos/FluentUI/AvatarView/AvatarView.swift
sophialee0416: The params portion is a little complicated right now because our params don"t even align with web. Documentation might be the better option as of now until that gets resolved.

sophialee0416 (approved):

macos/FluentUITestViewControllers/TestAvatarViewController.swift
huwilkes: 1. I think you'd be fine just adding a new entry to the list of sample personas from the persona list view demo controller?
2. That's fair, and something we should bring up with the other platforms, probably.

macos/FluentUI/AvatarView/AvatarView.swift
mischreiber: Commented!

macos/FluentUITestViewControllers/TestAvatarViewController.swift
mischreiber: Fun fact: the iOS algorithm does behave differently here than macOS or web -- it only ever checks the first letter of a string for unicode-ness, as opposed to the other platforms that check iteratively through the whole string.

I'm gonna go ahead and _not_ change that now because this PR is susceptible to a whole lot of bloat, but we should probably revisit that at some point.

huwilkes (approved):

PR #1425, opened 2022-12-02
Fix scope of window check in Segmented Control

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| Before | After |
|--------|-------|
| 31,640,362 bytes | 31,640,362 bytes |

Mirror of #1423 in main, since it's a bit more involved than a cherry pick.
The updated SegmentedControl wasn't having the issue, but, the logic is still bad.

### Verification

| SegmentedControl_WindowCheck_Main_Before | SegmentedControl_WindowCheck_Main_After |
| SegmentedControl_WindowCheck_Main_iPhone_Before | SegmentedControl_WindowCheck_iPhone_Before |
| SegmentedControl_WindowCheck_iPad_Before | SegmentedControl_WindowCheck_iPad_After |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1425)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentedControl.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentedControl.swift
@@ -359,9 +359,6 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
     }
 
     open override func sizeThatFits(_ size: CGSize) -> CGSize {
-        guard let window = window else {
-            return CGSize.zero
-        }
         var height: CGFloat = 0.0
         var width: CGFloat = 0.0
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -373,32 +370,33 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
+        return CGSize(width: min(width, size.width),
+                      height: min(height, size.height))
     }
 
     open override func didMoveToWindow() {
         super.didMoveToWindow()
 
         tokenSet.update(fluentTheme)
-        updateColors()
-        updateButtons()
```
laminesm, line 400: Why are the update calls removed?

ios/FluentUI/SegmentedControl/SegmentedControl.swift
```diff
@@ -373,32 +370,33 @@ open class SegmentedControl: UIView, TokenizedControlInternal {
...
         }
 
-        let windowSafeAreaInsets = window.safeAreaInsets
-        let windowWidth = window.bounds.width - windowSafeAreaInsets.left - windowSafeAreaInsets.right
         if isFixedWidth {
-            if traitCollection.userInterfaceIdiom == .pad {
-                width = max(windowWidth / 2, Constants.iPadMinimumWidth)
-            } else {
-                width = windowWidth
+            if let window = window {
```
mischreiber, line 374: Can you combine these two `if` statements?
```swift
if isFixedWidth, let window = window {
```

ios/FluentUI/SegmentedControl/SegmentedControl.swift
huwilkes: Because they're in `tokenSetSink`, so if the theme changes they'll get updated. We don't need to update the colors twice.

ios/FluentUI/SegmentedControl/SegmentedControl.swift
huwilkes: That's not really the same logic, because right now if it fails the window check it doesn't go through the else of isFixedWidth.

mischreiber: Ahhhhhh nested `if` statements make my brain go 🤯

mischreiber (approved):

PR #1434, opened 2022-12-06
Merge tokenized Tooltip into main

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Merging tokenized Tooltip and documentation into main 
- #1299 
- #1333
- #1341
- #1392

Screenshots:
| Before (main) | After (main) | Current (fluent2-tokens) |
|---------------|-------------|--------------------------|
| IMG_0039 | IMG_0051 | IMG_0052 |
| IMG_0037 | IMG_0050 | IMG_0053 |

Binary change:
| File | Before | After | Delta |
|------|--------|-------|-------|
| libFluentUI.a | 28,843,088 bytes (28.8 MB on disk) | 29,139,232 bytes (29.1 MB on disk) | 296,144 bytes |
| TooltipView.swift | 85,712 bytes (86 KB on disk) | 247,728 bytes (250 KB on disk) | 162,016 bytes |
| TooltipViewController.swift | N/A | 47,472 bytes (49 KB on disk) | 47,472 bytes |
| Tooltip.swift | 99,832 bytes (102 KB on disk) | 141,480 bytes (143 KB on disk) | 41,648 bytes |
| TooltipTokenSet.swift | N/A | 64,008 bytes (66 KB on disk) | 64,008 bytes |
| TooltipPositionController.swift | 62,824 bytes (66 KB on disk) | N/A | -62,824 bytes |

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1434)

PR #1439, opened 2022-12-07
Fix Tooltip Theming

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
@sophialee0416 noticed I didn't add theming support validation in #1434 (thank you!) and doing so uncovered that theming wasn't working due to nil tooltipView, so let's use anchorView instead. We also don't need to be passing tokenSet everywhere when it's exposed in Tooltip.shared, so let's simplify that logic.

Binary change:
| libFluentUI.a | 32,840,160 bytes (32.8 MB on disk) | 32,843,280 bytes (32.8 MB on disk) | 3,120 bytes |
| Tooltip.swift | 141,368 bytes (143 KB on disk) | 140,816 bytes (143 KB on disk) | -552 bytes |
| TooltipTokenSet.swift | 64,008 bytes (66 KB on disk) | 64,008 bytes (66 KB on disk) | 0 bytes |
| TooltipView.swift | 233,848 bytes (238 KB on disk) | 236,336 bytes (238 KB on disk) | 2,488 bytes |
| TooltipViewController.swift | 47,040 bytes (49 KB on disk) | 47,824 bytes (49 KB on disk) | 784 bytes |

| IMG_0046 | IMG_0049 |
| IMG_0047 | IMG_0048 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1439)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
@@ -155,8 +155,8 @@ extension TooltipDemoController: DemoAppearanceDelegate {
         return [
             .tooltipColor: .dynamicColor {
                 // "Berry"
-                return DynamicColor(light: GlobalTokens.sharedColors(.berry, .tint50),
-                                    dark: GlobalTokens.sharedColors(.berry, .shade40))
+                return DynamicColor(light: GlobalTokens.sharedColors(.berry, .shade30),
+                                    dark: GlobalTokens.sharedColors(.berry, .tint20))
             }
         ]
     }
@@ -165,13 +165,13 @@ extension TooltipDemoController: DemoAppearanceDelegate {
         return [
             .tooltipColor: .dynamicColor {
                 // "Brass"
-                return DynamicColor(light: GlobalTokens.sharedColors(.brass, .tint50),
-                                    dark: GlobalTokens.sharedColors(.brass, .shade40))
+                return DynamicColor(light: GlobalTokens.sharedColors(.brass, .tint40),
+                                    dark: GlobalTokens.sharedColors(.brass, .shade30))
             },
             .textColor: .dynamicColor {
                 // "Forest"
-                return DynamicColor(light: GlobalTokens.sharedColors(.forest, .tint10),
-                                    dark: GlobalTokens.sharedColors(.forest, .shade40))
+                return DynamicColor(light: GlobalTokens.sharedColors(.forest, .shade30),
+                                    dark: GlobalTokens.sharedColors(.forest, .tint40))
             },
             .backgroundCornerRadius: .float {
                 return 0
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
@@ -17,7 +17,7 @@ class TooltipDemoController: DemoController {
         titleView.setup(title: title ?? "", interactivePart: .title)
         titleView.delegate = self
         navigationItem.titleView = titleView
-        navigationItem.rightBarButtonItem = UIBarButtonItem(title: "Show on title", style: .plain, target: self, action: #selector(showTitleTooltip))
+        navigationItem.rightBarButtonItems?.append(UIBarButtonItem(title: "Show on title", style: .plain, target: self, action: #selector(showTitleTooltip)))
 
         container.addArrangedSubview(createButton(title: "Show single-line tooltip below", action: { sender in
             Tooltip.shared.show(with: "This is pointing up.", for: sender, preferredArrowDirection: .up)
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -242,14 +241,15 @@ open class Tooltip: NSObject, TokenizedControlInternal {
     public var tokenSet: TooltipTokenSet = .init()
     var tokenSetSink: AnyCancellable?
     var fluentTheme: FluentTheme {
-        guard let tooltipView = tooltipViewController?.view else {
+        // Use anchor view to get theme since tooltip view will most likely be nil
```
jeaniehuynh, line 244: @mischreiber can you confirm if this is an okay workaround? are there any side effects to this/is it hacky? I assumed it'd be okay since the anchor view and tooltip view will always be in the same window, but please let me know if that isn't correct :)

sophialee0416 (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
```diff
@@ -131,3 +131,51 @@ extension TooltipDemoController: TwoLineTitleViewDelegate {
...
+        return [
+            .tooltipColor: .dynamicColor {
+                // "Berry"
+                return DynamicColor(light: GlobalTokens.sharedColors(.berry, .tint50),
+                                    dark: GlobalTokens.sharedColors(.berry, .shade40))
+            }
+        ]
+    }
+
+    private var perControlOverrideTooltipTokens: [TooltipTokenSet.Tokens: ControlTokenValue] {
```
edjamesmsft, line 164: This combination of colors looks like it might not satisfy accessibility contrast requirements. Have you checked it with the colour contrast analyser? Otherwise we might get accessibility bugs filed on us!

ios/FluentUI/Tooltip/Tooltip.swift
harrieshin: this is also concerning that Tooltip.swift holds on the the anchorView and there is additional private let of anchorview inside the TooltipView?

ios/FluentUI.Demo/FluentUI.Demo/Demos/TooltipDemoController.swift
jeaniehuynh: Even if it's just demo colors to test the override? I just used the same colors Sophia used for the TableViewCell demo

jeaniehuynh: Gotcha, good catch. I'll update it! Thanks 😊

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -236,20 +234,22 @@ open class Tooltip: NSObject, TokenizedControlInternal {
     @objc public var textAlignment: NSTextAlignment = .natural
     /// Whether a tooltip is currently showing.
     @objc public private(set) var isShowing: Bool = false
+    public var anchorView: UIView?
```
mischreiber, line 237: * Does this need to be `@objc` exposed?
* Does this need to be public, or just internal? I don't see any users outside of Fluent (e.g. the test app).

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -9,21 +9,22 @@ import UIKit
...
-         message: String,
+    init(message: String,
          title: String? = nil,
          textAlignment: NSTextAlignment,
          preferredArrowDirection: Tooltip.ArrowDirection,
          offset: CGPoint,
-         arrowMargin: CGFloat,
-         tokenSet: TooltipTokenSet) {
+         arrowMargin: CGFloat) {
+        guard let anchorView = Tooltip.shared.anchorView else {
```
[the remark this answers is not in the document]
jeaniehuynh: It was to address Harrie's comment above: "_this is also concerning that Tooltip.swift holds on the the anchorView and there is additional private let of anchorview inside the TooltipView?_"

If I understand correctly, since Tooltip itself is a singleton, the change is essentially doing the same thing and simplifies the logic?

jeaniehuynh:
Apologies for the stream of consciousness commits, things kept coming to mind!

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -56,8 +58,8 @@ class TooltipViewController: UIViewController {
         }
     }
 
-    func updateAppearance(tokenSet: TooltipTokenSet) {
-        tooltipView.updateAppearance(tokenSet: tokenSet)
+    func updateAppearance() {
+        tooltipView.updateAppearance(tokenSet: Tooltip.shared.tokenSet)
```
mischreiber, line 62: Still hoping to avoid any reference to singletons if possible.

ios/FluentUI/Tooltip/TooltipViewController.swift
```diff
@@ -9,27 +9,29 @@ import UIKit
...
+    init(anchorView: @escaping () -> UIView?,
          message: String,
          title: String? = nil,
          textAlignment: NSTextAlignment,
          preferredArrowDirection: Tooltip.ArrowDirection,
          offset: CGPoint,
-         arrowMargin: CGFloat,
-         tokenSet: TooltipTokenSet) {
+         arrowMargin: CGFloat) {
+        guard let anchorView = anchorView() else {
```
mischreiber, line 19: Could you talk through the advantages of using a callback here instead of just passing the anchorView directly through?

jeaniehuynh: I was using your closure slides from the PPT Harrie sent as reference when doing this and thought it was needed to keep anchorView weak! But if that's not the case, I can change it back to just passing anchorView

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -324,6 +338,10 @@ class TooltipView: UIView {
     }
 
     private var sourcePointInAnchorView: CGPoint {
+        guard let anchorView = anchorView else {
+            preconditionFailure("Can't find anchorView")
```
mischreiber, line 342: Potential concern -- do we want to crash in this case? I'd lean towards asserting (so we catch it in debug builds) and then return a default `(0,0)` value.

mischreiber: Same comment for the other `preconditionFailure` cases below (`sourcePointInWindow`, `boundingRect`, `anchorViewInset(for:)`.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -368,24 +394,25 @@ class TooltipView: UIView {
         return inset
     }
 
+    private weak var anchorView: UIView?
+    private weak var tokenSet: TooltipTokenSet?
```
mischreiber, line 398: Thinking out loud: it's probably important to keep `anchorView` as a `weak` pointer, but `tokenSet` might be fine staying strong? The ownership model is that views own `tokenSet`s, but there should never ever be strong references in the reverse direction, so it's presumably safe to keep strong and not have to unwrap? But I don't personally have a strong opinion one way or the other.

ios/FluentUI/Tooltip/TooltipView.swift
```diff
@@ -132,48 +132,48 @@ class TooltipView: UIView {
...
-        var usePreferred = true
-        if (preferredArrowDirection.isVertical &&
-            preferredBoundingRect.height < preferredSize.height &&
-            backupBoundingRect.height >= backupSize.height) ||
-            (!preferredArrowDirection.isVertical &&
-             preferredBoundingRect.width < preferredSize.width &&
-             backupBoundingRect.width >= backupSize.width) {
-            usePreferred = false
-        }
+        if let window = anchorView?.window {
```
mischreiber, line 135: Per the Microsoft Swift guide, these large `if let x = z?.x` statements should instead be structured as a `guard let`, which will also remove the need to change the indentation of the below block.

ios/FluentUI/Tooltip/Tooltip.swift
```diff
@@ -242,14 +242,15 @@ open class Tooltip: NSObject, TokenizedControlInternal {
     public var tokenSet: TooltipTokenSet = .init()
     var tokenSetSink: AnyCancellable?
     var fluentTheme: FluentTheme {
-        guard let tooltipView = tooltipViewController?.view else {
+        // Use anchor view to get theme since tooltip view will most likely be nil
+        guard let anchorView = anchorView else {
```
mischreiber, line 246: Last thought: could the var pointer to `anchorView` down at the bottom of the file also be `weak` for safety? Shouldn't result in any additional changes to the class.

PR #1446, opened 2022-12-08
Adding additional leading view to TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adding a new "unread dot" view to the TableView Cell.

Binary change:
| libFluentUI.a | 28,593,248 bytes | 28,617,896 bytes | +24,648 bytes |
| TableViewCell | 818,360 bytes | 835,600 bytes | +17,240 bytes |
| TableViewCellTokenSet | 118,528 bytes | 118,424 bytes | -104 bytes |

### Verification

| Default                                       | In Selection Mode                                      |
| image | image |
| image | image |
| image | image |

Testing RTL: 
image

Testing unread/read dot changes in Selection mode:
Simulator Screen Recording - iPhone 14 Pro Max - 2022-12-09 at 17 41 56

Testing theming:
Simulator Screen Recording - iPhone 14 Pro Max - 2022-12-09 at 17 43 37

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1446)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1506,10 +1529,24 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
             )
         }
 
+        if let cellLeadingView = cellLeadingView {
+            let customViewDimensions: CGFloat = tokenSet[.cellLeadingViewDimensions].float
+            let customViewYOffset = ceil((contentView.frame.height - customViewDimensions) / 2)
+            let customViewXOffset = TableViewCell.customViewLeadingOffset(isInSelectionMode: isInSelectionMode,
```
sophialee0416, line 1535: So I totally could have made my own calculation here instead of using the `customViewLeadingOffset` static method here, but I chose this route because of code consistency + cleanliness. I do think that it might have compromised the binary size tho, because now tokens are being called + passed in every where else it's used in this file. Any thoughts would be appreciated :)

sophialee0416:
I also made this little property take in any UIView, but if we want to be more strict about this particular property and *only* make it as a dot, I can happily make that change as well.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -882,6 +892,19 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         }
     }
 
+    /// The custom view on the leading edge of the `customView` UIView. Does not display if layoutType is `.oneLine`.
```
harrieshin, line 895: why not in .oneline?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1506,10 +1529,24 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
             )
         }
 
+        if let cellLeadingView = cellLeadingView {
+            let cellLeadingViewDimensions: CGFloat = tokenSet[.cellLeadingViewDimensions].float
+            let cellLeadingViewYOffset = ceil((contentView.frame.height - cellLeadingViewDimensions) / 2)
+            let cellLeadingViewXOffset = TableViewCell.customViewLeadingOffset(isInSelectionMode: isInSelectionMode,
```
sophialee0416, line 1535: So I totally could have made my own calculation here instead of using the customViewLeadingOffset static method here, but I chose this route because of code consistency + cleanliness. I do think that it might have compromised the binary size tho, because now tokens are being called + passed in every where else it's used in this file. Any thoughts would be appreciated :)

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Specs only want it in multi-line

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -882,6 +892,19 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         }
     }
 
+    /// The custom view on the leading edge of the `customView` UIView. Does not display if layoutType is `.oneLine`.
+    @objc open var cellLeadingView: UIView? {
+        didSet {
+            if layoutType == .oneLine {
```
harrieshin, line 898: having this customizable cellleadinView is awesome! but i think we should also havea default circle drawn if it is part of fluent library now?

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Actually I suppose if I allow flexibility for the UIView, I can allow flexibility in any layoutType?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -409,6 +412,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     ///   - title: The title string
     ///   - subtitle: The subtitle string
     ///   - footer: The footer string
+    ///   - cellLeadingView: The view on the leading edge of the `customView`
```
harrieshin, line 415: something about this name is repetitive. you are already in cell class, so calling cell again seem redundant. you don't say "cellTitle" for example. customLeadingView maybe?

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: I was thinking this as well but wanted to leave options open just in case! It's a quick fix, so I'll go ahead and make that change.

harrieshin: follow example of badges how we use CALayer to draw tiny dots.  layers are more efficient than imageview sometimes.

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: That was my original naming actually, but changed it because it felt also redundant with the existing customView. But maybe with making the view only a pre-set dot, I can just go ahead and name it as leadingDotView? (or the like)

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Wait sorry, should I have a fallback default circle, or circle only and turn this into a Bool? (etc. showCellLeadingView)

ios/FluentUI/Table View/TableViewCell.swift
harrieshin: yeah it feels little strange. "you have all the knobs! jk... only certain times"

edjamesmsft:
Nit: could you explain in the PR description a bit more thoroughly what this change is and why?

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Resolved offline: Passing in a bool and setting circle internally

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Resolved with prior convo

ios/FluentUI/Table View/TableViewCell.swift
mischreiber: This definitely feels like the right code design/architecture to be using. Is there a specific binary size concern?

sophialee0416: This code will be changing so resolving this!

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -172,7 +172,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     /// The default leading padding in the cell.
     @objc public static let defaultPaddingLeading: CGFloat = {
         let tokenSet = TableViewCellTokenSet(customViewSize: { .default })
-        return tokenSet[.paddingLeading].float
+        return TableViewCellTokenSet.paddingLeading
```
edjamesmsft, line 175: If you're referencing the TableViewTokenSet this way now, do you need the above let? Also what is the rationale for this change?

ios/FluentUI.Demo/FluentUI.Demo/TableViewSampleData.swift
```diff
@@ -25,9 +26,25 @@ class TableViewSampleData {
...
+             hasFooter: Bool = false,
+             footerText: String = "",
+             footerLinkText: String = "",
+             hasCustomLinkHandler: Bool = false,
+             hasCustomAccessoryView: Bool = false,
+             hasCustomLeadingView: Bool = false,
+             hasHandler: Bool = false) {
             self.title = title
             self.items = items
+            self.isUnreadDotVisible = isUnreadDotVisible
```
edjamesmsft, line 47: Nit: you call it the leadingDot elsewhere, should we be consistent with naming?

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -424,6 +428,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     @objc public class func preferredWidth(title: String,
                                            subtitle: String = "",
                                            footer: String = "",
+                                           hasLeadingDot: Bool = false,
```
edjamesmsft, line 431: Is this going to be a breaking change?

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: good catch! Also we are moving away from spacing/padding as customizable tokens, hence the change to constant tokens

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -645,16 +652,17 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
     }
 
     private static func customViewLeadingOffset(isInSelectionMode: Bool,
-                                                tokenSet: TableViewCellTokenSet) -> CGFloat {
-        return tokenSet[.paddingLeading].float + selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
-                                                                        selectionImageMarginTrailing: TableViewCellTokenSet.selectionImageMarginTrailing,
-                                                                        selectionImageSize: TableViewCellTokenSet.selectionImageSize)
+                                                leadingPadding: CGFloat) -> CGFloat {
+        return leadingPadding + selectionModeAreaWidth(isInSelectionMode: isInSelectionMode,
```
edjamesmsft, line 656: I like NOT having to pass in the entire token set for a single value. Good change.

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -882,6 +890,33 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         }
     }
 
+    private lazy var unreadDotLayer: CALayer = {
+        let unreadDotLayer = CALayer()
+        let unreadDotSize = TableViewCellTokenSet.leadingDotDimensions
```
edjamesmsft, line 895: Why are you referencing TableViewTokenSet directly here, rather than tokenSet?

sophialee0416: Any time I use TableViewTokenSet, that means I'm using a token constant! Which is what leadingDotDimensions happens to be.

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: It was necessary before we had token constants :') Luckily no longer!

ios/FluentUI/Table View/TableViewCell.swift
sophialee0416: Good catch, yes it is. I'll put the label on!

ios/FluentUI.Demo/FluentUI.Demo/TableViewSampleData.swift
sophialee0416: Yes I totally thought I had replaced everything but clearly not! Changed it!

sophialee0416:
Interestingly, I think changing to CALayer for the unread dot might have increased the binary size further as opposed to using a UIView.

edjamesmsft:
> Interestingly, I think changing to CALayer for the unread dot might have increased the binary size further as opposed to using a UIView.

Do you know by how much? But either way, if the CALayer is more expensive, go for the UIView unless there are clear functional benefits to using the CALayer.

sophialee0416:
> > Interestingly, I think changing to CALayer for the unread dot might have increased the binary size further as opposed to using a UIView.
> 
> Do you know by how much? But either way, if the CALayer is more expensive, go for the UIView unless there are clear functional benefits to using the CALayer.

It was by 4kB I think? It might be because we weren't drawing any views prior. I think sticking to CALayer will be better for now

edjamesmsft (approved):

PR #1449, opened 2022-12-08
Restore SegmentedControl init with custom colors

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,510,136 bytes | 28,560,792 bytes  | 50,656 bytes |
| SegmentedControl.o | 350,208 bytes | 378,456 bytes | 28,248 bytes |

I removed this init in #940 when our guidance was "All color customization must be through tokens always", but we've relaxed on that point. So I'm adding it back (and redirecting the custom colors through tokens) to align a bit more with the SegmentedControl from before it was tokenized.

### Verification

| Light | Dark | Theme Overrides |
|---|---|---|
| SegmentedControl_CustomColors_Light | SegmentedControl_CustomColors_Dark | SegmentedControl_CustomColors_Overrides |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1449)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
@@ -98,35 +98,30 @@ class SegmentedControlDemoController: DemoController {
     }
 
     func addPillControl(items: [SegmentItem], style: SegmentedControlStyle, equalSegments: Bool = true, enabled: Bool = true, isFixedWidth: Bool = true, customColors: Bool = false) {
-        let backgroundColor: UIColor?
-        let selectedBackgroundColor: UIColor?
-        let textColor: UIColor?
-        let selectedTextColor: UIColor?
-        let unreadDotColor: DynamicColor?
+        let pillControl: SegmentedControl
         if customColors {
-            backgroundColor = UIColor(dynamicColor: DynamicColor(light: GlobalTokens.sharedColors(.bronze, .shade30),
-                                                                 dark: GlobalTokens.sharedColors(.bronze, .tint40)))
-            selectedBackgroundColor = UIColor(dynamicColor: DynamicColor(light: GlobalTokens.sharedColors(.steel, .tint40),
-                                                                         dark: GlobalTokens.sharedColors(.steel, .shade30)))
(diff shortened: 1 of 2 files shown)


edjamesmsft:
Nit: can you list in the PR the commit that removed this initially? Not a big deal though.

edjamesmsft (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/SegmentedControlDemoController.swift
```diff
@@ -82,14 +82,51 @@ class SegmentedControlDemoController: DemoController {
...
+            unreadDotColor = DynamicColor(light: GlobalTokens.sharedColors(.charcoal, .tint40),
+                                          dark: GlobalTokens.sharedColors(.charcoal, .shade30))
+        } else {
+            backgroundColor = nil
+            selectedBackgroundColor = nil
+            textColor = nil
+            selectedTextColor = nil
+            unreadDotColor = nil
+        }
+        let pillControl = SegmentedControl(items: items,
```
edjamesmsft, line 124: Slight nit - this now skips the path of the standard init, reducing code coverage. Is there a way to re-write this to cover both initializers, so we have fewer opportunities for bugs to sneak in?

mischreiber (changes requested):
> we've relaxed on that point

We have not relaxed on this point overall. Especially now that the token customization API has been simplified to require no overrides/polymorphism, there's no reason to keep these old styles of customization around. Especially since they've already been removed, there's not an obvious reason to bring it back.

huwilkes:
> > we've relaxed on that point
> 
> We have not relaxed on this point overall. Especially now that the token customization API has been simplified to require no overrides/polymorphism, there's no reason to keep these old styles of customization around. Especially since they've already been removed, there's not an obvious reason to bring it back.

They've been removed, but only really in the tokens branch. No actual release has removed them. The reason to bring them back is that it's just not possible to use the tokens in Objective C to override those colors. I've talked with @harrieshin about it, though, and in this case we're going to abandon this PR and just change the colors around the segmented control where this init was being used.

Issue #1427, opened 2022-12-05
AvatarGroup - Thread Warning


jpeng-ms:
### Environment Information

- **Platform**:
  - [X] iOS
  - [ ] macOS
- **Package version(s)**: (fill this out, include which package manager you're using)
- **Xcode and OS versions**: (fill this out if relevant)

### Please provide a reproduction of the bug:

when using avatar group in my project, it would trigger a lot of thread warnings when avatar group gets updated:
<img width="424" alt="image" src="https://user-images.githubusercontent.com/109105353/205725525-b6f04ae5-ea9f-44ea-a3d5-f335e4f9e6f9.png">

I have created a simple project that reproduce this issue. All you need to do is run the project, and issue would be triggered (I have set up some dispatch group to update the content).

<img width="708" alt="image" src="https://user-images.githubusercontent.com/109105353/205725897-bc47f411-eecf-4c59-9d7d-c05edb95b5f1.png">

download simple code:
https://github.com/jpeng-ms/FluentTest

#### Actual behavior:

got the warnings printed on Xcode:
```
Publishing changes from within view updates is not allowed, this will cause undefined behavior.
```

#### Expected behavior:

no warnings

### Priorities and help requested:

Are you willing to submit a PR to fix? (No)

Requested priority: (Normal)

Products/applications affected: (if applicable)

jpeng-ms:
Hi @sophialee0416 would you mind to take a look when you got a chance?

sophialee0416:
Thanks for creating the issue @jpeng-ms! I will take a look at this at my soonest availability.

jpeng-ms:
Hi @laminesm, I am not sure why this was closed. I just tested it out the latest release (0.10.1), the issue is still reproducible:

https://github.com/jpeng-ms/FluentTest/tree/release-0.10.1

jpeng-ms:
<img width="808" alt="image" src="https://user-images.githubusercontent.com/109105353/211088282-d7cd9ea0-7a26-45fc-ba91-ceb18f3c53ff.png">
<img width="677" alt="image" src="https://user-images.githubusercontent.com/109105353/211088303-f27ee8ff-56ea-4ecc-b7f4-53a78cd01d0d.png">

jpeng-ms:
@sophialee0416 @harrieshin can we reopen this since it's still reproducible?

emlynmac:
Just confirming; this is still reproducible on 0.13.2

[...]

PR #1450, opened 2022-12-09
Move `tokenSet.update(_:)` to `.onChange(of:)`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This resolves issue #1427 by moving Fluent token set updates away from a mid-render update and into a `.onChange` block for the environment variable. This ensures that the view state cannot be interrupted mid-render, silencing a new warning introduced in Xcode 14.

Binary change:

| libFluentUI.a | 28,689,720 | 28,838,304 | 148,584 |
| FluentUITestApp | 31,734,955 | 31,798,411 | 63,456 |

### Verification

Verified that modified controls update as expected. See `CardNudge` demo below:

Simulator Screen Recording - iPad Pro (11-inch) (4th generation) - 2022-12-09 at 11 30 34

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1450)

### DIFF
--- a/ios/FluentUI/Core/Theme/Tokens/TokenizedControlView.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/TokenizedControlView.swift
@@ -52,7 +52,9 @@ extension View {
     /// - Returns: The rendered view after applying updates.
     func fluentTokens<TokenSetType: Hashable>(_ tokenSet: ControlTokenSet<TokenSetType>,
                                               _ fluentTheme: FluentTheme) -> some View {
-        tokenSet.update(fluentTheme)
         return self
+            .onChange(of: fluentTheme) { newTheme in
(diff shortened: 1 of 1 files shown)


Saadnajmi (approved):

edjamesmsft:
Are we ok with a 148k delta?

mischreiber:
> Are we ok with a 148k delta?

To be clear, I really really don't like it. _And_, I don't personally know of another way to write this code correctly. I'd be open to any alternative approaches/solutions, but AFAIK this is the "right" way to observe, and react to, changes to an environment variable.

edjamesmsft (approved):

PR #1456, opened 2022-12-15
[Draft] Add functionality to enable the usage of group style avatars in PersonaButtonCarousel

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There is a partner request to enable the usage of group style avatars in `PersonaButtonCarousel`. I added a boolean variable `isGroupStyle` in the `MSFPersonaButtonCarouselState` protocol. As the name suggests, this boolean is used to determine whether the carousel contains regular avatars or group avatars. 
I confirmed with design that there are no changes to the paddings (or any other values) for the group style carousel. 

Binary change:

| libFluentUI.a | 28,864,736 bytes | 28,880,512 bytes | 15,776 bytes |

### Verification

| Screenshot 2022-12-15 at 6 27 28 PM | Screenshot 2022-12-15 at 6 27 33 PM |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1456)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
@@ -33,11 +33,14 @@ class PersonaButtonCarouselDemoController: DemoTableViewController {
         var actionButtons: Bool = false
         var size: MSFPersonaButtonSize = .large
         var carousel: Bool = false
+        var isGroupStyle: Bool = false
     }
     let tableInfo: [TableSectionInfo] = [
         TableSectionInfo(actionButtons: true),
         TableSectionInfo(size: .large, carousel: true),
         TableSectionInfo(size: .small, carousel: true),
+        TableSectionInfo(size: .large, carousel: true, isGroupStyle: true),
+        TableSectionInfo(size: .small, carousel: true, isGroupStyle: true),
         TableSectionInfo(size: .large, carousel: false),
         TableSectionInfo(size: .small, carousel: false)
     ]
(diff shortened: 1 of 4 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
```diff
@@ -268,4 +279,16 @@ class PersonaButtonCarouselDemoController: DemoTableViewController {
         PersonaData(email: "carlos.slattery@contoso.com", subtitle: "Software Engineer")
     ]
 
+    private let groupPersonas: [PersonaData] = [
```
edjamesmsft, line 282: Can you not re-use the other persona data? I'm not sure I love the structure of this test data.

ios/FluentUI/PersonaButton/PersonaButton.swift
```diff
@@ -61,8 +61,8 @@ public struct PersonaButton: View, TokenizedControlView {
     /// Creates a new `PersonaButton` instance.
     /// - Parameters:
     ///   - size: The` MSFPersonaButtonSize` value used by the `PersonaButton`.
-    public init(size: MSFPersonaButtonSize) {
-        let state = MSFPersonaButtonStateImpl(size: size)
+    public init(size: MSFPersonaButtonSize, isGroupStyle: Bool = false) {
```
edjamesmsft, line 64: Should this be a new convenience init instead?

edjamesmsft: (We shouldn't have breaking APIs if we can avoid it)

ios/FluentUI.Demo/FluentUI.Demo/Demos/PersonaButtonCarouselDemoController.swift
laminesm: It's possible, but I created this data to resemble the specs design shared with me. I've shared the files with you on teams.

ios/FluentUI/PersonaButton/PersonaButton.swift
laminesm: Apparently you can't have convenience initializers in structs. 
Question: would this still be considered breaking changes even if the new parameter has a default value passed? It changes the API but it wouldn't break any of the existing code.

edjamesmsft: Try adding it to the objc demo and see if it breaks?

laminesm: PersonaButton doesn't seem to have objc exposure. Is there another way to add it to the objc demo?

PR #1458, opened 2022-12-16
[XCUITests] Initial tests for Activity Indicator

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 54,943,520 bytes | 54,965,576 bytes | 22,056 bytes |

Added initial tests for activity indicator. These include tests for size, color, and starting/stopping activity (both when hidden when stopped is turned on and off).

### Verification

| Simulator Screen Recording - Clone 2 of iPhone 14 Pro - 2022-12-16 at 11 36 21 | Simulator Screen Recording - Clone 2 of iPhone 14 Pro - 2022-12-16 at 11 39 11 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1458)

### DIFF
--- a/ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
+++ b/ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
@@ -13,11 +13,16 @@ class ActivityIndicatorTest: XCTestCase {
         continueAfterFailure = false
         app.launch()
 
-        if app.staticTexts["FluentUI DEV"].exists {
-            app.staticTexts["ActivityIndicator"].tap()
-        } else if !app.staticTexts["ActivityIndicator"].exists {
-            app.buttons["FluentUI DEV"].tap()
-            app.staticTexts["ActivityIndicator"].tap()
+        let onHomePage = app.staticTexts["FluentUI DEV"].exists
+        let onDifferentControlPage = !onHomePage && !app.staticTexts["ActivityIndicator"].exists
+        let activityIndicatorCell = app.staticTexts["ActivityIndicator"]
+        let backButton = app.buttons["FluentUI DEV"]
+
+        if onHomePage {
+            activityIndicatorCell.tap()
+        } else if onDifferentControlPage {
+            backButton.tap()
+            activityIndicatorCell.tap()
         }
     }
 
@@ -38,14 +43,16 @@ class ActivityIndicatorTest: XCTestCase {
     }
 
     func testStartStopHide() throws {
+        let startStopButton = app.buttons["Start / Stop activity"]
+
         XCTAssert(app.images["In progress"].exists)
-        app.buttons["Start / Stop activity"].tap()
+        startStopButton.tap()
         XCTAssert(!app.images["In progress"].exists)
 
         app.cells.containing(.staticText, identifier: "Hides when stopped").firstMatch.tap()
         XCTAssert(app.images["Progress halted"].exists)
 
-        app.buttons["Start / Stop activity"].tap()
+        startStopButton.tap()
         XCTAssert(app.images["In progress"].exists)
     }
 }
--- a/ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
+++ b/ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
@@ -10,42 +10,42 @@ class ActivityIndicatorTest: XCTestCase {
     let app = XCUIApplication()
 
     override func setUpWithError() throws {
+        try super.setUpWithError()
         continueAfterFailure = false
         app.launch()
-        
+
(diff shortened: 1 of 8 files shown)


ios/FluentUI/ActivityIndicator/ActivityIndicatorTokenSet.swift
```diff
@@ -15,6 +15,23 @@ import UIKit
...
+         case .medium:
+             return "medium"
+         case .large:
+             return "large"
+         case .xLarge:
+             return "xLarge"
+         }
+     }
+ }
+
```
huwilkes, line 34: I don't think this extension should be showing up in release builds. If possible, we should find a way to add the functionality you need with 0 increase to the lib.

laminesm (commented):
We use lint to enforce some stylistic conventions. I got it with homebrew (https://formulae.brew.sh/formula/swiftlint) and you can run it in the terminal before pushing PRs and it'll fix the warnings for you :)

ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
```diff
@@ -0,0 +1,52 @@
...
+        if app.staticTexts["FluentUI DEV"].exists {
+            app.staticTexts["ActivityIndicator"].tap()
+        } else if !app.staticTexts["ActivityIndicator"].exists {
+            app.buttons["FluentUI DEV"].tap()
+            app.staticTexts["ActivityIndicator"].tap()
+        }
+    }
+
+    func testSizes() throws {
+        XCTAssertEqual(app.images.element(boundBy: 0).identifier, "Activity Indicator that is In progress of size 4")
```
mischreiber, line 26: Interesting note: these tests fail if your simulator is set to any language other than `en-US`. Are we okay with that?

joannaquu: Oh I didn't even think to check that. Where do we change the language? I tested changing it in Edit scheme -> Run -> Options -> App language and that worked fine.

lyzhan7: I think any tests we write should eventually work for any language we configure the tests to run on. Accessibility label/accessibility names both get localized, can we build the accessibility identifier off of something else? 

Looks like the accessibility label is created from the isAnimating property on the activity indicator, maybe we can use that directly. Color seems a little more tricky - I wonder if we can use the RGB value

ios/FluentUI.Demo/FluentUIDemoTests/ActivityIndicatorTest.swift
```diff
@@ -0,0 +1,51 @@
...
+
+    override func setUpWithError() throws {
+        try super.setUpWithError()
+        continueAfterFailure = false
+        app.launch()
+
+        if app.staticTexts["FluentUI DEV"].exists {
+            app.staticTexts["ActivityIndicator"].tap()
+        } else if !app.staticTexts["ActivityIndicator"].exists {
+            app.buttons["FluentUI DEV"].tap()
```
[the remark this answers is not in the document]
joannaquu: There's no label for the back button so it defaulted to FluentUI Dev. I wasn't sure if I should go in and change it so I left if for now.

lyzhan7:
Some overall notes:
- What's getting tested
  - It looks like this PR is pretty much testing that every single variation of the activity indicator exists - it's great how thorough this is, but not sure how feasible it is to do this for every component. There are a couple components that have a ton of variations (ex. Avatar), that might take up a lot of time to do this for- it might be more impactful to first establish a simpler baseline of tests across all the components first (ex. just opening every test page for every component, and then maybe checking that one component on every page exists), rather than potentially getting hung up on the components that have a ton of variants. This is also the approach that the FURN tests have taken.
- In your PR description I think you could add a bit more context on what these changes do:
  - You mentioned adding "initial tests" for activity indicator - can you add more detail to what the tests are testing and what the reasoning is behind them? 
  - And related to the above - what is in scope for this PR and what is out of scope? Ex. something out of scope - I would expect the UI tests to eventually be testing more behavior - ex. does toggling the "animating" button actually stop/start the animation?
  - Optional - since XCUITests are new to this codebase, what are some of the resources you referenced?
- We talked about this offline - that you had tried to add the accessibility identifier to the components in the demo controller rather than in the component itself, but it didn't work. I wonder if that's contributing to the increase in binary size? Maybe we can take another look at why setting the identifier in the demo controller component isn't working

And also great work on getting the first XCUITest up and running in FUA! Setting up anything new can be really daunting so great work diving in this :)

joannaquu:
Closing to focus on launch tests first.

PR #1468, opened 2022-12-27
Initial standalone implementation of colors2.swift functionality

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This is a draft pull request to review the new colors2.swift implementation, which will stand in side-by-side with the current colors.swift functionality until we have everything replaced in devmain and all dependencies, and then we will remove colors.swift and do a large rename. This will allow us to maintain a building state while doing a massive find and replace effort.

### Verification

Added ColorProviding2 implementation to the objective C demo controller.

| objCBeforeLight | objcAfterLight |
| objCBeforeDark | objcAfterDark |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1468)

### DIFF
--- a/ios/FluentUI/Core/Colors2.swift
+++ b/ios/FluentUI/Core/Colors2.swift
@@ -219,7 +219,7 @@ public final class Colors2: NSObject {
 
         static var brandForeground1: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1])
 
-        static var brandForeground1Pressed: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1Selected])
+        static var brandForeground1Pressed: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1Pressed])
 
         static var brandForeground1Selected: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1Selected])
 
--- a/ios/FluentUI/Core/Colors2.swift
+++ b/ios/FluentUI/Core/Colors2.swift
@@ -58,6 +58,9 @@ private func brandColorOverrides(provider: ColorProviding2, for theme: FluentThe
     if let brandBackgroundTint = provider.brandBackgroundTint(for: theme)?.dynamicColor {
         brandColors[.brandBackgroundTint] = brandBackgroundTint
     }
+    if let brandBackgroundDisabled = provider.brandBackgroundDisabled(for: theme)?.dynamicColor {
+        brandColors[.brandBackgroundDisabled] = brandBackgroundDisabled
+    }
     if let brandForeground1 = provider.brandForeground1(for: theme)?.dynamicColor {
         brandColors[.brandForeground1] = brandForeground1
     }
@@ -152,6 +155,10 @@ public final class Colors2: NSObject {
         return colorProvidersMap.object(forKey: theme)?.brandBackgroundTint(for: theme) ?? FallbackThemeColor.brandBackgroundTint
     }
 
+    @objc public static func brandBackgroundDisabled(for theme: FluentTheme) -> UIColor {
+        return colorProvidersMap.object(forKey: theme)?.brandBackgroundDisabled(for: theme) ?? FallbackThemeColor.brandBackgroundDisabled
+    }
+
     @objc public static func brandForeground1(for theme: FluentTheme) -> UIColor {
         return colorProvidersMap.object(forKey: theme)?.brandForeground1(for: theme) ?? FallbackThemeColor.brandForeground1
     }
@@ -207,6 +214,8 @@ public final class Colors2: NSObject {
         static var brandBackground3: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackground3])
 
         static var brandBackgroundTint: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackgroundTint])
+
+        static var brandBackgroundDisabled: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackgroundDisabled])
 
         static var brandForeground1: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1])
 
--- a/ios/FluentUI/Core/Colors2.swift
+++ b/ios/FluentUI/Core/Colors2.swift
@@ -115,126 +115,20 @@ public final class Colors2: NSObject {
     /// - Parameters:
     ///   - window: The window that should have its `ColorProvider2` removed.
     @objc public static func removeProvider(for theme: FluentTheme) {
+        let removedProvider = colorProvidersMap.object(forKey: theme)
         colorProvidersMap.removeObject(forKey: theme)
-        // TODO: what is the equivalent function here?
-        // window.fluentTheme = FluentThemeKey.defaultValue
-    }
 
-    // MARK: Primary
-
-    /// Use these funcs to grab a color customized by a ColorProvidingobject for a specific window. If no colorProvider2 exists for the window, falls back to deprecated singleton theme color
-    @objc public static func brandBackground1(for theme: FluentTheme) -> UIColor {
-        return colorProvidersMap.object(forKey: theme)?.brandBackground1(for: theme) ?? FallbackThemeColor.brandBackground1
-    }
-
-    @objc public static func brandBackground1Pressed(for theme: FluentTheme) -> UIColor {
-        return colorProvidersMap.object(forKey: theme)?.brandBackground1Pressed(for: theme) ?? FallbackThemeColor.brandBackground1Pressed
-    }
-
-    @objc public static func brandBackground1Selected(for theme: FluentTheme) -> UIColor {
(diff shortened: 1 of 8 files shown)


ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,234 @@
...
+    if let brandBackground2Selected = provider.brandBackground2Selected(for: theme)?.dynamicColor {
+        brandColors[.brandBackground2Selected] = brandBackground2Selected
+    }
+    if let brandBackground3 = provider.brandBackground3(for: theme)?.dynamicColor {
+        brandColors[.brandBackground3] = brandBackground3
+    }
+    if let brandBackgroundTint = provider.brandBackgroundTint(for: theme)?.dynamicColor {
+        brandColors[.brandBackgroundTint] = brandBackgroundTint
+    }
+    if let brandForeground1 = provider.brandForeground1(for: theme)?.dynamicColor {
```
huwilkes, line 61: `brandBackgroundDisabled`?

edjamesmsft: Good catch!

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,234 @@
...
+
+        static var brandBackground2Selected: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackground2Selected])
+
+        static var brandBackground3: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackground3])
+
+        static var brandBackgroundTint: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandBackgroundTint])
+
+        static var brandForeground1: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1])
+
+        static var brandForeground1Pressed: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandForeground1Selected])
```
nodes11, line 222: Variable name is `brandForeground1Pressed` but `. brandForeground1Selected ` is used to access value in `colors`.

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,234 @@
...
+
+        static var brandForeground1Selected: UIColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.brandStroke1Selected])
```
nodes11, line 224: Variable name is `brandForeground1Selected` but `.brandStroke1Selected`  is used to access value in `colors`.

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,234 @@
...
+        colorProvidersMap.setObject(provider, forKey: theme)
+
+        // Create an updated fluent theme as well
+        let brandColors = brandColorOverrides(provider: provider, for: theme)
+        brandColors.forEach { token, value in
+            theme.aliasTokens.colors[token] = value
+        }
+    }
+
+    /// Removes any associated `ColorProvider` from the given `UIWindow` instance.
```
nodes11, line 113: super nit: Maybe update to `ColorProvider2` for consistency

laminesm (commented):
Spoke offline, we should probably test ColorProviding2 objc override works.

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,243 @@
...
+        colorProvidersMap.removeObject(forKey: theme)
+        // TODO: what is the equivalent function here?
+        // window.fluentTheme = FluentThemeKey.defaultValue
+    }
+
+    // MARK: Primary
+
+    /// Use these funcs to grab a color customized by a ColorProvidingobject for a specific window. If no colorProvider2 exists for the window, falls back to deprecated singleton theme color
+    @objc public static func brandBackground1(for theme: FluentTheme) -> UIColor {
+        return colorProvidersMap.object(forKey: theme)?.brandBackground1(for: theme) ?? FallbackThemeColor.brandBackground1
```
mischreiber, line 127: Proposal:
* Remove `colorProvidersMap` variable entirely (this is a holdover from the version of `Colors.swift` that predates `FluentTheme`).
* Remove all of these static methods, along with `FallbackThemeColor`, entirely. Now that `AliasTokens` are accessible via Objective-C, clients can use that approach to fetch values.

Before:
```objc
MSFFluentTheme *fluentTheme = [[self view] fluentTheme];
UIColor *brandBackgroundColor = [MSFColors2 brandBackground1For:fluentTheme];
```

After:
```objc
MSFFluentTheme *fluentTheme = [[self view] fluentTheme];
UIColor *brandBackgroundColor = [[fluentTheme aliasTokens] colorsForToken:MSFColorsAliasTokensBackground1];

File: ios/FluentUI/Core/Colors2.swift [hunk 6]
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    /// Removes any associated `ColorProvider2` from the given `UIWindow` instance.
+    ///
+    /// - Parameters:
+    ///   - window: The window that should have its `ColorProvider2` removed.
+    @objc public static func removeProvider(for theme: FluentTheme) {
+        colorProvidersMap.removeObject(forKey: theme)
+        // TODO: what is the equivalent function here?
+        // window.fluentTheme = FluentThemeKey.defaultValue
```
mischreiber, line 120: Probably something like this:
```swift
brandColors.forEach { token, _ in
    theme.aliasTokens.colors[token] = nil
}
```

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,243 @@
...
+    }
+
+    @objc public static func brandStroke1Selected(for theme: FluentTheme) -> UIColor {
+        return colorProvidersMap.object(forKey: theme)?.brandStroke1Selected(for: theme) ?? FallbackThemeColor.brandStroke1Selected
+    }
+
+    private static var colorProvidersMap = NSMapTable<FluentTheme, ColorProviding2>(keyOptions: .weakMemory, valueOptions: .weakMemory)
+
+    /// A namespace for holding fallback theme colors (empty enum is an uninhabited type)
+    private enum FallbackThemeColor {
```
laminesm, line 201: Curious why is FallbackThemeColor an enum?

ios/FluentUI/Core/Colors2.swift
edjamesmsft: Are you saying we don't need the getters any more, just the setters? I agree with you, if that's the case, and is actually how I'm doing it in the demo controller test code.

ios/FluentUI/Core/Colors2.swift
edjamesmsft: Please see my updated attempt at resetting values - this does not yet work, but is conceptually close to what I'm intending to do (iterate through and reset the brand tokens to their original values)

ios/FluentUI/Core/Colors2.swift
edjamesmsft: Done!

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,181 @@
...
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+// MARK: ColorProviding2 - temporary stand-in for ColorProviding so we can replace side by side
+
+/// Protocol through which consumers can provide colors to "theme" their experiences
+/// The view associated with the passed in theme will display the set colors to alllow apps to provide different experiences per each view
```
laminesm, line 11: typo on allow

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+}
+
+public extension FluentTheme {
+    /// Associates a `ColorProvider2` with a given `FluentTheme` instance.
+    ///
+    /// - Parameters:
+    ///   - provider: The `ColorProvider2` whose colors should be used for controls in this theme.
+    ///   - theme: The theme where these colors should be applied.
+    @objc static func setProvider(provider: ColorProviding2, for theme: FluentTheme) {
```
laminesm, line 164: Since this is an extension to `FluentTheme`, could we remove the theme parameter and call self inside the function instead?

laminesm (approved):

ios/FluentUI/Core/Colors2.swift
edjamesmsft: Good call - that makes this simpler to call, as well.

ios/FluentUI/Core/Colors2.swift
```diff
@@ -0,0 +1,181 @@
...
+    }
+    if let brandStroke1Selected = provider.brandStroke1Selected(for: theme)?.dynamicColor {
+        brandColors[.brandStroke1Selected] = brandStroke1Selected
+    }
+    return brandColors
+}
+
+// MARK: Colors
+
+public enum BrandColorsForOverriding: CaseIterable {
```
mischreiber, line 96: 1. Does this need to be public?
2. If so, does it need an `@objc` name?

edjamesmsft: Good catch, I don't think it needs to be public!

PR #1485, opened 2023-01-10
Resolving label height clipping in larger text mode for TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 40,446,736 bytes | 40,447,720 bytes | +984 bytes |
| TableViewCell.o | 1,433,712 bytes | 1,434,696 bytes | +984 bytes |

When an `NSAttributedString` (or even a normal `String`) has no specified font, the system font is used as default. There was a specific scenario where an `NSAttributedString` without a `.font` attribute would cause the cell's label to be clipped in larger text mode. It seems that the boundingRect does not redraw when the system font adjusts to the larger text. This change will take the max height of our `.preferredSize()` (used for strings) and `.preferredLabelSize` (used for attributed strings) methods to accommodate for this scenario.

Also removed some outdated comments in `.preferredLabelSize()`.

### Verification

| image | image |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1485)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -581,25 +581,26 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
                                                                               text: text, labelAccessoryViewMarginLeading: labelAccessoryViewMarginLeading)
 
         let availableWidth = textAreaWidth - (leadingAccessoryAreaWidth + trailingAccessoryAreaWidth + labelAccessoryViewMarginTrailing)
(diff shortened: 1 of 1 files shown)


huwilkes (approved):

harrieshin:
should be bring this changes to .10 branch?

PR #1490, opened 2023-01-12
Tokenize BottomSheet

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
This change tokenizes the BottomSheet control

- ios/FluentUI.Demo/FluentUI.Demo/Demos.swift
  - Moves BottomSheet to fluent 2 section
- ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomSheetDemoController.swift
  - Changes demo controller parent class from UIViewController to DemoController so that it can demo theme/token overrides
  - Adds BottomSheetTokenSet.swift file
- ios/FluentUI/Bottom Sheet/BottomSheetController.swift
  - Makes BottomSheet subclass TokenizedControlInternal
  - BottomSheet now uses tokens!
  - Refactor to organize access levels
- ios/FluentUI/Bottom Sheet/BottomSheetTokenSet.swift
  - Add shadow, backgroundColor, and cornerRadius tokens

Binary change:

| libFluentUI.a | 32,508,016 bytes (32.5 MB on disk) | 32,635,144 bytes (32.6 MB on disk) | 127,128 bytes (0.1 MB on disk) |
| BottomSheetController.swift | 466,680 bytes (467 KB on disk) | 514,736 bytes (516 KB on disk) | 48,056 bytes (49 KB on disk) |
| BottomSheetTokenSet.swift | N/A | 49,296 bytes (53 KB on disk) | 49,296 bytes (53 KB on disk) |

| IMG_0058 | IMG_0074 |
| N/A | IMG_0076 |
| N/A | IMG_0060 |

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1490)


sophialee0416:
Just pushed out #1419, once that's merged in you can probably bring those changes into your branch to fix the failing test!

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -379,8 +391,101 @@ public class BottomSheetController: UIViewController {
...
 
         NSLayoutConstraint.activate(constraints)
+
+        updateAppearance()
+    }
+
+    public override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+
+        tokenSet.update(fluentTheme)
```
laminesm, line 401: (Trying to get a hang of the tokenization process) - why do we do this update in viewDidAppear?

laminesm (commented):
I don't think the shadows are showing on the after screenshots. Having done shadows on fluent2-colors, I think you might need to update shadows in `viewDidLayoutSubviews` for them to render properly.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
jeaniehuynh: This is how we update the token set if the theme changed and it's usually done in didMoveToWindow, but since BottomSheet is a UIViewController (rather than a UIView), this is the closest equivalent method. Ideally, the theme could be changed immediately (in the case the view is already showing), but I don't think that's possible currently (there's this TODO commented in DemoAppearanceView.swift):
image

laminesm: I seee. Thanks for the explanation!

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -379,8 +391,101 @@ public class BottomSheetController: UIViewController {
...
+
+    private func updateBackgroundColor() {
+        let backgroundColor = UIColor(dynamicColor: tokenSet[.backgroundColor].dynamicColor)
+        bottomSheetView.subviews[0].backgroundColor = backgroundColor
+        overflowView.backgroundColor = backgroundColor
     }
 
+    private func updateShadows() {
+        let shadowInfo = tokenSet[.shadow].shadowInfo
+        if let ambientShadow = bottomSheetView.layer.sublayers?[1] {
```
sophialee0416, line 461: If we do have sublayers (not nil), is it guaranteed that there will always be 2 items in the layer?

ios/FluentUI/Bottom Sheet/BottomSheetTokenSet.swift
```diff
@@ -0,0 +1,29 @@
...
+        case backgroundColor
+        case cornerRadius
+        case shadow
+    }
+
+    init() {
+        super.init { token, theme in
+            switch token {
+            case .backgroundColor:
+                return .dynamicColor { DynamicColor(light: GlobalTokens.neutralColors(.white),
```
sophialee0416, line 19: Would returning `theme.aliasTokens.backgroundColors[.neutral1]` do the trick instead?

jeaniehuynh: It would! But I just realized I had the wrong color token (supposed to be gray900, not black)

laminesm (approved):

PR #1498, opened 2023-01-19
Updating popover background color to background4

edjamesmsft:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,834,200 | 28,833,888 | -312 |

Updating popover background color to be correct grey (background4, not background2).

### Verification

Built and ran simulator, verified color with colour contrast analyser.

| beforeLight | afterLight |
| beforeDark | afterDark |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -104,8 +104,8 @@ open class DrawerController: UIViewController {
     }
 
     @objc public static func popoverBackground(fluentTheme: FluentTheme) -> UIColor {
-        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background2].light,
-                                                  dark: fluentTheme.aliasTokens.colors[.background2].darkElevated))
+        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background4].light,
+                                                  dark: fluentTheme.aliasTokens.colors[.background4].dark))
     }
 
     private struct Constants {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -104,8 +104,8 @@ open class DrawerController: UIViewController {
     }
 
     @objc public static func popoverBackground(fluentTheme: FluentTheme) -> UIColor {
-        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background2].light,
-                                                  dark: fluentTheme.aliasTokens.colors[.background2].darkElevated))
+        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background4].light,
```
lcapkovic, line 107: Why do we recreate the DynamicColor? Doesn't the alias token dynamic color already include the light / dark variants?

edjamesmsft: Because the DynamicColor has DarkElevated as well, which can accidentally be hit and show the wrong background color when the popover/drawer is in Elevated mode (i.e., on ipad side-by-side)

mischreiber: But isn't that the intended behavior, actually? Or should we only use darkElevated values if the spec explicitly calls for it?

lcapkovic: +1 on Mike's question. It would suck if for each alias token we had to look inside and check if we need to drop the elevated color per usage.

lcapkovic: Reading Apple's documentation, the elevated variant seems to be precisely for things like popovers and sheets. If whatever surface presents the popover is BG4, and the popover is also BG4, in dark mode the elevated variant will ensure the popover doesn't visually clash with the presenting surface. Isn't that what we want?

laminesm: It's mostly the intended behavior with a few exceptions. For instance, for DateTimePicker design requested specifically we don't use darkElevated colors. And here, the figma of the popover determines the dark variant to be used for dark mode.

mischreiber: Can we verify with design whether this is the expected fallback? It's possible that the component design predates the existence of dark elevated, which is relatively newer for the toolkit.

laminesm: Confirmed with design this is the expected behavior for the popover's background color.

PR #1501, opened 2023-01-20
[Fluent2-Colors] Switch ColorProviding2 to use FluentThemeable

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,918,928 | 28,883,920 | -35,008 |
|  |  |  |  |

Moving the `ColorProviding2` API to align more closely with the existing `ColorProviding` API. In particular, the primary change is to move from changing `AliasTokens` _on_ the current `FluentTheme`, back to creating an entirely new `FluentTheme`. Among other things, this restores dynamic component re-rendering when calling the `ColorProviding2` API.

I also took the liberty to clean up the Objective-C API a bit, since it needs to change anyway.

Before (`ColorProviding1`):
```objc
@selector(setProviderWithProvider:for:)
```

After (`ColorProviding2`):
```objc
@selector(setProvider:forThemeable:)
```

### Verification

Works in UIKit, SwiftUI, and via the Objective-C APIs.

https://user-images.githubusercontent.com/4934719/213594220-7fd03b94-14a8-4950-b1a7-2a903dd0e433.mp4

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1501)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
@@ -98,17 +98,18 @@
 }
 
 - (void)overridesButtonPressed:(id)sender {
-    MSFFluentTheme *fluentTheme = [[self view] fluentTheme];
-    [fluentTheme setProviderWithProvider:[ObjectiveCDemoColorProviding2 alloc]];
+    [MSFFluentTheme setProvider:[ObjectiveCDemoColorProviding2 alloc]
+                   forThemeable:[self view]];
 
-    MSFAliasTokens *aliasTokens = [fluentTheme aliasTokens];
+    MSFAliasTokens *aliasTokens = [[[self view] fluentTheme] aliasTokens];
     MSFDynamicColor *primaryColor = [aliasTokens aliasColorForToken:MSFColorAliasTokensBrandForeground1];
 
     [self addLabelWithText:@"Test label with override brand color"
                  textColor:[[UIColor alloc] initWithDynamicColor:primaryColor]];
 
     // Remove the overrides
-    [fluentTheme removeProvider];
+    [MSFFluentTheme removeProviderForThemeable:[self view]];
+    aliasTokens = [[[self view] fluentTheme] aliasTokens];
     primaryColor = [aliasTokens aliasColorForToken:MSFColorAliasTokensBrandForeground1];
 
     [self addLabelWithText:@"Test label with override color removed"
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
@@ -98,8 +98,7 @@
 }
 
 - (void)overridesButtonPressed:(id)sender {
-    [MSFFluentTheme setProvider:[ObjectiveCDemoColorProviding2 alloc]
-                   forThemeable:[self view]];
+    [[self view] setColorProvider:[[ObjectiveCDemoColorProviding2 alloc] init]];
 
     MSFAliasTokens *aliasTokens = [[[self view] fluentTheme] aliasTokens];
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Core/Colors2.swift
```diff
@@ -12,166 +12,110 @@ import UIKit
...
-            return .brandStroke1Pressed
-        case .brandStroke1Selected:
-            return .brandStroke1Selected
-        }
-    }
-}
-
 public extension FluentTheme {
-    /// Associates a `ColorProvider2` with a given `FluentTheme` instance.
+    /// Associates a `ColorProvider2` with a given `FluentThemeable` instance.
```
huwilkes, line 98: Should this still be an extension of `FluentTheme`? Seems a bit odd to pass something with a theme into a theme to get colors from a theme.

edjamesmsft (approved):

ios/FluentUI/Core/Colors2.swift
```diff
@@ -12,166 +12,110 @@ import UIKit
...
 
-    /// Removes any associated `ColorProvider2` from the given `FluentTheme` instance.
-    @objc func removeProvider() {
-        for brandColor in BrandColorsForOverriding.allCases {
-            self.aliasTokens.colors.removeOverride(brandColor.equivalentAliasToken)
-        }
+    /// Removes any associated `ColorProvider` from the given `FluentThemeable` instance.
+    ///
+    /// - Parameters:
+    ///   - window: The `FluentThemeable` that should have its `ColorProvider2` removed.
```
laminesm, line 116: Did window carry on from an old comment?

ios/FluentUI/Core/Colors2.swift
```diff
@@ -12,166 +12,110 @@ import UIKit
...
+    @objc(removeProviderForThemeable:)
+    static func removeProvider(for themeable: FluentThemeable) {
```
laminesm, line 118: Should we consider renaming this function to something like `resetThemeable`? `removeProvider` seems a bit misleading since the body of the function doesn’t directly interact with a provider.

mischreiber: I was wondering that too - I think I'd actually like to update this entirely:
- Make the methods be explicit categories on UIView, instead of static methods.
- Rename the methods to be very clear:
```swift
@objc(setColorProvider:)
func setColorProvider(_ provider: ColorProviding2)

@objc(resetFluentTheme)
func resetFluentTheme()
```

Sound good?

ios/FluentUI/Core/Colors2.swift
mischreiber: It's now an extension on `UIView`, which matches with normal `fluentTheme` setting and also clarifies the calling code.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -98,17 +98,17 @@ - (void)tokensButtonPressed:(id)sender {
 }
 
 - (void)overridesButtonPressed:(id)sender {
-    MSFFluentTheme *fluentTheme = [[self view] fluentTheme];
-    [fluentTheme setProviderWithProvider:[ObjectiveCDemoColorProviding2 alloc]];
+    [[self view] setColorProvider:[[ObjectiveCDemoColorProviding2 alloc] init]];
```
mischreiber, line 101: Note: you should _never_ call `alloc` on an Objective-C object without also calling `init` -- it's the same as calling `malloc` in C without actually initializing the allocated memory. It _might_ be clean, usable memory, or it might be junk that immediately crashes when you access it.

ios/FluentUI/Core/Colors2.swift
laminesm: Sounds great, thanks!

laminesm (approved):

PR #1502, opened 2023-01-20
Adding Avatar activity badge

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,908,720 bytes | 28,977,656 bytes | +68,936 bytes |
| Avatar | 661,536 bytes | 590,176 bytes | -71,360 bytes |
| AvatarModifiers | 36,352 bytes | 41,688 bytes | +5,336 bytes |
| AvatarTokenSet | 120,768 bytes | 136,816 bytes | +16,048 bytes |
| AvatarGroup | 362,344 bytes | 362,520 bytes | +176 bytes |
| ShapeCutout | 36,448 bytes | 39,280 bytes | +2832 bytes |

- **AvatarDemoController:** Added toggle for displaying circle and square activity badges
- **AvatarDemoController_SwiftUI:** Added Picker for MSFAvatarActivityStyles and a toggle to add a custom image
- **Avatar:** Added new activityStyle and activityImage property in the Avatar state, removed unused PresenceCutout
- **AvatarModifiers:** Added new modifier to create an activity badge
- **AvatarTokenSet:** Moved a few existing tokens to constant, added new tokens for activity badge
- **ShapeCutout:** Renamed from CircleCutout to give more shape creation versatility
- **AvatarGroup:** Updated CircleCutout usage to ShapeCutout

Note: Activity is only displayed in `.size40` and `.size56`, and in the `.default` style with image/initials.
### Verification

| LTR                                       | RTL                                      |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |
| image | image |

Presence (no change)
| Before | After |
|-----|------|
| image | image |
| image | image |

TableViewCell (no change)
| Before | After |
|-----|------|
| image | image |

AvatarGroup (no change)
Simulator Screen Recording - iPhone 14 Pro - 2023-01-27 at 15 16 56

### Pull request checklist

This PR has considered:

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1502)


edjamesmsft:
There is a fair amount of binary size increase unaccounted for - do you know where the rest might be coming from?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController_SwiftUI.swift
```diff
@@ -50,6 +52,7 @@ struct AvatarDemoView: View {
                 .imageBasedRingColor(showImageBasedRingColor ? AvatarDemoController.colorfulCustomImage : nil)
                 .isTransparent(isTransparent)
                 .presence(presence)
+                .activity(activityStyle, showActivityImage ? UIImage(named: "excelIcon") : nil)
                 .isOutOfOffice(isOutOfOffice)
                 .hasPointerInteraction(hasPointerInteraction)
                 .isAnimated(isAnimated)
```
huwilkes, line 58: I can't quite get to it here, but can we change the `Color.gray` to like `Colors.surfacePrimary`, or something? Or an alias token.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -264,7 +297,34 @@ public struct Avatar: View, TokenizedControlView {
                                         .frame(width: presenceIconFrameSideRelativeToOuterRing, height: presenceIconFrameSideRelativeToOuterRing,
                                                alignment: .bottomTrailing),
                                      alignment: .topLeading)
-                            .frame(width: overallFrameSide, height: overallFrameSide, alignment: .topLeading)
+                            .frame(width: presenceOverallFrameSide, height: presenceOverallFrameSide, alignment: .topLeading)
+                    })
+                    .modifyIf(shouldDisplayActivity, { thisView in
```
huwilkes, line 302: The logic for presence and activity look pretty similar. Would it be possible to move them into a method, or something?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -264,7 +297,34 @@ public struct Avatar: View, TokenizedControlView {
...
+                        thisView
+                            .clipShape(ShapeCutout(xOrigin: activityCutoutOriginX,
+                                                   yOrigin: activityCutoutOriginY,
+                                                   cornerRadius: cornerRadius,
+                                                   cutoutSize: activityIconOutlineSize),
+                                       style: FillStyle(eoFill: true))
+                            .overlay(RoundedRectangle(cornerRadius: cornerRadius)
+                                .foregroundColor(Color(dynamicColor: tokenSet[.borderColor].dynamicColor).opacity(isTransparent ? 0 : 1))
+                                .frame(width: activityIconOutlineSize, height: activityIconOutlineSize, alignment: .center)
+                                .contentShape(Circle())
```
huwilkes, line 312: Should the contentShape still be a circle?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -196,6 +197,21 @@ class AvatarDemoController: DemoTableViewController {
...
+    private var isShowingActivity: Bool = false {
+        didSet {
+            if oldValue != isShowingActivity {
+                var activityStyle: MSFAvatarActivityStyle
+                var isEven: Bool
+                for index in 0 ..< allDemoAvatarsCombined.count {
+                    isEven = index % 2 == 0
+                    activityStyle = isEven ? .circle : .square
+                    allDemoAvatarsCombined[index].state.activityStyle = isShowingActivity ? activityStyle : .none
+                    allDemoAvatarsCombined[index].state.activityImage = isEven ? nil : UIImage(named: "excelIcon")
```
harrieshin, line 209: excel icon is only for the square style right? can we add another image for the circular one?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -53,6 +53,10 @@ import SwiftUI
     /// Presence is not displayed in the xsmall size.
     var presence: MSFAvatarPresence { get set }
 
+    var activityStyle: MSFAvatarActivityStyle { get set }
```
harrieshin, line 56: comments please

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -139,7 +158,20 @@ public struct Avatar: View, TokenizedControlView {
         let presenceCutoutOriginX: CGFloat = layoutDirection == .rightToLeft ? presenceCutoutOriginXRTL : presenceCutoutOriginXLTR
         let presenceCutoutOriginY = presenceCutoutOriginXLTR
         let presenceIconFrameSideRelativeToOuterRing: CGFloat = presenceIconFrameSideRelativeToInnerRing + outerGapAndRingThicknesCombined
-        let overallFrameSide = max(ringOuterGapSize, presenceIconFrameSideRelativeToOuterRing)
+        let presenceOverallFrameSide = max(ringOuterGapSize, presenceIconFrameSideRelativeToOuterRing)
+
+        let activityCutoutOriginXLTR: CGFloat = AvatarTokenSet.avatarSize(size) - activityIconOutlineSize + activityBorderThickness + ringInnerGap + outerGapAndRingThicknesCombined
```
harrieshin, line 163: should we comment this?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -139,7 +158,20 @@ public struct Avatar: View, TokenizedControlView {
...
+        let activityCutoutOriginXRTL: CGFloat = 0 - activityBorderThickness + ringInnerGap + outerGapAndRingThicknesCombined
+        let activityCutoutOriginX: CGFloat = layoutDirection == .rightToLeft ? activityCutoutOriginXRTL : activityCutoutOriginXLTR
```
harrieshin, line 165: should large portion of these calculation be part of the custom modifier so we don't necesarily have to do this calculation unless we actually need to calculate activity cutout?

ios/FluentUI/Avatar/AvatarTokenSet.swift
```diff
@@ -30,11 +30,14 @@ public class AvatarTokenSet: ControlTokenSet<AvatarTokenSet.Tokens> {
         /// The gap around the ring around the `Avatar`.
         case ringOuterGap
 
-        /// The thickness of the outline around the presence icon.
-        case presenceIconOutlineThickness
+        /// The thickness of the outline around the presence/activity.
+        case borderThickness
```
edjamesmsft, line 34: Is this going to be a breaking change? Is it necesary?

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
harrieshin: grab some official emoji https://github.com/microsoft/fluentui-emoji

ios/FluentUI/Avatar/AvatarTokenSet.swift
```diff
@@ -231,9 +239,49 @@ extension AvatarTokenSet {
...
+        case .size72:
+            return 32
+        }
+    }
+
+    /// The size of the activity icon background.
+    static func activityIconBackgroundSize(_ size: MSFAvatarSize) -> CGFloat {
+        switch size {
+        case .size16, .size20, .size24, .size32:
+            return 0
```
edjamesmsft, line 273: Does this mean we'll still be showing the activity, it's just of size 0?

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -111,6 +119,14 @@ public struct Avatar: View, TokenizedControlView {
                                                                                                            secondaryText: state.secondaryText))
         let shouldUseCalculatedColors = !initialsString.isEmpty && style != .overflow
 
+        let activityImage: Image = {
+            if let image = state.activityImage {
+                return Image(uiImage: image)
+            } else {
+                return Image("ic_fluent_presence_unknown_10_regular", bundle: FluentUIFramework.resourceBundle)
```
huwilkes, line 126: Should we even have a default activity image? Seems like it might be better to just not display anything if they haven't set something.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -100,8 +104,12 @@ public struct Avatar: View, TokenizedControlView {
 
     public var body: some View {
         let style = state.style
+        let size = state.size
+        let activityStyle = state.activityStyle
         let presence = state.presence
-        let shouldDisplayPresence = presence != .none
+        let shouldDisplayActivity = activityStyle != .none
```
huwilkes, line 110: Related to the comment @edjamesmsft made in the tokens, but we should be checking here if there's even anything to display.

ios/FluentUI/Avatar/AvatarTokenSet.swift
sophialee0416: Yes, this is how we handle the presence as well since we have to make sure all the cases are addressed

ios/FluentUI/Avatar/AvatarTokenSet.swift
sophialee0416: The presence outline thickness and the activity outline thickness is intentionally the same. I felt like it was redundant to add another token, so I decided to create a generic borderThickness to address the two

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -127,7 +128,9 @@ class AvatarDemoController: DemoTableViewController {
                 cell.contentView.bottomAnchor.constraint(equalTo: avatarContentView.bottomAnchor)
             ])
 
-            cell.backgroundConfiguration?.backgroundColor = self.isUsingAlternateBackgroundColor ? Colors.tableCellBackgroundSelected : Colors.tableCellBackground
```
sophialee0416, line 130: This line of code would not modify the background configuration, which is why the change here was made.

ios/FluentUI.Demo/FluentUI.Demo/Demos/AvatarDemoController.swift
```diff
@@ -127,7 +128,9 @@ class AvatarDemoController: DemoTableViewController {
...
+            var background = UIBackgroundConfiguration.clear()
```
harrieshin, line 131: nit: actually name it backgroundConfiguration?

ios/FluentUI.Demo/FluentUI.Demo/Resources/Assets.xcassets/TableViewCell/excelIcon.imageset/Contents.json
```diff
@@ -2,14 +2,11 @@
   "images" : [
     {
       "idiom" : "universal",
-      "filename" : "excelIcon.pdf"
+      "filename" : "excelIcon.png"
```
harrieshin, line 5: assuming it is the same size, no impact on the tableviewcell demo?

sophialee0416: no impact! not sure why it was png to begin with but I can include TVC screenshots as well. size remains the same.

harrieshin:
avatargroup animation verification?

ios/FluentUI/Avatar/Avatar.swift
harrieshin: you are inside fluent library  at this point. do you really need to specify the bundle name?

harrieshin: actually.... why can't activityImage can't be just nil? we only need image when we need impage. we shouldn't every show unknown image

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -100,46 +108,63 @@ public struct Avatar: View, TokenizedControlView {
...
-        let presenceIconFrameDiffRelativeToOuterRing: CGFloat = ringOuterGapSize - (presenceIconFrameSideRelativeToInnerRing + outerGapAndRingThicknesCombined)
-        let presenceCutoutOriginXLTR = ringOuterGapSize - presenceIconFrameDiffRelativeToOuterRing - presenceIconOutlineSize
-        let presenceCutoutOriginXRTL = presenceIconFrameDiffRelativeToOuterRing
-        let presenceCutoutOriginX: CGFloat = layoutDirection == .rightToLeft ? presenceCutoutOriginXRTL : presenceCutoutOriginXLTR
-        let presenceCutoutOriginY = presenceCutoutOriginXLTR
-        let presenceIconFrameSideRelativeToOuterRing: CGFloat = presenceIconFrameSideRelativeToInnerRing + outerGapAndRingThicknesCombined
-        let overallFrameSide = max(ringOuterGapSize, presenceIconFrameSideRelativeToOuterRing)
+
+        // Avatar accessory calculation
+        let accessoryImage: Image = {
```
harrieshin, line 144: dumb question, we can't just call Image(uiImage: state.activityImage) in line 308 not remove this accessoryImage variable all together?

edjamesmsft:
I notice on the dark background screenshots, the cutout color isn't the same color as the background of the image. Is that expected? I assume that's just an icon issue and outside the scope of this PR?

image

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -100,46 +108,63 @@ public struct Avatar: View, TokenizedControlView {
...
+
+        let initialsString: String = ((style == .overflow) ? state.primaryText ?? "" : Avatar.initialsText(fromPrimaryText: state.primaryText,
+                                                                                                           secondaryText: state.secondaryText))
+        let shouldUseCalculatedColors = !initialsString.isEmpty && style != .overflow
+        let shouldUseDefaultImage = (state.image == nil && initialsString.isEmpty && style != .overflow)
+
+        let activityStyle = state.activityStyle
         let presence = state.presence
-        let shouldDisplayPresence = presence != .none
+        let shouldDisplayActivity = state.activityImage != nil && activityStyle != .none && (style == .default && !shouldUseDefaultImage)
```
huwilkes, line 120: I think this is looking for the wrong image. I think we should be checking for if `activityImage != nil`, rather than `!shouldUseDefaultImage`

huwilkes, line 120: If the size of the presence or activity is zero, we should probably just say we shouldn't show them, rather than doing all the work to show something with zero size.

edjamesmsft: Seconded - this is more directly getting at the comment I made earlier.

sophialee0416: No this is correct. Design only wants the activity image to be shown in default style when it's not using default image. I'm already checking in the beginning of this line if the activityImage is nil

sophialee0416:
> I notice on the dark background screenshots, the cutout color isn't the same color as the background of the image. Is that expected? I assume that's just an icon issue and outside the scope of this PR?
> 
> image

Not an icon issue, intentional & by design

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: state.activityImage is an optional and we can't pass a nil image in SwiftUI

edjamesmsft:
> > I notice on the dark background screenshots, the cutout color isn't the same color as the background of the image. Is that expected? I assume that's just an icon issue and outside the scope of this PR?
> > image
> 
> Not an icon issue, intentional & by design

Ah I see the background of the icon is customizable. For the purposes of making that clear, I might choose a more different background color, so it looks more intentional. At the moment the colors are close enough that it looks potentially accidental. But not a blocker.

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: We can't make this non-optional unless we want to provide a default image, which was agreed that we shouldn't in above threads https://github.com/microsoft/fluentui-apple/pull/1502#discussion_r1082908905

sophialee0416:
> > > I notice on the dark background screenshots, the cutout color isn't the same color as the background of the image. Is that expected? I assume that's just an icon issue and outside the scope of this PR?
> > > image
> > 
> > 
> > Not an icon issue, intentional & by design
> 
> Ah I see the background of the icon is customizable. For the purposes of making that clear, I might choose a more different background color, so it looks more intentional. At the moment the colors are close enough that it looks potentially accidental. But not a blocker.

This is also default by design. On the demo controller you can choose to use an alternate background if you'd like to see the border.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -246,32 +270,63 @@ public struct Avatar: View, TokenizedControlView {
                                         )
                                 .contentShape(Circle()),
                              alignment: .center)
-                    .modifyIf(shouldDisplayPresence, { thisView in
+            }
+        }
+
+        @ViewBuilder
+        var avatar: some View {
+            if style != .group {
```
huwilkes, line 278: Would it be possible to add a check for if we're showing activity or presence here, and then also move all the `let presenceMath`s down here?

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: Added additional checks!

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: Moved!

huwilkes (approved):

edjamesmsft (approved):

harrieshin (approved):

harrieshin:
make sure the avatar rings are fine too for both avatargroup and just avatar please!

PR #1506, opened 2023-01-23
Add objc background color for NavigationBar

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

There are currently multiple calls to fluent 1's `Colors.navigationBarBackground` in devmain. This PR adds a static `navigationBarBackgroundColor` constant in NavigationBar instead of trying to access BG3 token directly in devmain. This ensures we don't need to go update the alias token in devmain in a scenario where design changes the navigationBarBackground token. 
Some of the objc files calling this color do not have a view to access the `FluentTheme` from. Instead of creating a new `FluentTheme` in those scenarios, we use `FluentTheme.shared` to initialized the token internally in fluent.

Binary change:

| libFluentUI.a | 29,073,680 bytes | 29,078,440 bytes | 4,760 bytes |

### Verification

The demo app built and ran without issues.

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -93,6 +93,8 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+    @objc public static let navigationBarBackgroundColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.background3])
+
     /// Describes the sizing behavior of navigation bar elements (title, avatar, bar height)
     @objc(MSFNavigationBarElementSize)
     public enum ElementSize: Int {


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -93,6 +93,8 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+    @objc public static let navigationBarBackgroundColor = UIColor(dynamicColor: FluentTheme.shared.aliasTokens.colors[.background3])
```
harrieshin, line 96: in mainbranch, we have this defined in FluentUIFramework.swift. is it purposefully moved to this location?

laminesm: Yes, it seemed to me that it would be more coherent for the nav bar background color to live inside the NavigationBar. Would you rather have it moved to FluentUIFramework?

harrieshin: just making sure you are making intentional changes. it is also useful to put previous commit hash number to track the history of changes. the only little odd thing now is that navigationbar colors are technically defined in 2 different places. 
        func backgroundColor(fluentTheme: FluentTheme) -> UIColor {
            switch self {
            case .normal:
                return UIColor(dynamicColor: fluentTheme.aliasTokens.colors[.background3])

mischreiber, line 96: We should not have a public static method that references the shared theme. Clients are expected to use their own theme information when accessing colors.

mischreiber (changes requested):

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: Just to understand better, would you mind clarifying why? Also, would that be any different from creating a theme just to access this color in client code?

laminesm: It's rewritten to return `Style.system.backgroundColor(fluentTheme: fluentTheme)`

edjamesmsft: Should you update the pr description?

harrieshin (approved):

ios/FluentUI/Navigation/NavigationBar.swift
harrieshin: should fluentframworke.swift refere to this then?

edjamesmsft: FluentUIFramework.swift has already had navigationbarbackgroundcolor fully removed. Do you want to fully undo that change and replace all references with the new constant added here?

mischreiber (approved):

ios/FluentUI/Navigation/NavigationBar.swift
mischreiber: Discussed offline, lemme know if you have any more questions!

PR #1527, opened 2023-01-30
Add support for two-line navigation bar titles

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:
| libFluentUI.a | 28,767,328 | 28,834,032 | +66,704 |
| BadgeLabelButton.o | 832,288 | 834,368 | +2,080 | 
| LargeTitleView.o | 155,328 | 171,928 | +16,600 |
| NavigationBar.o | 528,424 | 539,128 | +10,704 |
| TwoLineTitleView.o | 214,320 | 235,328 | +21,008 |
| UINavigationItem+Navigation.o | 51,616 | 56,480 | +4,864 |
| __.SYMDEF | 4,285,072 | 4,296,520 | +11,448 |

This adds support for two-line navigation bar title views.

First, we augmented `UINavigationItem` to add an optional `subtitle` field.

Second, we enhanced `TwoLineTitleView` in the following ways:
* Add the ability to set alignment to either center or leading. (We don't have any use for trailing at the moment).
* Add the ability to set the entire area to be tappable. For this, we expand the "title" button to cover the entire field so accessibility will only detect a single button rather than two.
* Add the ability to configure whether or not the text elements will animate when we tap them. This is mainly for consistency with how our centered title elements work.

### Verification

**New stuff**
* Added subtitle examples to the test page.
* Validated that colors show up properly for all styles, both in light mode and dark mode.
* Validated that a "leading" `TwoLineTitleView` aligns along the right-hand side in RTL mode.

| Screenshot 1 | Screenshot 2 | Screenshot 3 | Screenshot 4 |
|---|---|---|---|
|  |  |  |  |

**Old behaviors maintained**
* Validated that the appearance of one-line large-title headers hasn't changed.
* Validated that the large content viewer still works in two-line navigation bar title cases.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1527)


joannaquu:
For linting issues, you can download swiftlint from [homebrew](https://formulae.brew.sh/formula/swiftlint) and run it before PRs.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -56,12 +56,24 @@ open class TwoLineTitleView: UIView {
         case light
         case dark
     }
+
```
sophialee0416: See https://github.com/microsoft/fluentui-apple/actions/runs/4047627286/jobs/6961876332#step:5:298 (two whitespace errors here)

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -53,7 +53,12 @@ open class NavigationBarTopSearchBarAttributes: NavigationBarTopAccessoryViewAtt
 /// Contains the MSNavigationTitleView class and handles passing animatable progress through
 /// Custom UI can be hidden if desired
 @objc(MSFNavigationBar)
-open class NavigationBar: UINavigationBar {
+open class NavigationBar: UINavigationBar, TwoLineTitleViewDelegate {
+    public func twoLineTitleViewDidTapOnTitle(_ twoLineTitleView: TwoLineTitleView) {
+        // TODO: This won't be needed once we move it to LargeTitleView
+        NotificationCenter.default.post(name: .accessoryExpansionRequested, object: self)
```
Saadnajmi, line 59: Do we need to fix this before check-in or is it a post check in fix?

ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -158,6 +171,8 @@ class LargeTitleView: UIView {
     private func initBase() {
         setupLayout()
         setupAccessibility()
+
```
joannaquu, line 174: Nit: whitespace

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -57,11 +57,23 @@ open class TwoLineTitleView: UIView {
...
     @objc(MSFTwoLineTitleViewInteractivePart)
     public enum InteractivePart: Int {
-        case none
-        case title
-        case subtitle
+        case none = 0
+        case title = 1
+        case subtitle = 2
+        case all = 3
```
Saadnajmi, line 71: Any reason for the explicit enum integers? Is this meant to be a bit mask?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -57,11 +57,23 @@ open class TwoLineTitleView: UIView {
...
+        case center
+        case leading
+    }
+
     @objc(MSFTwoLineTitleViewInteractivePart)
     public enum InteractivePart: Int {
-        case none
-        case title
-        case subtitle
+        case none = 0
```
joannaquu, line 68: Nit: I believe you can remove the ints (= 0, = 1). By default, objc enums will use ints.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -719,6 +725,19 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+    private func updateSubtitleView(for navigationItem: UINavigationItem?) {
+        guard let navigationItem = navigationItem, let subtitle = navigationItem.subtitle else {
+            // Use the default title view
+            navigationItem?.titleView = nil
```
laminesm, line 731: How does setting the titleView to nil here equate to using the default title view?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -719,6 +725,19 @@ open class NavigationBar: UINavigationBar {
...
+            return
+        }
+
+        let customTitleView = TwoLineTitleView(style: style == .primary ? .light : .dark)
+        customTitleView.setup(title: navigationItem.title ?? "", subtitle: subtitle, interactivePart: .title)
```
laminesm, line 736: Shouldn't we exit this function or have an assertionFailure in guard if we don't have a title in a TwoLineTitleView? Is it ever acceptable to only display a subtitle in a TwoLineTitleView?

ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -354,6 +386,18 @@ class LargeTitleView: UIView {
...
         })
     }
+
+    // MARK: - TwoLineTitleViewDelegate
+
+    func twoLineTitleViewDidTapOnTitle(_ twoLineTitleView: TwoLineTitleView) {
+        guard twoLineTitleView == self.twoLineTitleView else {
+            return
+        }
+        guard respondsToTaps else {
```
laminesm, line 396: can the two guard statements be merged? Something like:
```
 guard twoLineTitleView == self.twoLineTitleView, respondsToTaps else {
          return;
}
```

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
laminesm, line 71: Is all for both title & subtitle? If so, is that any different from subtitle since we typically have a subtitle with a title?

sophialee0416:
Please correct me if I'm wrong, but I think the spacing between the title/subtitle may not be aligned with design. Will that be addressed separately?

ios/FluentUI/Navigation/NavigationBar.swift
edjamesmsft: Seconded - if there's a way to make the behavior explicit (setting a defaultView bool, assigning a view that is constructed as a default view, etc), please do so. Side effects are an anti-pattern.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -719,6 +725,19 @@ open class NavigationBar: UINavigationBar {
         }
     }
 
+    private func updateSubtitleView(for navigationItem: UINavigationItem?) {
+        guard let navigationItem = navigationItem, let subtitle = navigationItem.subtitle else {
+            // Use the default title view
```
edjamesmsft, line 730: Should there be an enforcement here that subtitles are only added when there is also a title present?

ios/FluentUI/Navigation/Views/LargeTitleView.swift
```diff
@@ -196,8 +211,10 @@ class LargeTitleView: UIView {
 
         avatarView.centerYAnchor.constraint(equalTo: contentStackView.centerYAnchor).isActive = true
 
-        // title button setup
-        contentStackView.addArrangedSubview(titleButton)
+        // title container setup
```
edjamesmsft, line 214: Super tiny nit - I don't think this comment is useful, consider removing.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -202,25 +216,45 @@ open class TwoLineTitleView: UIView {
     // MARK: Setup
 
     /// Sets the relevant strings and button styles for the title and subtitle.
+    /// This is a legacy API that maintains Objective C compatibility with older versions of fluentui-apple.
```
edjamesmsft, line 219: Should this be marked as deprecated?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -202,25 +216,45 @@ open class TwoLineTitleView: UIView {
...
+    }
+
+    /// Sets the relevant strings and button styles for the title and subtitle.
+    ///
+    /// - Parameters:
+    ///   - title: A title string.
+    ///   - subtitle: An optional subtitle string. If nil, title will take up entire frame.
+    ///   - alignment: How to align the title and subtitle. Ignored if `subtitle` is nil.
+    ///   - interactivePart: Determines which line, if any, of the view will have interactive button behavior.
+    ///   - animatesTextColor: If true, the text color will flash when pressed. Ignored if `interactivePart` is `.none`.
```
edjamesmsft, line 237: Does design specify that this is the default behavior?

edjamesmsft:
Can you please add before screenshots as well as dark mode before and after screenshots?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
edjamesmsft: Also if there's an animated component, could you take a gif screenshot to show the behavior before and after?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
sophialee0416: I think this is more like a convenience initializer, so we probably (?) might not want to get rid of it?

sophialee0416:
> Can you please add before screenshots as well as dark mode before and after screenshots?

This is an addition to the Nav so there's nothing really to compare to the "before" 😆

edjamesmsft:
> > Can you please add before screenshots as well as dark mode before and after screenshots?
> 
> This is an addition to the Nav so there's nothing really to compare to the "before" 😆

I know this is an addition! I would still like to see the no-subtitle before and after to prove there's been no regression in behavior :)

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: Yes, this is supposed to be a bit mask. I wanted to use `OptionSet`, but that doesn't appear to play well with Objective C, so I figure this might be the next best thing.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: I want to be explicit here because this is supposed to be a bit mask. (See Saad's comment [here](https://github.com/microsoft/fluentui-apple/pull/1527/files/ecc87d8879573eb4403e3447a21d083eadf8a8d7#r1091178750).)

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: This is the "old" way to setup a `TwoLineTitleView`. It only exists here as a temporary measure to maintain backwards compatibility in case any of our clients are using this view. That being said, I don't mind deprecating it if we really want to, especially since we can point people to the new API. It's more a question of how much additional work we want to make for clients upgrading FUA to the next version.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: This is how Apple's documentation says you can specify a custom view for a navigation bar.
https://developer.apple.com/documentation/uikit/uinavigationitem/1624935-titleview

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Turns out we still do need it if we want to put a `TwoLineTitleView` in a vanilla navigation bar. Removing comment.

ios/FluentUI/Navigation/Views/LargeTitleView.swift
amgleitman: Done.

ios/FluentUI/Navigation/Views/LargeTitleView.swift
amgleitman: Is that best practice for checks that aren't closely related? Whether we're receiving a notification from the correct view and whether or not we respond to user taps seem like two different questions.

laminesm: I think we typically just group them regardless of context

amgleitman:
> Please correct me if I'm wrong, but I think the spacing between the title/subtitle may not be aligned with design. If so, will that be addressed separately?

This will be addressed when we're tokenizing this component.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
edjamesmsft: If this is only here as a temporary measure, and the plan is to remove it, it should be marked as deprecated (which is not the same as removing it yet! I can show you examples if you need them). If it's expected to remain as a convenience initializer, then the comment should be amended to indicated that it is a known, usable API, and not something that is going to be removed in the future.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: I'm not sure if we should do that. It definitely feels wrong to have a subtitle without a title, and we can definitely raise some sort of warning if that's the case, but I'm not sure if we should make this a requirement.

ios/FluentUI/Navigation/Views/LargeTitleView.swift
amgleitman: Fair.

ios/FluentUI/Navigation/NavigationBar.swift
edjamesmsft: It would be good to confirm with design about the intended behavior. If the intended design is to always have a title when a subtitle is present, we should enforce it. It's important to ensure our customers are guided to correct usage whenever possible.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: Simply deprecating this will cause other issues as is. For example, consider this line that's part of `updateSubtitleView` in `NavigationBar`:

```swift
let customTitleView = TwoLineTitleView(style: style == .primary ? .light : .dark)
customTitleView.setup(title: navigationItem.title ?? "", subtitle: subtitle, interactivePart: .title)
```

This causes an issue: **Which version of `TwoLineTitleView.setup` do we use?** As it is, the Swift compiler seems to gravitate towards this shorter version that we're looking to get rid of further down the road. The problem is, there doesn't seem to be a good way to make a function only available to Objective C, which is the ultimate problem since Objective C doesn't support default argument values like Swift does. I've looked [here](https://stackoverflow.com/questions/32384715/how-to-expose-swift-functions-only-to-objc) and [here](https://developer.apple.com/forums/thread/52340), but neither of these solutions worked for me.

The other solution is to get rid of this altogether and force Objective C consumers of FUA's `TwoLineTitleView` to update their constructors, although this is also less than ideal.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: As far as the new `alignment` and `animatesTextColor` parameters go, the default values are their current behaviors.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -53,7 +53,11 @@ open class NavigationBarTopSearchBarAttributes: NavigationBarTopAccessoryViewAtt
 /// Contains the MSNavigationTitleView class and handles passing animatable progress through
 /// Custom UI can be hidden if desired
 @objc(MSFNavigationBar)
-open class NavigationBar: UINavigationBar {
+open class NavigationBar: UINavigationBar, TwoLineTitleViewDelegate {
+    public func twoLineTitleViewDidTapOnTitle(_ twoLineTitleView: TwoLineTitleView) {
```
harrieshin, line 57: we should have comments for all public functions for documentation

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
harrieshin: no i think the bit operation is worse for readability. but if we want to use bit operation we should follow more the style of how apple does its like uiaccessibilitytraits. That way we can still use insert/remove bit operator

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -121,12 +133,14 @@ open class TwoLineTitleView: UIView {
 
     @objc public weak var delegate: TwoLineTitleViewDelegate?
 
+    private var alignment: Alignment = .center
     private var interactivePart: InteractivePart = .none
+    private var animatesTextColor: Bool = true
```
harrieshin, line 138: its bit unclear to me what animate text color means?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: `UIAccessibilityTraits` uses `OptionSet`, which doesn't play nicely in Objective C. This might be doable by defining the enum in Objective C using `NS_OPTIONS`, although this leaves the open question of whether the generated Swift version has the desired name or not.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Arguably, said documentation belongs on `TwoLineTitleViewDelegate` itself. That can be added.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: If you tap the title view in the "Objective-C Demos" section of the iOS test app, the text "flashes" briefly to indicate a press. Our existing "large title" doesn't flash, so we're adding the option to disable the flashing.

Simulator Screen Recording - iPhone 14 Pro - 2023-02-01 at 12 28 08

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
harrieshin: i am confused.  we can access uiaaccessbilitytraits in both objc and swift currently.

amgleitman: `UIAccessibilityTraits` was originally written in Objective C and has been made available in Swift. We're trying to do the opposite; that is, taking an enum written in Swift and making it available in Objective C.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: Perhaps a better name would be something like `textFlashesWhenPressed`.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: > TableViewCell has a similar structure where we kept the old API, but added a new API with more parameters, and the old API calls the new API with the values passed into the parameters. If the comment is confusing, we can either remove it or edit it?

Given this, I think we should just remove it.

laminesm (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -66,18 +74,30 @@ class NavigationControllerDemoController: DemoController {
...
 
+    @objc func showLargeTitleWithSystemStyleShyAccessoryAndSubtitle() {
+        presentController(withLargeTitle: true, subtitle: "Subtitle goes here", style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
+    }
+
     @objc func showLargeTitleWithSystemStyleAndFixedAccessory() {
         presentController(withLargeTitle: true, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: false)
     }
 
-    @objc func showRegularTitleWithShyAccessory() {
```
harrieshin, line 73: maybe the better way to handle this is to do a deprecation warning to ease people to moveto the new api

ios/FluentUI/Navigation/NavigationBar.swift
harrieshin: is this overriable for the clients who override navigationbar? still have comments here please

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -57,11 +59,23 @@ open class TwoLineTitleView: UIView {
...
-        case none
-        case title
-        case subtitle
+        // The @objc requirement doesn't let us use OptionSet, so we provide the bitmasks and the `contains` method ourselves
+        case none = 0
+        case title = 0b01
+        case subtitle = 0b10
+        case all = 0b11
+
+        func contains(_ other: InteractivePart) -> Bool {
```
harrieshin, line 76: similar to optionset, do we need helper function of  add and remove these conditions for objc clients?

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
amgleitman: This function is internal to the FUA test app. I'm just correcting a typo. :-p

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
amgleitman: Unfortunately this would require turning `InteractivePart` into a class that extends `NSObject`, which is a LOT more heavyweight than necessary. I want to rework `TwoLineTitleView` anyway to be simpler to use, anyway, so I don't think this is worth sweating at the moment.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Comment added. For what it's worth, clients who override `NavigationBar` can't override this because this method is public, not open.

PR #1529, opened 2023-01-30
 [XCUITests] Card Nudge tests

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,869,920 bytes | 28,869,920 bytes | 0 bytes |

Added targeted UI tests for Card Nudge. An accessibility identifier was added to `CardNudge` in debug that tracks the state's title, subtitle, accent text, icon, accent icon, action button, dismiss button, and style. Tests were then designed to test each part of Card Nudge. For example, tapping the action/dismiss button should prompt an alert only when the action/dismiss button are enabled.

### Verification

card nudge

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1529)


ios/FluentUI/Card Nudge/CardNudge.swift
```diff
@@ -158,6 +158,44 @@ public struct CardNudge: View, TokenizedControlView {
...
+
+            if elements.count == 1 {
+                identifier += " and \(elements[0])"
+            } else if elements.count > 1 {
+                for i in 0...elements.count - 2 {
+                    identifier += ", \(elements[i])"
+                }
+                identifier += ", and \(elements[elements.count - 1])"
+            }
+//
```
Saadnajmi, line 193: extraneous comment?

joannaquu: Nice catch, thanks!

huwilkes:
This might just be me showing how little I know about XCUITests, but is it possible to just keep running new tests while staying on the control's screens? Seems a bit silly to relaunch the app, go back to the home page, and then just go back to the control we started at.

joannaquu:
> This might just be me showing how little I know about XCUITests, but is it possible to just keep running new tests while staying on the control's screens? Seems a bit silly to relaunch the app, go back to the home page, and then just go back to the control we started at.

So for relaunching the app, there's no way around it. All tests start with `setUpWithError`, where the app is launched. There's no place where a chunk of code can be run only once before all tests. As for the going to the home page, and then back to the same page, I attempted to get rid of this in another PR but it caused some weird failures. Now that I've added test reruns, I'll open a new PR once all tests are merged to try this again but if the failures persist, I might have to leave it as is.

ios/FluentUI.Demo/FluentUIDemoTests/CardNudgeTest.swift
```diff
@@ -12,4 +12,29 @@ class CardNudgeTest: BaseTest {
     func testLaunch() throws {
         XCTAssertTrue(app.navigationBars[controlName].exists)
     }
+
+    func testElements() throws {
+        let mainIconSwitch: XCUIElement = app.tables.element(boundBy: 0).cells.element(boundBy: 2)
+        let subtitleSwitch: XCUIElement = app.tables.element(boundBy: 0).cells.element(boundBy: 3)
```
lyzhan7, line 18: Just for consistency, looks like we're testing all the switches except accent icon and accent. Should we test those too?

Also looks like we're testing just the main icon/subtitle switches in this method, whereas dismiss button/action button are in the other method. Maybe makes more sense if checking that all 6 switches work as expected happen in one method, while checking that actually clicking on the buttons work in the other method?

joannaquu: The buttons have been very finicky so I only tested the reliable ones but I'll add them in to see if they run more reliably on Github.

huwilkes (approved):

PR #1531, opened 2023-01-30
[XCUITests] HUD tests

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,875,336 bytes | 28,875,336 bytes | 0 bytes |

Added targeted UI tests for HUD. An accessibility identifier was added to `HUD` in debug that tracks the HUD's label and type (activity, success, failure, custom). Tests were then designed to test each part of HUD. For example, if the HUD is a blocking HUD, you should not be able to interact with the demo app while the HUD is presented, changes to the style should be reflected in the accessibility label, etc.

Note: since HUDs in the demo app disappear after 3 seconds, `setUpWithError` contains `sleep(5)` to ensure that the presented HUD disappears before testing the next element.

### Verification

hud

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1531)


ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -11,10 +11,55 @@ class HUDTestSwiftUI: BaseTest {
     override func setUpWithError() throws {
         try super.setUpWithError()
         app.staticTexts["SwiftUI Demo"].tap()
+        app.buttons["Present HUD for 3 seconds"].tap()
+        // sleeps for 5 seconds to wait for presented HUD to disappear
+        sleep(5)
```
Saadnajmi, line 16: Move magic number to constant?

ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -11,10 +11,55 @@ class HUDTestSwiftUI: BaseTest {
...
     }
 
     // launch test that ensures the demo app does not crash and is on the correct control page
     func testLaunch() throws {
         XCTAssertTrue(app.navigationBars.element(matching: NSPredicate(format: "identifier CONTAINS %@", controlName)).exists)
     }
+
+    func testPresentation() throws {
```
lyzhan7, line 24: Optional maybe this test would be better named as testBlockInteraction?

ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -11,10 +11,55 @@ class HUDTestSwiftUI: BaseTest {
...
+        presentHUDButton.tap()
+        // attempts to interact with Blocks interaction switch
+        blocksInteractionSwitch.tap()
+        // switch value should remain 1
+        XCTAssertTrue(blocksInteractionSwitch.value as? String == "1")
+
+        // sleeps for 5 seconds to wait for presented HUD to disappear
+        sleep(5)
+
+        blocksInteractionSwitch.tap()
```
lyzhan7, line 37: Maybe check here that the interaction switch value has changed? If interaction was just always blocked I think this test would still pass

lyzhan7 (commented):
Noticed that only the swiftUI HUD is being tested - are there plans to also add tests for the UIKit one or have we decided to only test swiftui for now?

joannaquu:
> Noticed that only the swiftUI HUD is being tested - are there plans to also add tests for the UIKit one or have we decided to only test swiftui for now?

There's not much to test on the UIKit page (it's just a list of show buttons to show each HUD) so we've decided to only test the pages where there is some sort of interaction possible.

ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -11,10 +11,55 @@ class HUDTestSwiftUI: BaseTest {
...
+
+    func testPresentation() throws {
+        let blocksInteractionSwitch: XCUIElement = app.switches["Blocks interaction"]
+        let presentHUDButton: XCUIElement = app.buttons["Present HUD for 3 seconds"]
+
+        presentHUDButton.tap()
+        // attempts to interact with Blocks interaction switch
+        blocksInteractionSwitch.tap()
+        // switch value should remain 1
+        XCTAssertTrue(blocksInteractionSwitch.value as? String == "1")
```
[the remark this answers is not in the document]
lyzhan7: I was thinking you could declare letBlocksInteractionSwitchInitialValue - blocksInteractionSwitch.value at the top of this test block, then you could just check whether or not the new values are equal to it. Just thought it might be cleaner but the comments you added also help!

lyzhan7 (approved):

ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -8,13 +8,68 @@ import XCTest
...
+        let presentHUDButton: XCUIElement = app.buttons["Present HUD for 3 seconds"]
+        var blocksInteractionSwitchInitialValue = blocksInteractionSwitch.value
+
+        // switch value should start as 1
+        XCTAssert(blocksInteractionSwitchInitialValue as? String == "1")
+        presentHUDButton.tap()
+        // attempts to interact with "Blocks interaction" switch
+        blocksInteractionSwitch.tap()
+        // switch value should not change (remains 1)
+        XCTAssert(blocksInteractionSwitch.value as? String == blocksInteractionSwitchInitialValue as? String)
```
lyzhan7, line 37: Hmm does blocksInteractionSwitch.value == blocksInteractionSwitchInitialValue not work? Do we have to cast them both to strings? Or maybe blocksInteractionSwitch.value.equalTo(blocksInteractionSwitchInitialValue)?

joannaquu: Unfortunately not because `blocksInteractionSwitch.value` is of type `Any?` so the only way to compare it is to cast it as a String.

ios/FluentUI.Demo/FluentUIDemoTests/HUDTest_SwiftUI.swift
```diff
@@ -8,13 +8,68 @@ import XCTest
...
+
+        // sleeps to wait for presented HUD to disappear
+        sleep(sleepSeconds)
+
+        // turns off block interaction functionality
+        blocksInteractionSwitch.tap()
+        blocksInteractionSwitchInitialValue = blocksInteractionSwitch.value
```
lyzhan7, line 44: I think it's a little confusing to change what the initial value is - I would only set it once at the top (line 29), if possible change it to a let - on the next line you could check that blocksInteractionSwitch.value != blocksInteractionSwitchInitialValue, and on the last line (51) check that the value has gone back to the initialvalue

joannaquu: I added the new asserts but I think it actually makes the code more confusing so I'm going to change back to just the comments.

PR #1532, opened 2023-01-30
[XCUITests] Notification View tests

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,875,336 bytes | 28,875,336 bytes | 0 bytes |

Added targeted UI tests for Notification View. An accessibility identifier was added to `NotificationView` in debug that tracks the Notification's title, message, image, action button, style, and flexibility (for toast notifications only). Tests were then designed to test each part of Notification View. For example, when show default dismiss button is enabled and there is no action button title, there should be a dismiss default button, if set trailing image is enabled and there is no action button title, there should be a trailing image, etc.

### Verification

notification

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1532)


ios/FluentUI.Demo/FluentUI.Demo/Demos/NotificationViewDemoController_SwiftUI.swift
```diff
@@ -59,6 +59,9 @@ struct NotificationDemoView: View {
         VStack {
             Rectangle()
                 .foregroundColor(.clear)
+#if DEBUG
+                .accessibilityIdentifier("Notification View with \(hasTitle ? hasAttribute ? "attributed title \"\(title)\"" : "title \"\(title)\"" : "no title"), \(hasMessage ? hasAttribute ? "attributed message \"\(message)\"" : "message \"\(message)\"" : "no message"), \(image != nil ? "an image" : "no image"), and \(actionButtonTitle != "" ? "an action button titled \"\(actionButtonTitle)\"" : trailingImage != nil ? "a trailing image" : "a dismiss button") in style \(style.rawValue) that is \(isFlexibleWidthToast ? "flexible in width" : "not flexible in width")")
+#endif
```
Saadnajmi, line 64: caann we make this easier to read?

lyzhan7: Personally I also think it's fine if the accessibility identifier isn't an actual sentence, ex. could do "Notification view, hasImage: false, actionButtonTitle: nil, isFlexibleWidthToast: true, etc. if that makes it easier

Saadnajmi: I was thinking more like pretty printing it so that things line up better

harrieshin (approved):

PR #1534, opened 2023-01-30
[XCUITests] Tooltip tests

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,869,920 bytes | 28,869,920 bytes | 0 bytes |

Added targeted UI tests for Tooltip. An accessibility identifier was added to `Tooltip` in debug that tracks the Tooltip's title, message, and direction. Tests were then designed to test each part of Tooltip. For example, the presence/lack of title should be reflected in the accessibility identifier, the correct orientation of the tooltip should also be reflected in the accessibility identifier, etc.

### Verification

Simulator Screen Recording - iPhone 14 Pro - 2023-02-01 at 10 33 01

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1534)


ios/FluentUI.Demo/FluentUIDemoTests/TooltipTest.swift
```diff
@@ -12,4 +12,29 @@ class TooltipTest: BaseTest {
     func testLaunch() throws {
         XCTAssertTrue(app.navigationBars[controlName].exists)
     }
+
+    func testMessage() throws {
+        app.buttons["Show single-line tooltip below"].tap()
+        XCTAssert(app.otherElements.element(matching: NSPredicate(format: "identifier CONTAINS %@", "Tooltip with message \"This is pointing up.\"")).exists)
```
lyzhan7, line 18: Optional, could move some of the NSPredicates into constants, similar to ActivityIndicator? Personally found that a bit easier to read

lyzhan7 (approved):
Looking at the tooltip test page in FUA there are a couple other controls that it looks like we aren't interacting with (ex. corner buttons for offset tooltips). Was wondering if there was a reason why you choose to test this subset of the components?

joannaquu:
> Looking at the tooltip test page in FUA there are a couple other controls that it looks like we aren't interacting with (ex. corner buttons for offset tooltips). Was wondering if there was a reason why you choose to test this subset of the components?

Nice catch! I'll add in a new test.

lyzhan7:
> > Looking at the tooltip test page in FUA there are a couple other controls that it looks like we aren't interacting with (ex. corner buttons for offset tooltips). Was wondering if there was a reason why you choose to test this subset of the components?
> 
> ~Nice catch! I'll add in a new test.~ Now I'm remembering why I omitted the test. There's not a great way to test for offset without either 1. adding in quite a bit of code duplication or 2. changing the existing functions just for this test.

Gotcha okay. What about the show tooltip on title component? Or the tap on tooltip dismissal component? The second one at least has some interesting interaction that might be useful to test

PR #1541, opened 2023-02-03
Fix DatePicker testSelectionManagerEnd test

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,977,864 bytes | 28,977,864 bytes | 0 bytes |

Fixed the failing DatePicker test (testSelectionManagerEnd) and reenabled the test in CI. When run locally, testSelectionManagerEnd had only been failing on Saturdays. This was because `previousIndexPath` and `nextIndexPath` were being set incorrectly when transitioning from Saturday into Sunday and vice versa (indexPath is a pair (week, day of week[0, 6])). Instead of updating day of week back to 0 for Sunday, it was updating to 7 and instead of updating day of week back to 6 for Saturday, it was updating to -1. This change moves the new `previousIndexPath` and `nextIndexPath` logic to external helper functions.

Note: I had to change my system date settings while working on this fix so the commit dates are wrong.

### Verification

| Screenshot 2023-01-28 at 10 27 48 AM | Screenshot 2023-01-28 at 10 28 06 AM |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1541)


ios/FluentUI.Tests/DatePickerControllerTests.swift
```diff
@@ -84,6 +84,22 @@ class DatePickerControllerTests: XCTestCase {
             return
         }
     }
+
+    // MARK: - helper functions
+
+    func getNextIndexPath(indexPath: IndexPath) -> IndexPath {
+        if indexPath.item == 6 {
```
huwilkes, line 91: Looks like you could probably `% 7` here, but I'm curious if this is something that should actually be fixed in the control, rather than the test?

joannaquu: Yes! That was my first thought as well but in swift -1%7 = -1 so while it'd work for `getNextIndexPath`, it wouldn't for `getPreviousIndexPath`. Would you prefer I just do it for `getNextIndexPath` or keep as is for consistency. But I can move this to the control.

huwilkes: Hm. I guess I can see why -1 would be 7*0 + -1 instead of 7*-1 + 6. I still think this is probably a fix that needs to happen in the control, rather than in the test, so I'll look into it some more.

ios/FluentUI/Date Time Pickers/Date Picker/DatePickerSelectionManager.swift
```diff
@@ -82,6 +82,20 @@ class DatePickerSelectionManager {
         }
     }
 
+    func getNextIndexPath(indexPath: IndexPath) -> IndexPath {
```
harrieshin, line 85: we should add comments similar to your pr descriptions why

harrieshin (approved):

huwilkes (approved):

PR #1542, opened 2023-02-03
Support Badge in the BottomCommandingContoller demo page

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 28,869,912 bytes | 28,927,608 bytes | 57,696 bytes |

Added new item in `BottomCommandingContoller` demo page that supports a Badge. To do so, `CommandingItem` was updated to have a new convenience init that takes in an optional badge. This badge is then provided to `TableViewCell`'s `customAccessoryView`. Although `BadgeView` is a `UIView`, using badge directly with `customAccessoryView` did not work so the work around was to add it to a `UIStackView` first.

### Verification

| Before                                       | Light                                      | Dark                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-02 at 10 58 42 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-02 at 10 54 21 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-03 at 11 50 44 |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1542)


ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -660,7 +660,11 @@ open class BottomCommandingController: UIViewController {
                 item.action?(item)
             }
         } else {
-            cell.setup(title: item.title ?? "", customView: iconView)
+            if let badge = item.badge {
```
harrieshin, line 663: i don't think we should make specifics about badge. tomorrow it might be avatar group or checkbox etc. we need to make it more generic in my opinion

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -56,11 +56,18 @@ class BottomCommandingDemoController: UIViewController {
         CommandingItem(title: "Boolean Item " + String($0), image: homeImage, action: commandAction, isToggleable: true)
     }
 
+    private lazy var badgeCommand: CommandingItem = {
+        let badge = BadgeView(dataSource: BadgeViewDataSource(text: "Badge"))
+        let stackView = UIStackView(frame: CGRect(x: 0, y: 0, width: 60, height: 25))
```
harrieshin, line 61: this seems wierd. even the tableviewcell demo cases, we pass in random things like label right? is there a different bug in cell rendering? or is the badgeView intrinsic content size isn't set correctly taht the calculation of the cell not working right? @sophialee0416

joannaquu: While I was testing, there was definitely something weird going on with both TableViewCell and BadgeView, I'll followup with Sophia.

joannaquu: Offline we agreed to merge these changes and file a separate bug for TableViewCell.

ios/FluentUI/Bottom Commanding/CommandingItem.swift
```diff
@@ -53,6 +53,15 @@ open class CommandingItem: NSObject {
         }
     }
 
+    /// A `UIView` to be displayed on the trailing end.
```
edjamesmsft, line 56: Would be nice to see this used in the objc demo as well, somehow, given it is new objc exposure.

ios/FluentUI/Bottom Commanding/CommandingItem.swift
```diff
@@ -86,6 +95,11 @@ open class CommandingItem: NSObject {
         self.isToggleable = isToggleable
     }
 
+    @objc public convenience init(title: String, image: UIImage, trailingView: UIView, action: @escaping (CommandingItem) -> Void, isToggleable: Bool = false) {
```
harrieshin, line 98: comments

harrieshin, line 98: do we need this init method? the trailingView feels like a transient thing clients will update as they need to highlight something in the cell. does it make sense to have it in init if we already have a public objc var that they can easily customize? trying to be conservative with our binary impact.

harrieshin (approved):

ios/FluentUI/Bottom Commanding/CommandingItem.swift
joannaquu: Offline Harrie mentioned adding to the BottomCommandingController demo page only so resolving for now but could be added going forward!

PR #1554, opened 2023-02-07
Avatar accessory image changed to accommodate multiple rendering modes

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In some scenarios, the activity image will not show even if the cutout is made within the Avatar if the image file is an SVG and the asset image is rendered as a Template Image. Added an SVG example in the demo controller and included hard-coded foreground color that will later be exposed as a token in a follow-up PR to unblock downstream dependencies.

### Verification

| image | image |
| image | image |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://api.github.com/repos/microsoft/fluentui-apple/pulls/1554)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -339,6 +339,9 @@ public struct Avatar: View, TokenizedControlView {
                                     .overlay(accessoryImage
                                         .interpolation(.high)
                                         .resizable()
+                                         // Rendering mode should be original for now to allow any image to show,
+                                         // but eventually should be changed to template when we add foreground tokens.
+                                        .renderingMode(.original)
                                         .frame(width: shouldDisplayActivity ? activityImageSize : accessoryIconSize,
                                                height: shouldDisplayActivity ? activityImageSize : accessoryIconSize,
                                                alignment: .center)
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -341,7 +341,7 @@ public struct Avatar: View, TokenizedControlView {
                                         .resizable()
                                          // Rendering mode should be original for now to allow any image to show,
                                          // but eventually should be changed to template when we add foreground tokens.
-                                        .renderingMode(.original)
+                                        .renderingMode(shouldDisplayPresence ? .template : .original)
                                         .frame(width: shouldDisplayActivity ? activityImageSize : accessoryIconSize,
                                                height: shouldDisplayActivity ? activityImageSize : accessoryIconSize,
                                                alignment: .center)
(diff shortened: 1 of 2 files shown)


edjamesmsft (approved):

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -339,6 +339,9 @@ public struct Avatar: View, TokenizedControlView {
                                     .overlay(accessoryImage
                                         .interpolation(.high)
                                         .resizable()
+                                         // Rendering mode should be original for now to allow any image to show,
+                                         // but eventually should be changed to template when we add foreground tokens.
+                                        .renderingMode(shouldDisplayPresence ? .template : .original)
```
harrieshin, line 344: i feel like we should respect accessoryImage.renderingMode? just because it is cutout doesn't mean it has to be original

harrieshin:
"Note: The images are showing but are currently not the correct color. This will come in a later PR." even if we aren't introducing the public token right away for .12 patch, we can still define the static constant value?

harrieshin (changes requested):
2 comments

ios/FluentUI/Avatar/Avatar.swift
sophialee0416: Agreed, and it was going to be a change in a follow-up PR for the following reasons:
- Presence should always be template if we want to stick to FluentUI design -- anything other than template will look like this: image
- Following the accessoryImage.renderingMode(<insert rendering mode here>) just means I need to expose another API anyway, which is what we want to avoid in this change here. This was going to be introduced in the next change where I can expose a property that allows user to set which rendering mode they'd like to use

sophialee0416:
> "Note: The images are showing but are currently not the correct color. This will come in a later PR." even if we aren't introducing the public token right away for .12 patch, we can still define the static constant value?

Seems like SwiftUI won't allow images to take on custom foreground colors, which I can't include in this change because that would mean exposing a new API (mentioned in thread above)

ios/FluentUI/Avatar/Avatar.swift
harrieshin: what is the exposed api? isn't it just part of UIImage?

sophialee0416: Synced offline -- I was talking about the SwiftUI context, not the UIKit context. Setting the property on the UIKit side will reduce over-complicating logic in the component side

harrieshin:
please update PR description screenshots? i think darkmode bug should be fixed

harrieshin (approved):

PR #1050, opened 2022-07-01
SearchBar Colors Update

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The SearchBar's colors were changed to match Fluent 2 colors. A new `brandContent` style was added.

### Verification

The changes were tested on the Demo app. The brand colors were tested by changing the SearchBar's style to `brandContent` in the `SearchBarDemoController`. 

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before_rest_light | after_rest_light |
| before_typing_light | after_typing_light |
| before_rest_dark | after_rest_dark |
| before_typing_dark | after_typing_dark |

**Brand**

| Rest                                       | Typing                                      |
|----------------------------------------------|--------------------------------------------|
| brand_resting | brand_typing |

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1050)

PR #1564, opened 2023-02-10
Get theme from parent to update ShyHeaderController colors

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There's a bug where the `ShyHeaderController` wasn't using the correct theme to update colors. This PR fixes the bug by passing down a weak reference of the view context to make sure `ShyHeaderController` uses the correct theme to update its colors. Also, `viewDidAppear` has been removed and the color updates will be called in `viewWillAppear`.
This will be a short term fix, as we will be looking into ways to delegate `ShyHeaderController` color updates to `NavigationController` in the longer term.
This PR also fixes a bug introduced by #1050 where the `SearchBar` doesn't use the right branding colors. The `brandContent` style for the `SearchBar` has been removed to partially revert the change and fix the bug.

Binary change:

| libFluentUI.a | 29,075,768 bytes | 29,085,640 bytes | 9,872 bytes |

### Verification

| before | after |
| before_blue | after_blue |
| before_green | after_green |
| before_dark | after_dark |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1564)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -66,10 +66,15 @@ class ShyHeaderController: UIViewController {
     private var contentScrollViewObservation: NSKeyValueObservation?
     private var previousContentScrollViewTraits = ContentScrollViewTraits() //properties of the scroll view at the last scrollDidOccurIn: update. Used with current traits to understand user action
 
-    init(contentViewController: UIViewController) {
+    // The context of the parent controller used to pull the correct FluentTheme to update visuals
+    weak var containingView: UIView?
+
+    init(contentViewController: UIViewController, containingView: UIView?) {
         self.contentViewController = contentViewController
         shyHeaderView.accessoryView = contentViewController.navigationItem.accessoryView
         shyHeaderView.navigationBarShadow = contentViewController.navigationItem.navigationBarShadow
+
+        self.containingView = containingView
 
         super.init(nibName: nil, bundle: nil)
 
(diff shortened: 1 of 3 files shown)


edjamesmsft (approved):

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -66,11 +66,16 @@ class ShyHeaderController: UIViewController {
     private var contentScrollViewObservation: NSKeyValueObservation?
     private var previousContentScrollViewTraits = ContentScrollViewTraits() //properties of the scroll view at the last scrollDidOccurIn: update. Used with current traits to understand user action
 
-    init(contentViewController: UIViewController) {
+    // The context of the parent controller used to pull the correct FluentTheme to update visuals
+    weak var containingView: UIView?
```
harrieshin, line 70: just for my understanding, do we need the entire containingView reference? or just the fluenttheme reference from navigationcontroller?

laminesm: Passing a fluentTheme reference is not guaranteed to work. If for whatever reason the parent view doesn't have a custom theme when we are grabbing its theme, it will return a reference to the shared theme and we will be stuck with that even when the parent view's theme changes.

harrieshin (approved):

PR #1577, opened 2023-02-15
communicationTextColor should use global tokens

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

`ActionCell`'s `communicationTextColor` is used in partner repos but it currently has a wrong value. It shouldn't be a brand color but comm80/100 instead.

Binary change:

| libFluentUI.a | 29,120,088 bytes | 29,123,568 bytes | 3,480 bytes |

### Verification

### Pull request checklist

This PR has considered:


harrieshin:
Is that the behavor of fluent 0.12? This change may need to be in .13 branch?

laminesm:
> Is that the behavor of fluent 0.12? This change may need to be in .13 branch?

Yes, I'll cherry pick it to the 0.13 branch

PR #1579, opened 2023-02-15
Enable Fluent-style back button for large/leading titles

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This allows a back button to be shown on navigation bars with large/leading titles. We also ensure that we use the Fluent-style back button (which has a thinner stroke than vanilla UIKit's back button) and ensure that leading spacing is correct.

### Verification

Testing has been done in the following areas:
* Ensure that we can navigate between a root view controller with both leading and centered titles.
* Ensure that the back button text is labeled appropriately when navigating from either direction (i.e., from root to child and from grandchild to child).
* Ensure that the back button faces the correct direction in RTL languages.
* Ensure that the layout is correct on iPhones running iOS 14, iOS 15, and iOS 16. (As of 3fe3ed738403b889244e26fc6781d08af71fef59, there's a known issue where the back button on centered titles is misaligned when we have a full-width app on iPad. We're still investigating this.)

| Leading | Centered | Leading (RTL) | Centered (RTL) |
|-|-|-|-|
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 11 44 55 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 11 45 04 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 11 40 59 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 11 41 03 |

### Test App Changes

* Added a new test case where we show a custom leading button in the root view instead of the avatar.
* The disclosure button in child cell 6 now changes the style of the title bar. This is more visible in a standalone app than `NSLog`. (I would've shown a `UIAlert`, but since the navigation controller we're demoing is already detached, it's considered bad practice to show another modal dialog.)

| Custom leading button | Disclosure action |
|-|-|
|  | Simulator Screen Recording - iPhone 14 Pro - 2023-02-15 at 13 33 14 |

### Pull request checklist

This PR has considered:


ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -169,6 +176,10 @@ class NavigationControllerDemoController: DemoController {
         if showAvatar {
             controller.msfNavigationBar.personaData = content.personaData
             controller.msfNavigationBar.onAvatarTapped = handleAvatarTapped
+        } else if showStarButton {
```
huwilkes, line 179: Are `showAvatar` and `showStarButton` supposed to be exclusive?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -332,6 +340,13 @@ open class NavigationBar: UINavigationBar, TwoLineTitleViewDelegate {
     private var navigationBarShadowObserver: NSKeyValueObservation?
     private var usesLargeTitleObserver: NSKeyValueObservation?
 
+    private var backButtonItem: UIBarButtonItem?
```
huwilkes, line 343: Looks like this is never set outside of init. Should this be a `let`, and non-optional?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -393,6 +408,8 @@ open class NavigationBar: UINavigationBar, TwoLineTitleViewDelegate {
         rightBarButtonItemsStackView.setContentHuggingPriority(.defaultHigh, for: .horizontal)
         rightBarButtonItemsStackView.setContentCompressionResistancePriority(.defaultHigh, for: .horizontal)
 
+        backButtonItem = UIBarButtonItem(image: UIImage.staticImageNamed("back-24x24"), style: .plain, target: nil, action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
```
huwilkes, line 411: Do you need to set the accessibility label? I think right now VoiceOver will just read the image name, rather than a localized description.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -656,9 +675,9 @@ open class NavigationBar: UINavigationBar, TwoLineTitleViewDelegate {
             if isLeftItem {
                 let isRTL = effectiveUserInterfaceLayoutDirection == .rightToLeft
                 button.contentEdgeInsets = UIEdgeInsets(top: 0,
-                                                        left: isRTL ? 0 : Constants.leftBarButtonItemLeadingMargin,
+                                                        left: isRTL ? Constants.leftBarButtonItemTrailingMargin : Constants.leftBarButtonItemLeadingMargin,
                                                         bottom: 0,
-                                                        right: isRTL ? Constants.leftBarButtonItemLeadingMargin : 0)
+                                                        right: isRTL ? Constants.leftBarButtonItemLeadingMargin : Constants.leftBarButtonItemTrailingMargin)
```
huwilkes, line 680: Could you factor the `isRTL` logic out? Maybe something like 
```
let leftInset: CGFloat
let rightInset: CGFloat
if effectiveUserInterfaceLayoutDirection == .rightToLeft {
  leftInset = Constants.leftBarButtonItemTrailingMargin
  rightInset = Constants.leftBarButtonLeadingMargin
} else {
  leftInset = Constants.leftBarButtonItemLeadingMargin
  rightInset = Constants.leftBarButtonTrailingMargin
}
button.contentEdgeInsets = UIEdgeInsets(top: 0, left: leftInset, bottom: 0, right: rightInset)
```

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -86,6 +86,13 @@ open class NavigationBarTitleAccessory: NSObject {
     }
 }
 
+// MARK: - NavigationBarBackButtonDelegate
+/// Handles presses from a `NavigationBar`'s makeshift back button.
```
huwilkes, line 90: nit: maybe "custom" instead of "makeshift"

amgleitman: Reworded.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Will fix.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: At this point we might as well factor out the `isLeftItem` check too. It actually saves on binary size. 😁

huwilkes (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
amgleitman: Yes, this is by design. As per `LargeTitleView`'s [`updateProfileButtonVisibility`](https://github.com/microsoft/fluentui-apple/blob/main/ios/FluentUI/Navigation/Views/LargeTitleView.swift#L310-L312) method, we don't support showing a custom leading button and an avatar at the same time. We can show a back button with an avatar at the same time, but this is only because the back button is REALLY important. Clients have explicitly requested the "avatar + back button + two-line title" case, and I think that looks okay, but swapping in a single large title looks somewhat gross in my opinion.

| Back button and avatar, large title | Back button and avatar, two-line title |
|-|-|
| Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 15 37 18 | Simulator Screen Shot - iPhone 14 Pro - 2023-02-15 at 15 37 22 |

amgleitman: That being said, I think we can clean up `presentController` to make this a bit more clear.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: VoiceOver appears to read this by its title, so I think we're okay for now. It might be nicer if we could get VoiceOver to say "back button," but that's a separate can of worms.

PR #1581, opened 2023-02-16
Fixing initial theme setting for `willMove(toWindow:)`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

We recently made a change where we would respect `FluentTheme` changes during a view's `willMove(toWindow:)` call, which is early enough that the correct theme color can be applied before the view appears on screen.

In order to support this, we need to cache the incoming `FluentTheme` instance off the `newWindow` parameter. Easy enough for a tokenized control: we simply add a call to `tokenSet.update(newWindow.fluentTheme)` and everything works as expected. 

It's a bit trickier for controls like `BadgeView` that are not yet tokenized: up until now they've been pulling alias tokens directly from their current window's `fluentTheme`, which unfortunately will not have been set yet during `willMove(toWindow:)`. So, I need a place to cache the theme and fetch it during render.

The most future-proofed solution I could come up with: partial tokenization. Each non-tokenized control that references a brand token now conforms to `TokenizedControlInternal` and boasts a `tokenSet` property, which will manage the `fluentTheme` itself.

However, since we don't _have_ tokens for all those controls yet, they will temporarily make use of the new `EmptyTokenSet`.

This did require some heavier changes in a few components, but in most cases it was as simple as the above.

Binary change:

| libFluentUI.a | 28,933,360 | 29,245,256 | +311,896 |
| FluentUI.Demo.app | 32,094,967 | 32,219,751 | +124,784 |

(a summary of the changes made, often organized by file)

### Verification

#### Before: 
https://user-images.githubusercontent.com/4934719/219449677-204d17e0-64d4-4622-8f3b-1d6e09f93840.mp4

#### After:
https://user-images.githubusercontent.com/4934719/219449686-4c0b9d57-eedb-4f34-aad4-9d274b1d4362.mp4

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1581)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1581)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
@@ -36,6 +36,13 @@ class ColoredPillBackgroundView: UIView {
 
     override func didMoveToWindow() {
         super.didMoveToWindow()
+
+        // Note: We usually do updates during `willMove(toWindow:)` to ensure that there's no "flash" of the
+        // old color in cases whre the view is briefly visible before this API is called. However, the
+        // public APIs for easily hooking into theme changes have not yet been exposed, so this demo
+        // controller is not in a position to easily follow those rules. This will be sufficient for our
+        // current needs, but it's technically less correct than I'd like.
+        // TODO: update this to use proper theme updating hooks once they're built
         updateBackgroundColor()
     }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
@@ -34,6 +34,11 @@ class ColoredPillBackgroundView: UIView {
         updateBackgroundColor()
     }
 
+    override func didMoveToWindow() {
+        super.didMoveToWindow()
+        updateBackgroundColor()
+    }
+
     func updateBackgroundColor() {
         switch style {
         case .neutral:
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -102,10 +102,10 @@ public class CommandBar: UIView, TokenizedControlInternal {
 
     public override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-		guard let newWindow else {
-			return
-		}
-		tokenSet.update(newWindow.fluentTheme)
+        guard let newWindow else {
+            return
+        }
+        tokenSet.update(newWindow.fluentTheme)
         updateButtonTokens()
     }
 
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -102,8 +102,10 @@ public class CommandBar: UIView, TokenizedControlInternal {
 
     public override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-
-        tokenSet.update(fluentTheme)
+		guard let newWindow else {
+			return
+		}
+		tokenSet.update(newWindow.fluentTheme)
         updateButtonTokens()
     }
 
--- a/ios/FluentUI/Core/Theme/Tokens/EmptyTokenSet.swift
+++ b/ios/FluentUI/Core/Theme/Tokens/EmptyTokenSet.swift
@@ -3,9 +3,13 @@
 //  Licensed under the MIT License.
 //
 
-// MARK: TokenSet
+/// An empty `ControlTokenSet` for components that want to use some of the perks
+/// of being tokenized, but are not fully at that stage yet.
 public class EmptyTokenSet: ControlTokenSet<EmptyTokenSet.Tokens> {
+
+    /// The set of tokens associated with this `EmptyTokenSet`.
     public enum Tokens: TokenSetKey {
+        /// A default token, which only exists because Swift requires at least one value in this enum.
         case none
     }
     init() {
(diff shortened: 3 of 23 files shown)


ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
```diff
@@ -34,6 +34,11 @@ class ColoredPillBackgroundView: UIView {
         updateBackgroundColor()
     }
 
+    override func didMoveToWindow() {
```
harrieshin, line 37: and this is because, there isn't a listener machanism for demo controllers? should we have todos?

ios/FluentUI/Other Cells/ActionsCell.swift
```diff
@@ -228,8 +228,10 @@ open class ActionsCell: UITableViewCell, TokenizedControlInternal {
 
     open override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-
-        tokenSet.update(fluentTheme)
+		guard let newWindow else {
+			return
+		}
+		tokenSet.update(newWindow.fluentTheme)
```
harrieshin, line 234: nice catch thanks!

harrieshin (approved):

sophialee0416:
I'm trusting that the theming validation here works? (since I don't see it here in the description)

ios/FluentUI/Core/Theme/Tokens/EmptyTokenSet.swift
```diff
@@ -0,0 +1,16 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+// MARK: TokenSet
+public class EmptyTokenSet: ControlTokenSet<EmptyTokenSet.Tokens> {
```
sophialee0416, line 7: need doc for this later?

sophialee0416:
This can be resolved after the merge, but I'm def curious to see what the binary size delta looks here with this change

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -102,8 +102,10 @@ public class CommandBar: UIView, TokenizedControlInternal {
 
     public override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-
-        tokenSet.update(fluentTheme)
+		guard let newWindow else {
```
laminesm, line 105: Spacing issue?

ios/FluentUI/Other Cells/ActionsCell.swift
```diff
@@ -228,8 +228,10 @@ open class ActionsCell: UITableViewCell, TokenizedControlInternal {
 
     open override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-
-        tokenSet.update(fluentTheme)
+		guard let newWindow else {
```
laminesm, line 231: Spacing issue?

ios/FluentUI/Other Cells/ActivityIndicatorCell.swift
```diff
@@ -74,7 +74,10 @@ open class ActivityIndicatorCell: UITableViewCell, TokenizedControlInternal {
 
     open override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
-        tokenSet.update(fluentTheme)
+		guard let newWindow else {
```
laminesm, line 77: I'm sorry if all these aren't actual spacing issues but they look off on my end haha

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -539,22 +554,10 @@ private class TableViewHeaderFooterTitleView: UITextView {
         self.textContainer.lineFragmentPadding = 0
         textContainerInset = .zero
         layoutManager.usesFontLeading = false
-        updateLinkTextColor()
-
-        NotificationCenter.default.addObserver(self,
```
laminesm, line 544: ooc - why don't we need this anymore?

laminesm (approved):

ios/FluentUI/Command Bar/CommandBar.swift
mischreiber: Yikes good catch

mischreiber: Moved all tabs to spaces

ios/FluentUI/Other Cells/ActivityIndicatorCell.swift
mischreiber: Nope it was very wrong, thank you!

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
mischreiber: This view was independently observing color changes, but it's a private subview of its parent. I updated it so that the parent `TableHeaderFooterView` would just inject the current `fluentTheme` during updates

ios/FluentUI.Demo/FluentUI.Demo/ColoredPillBackgroundView.swift
mischreiber: Yeah, I can add a quick todo. It's "good enough" for our demo needs, and a real fix would be more expensive.

edjamesmsft (approved):

PR #1590, opened 2023-02-17
[FHL] Add macOS Notification

joannaquu:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Binary change:

| libFluentUITestViewControllers.a | 2,838,064 bytes | 2,989,512 bytes | 151,448 bytes |

Added a simplified SwiftUI Notification to macOS. Leveraged the existing iOS SwiftUI Notification to create the macOS version with a few caveats.
1. The mac notification is using fluent 1 colors. 
2. It does not support some of the more complicated features that the iOS one does (attributed string, trailing image, background gradient, shadows). 
3. Presentation and width need to be set externally (similar to when `shouldSelfPresent` is false in iOS).
4. Since there are no existing design specs, notification has been trimmed down to just the bar variants.

### Verification

Screenshot 2023-03-01 at 9 53 40 AM

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1590)

### DIFF
--- a/macos/FluentUI/Notification/Notification.swift
+++ b/macos/FluentUI/Notification/Notification.swift
@@ -8,13 +8,13 @@ import SwiftUI
 /// Pre-defined styles of the notification
 @objc public enum MSFNotificationStyle: Int, CaseIterable {
 	/// Bar notification with brand colored text and background.
-	case primaryBar
+	case accent
 
 	/// Bar notification with brand colored text and neutral colored background.
-	case primaryOutlineBar
+	case subtle
 
 	/// Bar notification with neutral colored text and brackground.
-	case neutralBar
+	case neutral
 }
 
 /// Properties that can be used to customize the appearance of the `Notification`.
@@ -61,21 +61,21 @@ public struct Notification: View {
 		self.state = state
 		self.backgroundColor = {
 			switch state.style {
-			case .primaryBar:
-				return Constants.primaryBackgroundColor
-			case .primaryOutlineBar:
-				return Constants.primaryOutlineBackgroundColor
-			case .neutralBar:
+			case .accent:
+				return Constants.accentBackgroundColor
+			case .subtle:
+				return Constants.subtleBackgroundColor
+			case .neutral:
 				return Constants.neutralBackgroundColor
 			}
 		}()
 		self.foregroundColor = {
 			switch state.style {
-			case .primaryBar:
-				return Constants.primaryForegroundColor
-			case .primaryOutlineBar:
-				return Constants.primaryOutlineForegroundColor
-			case .neutralBar:
+			case .accent:
+				return Constants.accentForegroundColor
+			case .subtle:
+				return Constants.subtleForegroundColor
+			case .neutral:
 				return Constants.neutralForegroundColor
 			}
 		}()
@@ -156,7 +156,7 @@ public struct Notification: View {
 				.background(
 					backgroundColor
 				)
-				.overlay(Rectangle().frame(width: nil, height: state.style == .primaryOutlineBar ? Constants.outlineWidth : 0, alignment: .top).foregroundColor(Constants.neutralBackgroundColor), alignment: .top)
+				.overlay(Rectangle().frame(width: nil, height: state.style == .subtle ? Constants.outlineWidth : 0, alignment: .top).foregroundColor(Constants.neutralBackgroundColor), alignment: .top)
 				.onTapGesture {
 					if let messageAction = messageButtonAction {
 						messageAction()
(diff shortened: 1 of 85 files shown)


macos/FluentUI/Notification/Notification.swift
```diff
@@ -0,0 +1,346 @@
...
+		}
+
+		@ViewBuilder
+		var button: some View {
+			if let buttonAction = state.actionButtonAction ?? nil {
+				if let actionTitle = state.actionButtonTitle, !actionTitle.isEmpty {
+					SwiftUI.Button(action: {
+						buttonAction()
+					}, label: {
+						if #available(macOS 13.0, *) {
```
harrieshin, line 185: fluent framework supports only macOS15+ so you shouldn't need this

macos/FluentUITestViewControllers/TestNotificationController.swift
```diff
@@ -0,0 +1,46 @@
...
+
+		var constraints = [NSLayoutConstraint]()
+
+		for (i, notification) in notificationList.enumerated() {
+			let hostingView = NSHostingView(rootView: notification)
+			containerView.addView(hostingView, in: .center)
+			if i < 5 {
+				constraints.append(hostingView.widthAnchor.constraint(equalToConstant: 350))
+			} else {
+				constraints.append(hostingView.leadingAnchor.constraint(equalTo: containerView.leadingAnchor))
```
harrieshin, line 35: weird you don't get any constraint warning that the position is ambiguous if you only have leading anchor?

macos/FluentUI/Notification/Notification.swift
```diff
@@ -0,0 +1,346 @@
...
+		static let dangerToastForegroundColor: Color = Color(FluentUI.Colors.Palette.dangerShade20.color)
+		static let warningToastBackgroundColor: Color = Color(FluentUI.Colors.Palette.warningTint30.color)
+		static let warningToastForegroundColor: Color = Color(FluentUI.Colors.Palette.warningShade30.color)
+
+		static let toastCornerRadius: CGFloat = 12
+		static let barCornerRadius: CGFloat = 0
+		static let horizontalPadding: CGFloat = 16
+		static let verticalPadding: CGFloat = 12
+		static let horizontalSpacing: CGFloat = 16
+		static let minimumHeight: CGFloat = 52
```
harrieshin, line 293: because we don't have redlines for this i was curious if we could follow similar sizes of existing apps 
image

harrieshin:
we need a bar style with a action button on it for the feature request we have. which i don't think we have in the demo. so like toast style but longer.

macos/FluentUI/Notification/Notification.swift
joannaquu: The readme, as well as `FluentUI_common.xcconfig` still have macOS 10.15, should they be updated?

macos/FluentUITestViewControllers/TestNotificationController.swift
joannaquu: Nope, but I'll add trailing anchor as well just in case.

macos/FluentUI/Notification/Notification.swift
harrieshin: clearly i am dumb. and there is difference between 10.15 vs macos13

mischreiber (commented):
Very interesting, and exciting that it was relatively easy to move an iOS SwiftUI component to macOS!

The next step I'd like to see before merging: how can we use the same SwiftUI file for both platforms, without having to make a copy?

joannaquu:
> Very interesting, and exciting that it was relatively easy to move an iOS SwiftUI component to macOS!
> 
> The next step I'd like to see before merging: how can we use the same SwiftUI file for both platforms, without having to make a copy?

So I did have to rip out a bunch of code to get it working on macOS so I'm assuming it'll mean adding a bunch of ifdefs and restructuring the files since iOS and macOS are in two different projects. Any thoughts @harrieshin?

macos/FluentUI/Notification/Notification.swift
```diff
@@ -0,0 +1,241 @@
...
+                  messageButtonAction: nil,
+                  showFromBottom: true)
+	}
+
+    init(style: MSFNotificationStyle,
+         message: String? = nil,
+         actionButtonTitle: String? = nil,
+         actionButtonAction: (() -> Void)? = nil,
+         messageButtonAction: (() -> Void)? = nil,
+         showFromBottom: Bool = true) {
```
harrieshin, line 231: we should make it enum. just incase  like notiication animates from the side like system notfication

macos/FluentUI/Notification/Notification.swift
```diff
@@ -0,0 +1,241 @@
...
+
+	/// Action to be dispatched by tapping on the toast/bar notification.
+	var messageButtonAction: (() -> Void)? { get set }
+	
+	/// Defines whether the notification shows from the bottom or the top.
+	var showFromBottom: Bool { get set }
+}
+
+/// View that represents the Notification.
+public struct NotificationView: View {
```
harrieshin, line 46: NotificationBarView?

harrieshin (approved):

PR #1592, opened 2023-02-21
Make BadgeView's accessibilityLabel customizable

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Binary change:

| libFluentUI.a | 29,120,456 bytes | 29,121,224 bytes | 768 bytes |

Updated BadgeView's `accessibilityLabel` to allow for a custom label. If a custom accessibility label is not provided, it defaults to `label.text`.

### Verification

Screenshot 2023-02-21 at 1 20 08 PM

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1592)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -523,6 +523,6 @@ open class BadgeView: UIView, TokenizedControlInternal {
 
     open override var accessibilityLabel: String? {
         get { return dataSource?.accessibilityLabel ?? label.text }
-        set { dataSource?.accessibilityLabel = newValue }
+        set { }
     }
 }
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -521,5 +521,10 @@ open class BadgeView: UIView, TokenizedControlInternal {
 
     // MARK: Accessibility
 
-    open override var accessibilityLabel: String? { get { return label.text } set { } }
+    private var _accessibilityLabel: String?
+
+    open override var accessibilityLabel: String? {
+        get { return _accessibilityLabel ?? label.text }
+        set { _accessibilityLabel = newValue }
+    }
 }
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -521,5 +521,10 @@ open class BadgeView: UIView, TokenizedControlInternal {
 
     // MARK: Accessibility
 
-    open override var accessibilityLabel: String? { get { return label.text } set { } }
+    private var _accessibilityLabel: String?
```
huwilkes, line 524: Should this be on the datasource instead? Seems a bit odd to set the text and the accessibility label in different places.

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -521,5 +521,8 @@ open class BadgeView: UIView, TokenizedControlInternal {
 
     // MARK: Accessibility
 
-    open override var accessibilityLabel: String? { get { return label.text } set { } }
+    open override var accessibilityLabel: String? {
+        get { return dataSource?.accessibilityLabel ?? label.text }
```
edjamesmsft, line 525: Where does this accessibilityLabel come from in the dataSource? I'm trying to find it in code and I can't track it down.

joannaquu: It's the NSObject's `accessibilityLabel`. Originally I tried to add a new variable in dataSource but the compiler didn't like that.

ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -521,5 +521,8 @@ open class BadgeView: UIView, TokenizedControlInternal {
...
+        set { dataSource?.accessibilityLabel = newValue }
```
huwilkes, line 526: I'm not sure we need the set anymore. People should be able to update the accessibility label through the dataSource, so I don't think we need it exposed in two places.

huwilkes (approved):

PR #1596, opened 2023-02-23
Delegate ShyHeaderController's colors update to NavigationController

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

There is a bug in partner repos where the shy header would have the wrong color when backing out of a file or switching from dark to light mode. This issue was irreproducible on the demo app, but seemed to be caused by the `ShyHeaderController` fetching the wrong theme when updating its colors.
To fix the bug, we delegate shy header color updates to `NavigationController` which updates `ShyHeaderController`'s colors by injecting the correct FluentTheme from the navigation bar's tokenSet. 

Binary change:

Total increase: 26,384 bytes
Total decrease: -1,592 bytes
| Total | 29,425,960 bytes | 29,450,752 bytes | ⚠️ 24,792 bytes |
<details>

| NavigationController.o | 163,112 bytes | 185,120 bytes | ⚠️ 22,008 bytes |
| __.SYMDEF | 4,604,160 bytes | 4,607,632 bytes | ⚠️ 3,472 bytes |
| BadgeLabelButton.o | 852,160 bytes | 853,064 bytes | ⚠️ 904 bytes |
| ShyHeaderController.o | 251,216 bytes | 249,624 bytes | 🎉 -1,592 bytes |
</details>

### Verification

<details>

| before_brand | after_brand |
| before_custom | after_custom |
| before_system | after_system |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1596)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -125,6 +125,10 @@ class ShyHeaderController: UIViewController {
 
         updatePadding()
         setupNotificationObservers()
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
         updateNavigationBarStyle(theme: containingView?.fluentTheme ?? view.fluentTheme)
     }
 
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -259,14 +259,14 @@ class ShyHeaderController: UIViewController {
         return true
     }
 
-    private func updateBackgroundColor(with item: UINavigationItem) {
-        let color = item.navigationBarColor(fluentTheme: containingView?.fluentTheme ?? view.fluentTheme)
+    public func updateBackgroundColor(with item: UINavigationItem, theme: FluentTheme) {
+        let color = item.navigationBarColor(fluentTheme: theme)
         shyHeaderView.backgroundColor = color
         view.backgroundColor = color
         paddingView.backgroundColor = color
 
         navigationBarColorObservation = item.observe(\.customNavigationBarColor) { [weak self] item, _ in
-            self?.updateBackgroundColor(with: item)
+            self?.updateBackgroundColor(with: item, theme: theme)
         }
     }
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -523,10 +519,10 @@ class ShyHeaderController: UIViewController {
     }
 
     /// Updates based on the current navigation bar style.
-    private func updateNavigationBarStyle() {
+    public func updateNavigationBarStyle(theme: FluentTheme) {
```
harrieshin, line 522: why is this public? shyheader isn't a public class?

laminesm: Doesn't need to be public, I'll fix it!

mischreiber (commented):
Are you sure you uploaded the right Before images in the PR description? I don't see the bug reproing.

laminesm:
> Are you sure you uploaded the right Before images in the PR description? I don't see the bug reproing.

Yep that's right, the issue was not reproducible in our demo app, it was only present on the partner apps.

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -126,7 +125,11 @@ class ShyHeaderController: UIViewController {
 
         updatePadding()
         setupNotificationObservers()
-        updateNavigationBarStyle()
+    }
+
+    override func viewDidAppear(_ animated: Bool) {
```
mischreiber, line 130: Isn't viewDidAppear too late for doing appearance changes? As in, won't there be a blue flicker if we do the update at this point?

laminesm: Updating in viewWillAppear doesn't always reliably access the right theme, so updating colors in viewDidAppear is safer (which was probably why it was there beforehand). Also, I wasn't able to reproduce the issue you mentioned on this branch since the NavigationController handles the color updates with the correct theme from now on.

harrieshin (approved):

PR #1601, opened 2023-02-24
PillButton text should not scale with DynamicType.

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

The titleLabel text within the `PillButton` was scaling with Dynamic Type. This is not support behavior of the PillButton. Before Fluent2, the PillButton used a fixed size font. We should match that here.

To fix this, set the `shouldScale` property to `false`. This will prevent the text from scaling.

### Verification

- Verified in the PillButton and PillButtonBar demo controllers
- Verified in the devmain applications

Before:
before2

After:
after3

### Pull request checklist

This PR has considered:
- [] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1601)

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -151,7 +151,7 @@ open class PillButton: UIButton, TokenizedControlInternal {
             }
         } else {
             setTitle(pillBarItem.title, for: .normal)
-            titleLabel?.font = UIFont.fluent(titleFont)
+            titleLabel?.font = UIFont.fluent(titleFont, shouldScale: false)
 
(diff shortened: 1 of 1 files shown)


harrieshin:
should it scale upto certain point? or should it not scale at all?

harrieshin (approved):

nodes11:
I am not aware of any spec that defines a maximum font size for the control. This behavior matches the pre-fluent2 behavior.

PR #1613, opened 2023-02-28
Update NavigationController's padding view background color in ShyHeaderView's willMove(to: window)

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

When changing to landscape mode after backing out of a file in partner apps, the `NavigationController`'s padding view gets its background color from the shared theme fetched during `ShyHeaderController`'s `viewWillAppear`. 
To remedy this bug, the padding view background color now gets updated together with the `ShyHeaderView`'s background color in its `willMove(to: window)` function.

Binary change:

Total increase: 4,544 bytes
Total decrease: 0 bytes
| Total | 30,085,456 bytes | 30,090,000 bytes | ⚠️ 4,544 bytes |
<details>

| ShyHeaderView.o | 119,440 bytes | 121,336 bytes | ⚠️ 1,896 bytes |
| ShyHeaderController.o | 251,624 bytes | 253,056 bytes | ⚠️ 1,432 bytes |
| __.SYMDEF | 4,692,184 bytes | 4,693,400 bytes | ⚠️ 1,216 bytes |

### Verification

The bug was not reproducible on the fluent demo app since `NavigationControllerDemoController` doesn't have a page implementation that allows to move in and out of a file.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1613)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1613)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -71,17 +71,11 @@ class ShyHeaderController: UIViewController {
 
     init(contentViewController: UIViewController, containingView: UIView?) {
         self.contentViewController = contentViewController
-        shyHeaderView.accessoryView = contentViewController.navigationItem.accessoryView
-        shyHeaderView.navigationBarShadow = contentViewController.navigationItem.navigationBarShadow
-
         self.containingView = containingView
 
         super.init(nibName: nil, bundle: nil)
 
-        shyHeaderView.parentController = self
-        shyHeaderView.maxHeightChanged = { [weak self] in
-            self?.updatePadding()
-        }
+        setupShyHeaderView()
 
         loadViewIfNeeded()
         addChild(contentViewController)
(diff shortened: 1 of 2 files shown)


edjamesmsft (approved):

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -210,6 +204,16 @@ class ShyHeaderController: UIViewController {
         view.bringSubviewToFront(paddingView)
     }
 
+    private func setupShyHeaderView() {
```
harrieshin, line 207: why is this a neccesary chnage?

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -210,6 +204,16 @@ class ShyHeaderController: UIViewController {
...
+        shyHeaderView.accessoryView = contentViewController.navigationItem.accessoryView
+        shyHeaderView.navigationBarShadow = contentViewController.navigationItem.navigationBarShadow
+        shyHeaderView.paddingView = paddingView
```
harrieshin, line 210: it feels like all the paddingView logic should be in shyheaderView directly and not in shyheadercontroller? if the padding view is the a subview of shyheaderview? the ownership of the view seems really bad. I think maybe we need to have different solution for 0.14 release vs 0.13 release.

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
laminesm: Yep, it's not I was just cleaning up a bit by putting all the shy header setup in a function. Should I revert?

PR #1602, opened 2023-02-24
Tokenize Label

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
Finish tokenization of Label control

Binary change:
Total increase: 81,760 bytes
Total decrease: -5,000 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 29,231,968 bytes | 29,308,728 bytes | 🛑 76,760 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| LabelTokenSet.o | 0 bytes | 60,136 bytes | 🛑 60,136 bytes |
| __.SYMDEF | 4,529,576 bytes | 4,540,048 bytes | ⚠️ 10,472 bytes |
| ShimmerLinesView.o | 232,736 bytes | 237,432 bytes | ⚠️ 4,696 bytes |
| ActionsCell.o | 218,616 bytes | 222,024 bytes | ⚠️ 3,408 bytes |
| BadgeLabelButton.o | 847,352 bytes | 848,904 bytes | ⚠️ 1,552 bytes |
| BadgeView.o | 335,840 bytes | 337,336 bytes | ⚠️ 1,496 bytes |
| Label.o | 109,920 bytes | 104,920 bytes | 🎉 -5,000 bytes |
</details>

### Verification

(how the change was tested, including both manual and automated tests)

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| IMG_0095 | IMG_0090 |
| IMG_0096 | IMG_0091 |
| N/A | IMG_0098 |
| N/A | IMG_0099 |
</details>

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [x] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [x] Objective-C exposure (provide it only if needed)
[...]

PR #1629, opened 2023-03-07
Make Label font overridable

jeaniehuynh:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes
#1602 introduced a regression where Label fonts weren't overriding default tokens unless you explicitly provided it in the init. Let's fix this by making font overridable.

Binary change:
Total increase: 6,128 bytes
Total decrease: 0 bytes
| Total | 29,888,720 bytes | 29,894,848 bytes | ⚠️ 6,128 bytes |
<details>

| Label.o | 109,768 bytes | 113,584 bytes | ⚠️ 3,816 bytes |
| __.SYMDEF | 4,617,424 bytes | 4,619,136 bytes | ⚠️ 1,712 bytes |
| BadgeLabelButton.o | 866,584 bytes | 867,184 bytes | ⚠️ 600 bytes |
</details>

<details>

| IMG_0101 2 | IMG_0102 2 |
| IMG_0104 | IMG_0103 |
</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Label/Label.swift
+++ b/ios/FluentUI/Label/Label.swift
@@ -19,6 +19,7 @@ open class Label: UILabel, TokenizedControlInternal {
     }
     @objc open var style: AliasTokens.TypographyTokens = .body1 {
         didSet {
+            _font = nil
             updateFont()
         }
     }
@@ -37,6 +38,13 @@ open class Label: UILabel, TokenizedControlInternal {
         didSet {
             _textColor = textColor
             updateTextColor()
+        }
+    }
+
+    open override var font: UIFont! {
+        didSet {
+            _font = font
+            updateFont()
         }
     }
 
@@ -76,8 +84,9 @@ open class Label: UILabel, TokenizedControlInternal {
     }
 
     private func initialize() {
-        // textColor is assigned in super.init to a default value and so we need to reset our cache afterwards
+        // textColor and font are assigned in super.init to a default value and so we need to reset our cache afterwards
         _textColor = nil
+        _font = nil
 
         updateFont()
         updateTextColor()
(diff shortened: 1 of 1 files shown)


amgleitman (approved):

ios/FluentUI/Label/Label.swift
```diff
@@ -137,6 +146,7 @@ open class Label: UILabel, TokenizedControlInternal {
     }
 
     private var _textColor: UIColor?
+    private var _font: UIFont?
```
huwilkes, line 149: Is the issue that we weren't setting the tokens correctly where it's being used by other controls? Adding a new private var seems odd.

jeaniehuynh: Yeah, so if other controls wanted to override the font token, they would need to pass in the style in the Label init, but that would require a lot of changes in both our own controls and consumers directly using Label. Instead, we follow what text color is doing and override the font property of the UILabel.

harrieshin: this would affect all attributedstring usuage as well? right?

ios/FluentUI/Label/Label.swift
```diff
@@ -40,6 +41,13 @@ open class Label: UILabel, TokenizedControlInternal {
         }
     }
 
+    open override var font: UIFont! {
+        didSet {
+            _font = font
```
harrieshin, line 46: i am confused why we need ivar here when probably app has its own _font?

ios/FluentUI/Label/Label.swift
harrieshin: @sophialee0416  this would be problematic to your tableviewcell attributedstring largertext flow right?

jeaniehuynh: Attributed strings were still able to override the font as expected:
IMG_0105

sophialee0416: Yea so I can see that we're following the same pattern for font as we are for the textColor in this Label component here, which to me looks like it should be a safe fix. I remember we weren't super happy with the way textColor/font is handled here in general but not sure if that problem should be addressed in this PR

ios/FluentUI/Label/Label.swift
jeaniehuynh: By "app has its own _font," do you mean UILabel's font (i.e. super.font)? We need the ivar so that we know when the token is overridden by a different font. The UILabel's font can't be set to nil (it can, but it really just resets it to the default font which isn't actually nil)

sophialee0416 (approved):

ios/FluentUI/Label/Label.swift
harrieshin: sorry typo. Apple will have _font most likely for their font vairable in uilable. didSet shouldn't be actually setting the _font

jeaniehuynh: Hm, I don't see a native _font in UILabel

PR #1635, opened 2023-03-09
Use the right address sanitizer flag

Saadnajmi:
### Platforms Impacted
- [x] iOS
- [x] macOS

### Description of changes

Binary change: Working on it...

Based on these links, I think we were using the wrong flag before :S 
- https://xcodebuildsettings.com/#clang_address_sanitizer (exists)
- https://xcodebuildsettings.com/#enable_address_sanitizer (doesn't exist)
- https://twitter.com/steipete/status/775089634233769984 (Seems to confirm that one `CLANG_ADDREESS_SANITIZER` is the right one)

### Verification

CI should pass

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1635)


Saadnajmi:
Xcode 8 release notes seem to conflict: https://developer.apple.com/library/archive/releasenotes/DeveloperTools/RN-Xcode/Chapters/Introduction.html#//apple_ref/doc/uid/TP40001051-CH1-SW876

> The build setting ENABLE_ADDRESS_SANITIZER controls whether the address sanitizer is enabled across both the clang and swiftc compilers. Previously only CLANG_USE_ADDRESS_SANITIZER existed; that setting is now subordinate to ENABLE_ADDRESS_SANITIZER and should not be used by itself. If a project was previously using CLANG_USE_ADDRESS_SANITIZER as a build setting to enable the address sanitizer separate from the scheme option, then ENABLE_ADDRESS_SANITIZER should now be used instead. (24912445)

Though, I'm passing in `CLANG_ADDRESS_SANITIZER`, not `CLANG_USE_ADDRESS_SANITIZER`. Will keep looking around...

PR #1640, opened 2023-03-10
Realign Drawer to fluent 2 design 

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Tokenized drawer on the tokens branch needed to be realigned to match design and better fit our fluent 2 tokenization philosophy.
- Padding and shadow offset tokens have been moved to a DrawerTokenSet extension to make them not overridable.
- The objc colors now pull the colors from the DrawerTokenSet instead of accessing the tokens from a passed in FluentTheme
- DrawerShadowView now conforms to Shadowable to use the fluent 2 shadows API. Unused functions animate(withDuration duration:, animations:) and shadowOffset(for shadowDirection:, isFirst:) have been removed.
- ShadowInfo's initializeShadowLayer function has been changed to public since it's now used in DrawerShadowView

Note on DrawerShadowView

The way shadows are applied on the drawer is quite messy and presents a few issues. 
- DrawerShadowView gets an owner view from DrawerPresentationController and observes its bounds and frame. Every time either changes, DrawerShadowView updates its frame and redraws the shadow paths. Therefore, the updateFrame function gets called many many times whenever we open or close a drawer. 
- Also, for some obscure reason, applying the keyShadow to a vertical drawer seems to get clipped and makes the drawer look as if its shadow doesn't have rounded corners. I couldn't find a quick workaround so I left the shadow logic as is meaning horizontal drawer gets both key and ambient shadows whereas vertical drawer gets only the ambient shadow applied.
Screenshot 2023-03-10 at 1 08 08 AM

Dealing with these DrawerShadowView issues is a bit out of scope for this tokens -> main merge but I think we should be addressing them as soon as we can after the merge. A more efficient way to apply shadows would be to do what BottomSheet does and embed the drawer's content on a shadow view. This would remove the need for having observers for frame changes, allow us to have both key and ambient shadows on vertical drawer and also get rid of DrawerShadowView which could save us some binary size.

### Binary change

Total increase: 48 bytes
Total decrease: -38,016 bytes
| Total | 33,687,560 bytes | 33,649,592 bytes | 🎉 -37,968 bytes |
<details>

| BadgeLabelButton.o | 865,056 bytes | 865,104 bytes | ⚠️ 48 bytes |
| PopupMenuController.o | 366,872 bytes | 366,832 bytes | 🎉 -40 bytes |
| ShadowInfo.o | 54,704 bytes | 54,216 bytes | 🎉 -488 bytes |
| DrawerController.o | 530,552 bytes | 528,048 bytes | 🎉 -2,504 bytes |
| DrawerTokenSet.o | 59,208 bytes | 53,456 bytes | 🎉 -5,752 bytes |
| __.SYMDEF | 5,282,152 bytes | 5,275,384 bytes | 🎉 -6,768 bytes |
| DrawerShadowView.o | 88,664 bytes | 78,152 bytes | 🎉 -10,512 bytes |
| DrawerPresentationController.o | 275,040 bytes | 263,088 bytes | 🎉 -11,952 bytes |
</details>

### Verification

<details>

| before_bot | after_bot |
| before_bot_clear | after_bot_clear |
| before_bot_clear_dark | after_bot_clear_dark |
| before_bot_dark | after_bot_dark |
| before_left | after_left |
| before_left_dark | after_left_dark |
| before_right | after_right |
| before_right_dark | after_right_dark |
| before_top | after_top |
| before_top_clear | after_top_clear |
| before_top_clear_dark | after_top_clear_dark |
| before_top_dark | after_top_dark |
| before_anim | after_anim |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1640)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -99,14 +99,12 @@ public protocol DrawerControllerDelegate: AnyObject {
 
 @objc(MSFDrawerController)
 open class DrawerController: UIViewController, TokenizedControlInternal {
-    @objc public static func drawerBackground(fluentTheme: FluentTheme) -> UIColor {
-        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background2].light,
-                                                  dark: fluentTheme.aliasTokens.colors[.background2].dark))
+    /// DrawerController colors with obj-c support
+    @objc public static var drawerBackgroundColor: UIColor {
+        UIColor(dynamicColor: DrawerTokenSet()[.drawerContentBackground].dynamicColor)
     }
-
-    @objc public static func popoverBackground(fluentTheme: FluentTheme) -> UIColor {
-        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background4].light,
-                                                  dark: fluentTheme.aliasTokens.colors[.background4].dark))
+    @objc public static var popoverBackgroundColor: UIColor {
+        UIColor(dynamicColor: DrawerTokenSet()[.popoverContentBackground].dynamicColor)
     }
 
     private struct Constants {
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -391,7 +391,7 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
 
     /// Shadow is required if background is transparent
     var shadowOffset: CGFloat {
-        return presentationBackground == .none ? tokenSet[.shadowOffset].float : 0
+        return presentationBackground == .none ? DrawerTokenSet.shadowOffset : 0
     }
 
     private let sourceView: UIView?
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -451,7 +451,7 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
               return
         }
         tokenSet.update(fluentTheme)
-        updateAppearance()
+        updateBackgroundColor()
     }
 
     /**
@@ -486,7 +486,7 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         tokenSetSink = tokenSet.objectWillChange.sink { [weak self] _ in
             // Values will be updated on the next run loop iteration.
             DispatchQueue.main.async {
-                self?.updateAppearance()
+                self?.updateBackgroundColor()
             }
         }
     }
@@ -545,7 +545,7 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         updateResizingHandleView()
         resizingGestureRecognizer?.isEnabled = false
 
-        updateAppearance()
+        updateBackgroundColor()
     }
 
     open override func viewDidAppear(_ animated: Bool) {
@@ -795,15 +795,6 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
             resizingHandleView?.accessibilityLabel = "Accessibility.Drawer.ResizingHandle.Label.Expand".localized
             resizingHandleView?.accessibilityHint = "Accessibility.Drawer.ResizingHandle.Hint.Expand".localized
         }
-    }
-
-    private func updateAppearance() {
-        updateBackgroundColor()
-
-        guard let presentationController = (presentationController as? DrawerPresentationController) else {
-            return
-        }
-        presentationController.updateApperance()
     }
 
     private func offset(forResizingGesture gesture: UIPanGestureRecognizer) -> CGFloat {
--- a/ios/FluentUI/Drawer/DrawerShadowView.swift
+++ b/ios/FluentUI/Drawer/DrawerShadowView.swift
@@ -48,7 +48,6 @@ class DrawerShadowView: UIView, Shadowable {
         self.drawerTokenSet = tokenSet
         super.init(frame: .zero)
         updateShadow()
-        layer.cornerRadius = DrawerTokenSet.shadowOffset
         backgroundColor = .clear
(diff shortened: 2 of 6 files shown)


ios/FluentUI/Drawer/DrawerTokenSet.swift
```diff
@@ -8,86 +8,61 @@ import UIKit
...
-                               yAmbient: 0)
-                })
-
-            case .backgroundDimmedColor:
-                return .dynamicColor({
-                    DynamicColor(light: ColorValue(r: 0.0, g: 0.0, b: 0.0, a: 0.66),
-                                 dark: ColorValue(r: 0.0, g: 0.0, b: 0.0, a: 0.99))
-                })
+            case .cornerRadius:
+                return .float({ 14 })
```
huwilkes, line 31: Does the design not use a corner radius from global tokens?

laminesm: Yep just fixed it

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -799,15 +797,6 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         }
     }
 
-    private func updateAppearance() {
-        updateBackgroundColor()
-
-        guard let presentationController = (presentationController as? DrawerPresentationController) else {
-            return
-        }
-        presentationController.updateApperance()
```
huwilkes, line 808: Does the presentation controller still get updates?

ios/FluentUI/Drawer/DrawerShadowView.swift
```diff
@@ -5,7 +5,9 @@
 
 import UIKit
 
-class DrawerShadowView: UIView {
+class DrawerShadowView: UIView, Shadowable {
+    var ambientShadow: CALayer?
+    var keyShadow: CALayer?
```
huwilkes, line 10: Could you test shadow overrides? Looks like the overrides we have right now are just for the background color, but we should make sure that the other ones are working, too.

ios/FluentUI/Drawer/DrawerController.swift
laminesm: The updates in this function are to the shadowView, the dimmingView which are already handled separately.

ios/FluentUI/Drawer/DrawerShadowView.swift
laminesm: Yep just tested and shadow overrides work. Do you want me to add the changes as part of this PR?

huwilkes: It'd probably be good to add them to make it easier to test in the future.

huwilkes (approved):

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -99,14 +99,12 @@ public protocol DrawerControllerDelegate: AnyObject {
 
 @objc(MSFDrawerController)
 open class DrawerController: UIViewController, TokenizedControlInternal {
-    @objc public static func drawerBackground(fluentTheme: FluentTheme) -> UIColor {
```
edjamesmsft, line 102: Are there any usages of drawerBackground in our objc callers? If so, might be better to deprecate this API than replace it in one go.

ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -99,14 +99,12 @@ public protocol DrawerControllerDelegate: AnyObject {
...
-        return UIColor(dynamicColor: DynamicColor(light: fluentTheme.aliasTokens.colors[.background2].light,
-                                                  dark: fluentTheme.aliasTokens.colors[.background2].dark))
+    /// DrawerController colors with obj-c support
+    @objc public static var drawerBackgroundColor: UIColor {
+        UIColor(dynamicColor: DrawerTokenSet()[.drawerContentBackground].dynamicColor)
     }
-
-    @objc public static func popoverBackground(fluentTheme: FluentTheme) -> UIColor {
```
edjamesmsft, line 107: Same here - let's try to limit breaking changes if we can

ios/FluentUI/Drawer/DrawerShadowView.swift
```diff
@@ -71,27 +63,20 @@ class DrawerShadowView: UIView {
         owner = nil
     }
 
-    func animate(withDuration duration: TimeInterval, animations: () -> Void) {
```
edjamesmsft, line 74: Was this just not being called anywhere? Is that a mistake? I.e., was the animate() call accidentally removed in a merge somehow? Just making sure we're removing functionality intentionally.

ios/FluentUI/Drawer/DrawerController.swift
laminesm: Yes there are. But this is a change we need to make for tokenized controls.

ios/FluentUI/Drawer/DrawerShadowView.swift
laminesm: It wasn't called anywhere and I couldn't find any calls to it in the file history either. I think whoever wrote the script just used it during development and forgot to remove it.

ios/FluentUI/Drawer/DrawerController.swift
edjamesmsft: Ok sounds good! This should hold off going into main until after we cut 0.14 probably, but that's not a huge deal.

edjamesmsft (approved):

ios/FluentUI/Drawer/DrawerController.swift
laminesm: Yep tokens is getting merged after for sure

PR #1644, opened 2023-03-14
Let the OS do Dynamic Type scaling when possible

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Our previous method of Dynamic Type scaling depended on `UIFontMetrics`. While this is recommended for custom fonts, it doesn't quite give us the same values as a pure native OS would. This change allows us to lean on the OS to scale system fonts as it pleases to keep them exactly the same.

The reason is because `UIFontMetrics` scales fonts by line height, not point size. The new font size is calculated by the following formula, where the "original" font sizes are based on iOS's default [`.large`](https://developer.apple.com/documentation/uikit/uicontentsizecategory/1622934-large) content size category:

`scaledPointSize = round(originalPointSize * scaledLineHeight / originalLineHeight)`

For example, if we resize Fluent's Body 1 text (which lines up with [`.body`](https://developer.apple.com/documentation/uikit/uifont/textstyle/1616682-body) in UIKit) for [3XL](https://developer.apple.com/documentation/uikit/uicontentsizecategory/1623109-extraextraextralarge) and [AX5](https://developer.apple.com/documentation/uikit/uicontentsizecategory/1623086-accessibilityextraextraextralarg) content sizes as defined on [Apple's typography HIG page](https://developer.apple.com/design/human-interface-guidelines/foundations/typography/#:~:text=Dynamic-,Type,-sizes%20(iOS)), we get the following values:

3XL: `round(17 * 29 / 22) = 22`
AX5: `round(17 * 62 / 22) = 48`

With this change, these formulas would apply to old text, but they wouldn't These differ from the OS-curated point sizes of 23 and 53, respectively.

While this may not seem like a big deal, it does prevent Fluent from living perfectly seamlessly with UIKit-based controls. In fact, I found this to be an issue while testing the improved navigation bar against changing content size categories. I originally made this change in the navigation bar's feature branch, but considering how essential typography is, I feel like this is enough of a change to warrant its own PR.

### Binary change

For some weird reason the binaries I generated have both arm64 and x86_64 builds built into them. Either way, here's the data for both:

#### arm64
Total increase: 52,136 bytes
Total decrease: -5,848 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 29,599,336 bytes | 29,645,624 bytes | ⚠️ 46,288 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
[...]

PR #1646, opened 2023-03-14
Navbar appearance tweaks plus font stuff

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR consists of a bunch of minor tweaks and bug fixes that may not necessarily warrant PRs on their own.

Some focus areas here include:
* Dynamic Type compatibility (including #1644)
* Using token values instead of hard-coded constants
* Minor fixes to the NavBar demo page

### Verification

Validated by playing around with the preferred content size as seen below.

<details>

| Example 1 | Example 2 | Example 3 |
|-|-|-|
| Simulator Screen Recording - iPhone 14 Pro - 2023-03-14 at 11 58 19 | Simulator Screen Recording - iPhone 14 Pro - 2023-03-14 at 12 02 16 | Simulator Screen Recording - iPhone 14 Pro - 2023-03-14 at 12 06 41 |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1646)

### DIFF
--- a/ios/FluentUI/Navigation/BadgeLabelButton.swift
+++ b/ios/FluentUI/Navigation/BadgeLabelButton.swift
@@ -33,6 +33,10 @@ class BadgeLabelButton: UIButton {
                                                selector: #selector(badgeValueDidChange),
                                                name: UIBarButtonItem.badgeValueDidChangeNotification,
                                                object: item)
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(contentSizeCategoryDidChange(notification:)),
+                                               name: UIContentSizeCategory.didChangeNotification,
+                                               object: nil)
     }
 
     required init?(coder aDecoder: NSCoder) {
@@ -53,6 +57,9 @@ class BadgeLabelButton: UIButton {
         static let badgeBorderWidth: CGFloat = 2
         static let badgeHorizontalPadding: CGFloat = 10
         static let badgeCornerRadii: CGFloat = 10
+
+        static let maximumContentSizeCategory: UIContentSizeCategory = .extraExtraLarge
+        static let minimumContentSizeCategory: UIContentSizeCategory = .large
     }
 
     private let badgeLabel = BadgeLabel()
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -53,6 +57,9 @@ class BadgeLabelButton: UIButton {
         static let badgeBorderWidth: CGFloat = 2
         static let badgeHorizontalPadding: CGFloat = 10
         static let badgeCornerRadii: CGFloat = 10
+
+        static let maximumContentSizeCategory: UIContentSizeCategory = .extraExtraLarge
+        static let minimumContentSizeCategory: UIContentSizeCategory = .large
```
laminesm, line 62: Where do these values come from?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -627,6 +635,15 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         titleObserver = navigationItem.observe(\UINavigationItem.title) { [unowned self] item, _ in
             self.navigationItemDidUpdate(item)
         }
+        subtitleObserver = navigationItem.observe(\UINavigationItem.subtitle) { [unowned self] item, _ in
+            self.navigationItemDidUpdate(item)
+        }
+        titleAccessoryObserver = navigationItem.observe(\UINavigationItem.titleAccessory) { [unowned self] item, _ in
+            self.navigationItemDidUpdate(item)
+        }
+        titleImageObserver = navigationItem.observe(\UINavigationItem.titleImage) { [unowned self] item, _ in
```
laminesm, line 644: What are these observers used for or is that coming in a different PR?

ios/FluentUI/Navigation/BadgeLabelButton.swift
amgleitman: These were found through observation to be consistent with what UIKit does by default. I'll add a comment for this.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: These were added because they correspond to the new fields we've added to UINavigationItem to be consistent with the other listeners. This was just something on my backlog, and I needed a PR to get them in with.

ios/FluentUI/Navigation/BadgeLabelButton.swift
edjamesmsft: Do we want to be behaving the same as UIKit? What does design say about what these defaults should be?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -483,6 +486,12 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         bottom.priority = .defaultHigh
 
         NSLayoutConstraint.activate([leading, trailing, top, bottom])
+
+        if #available(iOS 15.0, *) {
+            // These are consistent with UIKit's default navigation bar
```
edjamesmsft, line 491: Same question as the other comment - do these values align with what design wants NavBar to look like? Consistency with UIKit is fine, but it's more important for us to be consistent across Fluent 2 designs.

amgleitman: Design hasn't given us any specs for text sizes in any content size categories besides [`.large`](https://developer.apple.com/documentation/uikit/uicontentsizecategory/1622934-large), so keeping consistency with UIKit is the best we can do.

ios/FluentUI/Navigation/BadgeLabelButton.swift
amgleitman: Design hasn't given us any specs for text sizes in any content size categories besides [`.large`](https://developer.apple.com/documentation/uikit/uicontentsizecategory/1622934-large), so keeping consistency with UIKit is the best we can do.

ios/FluentUI/Navigation/NavigationBar.swift
harrieshin: this is completely different direction than what we had before. We usually cap and min based on font size not the size category because most of our padding is fixed and not tokenize. i am not sure we should set min and max category necessarily.

amgleitman: I think this direction is worth pursuing. Peeking into a `UIView`'s [`appliedContentSizeCategoryLimits`](https://developer.apple.com/documentation/uikit/uiview/3750922-appliedcontentsizecategorylimits) property, we can see that UIKit's navigation bar has category caps set accordingly. Plus, this is arguably better than the alternative of specifying a "magic number" of 21pt somewhere.

PR #1672, opened 2023-03-24
Action clean up

huwilkes:
### Description of changes

Our localization pipeline broke because it isn't allowed to run actions from 3rd parties anymore. Updating it to use the GitHub terminal client to try and make a PR while I work on making a GitHub app to run it properly. Based on https://github.com/huwilkes/fluentui-apple/pull/6, this update to the localize action will probably work. Also removed the size comparison action since it's broken, and we maybe shouldn't be using artifacts.
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1672)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1672)


.github/workflows/localize.yml
```diff
@@ -26,12 +26,6 @@ jobs:
...
-      uses: peter-evans/create-pull-request@v3.5.0
-      with:
-        commit-message: Latest localized resource files from Touchdown Build
-        title: '[Localization] Localized Resource Files'
-        body: 'Latest localized resource files from Touchdown Build'
-        branch: tdbuild
-        labels: Automated PR
-        branch-suffix: timestamp
-    
+      run: git checkout -b 'tdbuild`date +%s`'; git push --set-upstream origin huwilkes/PRActions`date +%s`; gh pr create -B main -H tdbuild`date +%s` --title '[Localization] Localized Resource Files' --body  'Latest localized resource files from Touchdown Build' -l 'Automated PR'
```
mischreiber, line 29: Do you intend to be referencing `huwilkes/PRActions` branch?

mischreiber: Also wondering if this script is long enough that it should live in a dedicated file under `scripts/` instead of being inline here?

huwilkes: I sure didn't meant to leave that in there. Fixed, thanks!

huwilkes: I think it's fine inline for now, we can move it to a script if we even need it with the app.

huwilkes: Though I was slightly worried about the output from date changing while it ran, it seems to be fine. Probably good to move it out, though.

harrieshin (approved):

mischreiber (approved):

.github/workflows/localize.yml
huwilkes: script should work according to https://github.com/huwilkes/fluentui-apple/pull/7

PR #1673, opened 2023-03-27
Make objc colors take an optional FluentTheme parameter

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

All our objective-c colors need to take into account the theme of the context calling them in order to return the right colors. The colors take an optional FluentTheme to fetch the colors. If the theme is nil, we use the shared theme as fallback.

### Binary change

Total increase: 6,592 bytes
Total decrease: -17,296 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 33,451,304 bytes | 33,440,600 bytes | 🎉 -10,704 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| NavigationBar.o | 549,800 bytes | 551,928 bytes | ⚠️ 2,128 bytes |
| DrawerController.o | 513,944 bytes | 515,744 bytes | ⚠️ 1,800 bytes |
| __.SYMDEF | 5,234,808 bytes | 5,236,528 bytes | ⚠️ 1,720 bytes |
| Separator.o | 58,280 bytes | 59,224 bytes | ⚠️ 944 bytes |
| TableViewCell.o | 830,888 bytes | 813,592 bytes | 🎉 -17,296 bytes |
</details>

### Verification

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before_avatar | after_avatar |
| before_avatar_group | after_avatar_group |
| before_header_footer | after_header_footer |
| before_modal_view | after_modal_view |
| before_other_cells | after_other_cells |
| before_tvc | after_tvc |

</details>

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)
[...]

PR #1684, opened 2023-03-31
Tokenize NavigationBar

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This is a basic tokenization of NavigationBar.

The scope for this includes tokenizing elements of a basic navigation bar, including title colors and fonts.

As of aa9640c, this change does not yet apply tokenization to the accessory view (e.g., a pill button bar or a search bar) because each component has its own way of representing its style, and that logic doesn't necessarily belong in NavigationBar itself. We're still thinking about the best way to implement this.

### Verification

Verified using the test page. These screenshots show the token overrides in place.

One possible concern is that the light-mode override colors look weird on a brand background, but this is more of a pedagogical example to show that we can indeed override these tokens.

<details>

| Example 1 | Example 2 | Example 3 | Example 4 |
|-|-|-|-|
| Simulator Screen Shot - iPhone 14 Pro - 2023-03-31 at 11 44 03 | Simulator Screen Shot - iPhone 14 Pro - 2023-03-31 at 11 44 16 | Simulator Screen Shot - iPhone 14 Pro - 2023-03-31 at 11 44 50 | Simulator Screen Shot - iPhone 14 Pro - 2023-03-31 at 11 45 03 |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1684)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -107,43 +107,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
...
-                return defaultColor
-            case .system:
-                return fluentTheme.color(.background3)
-            case .custom:
-                return customColor ?? defaultColor
-            }
-        }
     }
 
+    @available(*, deprecated, renamed: "backgroundColor(forStyle:theme:)", message: "NavigationBar has been tokenized. Use the replacement method, passing in .system for the style.")
```
edjamesmsft, line 112: Unfortunately, since control tokens cannot be used in Objective C, and we don't have a plan for when they will be available, I don't think we can deprecate this API yet.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -301,7 +271,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         }
     }
 
-    var titleView = LargeTitleView() {
+    var titleView = AvatarTitleView() {
```
edjamesmsft, line 274: This change seems unrelated to tokenizing. Should this be done in a separate PR? Additionally, what is the rationale for moving from LargeTitleView to AvatarTitleView?

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -53,6 +53,22 @@ public class ControlTokenSet<T: TokenSetKey>: ObservableObject {
...
     }
 
+    /// Convenience method to override multiple tokens from another token set.
+    ///
+    /// This is useful if `otherTokenSet` belongs to a parent control and `self` belongs to a child control.
+    ///
+    /// - Parameter otherTokenSet: The token set we will be pulling values from.
+    /// - Parameter mapping: A `Dictionary` that maps our own tokens that we wish to override with
+    /// their corresponding tokens in `otherTokenSet`.
+    public func setOverrides<U>(from otherTokenSet: ControlTokenSet<U>, mapping: [T: U]) {
```
laminesm, line 63: Is this a safe addition? With our current token philosophy we want to add subcontrol tokens in the parent control and any overrides should happen from the parent. Wouldn't this open us up to potentially strange behaviors in situations where overrides occur from both parent and child?

ios/FluentUI/Navigation/NavigationBarTokenSet.swift
```diff
@@ -0,0 +1,89 @@
...
+                return .uiColor {
+                    switch style() {
+                    case .primary, .default, .custom:
+                        return UIColor(light: theme.color(.brandBackground1).light,
+                                       dark: theme.color(.background3).dark)
+                    case .system:
+                        return theme.color(.background3)
+                    }
+                }
+            case .buttonTintColor, .subtitleColor: // By default, these return the same color
```
edjamesmsft, line 46: Nit: I don't think this comment is necessary.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -107,43 +107,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         case primary
```
laminesm, line 107: Shouldn't `NavigationBar.Style` live in `NavigationBarTokenSet`?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -107,43 +107,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
...
-            case .system:
-                return fluentTheme.color(.background3)
-            case .custom:
-                return customColor ?? defaultColor
-            }
-        }
     }
 
+    @available(*, deprecated, renamed: "backgroundColor(forStyle:theme:)", message: "NavigationBar has been tokenized. Use the replacement method, passing in .system for the style.")
     @objc public static func navigationBarBackgroundColor(fluentTheme: FluentTheme) -> UIColor {
```
laminesm, line 113: Could you make this an an optional `FluentTheme` and use the shared theme as fallback (#1673 for reference)

ios/FluentUI/Navigation/Views/AvatarTitleView.swift
```diff
@@ -5,21 +5,19 @@
 
 import UIKit
 
-// MARK: LargeTitleView
```
edjamesmsft, line 8: Ohhh I see now. Given the scope of this change, it might be good to have broken this into two separate PRs. Easier to review, easier to roll back if something goes sideways.

ios/FluentUI/Navigation/NavigationBar.swift
laminesm, line 112: Could you please clarify what this message is for? This function is primarily for our objc callers and seems relatively unchanged.

ios/FluentUI/Navigation/NavigationBarTokenSet.swift
laminesm, line 46: nit: remove redundant comment

ios/FluentUI/Navigation/Views/AvatarTitleView.swift
```diff
@@ -158,18 +147,32 @@ class LargeTitleView: UIView, TwoLineTitleViewDelegate {
         setupAccessibility()
         twoLineTitleView.delegate = self
 
-        NotificationCenter.default.addObserver(self,
-                                               selector: #selector(themeDidChange),
-                                               name: .didChangeTheme,
-                                               object: nil)
+        tokenSet.registerOnUpdate(for: self) { [weak self] in
+            self?.updateAppearance()
```
edjamesmsft, line 151: Above, you setTitleColor after updateAppearance, but here you don't setTitleColor. That to me means either setTitleColor should be moved into updateAppearance, or also called here (preferably the former).

ios/FluentUI/Navigation/NavigationBarTokenSet.swift
laminesm: Just saw this was mentioned already by @edjamesmsft so I'm gonna resolve this

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: `LargeTitleView` is a bit of a misnomer now since it can also host a `TwoLineTitleView`. The defining characteristics of this view are that (a) it can show an avatar, and (b) its titles are leading-aligned. I chose the first one since the avatar is definitely an attention-grabber in practice.

It's tangentially related since we're tokenizing this component, so I think now's a fine time to rename it.

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -254,8 +254,8 @@ class ShyHeaderController: UIViewController {
             }
         }
 
-        // if the originator is a LargeTitleView, make sure it belongs to this heirarchy
```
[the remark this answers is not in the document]
amgleitman: See [this comment](https://github.com/microsoft/fluentui-apple/pull/1684#discussion_r1154847084).

ios/FluentUI/Navigation/NavigationBarTokenSet.swift
amgleitman: I don't think it hurts to keep it.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Since this method is publicly exposed, it's possible that a client might be calling this. This is for their benefit so they explicitly know we're deprecating this instead of just getting faced with a random compiler error.

amgleitman: Why can't we deprecate this? Even if we can't use control tokens in Objective-C, `NavigationBar.Style` and `FluentTheme` should be fair game.

laminesm: I guess I just don't understand why the deprecation message is placed on the `navigationBarBackgroundColor` function and not the `backgroundColor(forStyle:theme:)` function.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Ooh, good idea! Done with a7a741b.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: `navigationBarBackgroundColor` is the old function, while `backgroundColor(forStyle:theme:)` is the new one.

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
amgleitman: Do we have a preferred way to pass tokens to child controls? I don't see any examples of this currently in this repo.

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: I see. I don't think we should be attempting to replace this function with `backgroundColor(forStyle:theme:)`. We always want to simplify our APIs for our objc clients, and this new function does the opposite by adding a new parameter to the call.

amgleitman: I thought that we wanted to move all functionality regarding appearance to `NavigationBarTokenSet`. Is this not the case anymore?

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: The `Style` enum doesn't contain any information inherently about the token values themselves, so I think it's fine where it is now. Let the token set handle these translations.

ios/FluentUI/Navigation/Views/AvatarTitleView.swift
amgleitman: Upon closer inspection, `updateAppearance` already calls `setTitleColor`, so we can take a third option of removing the extra call.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Previously, the API would assume that we're using a `NavigationBar` with the `.system` style, which is not a very good assumption to make IMHO. At least we tell our clients to make sure that they absolutely want this style.

edjamesmsft: We want to avoid breaking API changes wherever possible. Is this a necessary change? Is it necessary right now?

ios/FluentUI/Navigation/NavigationBar.swift
edjamesmsft: Just to clarify - is it DIRECTLY related, or TANGENTALLY related? If it's tangental, as in, not very related, I really would rather it be in a separate PR.

In general, keeping PRs small and targeted is a good choice. Especially with a PR as big as this one has become, it might have been good to do this rename separately, after the large functionality change.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: Leaving it in is an unnecessary accumulation of debt. I'd rather nip it in the bud now than put it off for later, especially when the resulting cost for clients is very low (we're LITERALLY telling them how to update to the new API).

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: It's relevant here because of the addition of `AvatarTitleViewTokenSet`, and as per my previous comment, `LargeTitleViewTokenSet` is an unsuitable name.

If you absolutely insist on this being a separate PR, see #1686, but reverting it here just to immediately change it back after this PR gets merged into the feature branch is just plain silly.

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: The point of this function was to expose only the system style color of the nav bar. I was not aware of any situation in which we needed to access the colors from the other style while updating OMR to fluent 2. Also not sure how this counts as debt...

ios/FluentUI/Core/Theme/Tokens/ControlTokenSet.swift
amgleitman: Using [this bit of code](https://github.com/microsoft/fluentui-apple/blob/b7e84b4137ab808239940b2b37892285ca5f41e9/ios/FluentUI/Tab%20Bar/SideTabBar.swift#L381) as a reference point, I think what we really want here is to replace the RHS with `otherTokenSet.overrideValue(forToken: otherToken)`.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -280,15 +278,18 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
...
-        let subtitleIsNilOrEmpty = subtitle?.isEmpty ?? true
-        titleButtonLabel.maxFontSize = subtitleIsNilOrEmpty ? 0 : fluentTheme.aliasTokens.typography[.body1Strong].size
-
         setupButton(titleButton, label: titleButtonLabel, trailingImageView: titleButtonTrailingImageView, text: title, interactive: interactivePart.contains(.title), accessoryType: accessoryType)
         // Check for strict equality for the subtitle button's interactivity.
         // If the whole area is active, we'll stretch the title button to adjust the hit area
         // while still only keeping one button active from an accessibility standpoint.
         setupButton(subtitleButton, label: subtitleButtonLabel, trailingImageView: subtitleButtonImageView, text: subtitle, interactive: interactivePart == .subtitle, accessoryType: accessoryType)
 
+        if #available(iOS 15.0, *) {
```
mischreiber, line 287: This isn't necessary - the minimum SDK is already moved to 15.0

PR #1695, opened 2023-04-10
[Button] Add Inactive/Background Window State to all Button Styles

amoggo:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
**PROBLEM**
Fluent Buttons don't have any support for inactive/backgrounded window states and they appear just as prominent when the Window housing the Button view is inactive. This is especially prominent in the Primary (Accent colored background) Style buttons where the expectation is that it would get some treatment to indicate backgrounded-ness. 

**FIX**
- MacOS Ventura strips all the Button control's colors when its native app window is in the background. 
- To better align with this, changes have been made to transition all the button styles to its corresponding Inactive (Window Backgrounded) State (determined by closely approximating what that particular style would look like when backgrounded whilst aligning with the MacOS behavior viz stripping away all the colors)
-  Also added support to revert the Button back to its latest set Style, when the Window is made active again. 
- Only affects the Button's _Rest_ state. _Disabled_ and _Pressed_ states remain unchanged. 

### Verification

Test App - BEFORE: Window Backgrounded state
Screenshot 2023-04-10 at 13 01 44

Test App - AFTER: Window Backgrounded state
Screenshot 2023-04-11 at 12 42 39

### Pull request checklist
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1695)


anandrajeswaran:
Edge-case of all edge-cases, but what if someone programmatically changes the buttonStyle while the window is backgrounded. When the window becomes main we would then restore the previous style and ignore the new style?

That's kinda contrived, but I think a good illustration of why we should tend away from stateful patterns like this. Can we internally instead adjust the sites where we use properties to also consider the state of whether the window is inactive? That pattern lends itself more to the react-style `render` pattern and the SwiftUI `body` pattern which are all the rage. Store any state as discretely as possible in the model and then funnel usage of that state into a singular path so a change to any of those components of its states just involves a simple "invalidate" to get code to go down that singular path again.

huwilkes:
If we do add this, could we update the [documentation](https://github.com/microsoft/fluentui-apple/blob/main/macos/docs/Controls/Button.md) to include the new behavior?

huwilkes:
> Edge-case of all edge-cases, but what if someone programmatically changes the buttonStyle while the window is backgrounded. When the window becomes main we would then restore the previous style and ignore the new style?
> 
> That's kinda contrived, but I think a good illustration of why we should tend away from stateful patterns like this. Can we internally instead adjust the sites where we use properties to also consider the state of whether the window is inactive? That pattern lends itself more to the react-style `render` pattern and the SwiftUI `body` pattern which are all the rage. Store any state as discretely as possible in the model and then funnel usage of that state into a singular path so a change to any of those components of its states just involves a simple "invalidate" to get code to go down that singular path again.

We could probably just expose an optional public `inactiveWindowStyle`, and let the client choose which style they want, and default to something if it's nil. We have the `linkedPrimary` var, so I don't think it'd be too out there to expose this one, also.

macos/FluentUI/Button/Button.swift
```diff
@@ -253,6 +253,43 @@ open class Button: NSButton {
...
+			guard oldValue != isWindowInactive else {
+				return
+			}
+			// Re-compute the Button's color values for the latest Window State, and re-render it
+			setColorValues(forStyle: style, accentColor: accentColor, isWindowInactive: isWindowInactive ?? false)
+			needsDisplay = true
+		}
+	}
+
+	public override func viewDidMoveToWindow() {
```
anandrajeswaran, line 269: Should we update the value of isWindowInactive based on whether the window we're moving to is main?

macos/FluentUI/Button/Button.swift
```diff
@@ -253,6 +253,43 @@ open class Button: NSButton {
...
+
+		// Hook in Notification Handles to capture the Window's active and inactive states
+		NotificationCenter.default.addObserver(forName: NSWindow.didResignMainNotification,
+											   object: nil,
+											   queue: nil) {[weak self] _ in
+			guard let strongSelf = self else {
+				return
+			}
+			strongSelf.isWindowInactive = true
+		}
```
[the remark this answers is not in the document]
anandrajeswaran: Same below

macos/FluentUI/Button/Button.swift
```diff
@@ -253,6 +253,43 @@ open class Button: NSButton {
...
+			needsDisplay = true
+		}
+	}
+
+	public override func viewDidMoveToWindow() {
+		super.viewDidMoveToWindow()
+
+		// Hook in Notification Handles to capture the Window's active and inactive states
+		NotificationCenter.default.addObserver(forName: NSWindow.didResignMainNotification,
+											   object: nil,
```
anandrajeswaran, line 274: We ned to be listening to the actual window we're moving to. Otherwise we'll react to notifications for all windows

anandrajeswaran (changes requested):

macos/FluentUI/Button/Button.swift
```diff
@@ -369,9 +406,12 @@ open class Button: NSButton {
 			guard oldValue != accentColor else {
 				return
 			}
-			// Recompute relevant state-specific colors appropriate to the style
-			setColorValues(forStyle: style, accentColor: accentColor)
-			needsDisplay = true
+
+			if (isWindowInactive == nil || !(isWindowInactive ?? false)) {
```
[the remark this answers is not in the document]
amoggo: This was in response to your edge case of the edge cases where the style can be changed when the Window is in the inactive state and we might miss out on recording it

macos/FluentUI/Button/Button.swift
amoggo: Sorry clarify? because NSWindow.didBecomeMainNotification handler already takes care of that below? by we're moving to instead of "we've moved to", are you talking about viewWillMoveToWindow?

macos/FluentUI/Button/Button.swift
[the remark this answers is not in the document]
anandrajeswaran: Now that we don't unnecessarily modify the style, we don't have that bug so don't need this extra complexity/stat. There's nothing wrong with letting a setStyle cause the button to re-render when the window is inactive. That would happen before your change and is ultimately orthogonal to the change/

macos/FluentUI/Button/Button.swift
anandrajeswaran: We register for notifications so we do something when the window's main-ness changes. What if the window is already main when the button is added?

macos/FluentUI/Button/Button.swift
```diff
@@ -254,6 +254,39 @@ open class Button: NSButton {
 		path.fill()
 	}
 
+	public override func viewDidMoveToWindow() {
+		super.viewDidMoveToWindow()
+		isWindowInactive = !(window?.isMainWindow ?? false)
+
+		var resignMainWindowObserver: NSObjectProtocol?
+		var becomeMainWindowObserver: NSObjectProtocol?
+
```
anandrajeswaran, line 263: These are func local variables so go out of scope when the func completes. That means the `else` block below won't ever do anything

anandrajeswaran (changes requested):

brentpe (approved):

macos/FluentUI/Button/Button.swift
```diff
@@ -254,6 +254,39 @@ open class Button: NSButton {
...
+			// previously loaded into. Good opportunity to remove the Notification Observers here.
+			if let resignMainWindowObserver = resignMainWindowObserver {
+				NotificationCenter.default.removeObserver(resignMainWindowObserver)
+			}
+			resignMainWindowObserver = nil
+
+			if let becomeMainWindowObserver = becomeMainWindowObserver {
+				NotificationCenter.default.removeObserver(becomeMainWindowObserver)
+			}
+			becomeMainWindowObserver = nil
```
anandrajeswaran, line 285: We also need to remove observer from the old window if we're moving directly into a new window? So maybe the structure of viewDidMoveToWindow is:
if let observer1 {
removeObserver
}

if let observer2 {
removeObserver
}

if window != nil {
addObserver
}

brentpe (approved):

anandrajeswaran (approved):

Issue #1701, opened 2023-04-16
Swift Package Manager builds for macOS are broken


markavitale:
### Environment Information

- **Platform**:
  - [ ] iOS
  - [x] macOS
- **Package version(s)**: Swift Package Manager - works in release version 0.12, broken in versions 0.13-0.16
- **Xcode and OS versions**: macOS 13.2.1, Xcode 14.3

### Please provide a reproduction of the bug:

cd into the fluentui-apple directory and run `swift build`

#### Actual behavior:
```
% swift build
warning: 'fluentui-apple': found 1 file(s) which are unhandled; explicitly declare them as resources or exclude from the target
    /.../fluentui-apple/ios/FluentUI/TextField/TextField.resources.xcfilelist
Building for debugging...
error: emit-module command failed with exit code 1 (use -v to see invocation)
/.../fluentui-apple/macos/FluentUI/Core/FluentUIResources.swift:8:8: error: no such module 'FluentUIResources'
import FluentUIResources
       ^
/.../fluentui-apple/macos/FluentUI/Core/FluentUIResources.swift:8:8: error: no such module 'FluentUIResources'
import FluentUIResources
       ^
/.../fluentui-apple/macos/FluentUI/Core/FluentUIResources.swift:8:8: error: no such module 'FluentUIResources'
import FluentUIResources
       ^
[3/4] Compiling FluentUI_macos FluentUIResources.swift
```

then after fixing that
```
 % swift build
Building for debugging...
error: emit-module command failed with exit code 1 (use -v to see invocation)
/.../fluentui-apple/macos/FluentUI/Notification/NotificationBarView.swift:189:51: error: cannot find 'FluentUI' in scope
                static let accentBackgroundColor: Color = Color(FluentUI.Colors.primaryTint30)
                                                                ^~~~~~~~
/.../fluentui-apple/macos/FluentUI/Notification/NotificationBarView.swift:190:51: error: cannot find 'FluentUI' in scope
                static let accentForegroundColor: Color = Color(FluentUI.Colors.primaryShade20)
                                                                ^~~~~~~~
/.../fluentui-apple/macos/FluentUI/Notification/NotificationBarView.swift:192:51: error: cannot find 'FluentUI' in scope
                static let subtleForegroundColor: Color = Color(FluentUI.Colors.primary)
                                                                ^~~~~~~~
[...]

PR #1702, opened 2023-04-16
Fix Swift Package Manager builds on macOS

markavitale:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Issue describe in more detail in #1701 

Pull request #1544 removed the FluentUIResources module but didn't remove the code that conditionally imports it when compiling via Swift Package Manager. Fix that by just removing the import as it's no longer relevant.

When macOS Notification bar support was added in #1590, it references the FluentUI namespace directly. I'm not actually sure why this works in Xcode builds but not SPM builds, but removing the FluentUI namespace works to fix both.

While here, fix up a warning that `TextField/TextField.resources.xcfilelist` should be excluded from our SPM build on iOS

### Binary change

macOS changes don't need binary size comparisons.

### Verification

Ensure `swift build` works as expected with no warnings or errors. Sanity check that the notification bar view test app looks the same before and after, paying close attention to colors.

<details>

| image | image |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1702)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -41,6 +41,7 @@ let package = Package(
                 "Other Cells/OtherCells.resources.xcfilelist",
                 "Resources/Localization/CultureMapping.json",
                 "Table View/TableView.resources.xcfilelist",
+                "TextField/TextField.resources.xcfilelist",
                 "Tooltip/Tooltip.resources.xcfilelist",
                 "TwoLineTitleView/TwoLineTitleView.resources.xcfilelist",
             ]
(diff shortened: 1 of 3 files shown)


markavitale:
@microsoft-github-policy-service agree

huwilkes (approved):

PR #1711, opened 2023-04-20
Add NavigationBar gradient style

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This PR adds a new gradient variant for the `NavigationBar`. 

**Implementation**:
- a new `gradient` style is added 
- There are 2 new public objc variables: `gradient` and `gradientMask` which are both optional `CAGradientLayer`s. `gradient` must be set when using the `gradient` style of the nav bar. `gradientMask` is actually optional since it's not necessarily needed for axial gradients, though it will most likely have to be provided for conic gradients. 
- If set, the mask will be applied on the main gradient, which will be rendered into a UIImage that is set as the background image of the nav bar's standard appearance.
- The background view, the native title and right bar buttons are set to clear colors or hidden in order to not inhibit the gradient.

**Design tweaks**
The original design from the partner request had a gradient for the shy header when searching. However, since the nav bar gets hidden and pushes up the shy header controller, we can only achieve that by applying the gradient to the shy header which led to this clunky animation below. After discussions with partner designers, they made the decision to hide the gradient when searching. But if time allows, we should look into whether we can fix the hierarchy to accommodate for the gradient on search bar post fork.
<details>
<summary> Clunky shy header animation </summary>

shy

</details>

### Binary change

Total increase: 30,368 bytes
Total decrease: 0 bytes
| Total | 30,120,816 bytes | 30,151,184 bytes | ⚠️ 30,368 bytes |
<details>

| NavigationBar.o | 554,744 bytes | 579,136 bytes | ⚠️ 24,392 bytes |
| __.SYMDEF | 4,717,736 bytes | 4,722,312 bytes | ⚠️ 4,576 bytes |
| NavigationController.o | 163,560 bytes | 164,304 bytes | ⚠️ 744 bytes |
| FocusRingView.o | 799,488 bytes | 800,144 bytes | ⚠️ 656 bytes |
</details>

### Verification

<details>

| light_large_title | dark_large_title |
| light_search_bar | dark_search_bar |
| light_pill | dark_pill |
| light_collapsible | dark_collapsible |
| light_shy_search | dark_shy_search |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1711)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -179,12 +179,10 @@ class NavigationControllerDemoController: DemoController {
             changeStyleContinuously(in: content.navigationItem)
         }
 
-        let controller: NavigationController
-        if style == .gradient {
-            controller = NavigationController(rootViewController: content, gradient: gradient, mask: gradientMask)
-        } else {
-            controller = NavigationController(rootViewController: content)
-        }
+        let controller = NavigationController(rootViewController: content)
+        controller.msfNavigationBar.gradient = gradient
+        controller.msfNavigationBar.gradientMask = gradientMask
+
         if showAvatar {
             controller.msfNavigationBar.personaData = content.personaData
             controller.msfNavigationBar.onAvatarTapped = handleAvatarTapped
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -120,7 +120,7 @@ class NavigationControllerDemoController: DemoController {
                                    updateStylePeriodically: Bool = false) -> NavigationController {
         let content = RootViewController()
         content.navigationItem.usesLargeTitle = useLargeTitle
-        content.navigationItem.navigationBarStyle = style
+        content.navigationItem.navigationBarStyle = .customGradient
         content.navigationItem.navigationBarShadow = showShadow ? .automatic : .alwaysHidden
         content.navigationItem.accessoryView = accessoryView
         content.navigationItem.topAccessoryViewAttributes = NavigationBarTopSearchBarAttributes()
@@ -168,7 +168,7 @@ class NavigationControllerDemoController: DemoController {
                 newStyle = .primary
             case .primary:
                 newStyle = .system
-            case .system:
+            case .system, .customGradient:
                 newStyle = .custom
             }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -175,12 +175,13 @@ class NavigationControllerDemoController: DemoController {
         }
 
         let controller = NavigationController(rootViewController: content)
-        controller.msfNavigationBar.gradient = gradient
-        controller.msfNavigationBar.gradientMask = gradientMask
+        let navigationBar = controller.msfNavigationBar
+        navigationBar.gradient = gradient
+        navigationBar.gradientMask = gradientMask
 
         if showAvatar {
-            controller.msfNavigationBar.personaData = content.personaData
-            controller.msfNavigationBar.onAvatarTapped = handleAvatarTapped
+            navigationBar.personaData = content.personaData
+            navigationBar.onAvatarTapped = handleAvatarTapped
         } else {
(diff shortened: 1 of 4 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -43,6 +49,25 @@ class NavigationControllerDemoController: DemoController {
         container.addArrangedSubview(createButton(title: "Change the style every second", action: #selector(showSearchChangingStyleEverySecond)))
     }
 
+    let gradient: CAGradientLayer = {
+        let redColor = UIColor(colorValue: ColorValue(0xAE7EE1)).withAlphaComponent(0.4).cgColor
```
huwilkes, line 53: Since you want a `CGColor` with an alpha value anyway, might be worth just converting the hex value to the rgb floats and using `CGColor(red: green: blue: alpha:)`

huwilkes:
What's the plan for getting this into main and the other releases between?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -402,6 +421,31 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal {
         }
     }
 
+    private func updateGradient() {
+        if style != .gradient {
+            return
+        }
+
+        guard let gradient = gradient else {
+            assertionFailure("gradient cannot be nil when using the gradient style")
```
huwilkes, line 430: I don't think we want an assertion failure here. This would fail if the user sets the style to gradient, then sets the gradient mask before the gradient. I think just returning should be fine.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -516,18 +560,19 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal {
 
     func updateColors(for navigationItem: UINavigationItem?) {
         let color = navigationItem?.navigationBarColor(fluentTheme: tokenSet.fluentTheme)
+        let shouldHideRegularTitle: Bool = (style == .gradient) && showsLargeTitle
```
joannaquu, line 563: Nit: do we need ()?

mischreiber (changes requested):
I think the code looks good overall, but let's talk offline about our rollout strategy (I don't think I want to check this directly into the 0.14 branch, because it does slightly change the API and could be binary-incompatible in certain scenarios).

amgleitman:
It does seem a bit weird that we have `gradient` and `gradientMask` as separate properties. Can we combine these somehow?

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: Nope just added them to make it slightly easier to read the conditions around the = operators.

laminesm:
> It does seem a bit weird that we have `gradient` and `gradientMask` as separate properties. Can we combine these somehow?

We can combine them in a tuple but that would probably mean partners would have to set/reset both every time they want to change one. 
Another option would be to make the partners apply the mask to the layer themselves and only pass in the main gradient.

mischreiber:
> 

Also worth

mischreiber:
> > It does seem a bit weird that we have `gradient` and `gradientMask` as separate properties. Can we combine these somehow?
> 
> We can combine them in a tuple but that would probably mean partners would have to set/reset both every time they want to change one. Another option would be to make the partners apply the mask to the layer themselves and only pass in the main gradient.

Another point: if we make it a tuple then we lose Objective-C interop, since tuples are Swift-only. We could make a whole new class type for it, but that starts to feel a little heavy. I vote, in no particular order:
- Create a new function for applying a gradient that takes two arguments, though that still has the problem @laminesm mentioned about needing to set both
- Leave it as-is

mischreiber (approved):

laminesm:
> > > It does seem a bit weird that we have `gradient` and `gradientMask` as separate properties. Can we combine these somehow?
> > 
> > 
> > We can combine them in a tuple but that would probably mean partners would have to set/reset both every time they want to change one. Another option would be to make the partners apply the mask to the layer themselves and only pass in the main gradient.
> 
> Another point: if we make it a tuple then we lose Objective-C interop, since tuples are Swift-only. We could make a whole new class type for it, but that starts to feel a little heavy. I vote, in no particular order:
> 
> * Create a new function for applying a gradient that takes two arguments, though that still has the problem @laminesm mentioned about needing to set both
> * Leave it as-is

I'll leave it as is for now since the request is pressing. We can come back and optimize it later on.

PR #1712, opened 2023-04-20
Replace `NavigationBar`'s `usesLargeTitle` with `NavigationBar.TitleStyle`

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Previously, there were two different styles of navigation bar title: the system style and the "large title view" style. Nowadays there's a third style: a leading-aligned title view with a normal-sized title (or a `TwoLineTitleView`), and we should make it easier for consumers of FluentUI to pick exactly which style they want.

We previously used `usesLargeTitle` to represent whether we use an `AvatarTitleView` (formerly known as `LargeTitleView`), so the most natural translation of `usesLargeTitle` is `titleStyle != .system`. Eventually we'll want to deprecate `usesLargeTitle`.

### Verification

Validated that there are no changes to the different navigation bars in appearance or behavior, both for parent views and child views.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1712)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -49,89 +49,89 @@ class NavigationControllerDemoController: DemoController {
     }
 
     @objc func showLargeTitle() {
-        presentController(withLargeTitle: true)
+        presentController(withTitleStyle: .largeLeading)
     }
 
     @objc func showLargeTitleWithShyAccessory() {
-        presentController(withLargeTitle: true, accessoryView: createAccessoryView(), contractNavigationBarOnScroll: true)
+        presentController(withTitleStyle: .largeLeading, accessoryView: createAccessoryView(), contractNavigationBarOnScroll: true)
     }
 
     @objc func showLargeTitleWithFixedAccessory() {
-        presentController(withLargeTitle: true, accessoryView: createAccessoryView(), contractNavigationBarOnScroll: false)
+        presentController(withTitleStyle: .largeLeading, accessoryView: createAccessoryView(), contractNavigationBarOnScroll: false)
     }
 
     @objc func showLargeTitleWithFixedAccessoryAndSubtitle() {
-        presentController(withLargeTitle: true, subtitle: "Subtitle goes here", accessoryView: createAccessoryView(), contractNavigationBarOnScroll: false)
+        presentController(withTitleStyle: .largeLeading, subtitle: "Subtitle goes here", accessoryView: createAccessoryView(), contractNavigationBarOnScroll: false)
     }
 
     @objc func showLargeTitleWithSystemStyle() {
-        presentController(withLargeTitle: true, style: .system)
+        presentController(withTitleStyle: .largeLeading, style: .system)
     }
 
     @objc func showLargeTitleWithSystemStyleAndNoShadow() {
-        presentController(withLargeTitle: true, style: .system, contractNavigationBarOnScroll: false, showShadow: false)
+        presentController(withTitleStyle: .largeLeading, style: .system, contractNavigationBarOnScroll: false, showShadow: false)
     }
 
     @objc func showLargeTitleWithSystemStyleAndShyAccessory() {
-        presentController(withLargeTitle: true, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
+        presentController(withTitleStyle: .largeLeading, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
     }
 
     @objc func showLargeTitleWithSystemStyleShyAccessoryAndSubtitle() {
-        presentController(withLargeTitle: true, subtitle: "Subtitle goes here", style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
+        presentController(withTitleStyle: .largeLeading, subtitle: "Subtitle goes here", style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
     }
 
     @objc func showLargeTitleWithSystemStyleAndFixedAccessory() {
-        presentController(withLargeTitle: true, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: false)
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -109,6 +109,17 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         case custom
     }
 
+    @objc(MSFNavigationBarTitleStyle)
+    /// Describes the style in which the title is shown in a navigation bar.
+    public enum TitleStyle: Int {
+        /// Shows a center-aligned title and/or subtitle. Most closely aligned with UIKit's default. Not capable of showing an avatar. b
```
huwilkes, line 115: nit: "Not capable of showing an avatar. b"

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -585,7 +596,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         let (actualStyle, actualItem) = actualStyleAndItem(for: navigationItem)
         style = actualStyle
         updateColors(for: actualItem)
-        showsLargeTitle = navigationItem.usesLargeTitle
+        usesAvatarTitleView = navigationItem.titleStyle != .system
```
huwilkes, line 599: This logic should probably be
```
let titleStyle = navigationItem.titleStyle
usesAvatarTitleView = titleStyle == .leading || titleStyle == .largeLeading
```
to make it more resilient to future style additions.

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
```diff
@@ -101,12 +101,23 @@ import UIKit
...
+        }
+        set {
+            objc_setAssociatedObject(self, &AssociatedKeys.titleStyle, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        }
+    }
+
     var usesLargeTitle: Bool {
         get {
-            return objc_getAssociatedObject(self, &AssociatedKeys.usesLargeTitle) as? Bool ?? false
+            assertionFailure("`usesLargeTitle` will be deprecated in a future release. Use `titleStyle` instead.")
```
huwilkes, line 115: If the goal is to use an assert instead of a deprecation message, you should probably use just `assert` instead of `assertionFailure`. I'm not sure them trying to use `usesLargeTitle` is worth crashing over, especially in release builds.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: The B stands for "bane of my existence, these typos are." But yeah, good catch.

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
amgleitman: `assertionFailure` doesn't crash on release builds anyway.

https://www.hackingwithswift.com/plus/intermediate-swift/understanding-assertions#assertionfailure

huwilkes: Ah, so it doesn't. That's misleading of the documentation. Explicitly calling out that `assert` is just in debug, and then saying nothing about `assertionFailure`. I guess debug vs release is `assert` vs `precondition`. Regardless, we probably don't want to crash on our debug builds, either.

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: We might as well bake this into a computed property of `NavigationBar.TitleStyle` so we don't need to have this verbosity all over the place.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -178,7 +189,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
 
     /// Returns the first match of an optional view for a bar button item with the given tag.
     @objc public func barButtonItemView(with tag: Int) -> UIView? {
-        if showsLargeTitle {
+        if usesAvatarTitleView {
```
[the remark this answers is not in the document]
amgleitman: Fair. Makes me wonder if we should rename `AvatarTitleView` to `LeadingTitleView` or something like that, but there are definitely bigger issues to worry about right now.

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
amgleitman: Discussed offline. Let's just outright deprecate `usesLargeTitle` to help nudge consumers of this library to update to the newest API.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -49,89 +49,89 @@ class NavigationControllerDemoController: DemoController {
...
     }
 
     @objc func showLargeTitleWithSystemStyleAndShyAccessory() {
-        presentController(withLargeTitle: true, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
+        presentController(withTitleStyle: .largeLeading, style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
     }
 
     @objc func showLargeTitleWithSystemStyleShyAccessoryAndSubtitle() {
-        presentController(withLargeTitle: true, subtitle: "Subtitle goes here", style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
+        presentController(withTitleStyle: .largeLeading, subtitle: "Subtitle goes here", style: .system, accessoryView: createAccessoryView(with: .darkContent), contractNavigationBarOnScroll: true)
```
huwilkes, line 80: Does this still show a subtitle? The documentation comment says that `.largeLeading` should ignore subtitles.

amgleitman: Fixed with 179e1be.

huwilkes (approved):

PR #1722, opened 2023-04-28
Fix ShyHeaderController regression

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There is a bug in partner apps where the view of the `ShyHeaderController` gets comm blue colors. 
This is caused by the parent view not registering the custom theme in `viewWillAppear`. Since this only happens after backing out of a file, I added back the color update in `viewDidAppear` as a safety measure in case the update in `viewWillAppear` is too early and returns the shared theme. 

### Binary change

Total increase: 1,960 bytes
Total decrease: 0 bytes
| Total | 30,156,968 bytes | 30,158,928 bytes | ⚠️ 1,960 bytes |
<details>

| ShyHeaderController.o | 252,984 bytes | 254,504 bytes | ⚠️ 1,520 bytes |
| __.SYMDEF | 4,723,104 bytes | 4,723,544 bytes | ⚠️ 440 bytes |
</details>

### Verification

The bug was not reproducible on our demo app but it's been verified and fixed on all partner apps.

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1722)

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -124,6 +124,11 @@ class ShyHeaderController: UIViewController {
         updateNavigationBarStyle()
     }
 
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        updateNavigationBarStyle()
+    }
+
     // MARK: - Base Construction
 
     /// Constructs the UI for the gesture-based configuration
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -124,6 +124,7 @@ class ShyHeaderController: UIViewController {
         updateNavigationBarStyle()
     }
 
+    // This is needed as a safety measure to ensure the colors are updated with the correct theme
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -124,6 +124,12 @@ class ShyHeaderController: UIViewController {
         updateNavigationBarStyle()
     }
 
+    // This is needed as a safety measure to ensure the colors are updated with the correct theme
+    override func viewDidAppear(_ animated: Bool) {
```
huwilkes, line 128: Is there any lag in updating the color? Like, is there a flash of comm blue before going to the correct color for the theme?

huwilkes (approved):

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
laminesm: There is no visual lag because when we back out of a file the ShyHeaderController's view is hidden behind the NavigationBar, its background view and the shy header view. It only gets pushed up when searching.

PR #1725, opened 2023-05-02
[Fluent Button] Added support for Increase Contrast Accessibility

amoggo:
### Platforms Impacted
- [ ] iOS
- [x] macOS

**PROBLEM**
Fluent Button's don't support "Increase Contrast" Accessibility Setting. A default NSButton already subscribes to the HighContrast requirements but the Fluent Button which subclasses NSButton, doesn't have it.

**FIX**
- Determined that the Native NSButton's drawing code could not be leveraged for this setting. Only way to get this for free is to directly call any of NSButton's convenience initializers but because Fluent Button already has a designated init this wasn't possible - A super class's convenience init cannot be called from a subclassed designated init. Only other way to do this is to implement it ourselves..
- AppKit conveniently provides a `accessibilityDisplayShouldIncreaseContrast` property indicating if the HighContrast is enabled or not. This coupled with a Notification update API posted everytime this setting changes viz. [accessibilityDisplayOptionsDidChangeNotification](https://developer.apple.com/documentation/appkit/nsworkspace/1534227-accessibilitydisplayoptionsdidch) provides a straightforward solution to hook into the Fluent Button API
- chose to use the native OS ".textColor" NSColor for the Button border since this closely matches the button borders in native NSButton with Increase Contrast Enabled. Also this color responds correctly to Light/Dark modes as well as accessibility settings like Invert Colors and Increase Contrast.

**PRODUCT IMPACT**
Only impacts when "Increase Contrast" is enabled, where all the Fluent Buttons have darker borders.

**VERIFICATION**
[x] Light/Dark Mode
[x] Button's rest, pressed and disabled states
[x] Booting the test app when "increase contrast" is already enabled / Toggling increase contrast when test app is already booted
[x] All Button Style Formats and sizes - viz. the test app Button Pane
 
<details>

| Screenshot 2023-05-02 at 16 00 27 | Screenshot 2023-05-02 at 16 01 01 |

| After (Inactive State)                                       | After (Dark Mode)                                      |
| Screenshot 2023-05-02 at 16 00 53 | Screenshot 2023-05-02 at 16 01 16 |

</details>
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1725)

### DIFF
--- a/macos/FluentUI/Button/Button.swift
+++ b/macos/FluentUI/Button/Button.swift
@@ -113,6 +113,12 @@ open class Button: NSButton {
 		if size == defaultFormat.size {
 			setSizeParameters(forSize: size)
 		}
+
+		NSWorkspace.shared.notificationCenter.addObserver(forName: NSWorkspace.accessibilityDisplayOptionsDidChangeNotification,
+														  object: nil,
+														  queue: nil) {[weak self] _ in
+			self?.increaseContrastEnabled = NSWorkspace.shared.accessibilityDisplayShouldIncreaseContrast
+		}
 	}
 
 	open override class var cellClass: AnyClass? {
@@ -352,6 +358,14 @@ open class Button: NSButton {
 	}
 
 	private func setColorValues(forStyle: ButtonStyle, accentColor: NSColor) {
+		if increaseContrastEnabled {
+			// Use "UI Element Colors" System Colors which respond correctly to accessibility
+			// settings like increase contrast
+			borderColorRest = .textColor
+			borderColorPressed = .textColor
+			borderColorDisabled = .textColor
+		}
+
 		switch forStyle {
 		case .primary:
 			contentTintColorRest = isWindowInactive ? .textColor : ButtonColor.neutralInverted
@@ -360,9 +374,11 @@ open class Button: NSButton {
 			backgroundColorRest = isWindowInactive ? ButtonColor.neutralBackground2 : accentColor
 			backgroundColorPressed = accentColor.withSystemEffect(.pressed)
 			backgroundColorDisabled = ButtonColor.brandBackgroundDisabled
-			borderColorRest = isWindowInactive ? ButtonColor.neutralStroke2 : .clear
-			borderColorPressed = .clear
-			borderColorDisabled = .clear
+			if !increaseContrastEnabled {
+				borderColorRest = isWindowInactive ? ButtonColor.neutralStroke2 : .clear
+				borderColorPressed = .clear
+				borderColorDisabled = .clear
+			}
 		case .secondary:
 			contentTintColorRest = .textColor
 			contentTintColorPressed = ButtonColor.neutralInverted?.withSystemEffect(.pressed)
@@ -370,9 +386,11 @@ open class Button: NSButton {
 			backgroundColorRest = ButtonColor.neutralBackground2
 			backgroundColorPressed = accentColor.withSystemEffect(.pressed)
 			backgroundColorDisabled = ButtonColor.neutralBackground2?.withSystemEffect(.disabled)
-			borderColorRest = ButtonColor.neutralStroke2
-			borderColorPressed = .clear
-			borderColorDisabled = ButtonColor.neutralStroke2?.withSystemEffect(.disabled)
+			if !increaseContrastEnabled {
+				borderColorRest = ButtonColor.neutralStroke2
+				borderColorPressed = .clear
+				borderColorDisabled = ButtonColor.neutralStroke2?.withSystemEffect(.disabled)
+			}
 		case .acrylic:
 			contentTintColorRest = ButtonColor.neutralForeground3
 			contentTintColorPressed = ButtonColor.neutralForeground3?.withSystemEffect(.pressed)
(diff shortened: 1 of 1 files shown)


macos/FluentUI/Button/Button.swift
```diff
@@ -352,6 +358,14 @@ open class Button: NSButton {
 	}
 
 	private func setColorValues(forStyle: ButtonStyle, accentColor: NSColor) {
+		if increaseContrastEnabled {
+			// Use "UI Element Colors" System Colors which respond correctly to accessibility
+			// settings like increase contrast
+			borderColorRest = .textColor
+			borderColorPressed = .textColor
+			borderColorDisabled = .textColor
```
Saadnajmi, line 366: I'm not sure we want to use textColor for borderColor, but I guess you didn't find a better one?

macos/FluentUI/Button/Button.swift
```diff
@@ -352,6 +358,14 @@ open class Button: NSButton {
 	}
 
 	private func setColorValues(forStyle: ButtonStyle, accentColor: NSColor) {
+		if increaseContrastEnabled {
+			// Use "UI Element Colors" System Colors which respond correctly to accessibility
+			// settings like increase contrast
```
Saadnajmi, line 363: You could also link the Apple HIG and/or apple docs here. Or just say "Use system colors"

macos/FluentUI/Button/Button.swift
```diff
@@ -360,39 +374,47 @@ open class Button: NSButton {
...
 			backgroundColorPressed = accentColor.withSystemEffect(.pressed)
 			backgroundColorDisabled = ButtonColor.brandBackgroundDisabled
-			borderColorRest = isWindowInactive ? ButtonColor.neutralStroke2 : .clear
-			borderColorPressed = .clear
-			borderColorDisabled = .clear
+			if !increaseContrastEnabled {
+				borderColorRest = isWindowInactive ? ButtonColor.neutralStroke2 : .clear
+				borderColorPressed = .clear
+				borderColorDisabled = .clear
+			}
```
Saadnajmi, line 381: Nit (feel free to disagree), I'd rather move the check to per color, rather than one check where we set the colors again.

huwilkes: Wouldn't it also be possible to just move the initial check for `increasedContrastEnabled` to the end, and not check at all in the switch?

macos/FluentUI/Button/Button.swift
laminesm: Is there a particular reason we wouldn't want to use the textColor for the border? If anything, it should guarantee a strong contrast between the background and the border.

Saadnajmi: Mostly because this is a semantic color, who's name is "textColor" lol. But I can't think of a better color tbh 😅. 

What does Fluent 2 use? Are the text and border color the same on iOS?

macos/FluentUI/Button/Button.swift
```diff
@@ -483,6 +505,18 @@ open class Button: NSButton {
 			needsDisplay = true
 		}
 	}
+	
+	/// Indicates if the `Increase Contrast` Accessibility Setting is enabled
+	private var increaseContrastEnabled: Bool = NSWorkspace.shared.accessibilityDisplayShouldIncreaseContrast {
```
laminesm, line 510: The newline spacing of the content of this var seems off.

macos/FluentUI/Button/Button.swift
brentpe: Maybe controlTextColor vs. textColor?

macos/FluentUI/Button/Button.swift
brentpe: FWIW, my preference would be to continue the pattern already used and set the colors once inline. Makes future edits clear on where stuff happens vs. starting a new pattern.

macos/FluentUI/Button/Button.swift
amoggo: Didn't really find anything else that's border/outline specific.

macos/FluentUI/Button/Button.swift
amoggo: @huwilkes Yep I thought of that too - but felt a bit unnatural to set to one value and then overwrite it.

I just went with Saad's suggestion.

Saadnajmi (approved):
Given that we currently don't know anything better than textColor for the borderColor and IMO this PR makes our button way better, I think LGTM :)

macos/FluentUI/Button/Button.swift
```diff
@@ -352,6 +358,11 @@ open class Button: NSButton {
 	}
 
 	private func setColorValues(forStyle: ButtonStyle, accentColor: NSColor) {
+
+		// Use System Colors which respond correctly to accessibility settings like increase contrast
+		// https://developer.apple.com/documentation/appkit/nscolor/ui_element_colors
+		let increaseContrastBorderColor: NSColor = .textColor
+
```
anandrajeswaran, line 365: Thoughts on changing this to a private static let on Button so we get implicit laziness? And then each reference to it below would be Button.increaseContrastBorderColor inline.

Advantages:
- trivial perf - not calculated until needed (even though calculation is super cheap now, it might not stay that way as logic gets added.
- simplicity - someone reading through or debugging code doesn't need to rationalize this value unless it is actually needed

PR #1727, opened 2023-05-03
Use a minimum 44px height navigation bar everywhere

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

By default, "small" phones in landscape mode will show a compact navigation bar that's only 32px tall instead of the usual 44px. For the sake of consistency between different kinds of title views and not having a two-line title view look so cramped, we still want to use that 44px tall bar in this case.

### Verification

Most testing involved navigating between screens and constantly rotating the device. Testing was done on simulators for the iPhone 14 Pro ("small") and the iPhone 14 Pro Max ("large").

<details>

Screenshots are taken from an iPhone 14 Pro (which is considered a "small" phone).

| Before | After |
|-|-|
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 33 56 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 32 37 |
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 34 00 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 32 45 |
| Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 34 04 | Simulator Screen Shot - iPhone 14 Pro - 2023-05-03 at 16 32 51 |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1727)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -149,8 +149,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
 
     private static var defaultStyle: Style = .primary
 
+    // These two constants are based on OS default values
+    static let systemHeight: CGFloat = 44
+    static let compactSystemHeight: CGFloat = 32
+
     private struct Constants {
-        static let systemHeight: CGFloat = 44
         static let normalContentHeight: CGFloat = 44
         static let expandedContentHeight: CGFloat = 48
 
@@ -478,7 +481,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         contentStackView.directionalLayoutMargins = NSDirectionalEdgeInsets(
             top: 0,
             leading: contentLeadingMargin,
-            bottom: -(contentHeight - Constants.systemHeight),
+            bottom: -(contentHeight - Self.systemHeight),
             trailing: contentTrailingMargin
         )
     }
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -415,7 +432,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         if let topAccessoryView = self.topAccessoryView {
             topAccessoryView.translatesAutoresizingMaskIntoConstraints = false
 
-            let insertionIndex = contentStackView.arrangedSubviews.firstIndex(of: leadingSpacerView)! + 1
+            let insertionIndex = contentStackView.arrangedSubviews.firstIndex(of: postTitleSpacerView)! + 1
```
huwilkes, line 435: How much harder would it be to add a `topAccessoryHostView` and keep that in the same spot in the `contentStackView` and just add/remove subviews from that and use `.isHidden` to hide it if necessary? It would be great if we could remove the `!`.

amgleitman: I can see this being a nice-to-have, but I'd rather focus on getting this feature merged into the main branch in the short term. We can always create an issue for it later.

huwilkes (approved):

PR #1692, opened 2023-04-06
Fix Button's `sizeThatFits` calculation

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A client was having an issue where their button would break into two lines. The issue seems to be coming from `proposedTitleLabelWidth`, so we're removing that from the `sizeThatFits` calculation for now.

### Verification

Tested in their app, as well as in our test app for iOS 14, 15, and 16.

<details>
<summary>Visual Verification</summary>

| iOS Version | Before | After |
|---|---|---|
| 14 | Button_14_Before | Button_14_After |
| 15 | Button_15_Before | Button_15_After |
| 16 | Button_16_Before | Button_16_After |
</details>

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [x] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1692)

PR #1753, opened 2023-05-17
Update Button's large text support

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

#1692 broke large text support in the button by removing `proposedTitleLabelWidth` in an attempt to fix the title breaking in to two lines when there was still space for a single line. The root of that issue was that the button would get into a try to base the content size on the size of the bounds, and then shrink it to guess how much space was left for the button. By changing to just finding the `systemLayoutSizeFitting(size)`, where size was the max value for both height and width, we were getting returned a size of a height we wanted (for a single line), but the width wasn't correctly bound by the space available to the button.

The end goal of the overridden `sizeThatFits` was to provide a lower bound on the height of the button. However, `UIButton.ButtonConfiguration` doesn't provide any way to do so, it only allows the user to set the `contentInsets`. Our designers have opted for defining a minimum height to guarantee that the button will fit in to a 4pt grid, rather than try to react to changes in the font to get the correct insets. Since the goal is to add a minimum height, I've added a constraint on the height of the button to be greater than or equal to the minimum height defined in the token set. I also tried getting `super.sizeThatFits()` and adjusting the height of that to meet the minimum, but that lead to the buttons growing in height rather than width. The minimum height constraint will automatically be disabled if the edge insets are anything other than the default, allowing users to create custom buttons of a smaller size.

I also noticed that the floating button styles would sometimes load in with no shadow and as rectangles, so I've updated the logic to try and leverage the `.capsule` corner style from the configuration. I encountered an issue there with the configuration not updating the layer/view of the button in a way that our shadow layers could recognize, so I had to set the corner radius, corner style, and background color manually before applying shadows. We also didn't have large content viewer support, so I added that.

I couldn't figure out a good way to dynamically set `showsLargeContentViewer`, so the current implementation still requires the client to set `showsLargeContentViewer` on button they need to show the large content viewer. We needed to be able to support the large content viewer in the floating buttons, or when our button had either no text or text that wouldn't scale, but there weren't sufficient ways to listen for all of those settings.

### Binary change

Total increase: 7,104 bytes
Total decrease: 0 bytes
| Total | 31,811,536 bytes | 31,818,640 bytes | ⚠️ 7,104 bytes |
<details>

| Button.o | 202,608 bytes | 208,176 bytes | ⚠️ 5,568 bytes |
| __.SYMDEF | 4,780,208 bytes | 4,781,096 bytes | ⚠️ 888 bytes |
| FocusRingView.o | 788,352 bytes | 789,000 bytes | ⚠️ 648 bytes |
</details>

### Verification

<details>

| Button_DefaultTextSize_Before | Button_DefaultTextSize_After |
| Button_MaxTextSize_Before | Button_MaxTextSize_After |
| n/a | Button_LargeContentViewer_After |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1753)

### DIFF
--- a/ios/FluentUI/Button/Button.swift
+++ b/ios/FluentUI/Button/Button.swift
@@ -86,6 +86,7 @@ open class Button: UIButton, Shadowable, TokenizedControlInternal {
             return outgoing
         }
         configuration.titleTextAttributesTransformer = titleTransformer
+        configuration.contentInsets = edgeInsets
         self.configuration = configuration
 
         update()
@@ -137,9 +138,23 @@ open class Button: UIButton, Shadowable, TokenizedControlInternal {
     public override func traitCollectionDidChange(_ previousTraitCollection: UITraitCollection?) {
         super.traitCollectionDidChange(previousTraitCollection)
 
-        if previousTraitCollection?.userInterfaceStyle != traitCollection.userInterfaceStyle {
+        guard let previousTraitCollection else {
+            return
+        }
+
+        if previousTraitCollection.userInterfaceStyle != traitCollection.userInterfaceStyle {
             updateBorder()
         }
+
+        if previousTraitCollection.preferredContentSizeCategory != traitCollection.preferredContentSizeCategory {
+            invalidateIntrinsicContentSize()
+        }
+    }
+
+    public func defaultEdgeInsets() -> NSDirectionalEdgeInsets {
+        let leadingPadding = ButtonTokenSet.horizontalPadding(style: style, size: sizeCategory)
+        let trailingPadding = style.isFloating && titleLabel?.text != nil && image != nil ? ButtonTokenSet.fabAlternativePadding(sizeCategory) : ButtonTokenSet.horizontalPadding(style: style, size: sizeCategory)
+        return NSDirectionalEdgeInsets(top: 0, leading: leadingPadding, bottom: 0, trailing: trailingPadding)
     }
 
     public typealias TokenSetKeyType = ButtonTokenSet.Tokens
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Button/Button.swift
```diff
@@ -53,83 +53,48 @@ open class Button: UIButton, Shadowable, TokenizedControlInternal {
...
-                var configuration = self.configuration ?? UIButton.Configuration.plain()
-                configuration.contentInsets = edgeInsets
-                self.configuration = configuration
-            } else {
-                let left: CGFloat
-                let right: CGFloat
-                if effectiveUserInterfaceLayoutDirection == .leftToRight {
-                    left = edgeInsets.leading
-                    right = edgeInsets.trailing
-                } else {
```
laminesm, line 76: Do we not need to handle different insets for RTL anymore?

ios/FluentUI/Button/Button.swift
```diff
@@ -343,21 +290,24 @@ open class Button: UIButton, Shadowable, TokenizedControlInternal {
...
 
     private func updateShadow() {
+        // UIButton.Configuration doesn't set the layer by default, so in order
+        // for our shadow layers to work we need to update the layer.
+        let configuration = self.configuration ?? UIButton.Configuration.plain()
+        self.backgroundColor = configuration.background.backgroundColor
+        layer.cornerCurve = .continuous
+        layer.cornerRadius = style.isFloating ? frame.height / 2 : configuration.background.cornerRadius
+
         let shadowInfo = !isEnabled || isHighlighted || isFocused ? tokenSet[.shadowPressed].shadowInfo : tokenSet[.shadowRest].shadowInfo
```
laminesm, line 307: nit: parentheses around `!isEnabled || isHighlighted || isFocused`

ios/FluentUI/Button/Button.swift
huwilkes: We're using `NSDirectionalEdgeInsets` now, which uses leading + trailing, instead of `UIEdgeInset`, which used left + right, so RTL support is built in.

laminesm (approved):

PR #1758, opened 2023-05-25
Add new ColorProviding protocol for Mac Fluent colors

brentpe:
### Platforms Impacted
  - [ ] iOS
  - [x] macOS
  
### Description of changes
  
This change incrementally moves the Mac usage of Fluent Colors towards the iOS model, where a ColorProvider is used. This is an incremental changes where clients can optionally use a ColorProvider object vs. initializing all the Fluent v1 colors via the setters in Colors.swift, such as "Colors.primary = .blue" or "[MSFColors setPrimary:[NSColor blueColor]]" (Obj-C) type calls during application boot.

Clients make use of this by creating a class adhering to the ColorProviding protocol by adding overrides for returning the different primary FluentV1 colors. This can be implemented in simple sharedInstance (singleton) object, or a such as the application delegate. When the colors are called via "Colors.primary" or "[MSFColors primary]" (Obj-C), the colors are fetched on demand from the ColorProvider if specified.
  
It's worth noting that this change does not use Fluent v2 colors or the newer theming patterns on iOS as that will come separately with future work.
  
### Verification
  
This code was verified using the FluentUITestApp. The TestColorViewController was modified to use the newer ColorProvider by default, then fall back to the legacy methods when the "Use ColorProvider" switch is turned off.

When Use ColorProvider switch is On **custom** primary colors will be used:
image

When Use ColorProvider switch is Off **CommunicationBlue** primary colors will be used as a fallback:
image
  
### Pull request checklist
  
  This PR has considered:
  - [x] macOS supported versions (all major versions greater than or equal current target deployment version)
  - [x] Different resolutions (1x, 2x)

### DIFF
--- a/macos/FluentUI/Core/ColorProviding.swift
+++ b/macos/FluentUI/Core/ColorProviding.swift
@@ -1,0 +1,23 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import AppKit
+
+// MARK: ColorProviding
+
+/// Protocol through which consumers can provide colors to "theme" their experiences
+/// The view associated with the passed in theme will display the set colors to allow apps to provide different experiences per each view
+@objc(MSFColorProviding)
+public protocol ColorProviding {
+    /// If this protocol is not conformed to, communicationBlue variants will be used
+    @objc var primary: NSColor { get }
+    @objc var primaryTint10: NSColor { get }
+    @objc var primaryTint20: NSColor { get }
+    @objc var primaryTint30: NSColor { get }
+    @objc var primaryTint40: NSColor { get }
+    @objc var primaryShade10: NSColor { get }
+    @objc var primaryShade20: NSColor { get }
+    @objc var primaryShade30: NSColor { get }
+}
--- a/macos/FluentUI/Core/ColorProviding.swift
+++ b/macos/FluentUI/Core/ColorProviding.swift
@@ -13,11 +13,11 @@ import AppKit
 public protocol ColorProviding {
     /// If this protocol is not conformed to, communicationBlue variants will be used
     @objc var primary: NSColor { get }
+	@objc var primaryShade10: NSColor { get }
+	@objc var primaryShade20: NSColor { get }
+	@objc var primaryShade30: NSColor { get }
     @objc var primaryTint10: NSColor { get }
     @objc var primaryTint20: NSColor { get }
     @objc var primaryTint30: NSColor { get }
     @objc var primaryTint40: NSColor { get }
-    @objc var primaryShade10: NSColor { get }
-    @objc var primaryShade20: NSColor { get }
-    @objc var primaryShade30: NSColor { get }
 }
--- a/macos/FluentUI/Core/Colors.swift
+++ b/macos/FluentUI/Core/Colors.swift
@@ -438,22 +438,115 @@ public final class Colors: NSObject {
 	}
 
 	// MARK: Primary
+	@objc public static var primary: NSColor {
+		get {
+			if let colorProvider = colorProvider {
+				return colorProvider.primary
+			} else {
+				return _primary
+			}
+		}
+		set {
+			_primary = newValue
+		}
+	}
+	@objc public static var primaryTint10: NSColor {
+		get {
+			if let colorProvider = colorProvider {
+				return colorProvider.primaryTint10
+			} else {
+				return _primaryTint10
+			}
+		}
+		set {
+			_primaryTint10 = newValue
+		}
+	}
+	@objc public static var primaryTint20: NSColor {
+		get {
+			if let colorProvider = colorProvider {
+				return colorProvider.primaryTint20
+			} else {
+				return _primaryTint20
+			}
+		}
+		set {
+			_primaryTint20 = newValue
+		}
+	}
+	@objc public static var primaryTint30: NSColor {
+		get {
+			if let colorProvider = colorProvider {
+				return colorProvider.primaryTint30
+			} else {
+				return _primaryTint30
+			}
+		}
+		set {
+			_primaryTint30 = newValue
+		}
+	}
+	@objc public static var primaryTint40: NSColor {
+		get {
+			if let colorProvider = colorProvider {
+				return colorProvider.primaryTint40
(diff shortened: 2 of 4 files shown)


macos/FluentUI/Core/Colors.swift
```diff
@@ -438,22 +438,115 @@ public final class Colors: NSObject {
...
-
-	@objc public static var primaryShade10: NSColor = Colors.Palette.communicationBlueShade10.color
-
-	@objc public static var primaryShade20: NSColor = Colors.Palette.communicationBlueShade20.color
+	/// Color provider object. When a Color Provider is present, colors will be retrieved from the provider vs. internally.
+	@objc public static var colorProvider: ColorProviding?
 
-	@objc public static var primaryShade30: NSColor = Colors.Palette.communicationBlueShade30.color
+	/// These are initialized with the Communication Blue defaults.
+	private static var _primary = Colors.Palette.communicationBlue.color
```
laminesm, line 542: Actually do we need to have both private and public vars for these color values? Maybe we could just have public values, initialize them with default comm blue and have a setter in colorProvider that updates them to the passed in values in the provider?

macos/FluentUI/Core/ColorProviding.swift
```diff
@@ -0,0 +1,23 @@
...
+import AppKit
+
+// MARK: ColorProviding
+
+/// Protocol through which consumers can provide colors to "theme" their experiences
+/// The view associated with the passed in theme will display the set colors to allow apps to provide different experiences per each view
+@objc(MSFColorProviding)
+public protocol ColorProviding {
+    /// If this protocol is not conformed to, communicationBlue variants will be used
+    @objc var primary: NSColor { get }
```
anandrajeswaran, line 15: might be me not understanding GitHub PR experience, but is this a mix of tabs and spaces?

macos/FluentUI/Core/Colors.swift
anandrajeswaran: If I'm understanding correctly, the problem is that would then ruin the benefit of the provider model - lazy loading of colors when we actually use them. Because then we'd have to load every color when the colorProvider is set.

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -52,8 +52,9 @@ class TestColorViewController: NSViewController {
 
 		let switchButton = NSSwitch(frame: CGRect(x: 1, y: 1, width: 100, height: 50))
 		switchButton.target = self
+		switchButton.state = useColorProvider ? .on : .off
 		switchButton.action = #selector(toggleClicked)
-		toggleTextView.string = "Default"
+		toggleTextView.string = "Use ColorProvider"
```
anandrajeswaran, line 57: The problem is we now are not testing the case where the colors are overridden without a provider object, but maybe that's ok if the intent is soon to just remove the ability to do so?

To that point, should we mark usage of those setters as deprecated?

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -78,6 +79,19 @@ class TestColorViewController: NSViewController {
 		view = containerView
 	}
 
+	// MARK: ColorProviding Protocol
+
+	var primary: NSColor = useColorProvider ? (NSColor(named: "Colors/DemoPrimaryColor"))! : Colors.Palette.communicationBlue.color
```
anandrajeswaran, line 84: Should we just make this `var primary: NSColor = NSColor(named: "Colors/DemoPrimaryColor"))!`? If `useColorProvider` is false, won't these not get called?

macos/FluentUITestViewControllers/TestColorViewController.swift
```diff
@@ -154,3 +162,6 @@ class ColorRectView: NSView {
 }
 
 private let colorRowSpacing: CGFloat = 10.0
+
+// Default to using the new ColorProvider protocol for fetching the Fluent Primary Colors
+private var useColorProvider: Bool = true
```
anandrajeswaran, line 167: Tiny bit of brevity without loss of clarity - can use type inference for this one: `private var useColorProvider = true`

anandrajeswaran (approved):

macos/FluentUITestViewControllers/TestColorViewController.swift
brentpe: Yeah, that was my intention since we will be deprecating this. In terms of deprecating, was going to bring that up separately after we fix up callers.

macos/FluentUI/Core/Colors.swift
brentpe: That was my original intention, but this class isn't instantiated so Swift doesn't allow that. I can use static var's but it won't allow an assigned value and a get - I get errors like "Cannot find 'get' in scope.

macos/FluentUI/Core/ColorProviding.swift
brentpe: No. Not sure why it did that. Fixing.

macos/FluentUI/Core/Colors.swift
laminesm: I don't think the provider model does lazy loading of branding colors on the iOS side though. There is no `didSet` but we have a `setColorProvider` function that essentially works like a didSet and overrides all the branding colors as soon as a provider is set.

macos/FluentUITestViewControllers/TestColorViewController.swift
brentpe: OH, good catch. That was part of an earlier iteration where I was going to try and keep the existing "Default Colors" behavior, which I now believe is overkill since we'll be deprecating this.

macos/FluentUI/Core/Colors.swift
anandrajeswaran: Ooh that seems like a potential improvement we can make to iOS since it kinda defeats the purpose of the provider pattern?

macos/FluentUITestViewControllers/TestColorViewController.swift
brentpe: Added a follow-up task to track this.

anandrajeswaran (approved):

PR #1767, opened 2023-05-31
Add accessibility identifier to NavigationBar's back button

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

Some partner xcui tests fail because they cannot find the item with "Back" as its accessibility identifier. This is caused by the new `NavigationBar` using a custom `UIBarButtonItem` as its back button item without setting its accessibility identifier. To fix the test failures, we're setting the accessibility identifier of `backButtonItem` to "Back".

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1767)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -342,7 +342,14 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
     private var navigationBarShadowObserver: NSKeyValueObservation?
     private var titleStyleObserver: NSKeyValueObservation?
 
-    private let backButtonItem: UIBarButtonItem = UIBarButtonItem(image: UIImage.staticImageNamed("back-24x24"), style: .plain, target: nil, action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
+    private let backButtonItem: UIBarButtonItem = {
+        let backButtonitem = UIBarButtonItem(image: UIImage.staticImageNamed("back-24x24"),
+                                             style: .plain,
+                                             target: nil,
+                                             action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
+        backButtonitem.accessibilityIdentifier = "Back"
+        return backButtonitem
+    }()
 
     weak var backButtonDelegate: NavigationBarBackButtonDelegate? {
         didSet {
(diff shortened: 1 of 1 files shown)


joannaquu (approved):

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -342,7 +342,14 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
     private var navigationBarShadowObserver: NSKeyValueObservation?
     private var titleStyleObserver: NSKeyValueObservation?
 
-    private let backButtonItem: UIBarButtonItem = UIBarButtonItem(image: UIImage.staticImageNamed("back-24x24"), style: .plain, target: nil, action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
+    private let backButtonItem: UIBarButtonItem = {
+        let backButtonItem = UIBarButtonItem(image: UIImage.staticImageNamed("back-24x24"),
```
huwilkes, line 346: What does VoiceOver read?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -342,7 +342,14 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
...
+                                             style: .plain,
+                                             target: nil,
+                                             action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
+        backButtonItem.accessibilityIdentifier = "Back"
```
huwilkes, line 350: Accessibility Identifiers don't need to be localized, right?

laminesm: I don't think so since it's mostly used for automation. Also looking at our repo it seems only accessibility hints and labels need to be localized.

joannaquu: Nope!

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: VO just reads "button". We would probably need to add a label or hint on the `BadgeLabelButton` if we want it to read "back button".

huwilkes (approved):

PR #1796, opened 2023-06-20
Remove deprecated API `shouldAutorotate` in DrawerController

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

To help partners with their preparation for iOS 17 release, we're removing the override of `shouldAutorotate` since it's a deprecated API in iOS 16. 
For iOS 15 and below our drawer uses the default value of true for `shouldAutorotate`. Since we've already overridden `supportedInterfaceOrientations` to conform to the presenting view controller's orientations, the drawer will now always adapt to the presenting controller's orientation.
For iOS 16, Apple wants us to explicitly call `setNeedsUpdateOfSupportedInterfaceOrientations()` as well so we're doing that in `viewDidLoad`.

### Binary change

Total increase: 0 bytes
Total decrease: -2,944 bytes
| Total | 30,391,856 bytes | 30,388,912 bytes | 🎉 -2,944 bytes |
<details>

| __.SYMDEF | 4,727,112 bytes | 4,726,416 bytes | 🎉 -696 bytes |
| DrawerController.o | 490,800 bytes | 488,552 bytes | 🎉 -2,248 bytes |
</details>

### Verification

This was validated for iOS 15 and 16 on our demo app and partner apps.

<details>

| before_portrait | after_portrait |
| before_landscape | after_landscape |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1796)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -363,10 +363,6 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         return super.preferredContentSize.height == 0 && preferredContentHeight == 0 && (contentController?.preferredContentSize.height ?? 0) == 0
     }
 
-    open override var shouldAutorotate: Bool {
-        return presentingViewController?.shouldAutorotate ?? super.shouldAutorotate
-    }
-
     open override var supportedInterfaceOrientations: UIInterfaceOrientationMask {
         return presentingViewController?.supportedInterfaceOrientations ?? super.supportedInterfaceOrientations
     }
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -524,6 +520,10 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         tokenSet.registerOnUpdate(for: view) { [weak self] in
             self?.updateBackgroundColor()
         }
+
+        if #available(iOS 16, *) {
+            setNeedsUpdateOfSupportedInterfaceOrientations()
```
harrieshin, line 525: do we need this here? if we do need to notifiy it should be when we rotate not in viewdidload?

laminesm: Looking at the documentation it seems it needs to be called only when either of these variables `supportedInterfaceOrientations` or `preferredInterfaceOrientationForPresentation` changes. So I'm not sure if it would make sense for this to be called when rotation changes unless I'm misunderstanding something.

harrieshin (approved):

PR #1799, opened 2023-06-21
Add `forcesCompactBarExpansion` to `NavigationController`

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This adds a new argument for `NavigationController` to allow clients to opt out of our more eager approach of wrapping view controllers in `ShyHeaderController`. This allows for a client-side fix where a navigation controller on a drawer doesn't properly show its content.

We make `forcesCompactBarExpansion` false by default because hacking the appearance of the navigation bar has proven to be a rather messy affair, and we'd rather not touch it unless we absolutely have to or if the user explicitly asks for it.

We also add documentation, because documentation is nice.

A better fix might be to better separate the "add extra padding to the navigation bar" and "have a bit that can expand and contract" components of the shy header. However, this should still be workable in the short term.

### Verification

Validated by applying this change to the client apps that originally reported the bug and noticing that it no longer repros.

| Before | After |
|-|-|
| Screenshot 2023-07-07 at 1 43 54 PM | Screenshot 2023-07-07 at 1 46 43 PM |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1799)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -351,6 +351,9 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         return backButtonItem
     }()
 
+    // A default UINavigationItem doesn't have any traits that require an expanded navigation bar
+    private var itemCanBeCompact: Bool = true
+
     weak var backButtonDelegate: NavigationBarBackButtonDelegate? {
         didSet {
             backButtonItem.target = backButtonDelegate
@@ -516,12 +519,14 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
 
     private func updateContentStackViewMargins(forExpandedContent contentIsExpanded: Bool) {
         let contentHeight = contentIsExpanded ? TokenSetType.expandedContentHeight : TokenSetType.normalContentHeight
-        let systemHeight = systemWantsCompactNavigationBar ? TokenSetType.compactSystemHeight : TokenSetType.systemHeight
+
+        let desiredHeight = (systemWantsCompactNavigationBar && itemCanBeCompact) ? TokenSetType.compactSystemHeight : TokenSetType.systemHeight
+        print("desiredHeight is \(desiredHeight)")
 
         contentStackView.directionalLayoutMargins = NSDirectionalEdgeInsets(
             top: 0,
             leading: contentLeadingMargin,
-            bottom: systemHeight - contentHeight,
+            bottom: desiredHeight - contentHeight,
             trailing: contentTrailingMargin
         )
     }
@@ -657,6 +662,9 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         updateShadow(for: navigationItem)
         updateTopAccessoryView(for: navigationItem)
         updateSubtitleView(for: navigationItem)
+
+        itemCanBeCompact = navigationItem.canBeCompact
+        updateContentStackViewMargins(forExpandedContent: contentIsExpanded)
 
         titleView.update(with: navigationItem)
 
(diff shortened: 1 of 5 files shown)


laminesm (commented):
This logic was also problematic in the other partner app mentioned in the task. In that scenario, they use a navigation controller inside a drawer. The contained view controller gets wrapped in a ShyHeaderController resulting in its content being hidden. Was that validated as well?

amgleitman:
> This logic was also problematic in the other partner app mentioned in the task. In that scenario, they use a navigation controller inside a drawer. The contained view controller gets wrapped in a ShyHeaderController resulting in its content being hidden. Was that validated as well?

This has been verified with f39a012.

ios/FluentUI/Navigation/NavigationController.swift
```diff
@@ -131,6 +155,9 @@ open class NavigationController: UINavigationController {
         if viewController is ShyHeaderController {
             return false
         }
+        if prefersShyHeader {
```
laminesm, line 158: Not a fan of these API changes as this feels like something we need to handle internally. `ShyHeaderController` is also an internal control and we should probably try to make sure our clients never have to worry about dealing with it. Wouldn't it work to just check if the parent of the nav controller is a DrawerController and omit the shy header in that case?

laminesm: Also I don't see how this fixes the TwoLineTitleView on landscape since `viewControllerNeedsWrapping` is mostly unchanged barring the addition of `prefersShyHeader` check

amgleitman: Fair point about `ShyHeaderController` being internal. That being said, I feel like checking specifically for a DrawerController might be *too* specific, although it could be fine for now. I've been at somewhat of a loss for more elegant solutions.

And as far as fixing `TwoLineTitleView` on landscape, this happens because `prefersShyHeader` defaults to true. Of course, whether or not we want to keep this solution is a separate matter.

harrieshin: i agree with Lamine. we shouldn't expose this prefershyheader at all. the default behavior should be pre 0.17 version. can we do it more internally like if user passing in something in navigationItem to actuallly utilize twolinetitleview, then we we use shyheaderlogic?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -516,12 +519,14 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
 
     private func updateContentStackViewMargins(forExpandedContent contentIsExpanded: Bool) {
         let contentHeight = contentIsExpanded ? TokenSetType.expandedContentHeight : TokenSetType.normalContentHeight
-        let systemHeight = systemWantsCompactNavigationBar ? TokenSetType.compactSystemHeight : TokenSetType.systemHeight
+
+        let desiredHeight = (systemWantsCompactNavigationBar && itemCanBeCompact) ? TokenSetType.compactSystemHeight : TokenSetType.systemHeight
+        print("desiredHeight is \(desiredHeight)")
```
huwilkes, line 524: Did you mean to have this `print` here?

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
```diff
@@ -145,3 +145,14 @@ import UIKit
...
 }
+
+extension UINavigationItem {
+    var canBeCompact: Bool {
+        // Large leading titles already do the right thing in compact nav bars
+        if titleStyle == .largeLeading {
+            return true
+        }
+
+        return subtitle == nil && titleStyle != .largeLeading && accessoryView == nil
```
huwilkes, line 156: Do we need to check for the `titleStyle` being something other than `.largeLeading` here? We're already returning above if it is `.largeLeading`, so it should have to be something else to get to this point.

amgleitman: This was a remnant of when I was trying to figure out the correct logic to use.

ios/FluentUI/Navigation/NavigationController.swift
amgleitman: It feels weird to have to set two properties of a `UINavigationItem` to enable two-line title views. Doing this in the navigation controller itself might be a bit more natural.

PR #1800, opened 2023-06-22
Exposing the UIView for accessoryTypeView in TableViewCell

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Exposing a method that returns UIView of the internal `accessoryTypeView`.

### Binary change

### Verification

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1800)


ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -958,6 +958,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         }
     }
 
+    /// The UIView of the `accessoryView`.
+    @objc open var accessoryButtonView: UIView? { return accessoryTypeView }
```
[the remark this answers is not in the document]
sophialee0416: Yea I was stuck on naming too, I chose it based off of the delegate method. I can change it to that, since I couldn't use `accessoryTypeView` nor `accessoryView`

huwilkes (approved):

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -958,6 +958,9 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         }
     }
 
+    /// The UIView of the `accessoryView`.
+    @objc open var trailingAccessoryView: UIView? { return accessoryTypeView }
```
harrieshin, line 962: feels bit confusing. because we have like titletrailingview and subtitletrailingview. shouldn't it be like accessoryview? and if client has set custom accessory view, then it should return custom accessotryview instead of accessoryTypeView?

sophialee0416: Yea so I couldn't name accessoryView because that would mean I would have to override the property which I didn't have a problem with, but I cannot make it a get-only and would have to make it settable as well, which I don't think we want. And in terms of the custom accessory view, that can be displayed in conjunction with accessoryTypeView so that is a completely separate property

sophialee0416: Follow-up turn into function

harrieshin (approved):

ios/FluentUI/Table View/TableViewCell.swift
```diff
@@ -1387,6 +1387,12 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
         setNeedsLayout()
         invalidateIntrinsicContentSize()
     }
+
+    /// Returns the UIView of the `accessoryView`.
+    @objc open func getAccessoryView() -> UIView? {
+        return accessoryTypeView
```
mischreiber, line 1393: Any reason not to just make `accessoryTypeView` have a public getter?

sophialee0416: It's not possible because the `TableViewCellAccessoryType` is internal class so we aren't allowed to expose it unfortunately! The other approach was to create a public getter of the internal `accessoryTypeView` UIView but this approach would seem less confusing to anyone wondering why we have 2 properties of the same thing

PR #1809, opened 2023-06-27
Add internal init for PillButton that takes token overrides

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

A client app has been experiencing a bug since we added the tokenized `PillButtonBar` where the `PillButtonBar` would load in, and then flash with default colors. Since the `PillButtonBar` knows what overrides it wants to apply to new `PillButton`s, applying them in init means that the `PillButton` never has a chance to load the wrong color. It also has the added benefit of being able to apply the overrides to the `tokenSet` before calling `registerOnUpdate`, so we don't call the closure from `registerOnUpdate` again for each override value.

I also removed the `PillButtonBar`'s `didMoveToWindow` override. The `PillButton` already has an implementation of `willMove(toWindow)`, so the buttons take care of updating themselves.

### Verification

No visual difference in our test app, but verified in the client app that the `PillButtonBar` is no longer flashing.
The blue/green toggle in our test app still works.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1809&drop=dogfoodAlpha

### DIFF
--- a/ios/FluentUI/Pill Button Bar/PillButton.swift
+++ b/ios/FluentUI/Pill Button Bar/PillButton.swift
@@ -53,7 +53,9 @@ open class PillButton: UIButton, TokenizedControlInternal {
                                                name: PillButtonBarItem.titleValueDidChangeNotification,
                                                object: pillBarItem)
 
-        tokenSet.replaceAllOverrides(with: tokenOverrides)
+        if let tokenOverrides {
+            tokenSet.replaceAllOverrides(with: tokenOverrides)
+        }
         tokenSet.registerOnUpdate(for: self) { [weak self] in
             self?.updateAppearance()
         }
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Pill Button Bar/PillButton.swift
```diff
@@ -46,6 +53,7 @@ open class PillButton: UIButton, TokenizedControlInternal {
                                                name: PillButtonBarItem.titleValueDidChangeNotification,
                                                object: pillBarItem)
 
+        tokenSet.replaceAllOverrides(with: tokenOverrides)
```
harrieshin, line 56: for "normal" cases, is it doing unnecessary iteration within replacealloverride because we will pass nil and it will try to remove overrides when there isn't anything that got override?

huwilkes: Yeah, it would be. I'll add a nil check in the init.

harrieshin (approved):

PR #1819, opened 2023-07-05
Bump Xcode version to fix locally failing XCUITests

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

An [issue](https://stackoverflow.com/questions/76062670/swiftui-toggle-not-being-toggled-in-ui-test) with the SwiftUI toggle has been causing XCUITests to fail locally. Previously, `app.switches["Identifier"].tap()` would toggle the switch. However, you now need to access the first switch within the switch `app.switches["Identifier"].switches.firstMatch.tap()`. 

This change seems to have been caused by Xcode 14.3, which requires macOS 13.0 so bumping Xcode and macOS versions in `ci.yml`.

This PR also cherry-picks 9103720ecc818be5d8d2700cd12fdb3b2e5d0495, which was done on the tokens branch but not main.

### Binary change

Total increase: 0 bytes
Total decrease: 0 bytes
| Total | 46,346,152 bytes | 46,346,152 bytes | 🎉 0 bytes |
<details>

</details>

### Verification

<details>

| Screenshot 2023-07-05 at 11 40 48 AM | N/A |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1819&drop=dogfoodAlpha


harrieshin (approved):

harrieshin (approved):

macos/xcode/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -1376,13 +1376,15 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = 8F931A6C22BD593300311764 /* FluentUI_unittest.xcconfig */;
 			buildSettings = {
+				MACOSX_DEPLOYMENT_TARGET = 13.0;
```
harrieshin, line 1379: oh bit . surprised we need to update our deployment target. should we update our readme file then too? what about package.swift?

joannaquu: So since CI needs to be on macOS 13.0 for Xcode 14.3, I updated all the deployment targets to be 13 as well. But it's not absolutely necessary for tests to run. So would we prefer updating or leaving as is?

scripts/removeUnusedResourcesFromAssets.swift
```diff
@@ -1,114 +0,0 @@
-import Foundation
```
harrieshin: why?

joannaquu: This was accidentally removed, just added it back in the latest commit.

harrieshin (approved):

PR #1823, opened 2023-07-07
Add support for custom trailing image of TwoLineTitleView's subtitle

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

There's a partner request to be able to customize the trailing image of the `TwoLineTitleView`'s subtitle on our navigation bar. 
To that end I created a new `customSubtitleTrailingImage` property in `navigationItem`. There's also a new `custom` style in `NavigationBarTitleAccessory` that allows us to check if the client passed in a custom image, in which case we use that one instead of the default chevrons.

### Binary change
Total increase: 7,512 bytes
Total decrease: -1,816 bytes
| Total | 29,992,728 bytes | 29,998,424 bytes | ⚠️ 5,696 bytes |
<details>

| UINavigationItem+Navigation.o | 72,928 bytes | 75,584 bytes | ⚠️ 2,656 bytes |
| __.SYMDEF | 4,662,168 bytes | 4,664,192 bytes | ⚠️ 2,024 bytes |
| TwoLineTitleView.o | 255,168 bytes | 257,040 bytes | ⚠️ 1,872 bytes |
| FocusRingView.o | 792,512 bytes | 793,248 bytes | ⚠️ 736 bytes |
| TwoLineTitleView+Navigation.o | 27,720 bytes | 27,944 bytes | ⚠️ 224 bytes |
| FluentTheme+Tokens.o | 74,336 bytes | 74,320 bytes | 🎉 -16 bytes |
| AliasTokens.o | 282,616 bytes | 280,816 bytes | 🎉 -1,800 bytes |
</details>

### Verification

<details>

| after_nav_light | after_nav_dark |
| after_twoline_light | after_two_line_dark |

</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
+++ b/ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
@@ -137,7 +137,7 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
     // |  |--titleTrailingImageView (chevron, optional)
     // |--subtitleContainer
     // |  |--subtitleLabel
-    // |  |--subtitleImageView (chevron, custom, optional)
+    // |  |--subtitleImageView (optional)
 
     private lazy var containingStackView: UIStackView = {
         let stackView = UIStackView()
@@ -341,10 +341,10 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
         container.addArrangedSubview(label)
 
         if interactive {
-            let showCustomSubtitleImage = (accessoryType == .custom) && (container == subtitleContainer)
+            let isTitle = container == titleContainer
             container.accessibilityTraits.insert(.button)
             container.accessibilityTraits.remove(.staticText)
-            trailingImageView.image = showCustomSubtitleImage ? customSubtitleTrailingImage : accessoryType.image(isTitle: container == titleContainer)
+            trailingImageView.image = (!isTitle && accessoryType == .custom) ? customSubtitleTrailingImage : accessoryType.image(isTitle: isTitle)
         } else {
             container.accessibilityTraits.insert(.staticText)
             container.accessibilityTraits.remove(.button)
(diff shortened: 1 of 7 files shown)


harrieshin:
so does tapping on subtitle does different thing than tapping on the trailing icon? or is it the same? what does larger content viewer should do?

laminesm:
> so does tapping on subtitle does different thing than tapping on the trailing icon? or is it the same? what does larger content viewer should do?

The trailing icon is included in subtitle so it will register taps as well. Overall, the tapping behavior has remained unchanged. Both title and subtitle will register taps when `interactivePart` is different from `none`.

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -335,9 +341,10 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
         container.addArrangedSubview(label)
 
         if interactive {
+            let showCustomSubtitleImage = (accessoryType == .custom) && (container == subtitleContainer)
```
huwilkes, line 344: nit: Should we factor out the two container checks to just `let isTitle = container == titleContainer` and then check `!isTitle` here and pass in `isTitle` below?

laminesm: Yep that's better I'll do that!

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -134,7 +137,7 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
     // |  |--titleTrailingImageView (chevron, optional)
     // |--subtitleContainer
     // |  |--subtitleLabel
-    // |  |--subtitleImageView (chevron, optional)
+    // |  |--subtitleImageView (chevron, custom, optional)
```
huwilkes, line 140: I think `chevron, custom` is making this comment more confusing. Might be better of with just `(optional)`?

huwilkes (approved):

PR #1825, opened 2023-07-12
Support leading image for both title and subtitle in TwoLineTitleView

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

To support a leading image for both title and subtitle in `TwoLineTitleView`, I added a `isTitleImageLeadingForTitleAndSubtitle` check. When this is set to true, we modify the `containingStackView` to be horizontal with an 8.0 spacing (taken from the partner designs) where we add the leading image and the new vertical `titlesStackView` containing both the title and subtitle. 
We also hide the avatar on the `AvatarTitleView` on our navigation bar.

### Binary change

Total increase: 23,184 bytes
Total decrease: 0 bytes
| Total | 30,012,408 bytes | 30,035,592 bytes | ⚠️ 23,184 bytes |
<details>

| TwoLineTitleView.o | 258,184 bytes | 268,464 bytes | ⚠️ 10,280 bytes |
| __.SYMDEF | 4,666,568 bytes | 4,671,816 bytes | ⚠️ 5,248 bytes |
| UINavigationItem+Navigation.o | 75,584 bytes | 80,792 bytes | ⚠️ 5,208 bytes |
| AvatarTitleView.o | 247,192 bytes | 247,968 bytes | ⚠️ 776 bytes |
| TwoLineTitleViewTokenSet.o | 57,728 bytes | 58,440 bytes | ⚠️ 712 bytes |
| FocusRingView.o | 793,912 bytes | 794,600 bytes | ⚠️ 688 bytes |
| TwoLineTitleView+Navigation.o | 27,944 bytes | 28,184 bytes | ⚠️ 240 bytes |
| DatePickerController.o | 319,152 bytes | 319,168 bytes | ⚠️ 16 bytes |
| DateTimePickerController.o | 140,096 bytes | 140,112 bytes | ⚠️ 16 bytes |
</details>

### Verification

<details>

| after_nav_leading_light | after_nav_leading_dark |
| after_nav_center_light | after_nav_center_dark |
| after_twoline_light | after_twoline_dark |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1825&drop=dogfoodAlpha


ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
```diff
@@ -279,19 +292,39 @@ open class TwoLineTitleView: UIView, TokenizedControlInternal {
...
+                                 alignment: alignment.stackViewAlignment)
 
-        containingStackView.removeAllSubviews()
-        containingStackView.alignment = alignment.stackViewAlignment
-        containingStackView.addArrangedSubview(titleContainer)
-
-        if subtitle?.isEmpty == false {
+        if isTitleImageLeadingForTitleAndSubtitle {
+            guard subtitle?.isEmpty == false, titleImage != nil else {
+                preconditionFailure("A title image and a subtitle must be provided when `isTitleImageLeadingForTitleAndSubtitle` is set to true.")
```
huwilkes, line 300: Do we need the `preconditionFailure` here? If there's no subtitle, how different does it look from just a leading image on the title?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
```diff
@@ -135,7 +135,8 @@ - (void)setupTitleView {
                           subtitle:nil
                    interactivePart:MSFTwoLineTitleViewInteractivePartTitle
                      accessoryType:MSFTwoLineTitleViewAccessoryTypeNone
-       customSubtitleTrailingImage:nil];
+       customSubtitleTrailingImage:nil
+isTitleImageLeadingForTitleAndSubtitle:false];
```
joannaquu, line 139: Why do we need a bool? Can't we just show/hide the image based on whether the leading image is nil?

ios/FluentUI/TwoLineTitleView/TwoLineTitleView.swift
laminesm: The code inside this if statement block assumes the image and subtitle are provided so it shouldn't run if those are nil. We could potentially simply return if the guard fails but the preconditionFailure is useful to clients since it'll alert them when this specific api is being misused.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ObjectiveCDemoController.m
laminesm: We can't do that since titleImage being non nil doesn't tell us if it only gets applied to the title or to both title and subtitle

huwilkes (approved):

PR #1801, opened 2023-06-22
Add a trailing constraint for `CommandBar`'s scroll view

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

The `scrollView` inside of `CommandBar` was missing a trailing constraint, leading to scrolling being broken on RTL. This PR adds the missing constraint to fix the bug.

### Binary change

Total increase: 488 bytes
Total decrease: 0 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 30,388,912 bytes | 30,389,400 bytes | ⚠️ 488 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| CommandBar.o | 206,120 bytes | 206,608 bytes | ⚠️ 488 bytes |
</details>

### Verification

Fix was validated on our demo app.

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before | after |
</details>

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [x] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [x] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://portal.fabricbot.ms/api/codeflow?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1801)

PR #1826, opened 2023-07-13
Buttons in scrollable CommandBar are filling containing view equally, creating undesirable layout

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

An issue was observed with the scrollable `CommandBar` component where items would expand to fill the visible space when only a couple of items are shown in the bar.

This issue was caused by two constraints in `mainCommandGroupsViewConstraints` and the `UIStackView.distribution` property set on the `CommandBarButtonGroupView`.

The change fixes the behavior by:
- Replace the `widthAnchor` constraint with a `trailingAnchor` constraint which pairs with the existing `leadingAnchor` constraint
- Conditionally apply `centerX` constraint when the `CommandBar.isScrollableContentCentered` property is `true`
- Only set the `stackView.distribution = .fillEqually` on `CommandBarButtonGroupView` when the `CommandBar.isScrollable` is `false`. This value is set from `CommandBarCommandGroupsView.updateButtonsShown()` when needed.
- Update the `setEqualWidthGroups()` and `setEqualWidthButtons()` to properties with `didSet` implementations

### Binary change

### Verification

- Verified behavior in CommandBar demo
- Verified behavior in MultilineCommandBar demo
- Verified behavior in affected applications

<details>

**Before:**
Screenshot 2023-07-13 at 11 55 18 AM
_Above image shows that buttons are stretched to fill the view._

**After:**
Screenshot 2023-07-13 at 11 46 29 AM
_Above image shows that buttons are not stretched to fill the view._

Screenshot 2023-07-13 at 12 31 59 PM
_Above image shows that sibling buttons of different sizes do not stretch to fill their containing view equally (see "Body", undo, redo button group)._

Screenshot 2023-07-13 at 12 26 28 PM
_Above image shows that the Multiline CommandBar is unchanged with these modifications._

Screenshot 2023-07-18 at 10 36 03 AM
_Above image shows that the Multiline CommandBar is centered when in landscape._

Screenshot 2023-07-18 at 11 56 28 AM
_Above images that the CommandBar lays out correctly on iPad_

Screenshot 2023-07-18 at 11 56 47 AM
_Above image shows that the Multiline CommandBar lays out correctly on iPad._

Screenshot 2023-07-18 at 11 56 40 AM
_Above image shows that the Multiline CommandBar lays out correctly in landscape on iPad._

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1826&drop=dogfoodAlpha

### DIFF
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -324,11 +324,10 @@ public class CommandBar: UIView, TokenizedControlInternal {
         NSLayoutConstraint.deactivate(mainCommandGroupsViewConstraints)
         if isScrollable {
             mainCommandGroupsViewConstraints = [
-                mainCommandGroupsView.widthAnchor.constraint(equalTo: scrollView.contentLayoutGuide.widthAnchor),
                 mainCommandGroupsView.topAnchor.constraint(equalTo: scrollView.contentLayoutGuide.topAnchor),
                 mainCommandGroupsView.leadingAnchor.constraint(greaterThanOrEqualTo: scrollView.leadingAnchor),
                 mainCommandGroupsView.bottomAnchor.constraint(equalTo: scrollView.contentLayoutGuide.bottomAnchor),
-                mainCommandGroupsView.centerXAnchor.constraint(greaterThanOrEqualTo: scrollView.centerXAnchor)
+                mainCommandGroupsView.trailingAnchor.constraint(greaterThanOrEqualTo: scrollView.trailingAnchor)
             ]
         } else {
             mainCommandGroupsViewConstraints = [
--- a/ios/FluentUI/Command Bar/CommandBar.swift
+++ b/ios/FluentUI/Command Bar/CommandBar.swift
@@ -201,13 +201,24 @@ public class CommandBar: UIView, TokenizedControlInternal {
         }
     }
 
-    /// Whether or not the CommandBar is scrollable
+    /// Whether or not the `CommandBar` is scrollable
+    /// Defaults to `false`
     public var isScrollable: Bool = true {
         didSet {
             updateViewHierarchy()
             updateMainCommandGroupsViewConstraints()
             if !isScrollable {
-                mainCommandGroupsView.setEqualWidthGroups()
+                mainCommandGroupsView.equalWidthGroups = true
+            }
+        }
+    }
+
+    /// Whether or not the `CommandBar` scrollable content is centered in its container
+    /// Defaults to `false`
+    public var isScrollableContentCentered: Bool = false {
+        didSet {
+            if isScrollable {
+                updateMainCommandGroupsViewConstraints()
             }
         }
     }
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -324,11 +324,10 @@ public class CommandBar: UIView, TokenizedControlInternal {
         NSLayoutConstraint.deactivate(mainCommandGroupsViewConstraints)
         if isScrollable {
             mainCommandGroupsViewConstraints = [
-                mainCommandGroupsView.widthAnchor.constraint(equalTo: scrollView.contentLayoutGuide.widthAnchor),
                 mainCommandGroupsView.topAnchor.constraint(equalTo: scrollView.contentLayoutGuide.topAnchor),
                 mainCommandGroupsView.leadingAnchor.constraint(greaterThanOrEqualTo: scrollView.leadingAnchor),
                 mainCommandGroupsView.bottomAnchor.constraint(equalTo: scrollView.contentLayoutGuide.bottomAnchor),
-                mainCommandGroupsView.centerXAnchor.constraint(greaterThanOrEqualTo: scrollView.centerXAnchor)
```
joannaquu, line 331: How does the multilineCommandBar look in landscape mode? This constraint was added so that when the command bar was smaller than the screen width, it would be centered instead of leading aligned.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
```diff
@@ -42,12 +42,15 @@ class CommandBarButtonGroupView: UIView {
         isHidden = allViewsHidden
     }
 
+    func setEqualWidthButtons() {
```
anandrajeswaran, line 45: minor: sounds like a setter but isn't. Either update naming to make clear this is not a setter or add a parameter in some this can be set to true or false? To the latter point, are we ok that this is only one-way settable? We never need to reset it?

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
```diff
@@ -135,4 +138,8 @@ class CommandBarCommandGroupsView: UIView {
         sender.item.handleTapped(sender)
         sender.updateState()
     }
+
+    private func usesEqualWidthGroups() -> Bool {
+        return buttonGroupsStackView.distribution == .fillEqually
```
anandrajeswaran, line 143: PR notes suggest we should be checking whether we're scrollable here to determine whether to use equal width groups? Is this just an indirect way of doing so?

ios/FluentUI/Command Bar/CommandBar.swift
nodes11: Ah, I missed this in the initial pass. I've updated the PR to account for this. A screenshot is attached.

ios/FluentUI/Command Bar/CommandBarButtonGroupView.swift
nodes11: Good point on the naming. I updated this and the similar method in `CommandBarCommandGroupsView` to have property with a default value that updates the views in `didSet`.

ios/FluentUI/Command Bar/CommandBarCommandGroupsView.swift
nodes11: Removed in this in favor of the new property I added.

ios/FluentUI/Command Bar/CommandBar.swift
```diff
@@ -324,12 +333,16 @@ public class CommandBar: UIView, TokenizedControlInternal {
         NSLayoutConstraint.deactivate(mainCommandGroupsViewConstraints)
         if isScrollable {
             mainCommandGroupsViewConstraints = [
-                mainCommandGroupsView.widthAnchor.constraint(equalTo: scrollView.contentLayoutGuide.widthAnchor),
                 mainCommandGroupsView.topAnchor.constraint(equalTo: scrollView.contentLayoutGuide.topAnchor),
-                mainCommandGroupsView.leadingAnchor.constraint(greaterThanOrEqualTo: scrollView.leadingAnchor),
                 mainCommandGroupsView.bottomAnchor.constraint(equalTo: scrollView.contentLayoutGuide.bottomAnchor),
-                mainCommandGroupsView.centerXAnchor.constraint(greaterThanOrEqualTo: scrollView.centerXAnchor)
+                mainCommandGroupsView.leadingAnchor.constraint(greaterThanOrEqualTo: scrollView.leadingAnchor),
+                mainCommandGroupsView.trailingAnchor.constraint(greaterThanOrEqualTo: scrollView.trailingAnchor)
```
joannaquu, line 339: #1801 added a `trailingAnchor` to fix an issue with scrolling in RTL. Now that we're adding the anchor here, can that change be removed? (line 267)

joannaquu (approved):

anandrajeswaran (approved):

PR #1827, opened 2023-07-13
[Fluent Button] Added Border Shadows

amoggo:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes
- Added support to show 3 border shadows for the Primary and Secondary style Buttons. This more closely matches the MacOS System Native Buttons. 
(The shadows are subtle, so observe the secondary Buttons Before/After to see the effect)

### Verification via FluentUI Test App
[x] FluentUI Test App - verified that the shadows show up on all button states - active, inactive and disabled. 
[x] Dark/Light modes

<details>

MicrosoftTeams-image
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1827&drop=dogfoodAlpha


macos/FluentUI/Button/Button.swift
```diff
@@ -121,6 +122,37 @@ open class Button: NSButton {
 		}
 	}
 
+	func setupBorderShadowsIfNeeded() {
+		// At the moment border shadows only apply for the Primary and Secondary Button Styles.
+		// Border shadows being meant for buttons and their corresponding States, that have ~some~
+		// background color other than `clear` which is why they don't apply to the Borderless Style
+		// buttons. Acrylic Style buttons do possess a background color, yet they maintain transparency.
+		// This transparency allows any underlying border shadow layers to bleed through, consequently
+		// impacting the overall button background color. Adding shadows in this context presents a
```
Saadnajmi, line 131: I think this makes total sense, thanks for writing! And if I'm not mistaken, the acrylic style is meant to mimic some system buttons that have a visual effect and no shadow so makes sense.

macos/FluentUI/Button/Button.swift
```diff
@@ -121,6 +122,37 @@ open class Button: NSButton {
...
+		// challenge. Therefore, unless it becomes necessary, these buttons will remain without shadows.
+		// And perhaps shadowPath is the way to go, where you can have the CALayer have a Clear
+		// background color and still show a shadow by setting a hollow shadowPath around the button.
```
Saadnajmi, line 134: I think there's some fancy stuff you can do around clipping content to bounds to only have the outside of the shadow, but yeah for the future.

Saadnajmi: NVM, I see you're already doing some below.

Saadnajmi (approved):

macos/FluentUI/Resources/FluentUI-macos.xcassets/ButtonColors/secondOuterDropShadow.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
+{
+  "colors" : [
+    {
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "0.150",
```
harrieshin, line 7: can we make sure these are hex not cgfloat to match the figma files?

macos/FluentUI/Resources/FluentUI-macos.xcassets/ButtonColors/secondOuterDropShadow.colorset/Contents.json
```diff
@@ -0,0 +1,38 @@
...
+      "appearances" : [
+        {
+          "appearance" : "luminosity",
+          "value" : "dark"
+        }
+      ],
+      "color" : {
+        "color-space" : "srgb",
+        "components" : {
+          "alpha" : "0.250",
```
Saadnajmi, line 25: Alphas are still decimal, not hex? I think designers specify these using hex so there's still a lossy conversion?

amoggo: Erm, isn't alpha a %? How would you specify that in Hex? Also all the pre-existing ColorSets in ButtonColors have a decimal format alpha

Saadnajmi: An 8 digit hex string. An example:
https://github.com/microsoft/fluentui-design-tokens/blob/main/src/black.json#L14

If Xcode doesn't let you / your designs are in decimal, then that's fine.

brentpe (approved):

macos/FluentUI/Resources/FluentUI-macos.xcassets/ButtonColors/secondOuterDropShadow.colorset/Contents.json
amoggo: the example you linked in a color and not alpha value? I can't find out how to express alpha in hex through Xcode, and don't see examples of this used in Contents.json so just going with what we have for now

Saadnajmi: For `"#0000001f"`, the first 6 digits are the color, and the last 2 (1f) is the alpha. But no worries, if Xcode can't do it then that's fine :)

macos/FluentUI/Button/Button.swift
```diff
@@ -121,6 +122,37 @@ open class Button: NSButton {
...
+		self.usesBorderShadows = style == .primary || style == .secondary
```
anandrajeswaran, line 135: 2 minor things:
1. can omit `self` for `self.usesBorderShadows`
2. Tend to prefer switch/case statements when dealing with enums. That forces new enum values to explicitly make behavioral decisions at compile time, which generally makes that experience much easier.

macos/FluentUI/Button/Button.swift
```diff
@@ -121,6 +122,37 @@ open class Button: NSButton {
...
+
+	open override func layout() {
+		super.layout()
+		if usesBorderShadows {
+			// Set the frame of all the shadow layers to the same bounds as this Button View. This
+			// couldn't have been done in the init where the shadow layers are setup because the Button
+			// view doesn't yet have any bounds established, so we're forced to do it in the layout
+			// override where we're guaranteed to have bounds and is also invoked every time the view's
+			// size or position needs adjustments.
+			firstOuterDropShadowLayer?.frame = self.bounds
```
anandrajeswaran, line 150: should generally assign to a layer's bounds instead of frame since frame is a computed property. From the documentation, modifying frame could lead to side-effects by modifying the other properties which factor into it (position, anchorPoint)

macos/FluentUI/Button/Button.swift
```diff
@@ -247,6 +279,30 @@ open class Button: NSButton {
...
+			updateShadowLayer(shadowLayer: firstOuterDropShadowLayer,
+							  shadowColor: ButtonColor.firstOuterDropShadow?.cgColor)
+			updateShadowLayer(shadowLayer: secondOuterDropShadowLayer,
+							  shadowColor: ButtonColor.secondOuterDropShadow?.cgColor)
+			updateShadowLayer(shadowLayer: innerShadowLayer,
+							  shadowColor: ButtonColor.innerShadow?.cgColor)
+		}
+	}
+
+	func updateShadowLayer(shadowLayer: CALayer?, shadowColor: CGColor?) {
```
anandrajeswaran, line 297: In this particular func, don't we always want shadowLayer and shadowColor to be non-optional? Can use if/lets in the caller to enforce that

macos/FluentUI/Button/Button.swift
```diff
@@ -247,6 +279,30 @@ open class Button: NSButton {
 			layer.backgroundColor = backgroundColorRest?.cgColor
 			layer.borderColor = borderColorRest?.cgColor
 		}
+
+		if usesBorderShadows {
```
anandrajeswaran, line 283: Per below comment, rather than checking this, we can check if the specific layers we want to update are non null using if/lets. Just a tiny bit more direct - if the layers we want to modify exist, then we are free to operate on them

macos/FluentUI/Button/Button.swift
```diff
@@ -433,11 +489,62 @@ open class Button: NSButton {
...
+			}
+
+			if usesBorderShadows {
+				firstOuterDropShadowLayer = CALayer()
+				secondOuterDropShadowLayer = CALayer()
+				innerShadowLayer = CALayer()
+
+				// Ensure that the Shadow sublayers can in fact extend beyond the bounds of the Button View's
+				// layer so that the shadows don't get clipped
+				self.layer?.masksToBounds = false
```
anandrajeswaran, line 511: Probably want to `guard let layer` early to avoid needing to intersperse a bunch of optional unwraps throughout the code. Usually good for clarity to just do it in one place.

macos/FluentUI/Button/Button.swift
```diff
@@ -433,11 +489,62 @@ open class Button: NSButton {
...
+
+				if let firstOuterDropShadowLayer = firstOuterDropShadowLayer {
```
anandrajeswaran, line 513: can create local (non-optional) variables with a different name and then assign to firstOuterDropShadowLayer at the end of this func to avoid this if lets when we know the object is nonnull

anandrajeswaran, line 513: These blocks are all identical minus some specific values. Thoughts on using a private enum ShadowLayerType which has shadowOffset/shadowRadius/shadowOpacity properties and then use a func which takes in a ShadowLayerType and executes one of these blocks?

Then this repetitive code becomes
```
update(shadowLayer: firstOuterDropShadow, type: .firstOuter)
update(shadowLayer: secondOuterDropShadow, type: .secondOuter)
update(shadowLayer:innerShadow, type: .inner)
```

macos/FluentUI/Button/Button.swift
```diff
@@ -732,6 +839,9 @@ public struct ButtonFormat {
 
 @objc(MSFButtonColor)
 class ButtonColor: NSObject {
+	public static let firstOuterDropShadow = NSColor(named: "ButtonColors/firstOuterDropShadow", bundle: FluentUIResources.resourceBundle)
+	public static let secondOuterDropShadow = NSColor(named: "ButtonColors/secondOuterDropShadow", bundle: FluentUIResources.resourceBundle)
+	public static let innerShadow = NSColor(named: "ButtonColors/innerShadow", bundle: FluentUIResources.resourceBundle)
 	public static let brandForegroundDisabled = NSColor(named: "ButtonColors/brandForegroundDisabled", bundle: FluentUIResources.resourceBundle)
```
anandrajeswaran, line 845: Not new, but we should probably have a wrapper function in Color which returns colors from the bundle as non-optional and crashes if it doesn't exist so every caller doesn't need to handle nil

macos/FluentUI/Button/Button.swift
amoggo: When I do this, the layer's positioning gets messed up. After a quick bit of research, it looks like frame gets the correct result because we're positioning the layer within the view (the layer's parent), so we're working in the parent's coordinate system. But using layer.bounds instead possibly is getting messy, because this could be modifying the layer's own coordinate system instead of its size and position in the parent's coordinate system. 

Screenshot 2023-07-18 at 10 03 24

macos/FluentUI/Button/Button.swift
amoggo: defeats the purpose of using usesBorderShadows, which was added to keep code cleaner, but it sounds like it sacrifices correctness?

anandrajeswaran: I think it's just another axis of keeping code cleaner - from what I've seen, persisting optional-ness longer in swift codepaths has abnormally large impact on code cleanliness given all the unwraps/"if let"s it forces

macos/FluentUI/Button/Button.swift
anandrajeswaran: All this is reminding me is I love views and autolayout :). But yeah going with bounds seems correct enough because we, in practice, are not modifying anchorPoint/position. I think I'd like to set aside some time to go deep dive into this stuff again. Is it possible the "right" thing to do is set frame, but we also need to convert from bounds from our layer's coordinate system to the sublayer's coordinate system?

PR #1830, opened 2023-07-17
Add script to bump version strings

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

This PR adds a script `BumpVersion.swift` that bumps all relevant strings to the version provided. Essentially the script is using regex patterns, parsing through the podspec and plists to update the version strings.
The script will also check that the provided version follows semantic versioning rules.

### Verification

<details>

Screenshot 2023-07-17 at 12 50 35 PM

</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/scripts/BumpVersion.swift
+++ b/scripts/BumpVersion.swift
@@ -1,0 +1,29 @@
+import Foundation
+
+let currentDirectory = FileManager.default.currentDirectoryPath
+let filePath = "\(currentDirectory)/MicrosoftFluentUI.podspec"
+
+guard var fileContents = try? String(contentsOfFile: filePath, encoding: .utf8) else {
+    print("Failed to read MicrosoftFluentUI.podspec file.")
+    exit(0)
+}
+
+let regexPattern = "s.version\\s*=\\s*'([^']*)'"
+let newValue = "0.20.0"
+
+if let range = fileContents.range(of: regexPattern, options: .regularExpression) {
+    let oldValue = fileContents[range]
+    let updatedValue = oldValue.replacingOccurrences(of: oldValue, with: "s.version          = '\(newValue)'")
+    fileContents.replaceSubrange(range, with: updatedValue)
+} else {
+    print("Failed to find the field s.version in the MicrosoftFluentUI.podspec file.")
+    exit(0)
+}
+
+do {
+    try fileContents.write(toFile: filePath, atomically: true, encoding: .utf8)
+    print("Successfully updated the value of s.version in the MicrosoftFluentUI.podspec file.")
+} catch {
+    print("Failed to write to the MicrosoftFluentUI.podspec file.")
+}
+
--- a/scripts/BumpVersion.swift
+++ b/scripts/BumpVersion.swift
@@ -1,15 +1,27 @@
 import Foundation
 
-let currentDirectory = FileManager.default.currentDirectoryPath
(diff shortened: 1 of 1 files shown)


harrieshin:
can we write how to use this script in fluent wiki please?

joannaquu:
image
I've never seen this with the pr checks before, can you try rerunning?

scripts/BumpVersion.swift
```diff
@@ -0,0 +1,169 @@
...
+        return false
+    }
+
+    if newMajor - currentMajor > 1 {
+        print("\(yellowColor)Warning: New version major number should not jump by more than 1.\(redColor)")
+    }
+
+    if newMajor > currentMajor  {
+        if newMinor != 0 {
+            print("\(yellowColor)Warning: New major version should have a minor number of 0.\(resetColor)")
```
huwilkes, line 43: Why can't the new major version have a minor version of 0? Do we not want to be able to go to `1.0.0`?

scripts/BumpVersion.swift
```diff
@@ -0,0 +1,169 @@
...
+
+    if newMinor - currentMinor > 1 {
+        print("\(yellowColor)Warning: New version minor number should not jump by more than 1.\(resetColor)")
+    } else if newMinor > currentMinor && newPatch != 0 {
+        print("\(yellowColor)Warning: New minor version should have a patch number of 0.\(resetColor)")
+    }
+
+    if newPatch - currentPatch > 1 {
+        print("\(yellowColor)Warning: New version patch number should not jump by more than 1.\(resetColor)")
+    }
```
huwilkes, line 57: I wonder if it might be better to just take in `-M, --major`, `-m, --minor`, and `-p, --patch`, and just bump the correct number instead of doing all the work to parse input and validate it.

laminesm: That's a good point, I think it would've made the implementation easier from the beginning for sure. But since input validation and parsing is already implemented, I think just taking in the version as a single argument is the easier option for users.

huwilkes (approved):

scripts/BumpVersion.swift
```diff
@@ -0,0 +1,169 @@
...
+let resetColor = "\u{001B}[0;0m"
+let redColor = "\u{001B}[0;31m"
+let yellowColor = "\u{001B}[33m"
+
+guard CommandLine.arguments.count == 2 else {
+    print("\(redColor)Please provide a version in the format `major.minor.patch`\(resetColor)")
+    exit(1)
+}
+
+func isVersionBumped(_ currentVersion: String, _ newVersion: String) -> Bool {
```
Saadnajmi, line 13: Rather than implement this manually, we use a package called `semver` in React Native macOS to help. This is where we use it: https://github.com/microsoft/react-native-macos/blob/main/.ado/get-next-semver-version.js

A quick google says similar stuff may be available for Swift: https://swiftpackageindex.com/sersoft-gmbh/semver

laminesm: This seems pretty cool actually! Though I'm not not sure if it's worth adding this as a package in fluent

Saadnajmi: Yeah that makes sense. Thought I'd share anyway

laminesm: Totally, thanks for doing so!

PR #1842, opened 2023-07-21
Fix bug in which showProfileButton was being reset to true for AvatarTitleView

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

A [recent change](https://github.com/microsoft/fluentui-apple/pull/1825/files#diff-e6ef8e9f234a1b2b6135e4197109a637a24fb4e5f86f126ec6080f5f8e81c46f) was hiding the Avatar of the AvatarTitleView when the title image is the leading image for both the title and subtitle. 

The issue this created was that it updated the visibility of the avatar all the time. So for scenarios in which "showProfileButton" was false, it was reset to true. This meant that even if personData was nil for the AvatarTitleView we still showed a default avatar view.

The change makes it so we only set the value to false when updating.

### Binary change

Did not examine.

### Verification

Tested in the Fluent Test app.
- Opened TwoLineTitleView. Tested that the added scenarios from this [recent change](https://github.com/microsoft/fluentui-apple/pull/1825/files#diff-e6ef8e9f234a1b2b6135e4197109a637a24fb4e5f86f126ec6080f5f8e81c46f) still worked.
- Tested NSFNavigationController. 
    - When shown without Avatar, the avatar is not in the navigation bar
    - When shown with a custom leading button, the avatar is not in the navigation bar
    - When personData is set, the avatar appears 

| Scenario | Before                               | After                                      |
|----|-------------------------------|--------------------------------------------|
| "Show without an avatar" | image| image |
| "Show with a custom leading button": no person data provided | image| image|
| "Show without accessory": person data is provided | image| image|

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1842&drop=dogfoodAlpha


ios/FluentUI/Navigation/Views/AvatarTitleView.swift
```diff
@@ -322,7 +322,9 @@ class AvatarTitleView: UIView, TokenizedControlInternal, TwoLineTitleViewDelegat
         twoLineTitleView.setup(navigationItem: navigationItem)
 
         // Hide the avatar if TwoLineTitleView has a leading image for both title and subtitle.
-        showsProfileButton = !navigationItem.isTitleImageLeadingForTitleAndSubtitle
+        if navigationItem.isTitleImageLeadingForTitleAndSubtitle {
+            showsProfileButton = false
```
huwilkes, line 326: Would we ever need to set it to true again?

huwilkes (approved):
Since you're doing a squash merge it won't really matter, probably, but it looks like your working of a branch of your TVC branch, so there's a bunch of extra commits.

ios/FluentUI/Navigation/Views/AvatarTitleView.swift
alexanderboswell: good point. I think we will need to recalculate the visibility of the button for the "else" scenario. I have updated to reflect that.

PR #1843, opened 2023-07-24
Tokenize BadgeField

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Tokenized `BadgeField`. Previously, `labelView` and `placeholderView`'s colors were set through the label's `colorStyle`. To take advantage of our token system, they are now set through `textColor`. This means that instead of using Label's `secondary` style as the default color, it uses the `foreground2` token directly.

### Binary change

Total increase: 129,376 bytes
Total decrease: 0 bytes
| Total | 46,602,704 bytes | 46,732,080 bytes | 🛑 129,376 bytes |
<details>

| BadgeFieldTokenSet.o | 0 bytes | 105,216 bytes | 🛑 105,216 bytes |
| BadgeField.o | 950,208 bytes | 965,960 bytes | ⚠️ 15,752 bytes |
| __.SYMDEF | 2,885,544 bytes | 2,893,464 bytes | ⚠️ 7,920 bytes |
| PeoplePicker.o | 465,696 bytes | 466,184 bytes | ⚠️ 488 bytes |
</details>

### Verification

<details>

| Before                                       | After                                      | Theme Overrides                                      | Control Overrides                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|--------------------------------------------|
| before light | after light | theme light | per light |
| before dark | after dark | theme dark | per dark |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1843&drop=dogfoodAlpha

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -159,6 +159,11 @@ open class BadgeField: UIView, TokenizedControlInternal {
 
     @objc public init() {
         super.init(frame: .zero)
+        NotificationCenter.default.addObserver(self,
+                                               selector: #selector(themeDidChange),
+                                               name: .didChangeTheme,
+                                               object: nil)
+
         addSubview(labelView)
         addSubview(placeholderView)
 
--- a/ios/FluentUI/Badge Field/BadgeField.swift
+++ b/ios/FluentUI/Badge Field/BadgeField.swift
@@ -57,18 +57,13 @@ public protocol BadgeFieldDelegate: AnyObject {
  * voiceover and dynamic text sizing
  */
 @objc(MSFBadgeField)
-open class BadgeField: UIView {
+open class BadgeField: UIView, TokenizedControlInternal {
     private struct Constants {
-        static let badgeMarginHorizontal: CGFloat = 5
-        static let badgeMarginVertical: CGFloat = 5
         static let emptyTextFieldString: String = ""
         static let dragAndDropMinimumPressDuration: TimeInterval = 0.2
         static let dragAndDropScaleAnimationDuration: TimeInterval = 0.3
         static let dragAndDropScaleFactor: CGFloat = 1.10
         static let dragAndDropPositioningAnimationDuration: TimeInterval = 0.2
-        static let labelMarginRight: CGFloat = 5
-        static let labelColorStyle: TextColorStyle = .secondary
-        static let textFieldMinWidth: CGFloat = 100
     }
 
     @objc open var label: String = "" {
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Badge Field/BadgeFieldTokenSet.swift
```diff
@@ -0,0 +1,53 @@
...
+        case textFieldColor
+
+        /// The font of the BadgeField's label.
+        case labelFont
+
+        /// The font of the BadgeField's placeholder.
+        case placeholderFont
+
+        /// The font of the BadgeField's text field.
+        case textFieldFont
```
huwilkes, line 30: Should we add tokens for customizing the badges?

mischreiber: Fair question -- my inclination is to add it as an issue in GitHub and then wait to see if we have any client interest, since I don't currently know of anyone who's been asking for that functionality. Either way, definitely not a requirement for this PR.

ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -200,17 +188,36 @@ open class BadgeField: UIView {
...
 
     @objc private func themeDidChange(_ notification: Notification) {
         guard let themeView = notification.object as? UIView, self.isDescendant(of: themeView) else {
             return
         }
-        updateBackgroundColor()
+        updateTokenizedValues()
+    }
+
+    private func updateTokenizedValues() {
```
laminesm, line 205: nit: could this be renamed to `updateAppearance` instead?

ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -316,6 +327,15 @@ open class BadgeField: UIView {
         return CGSize(width: UIView.noIntrinsicMetric, height: contentHeight(forBoundingWidth: frame.width))
     }
 
+    open override func willMove(toWindow newWindow: UIWindow?) {
+        super.willMove(toWindow: newWindow)
+        guard let newWindow else {
+            return
+        }
+        tokenSet.update(newWindow.fluentTheme)
+        updateAppearance()
```
huwilkes, line 336: I think you shouldn't need the `updateAppearance()` here, since setting the theme through `tokenSet.update` should trigger what you added with `registerOnUpdate` if the theme changes.

huwilkes (approved):

ios/FluentUI/Badge Field/BadgeField.swift
```diff
@@ -200,17 +193,36 @@ open class BadgeField: UIView {
...
         // An accessible container must set isAccessibilityElement to false
         isAccessibilityElement = false
+
+        // Update appearance whenever `tokenSet` changes.
+        tokenSet.registerOnUpdate(for: self) { [weak self] in
+            self?.updateAppearance()
+        }
     }
 
     @objc private func themeDidChange(_ notification: Notification) {
```
laminesm, line 203: Also shouldn't this function be removed since we have `registerOnUpdate`?

joannaquu: That was my original thought but I noticed is was still in `Separator` so I wasn't sure if behind the scenes it was still necessary.

laminesm (approved):

PR #1845, opened 2023-07-25
ListItem initial MVP

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

This is the first of many PRs for creating a swiftUI native control for `MSFTableViewCell`. The goal of this PR is to provide the initial MVP that supports many of the common scenarios as a basis for supporting all of the functionality of `MSFTableViewCell`.

It builds upon the [initial investigations](https://github.com/microsoft/fluentui-apple/blob/fluent2-tokens/ios/FluentUI/SwiftUI%20Experiments/List/ListCell.swift) done to create a list control that is available to consume in UIKit and SwiftUI. But after discussion, the decision came that if the control is needed in UIKit `MSFTableViewCell` is sufficient, so the control should be built only for consumption in SwiftUI. It removes the UIKit wrapper that was done in the experimentation and uses SwiftUI conventions to allow for customization of the view.

Changes addressed in this PR:

- `ListItem.swift`
     -  Add support for a title, subtitle, and footer
     - Add support for a leadingView and trailingView.
         - Used `@ViewBuilder` to support passing in any view and follow the convention of other controls provided by apple. 
     - Create alias for tokens and enums used by `MSFTableViewCell` as the same values will be needed. This was to reduce binary size impact, ease of maintenance, and not requiring creating another set of tokens for the same control visually. As this is only a difference in engineering implementation (UIKit vs swiftUI), there shouldnt be a design difference.
 -  `ListItemModifiers.swift`
     - Follows the convention of other SwiftUI controls (Avatar, Notification, etc) by allowing convenience functions for changing state of the control. This makes it so the control doesnt have a massive initializer.
     - Exposes all of the customization of `ListItem` and follows the syntax of a swiftUI view. Example:
``` .swift

ListItem(title: "Title of my list item")
         .accessoryType(.checkmark)
         .titleLineLimit(2)
```` 
- `ListItemState.swift`
   -  Defines all of the properties that can be customized on the `ListItem` view.

### Example usage

``` .swift
List {
   ListItem(title: "Just a title")
   ListItem(title: "A title", subtitle: "And a subtitle too!") 
   ListItem(title: "A three line configuration", subtitle: "This is the second line", footer: "This is the third line")
   ListItem(title: "Just a custom leading view", leadingContent: {
      MyCustomLeadingView()
   })
   ListItem(title: "Just a custom trailing view", trailingContent: {
      MyCustomTrailingView()
   })
   ListItem(title: "A custom leading and trailing view", leadingContent: { 
      MyCustomLeadingView()
   }, trailingContent: {
      MyCustomTrailingView()
   })
   ListItem(title: "A disclosure indicator accessory")
            .accessoryType(.disclosureIndicator)
   ListItem(title: "A check mark accessory")
            .accessoryType(.checkmark)
   ListItem(title: "A detail button accessory")
            .accessoryType(.detailButton)
            .onAccessoryTapped {
                print("The button was tapped!")
            }
   ListItem(title: "A list item with everything", subtitle: "This is a subtitle", footer: "This is a footer", leadingContent: {
        MyCustomLeadingView()
   }, trailingContent: {
      MyCustomTrailingView()
   })
            .accessoryType(.checkmark)
}
```

### Binary change

The size impact was a known before making this change. Further analysis will be made once all of the changes are on the feature branch and before merging to main.

| libFluentUI.a | 30,118,904 bytes | 30,414,816 bytes | ⛔️ 295,912 bytes |
| FluentUI.Demo | 33,764,955 bytes | 33,884,203 bytes | 🛑 119,248 bytes |
<details>

| ListItem.o | 0 bytes | 194,392 bytes | 🛑 194,392 bytes |
| __.SYMDEF | 4,561,232 bytes | 4,605,136 bytes | ⚠️ 43,904 bytes |
| ListItemState.o | 0 bytes | 34,536 bytes | ⚠️ 34,536 bytes |
| ListItemModifiers.o | 0 bytes | 18,032 bytes | ⚠️ 18,032 bytes |
| FocusRingView.o | 793,488 bytes | 798,536 bytes | ⚠️ 5,048 bytes |
</details>

### Verification
Examined in the fluent test app. Adding a demo and XCUITests are out of scope for this PR and will be in the next set of changes.

| iPhone 14 Pro                                  | iPad Pro (12.9 inch)                                      |
| Simulator Screenshot - iPhone 14 Pro - 2023-07-25 at 14 51 45 | Simulator Screenshot - iPad Pro (12 9-inch) (6th xgeneration) - 2023-07-25 at 14 49 57 |

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -11,10 +11,10 @@ public typealias ListItemTokenSet = TableViewCellTokenSet
 
 /// View that represents an item in a List.
 public struct ListItem<LeadingView: View,
-					   TrailingView: View,
-					   Title: StringProtocol,
-					   Subtitle: StringProtocol,
-					   Footer: StringProtocol>: View {
+                       TrailingView: View,
+                       Title: StringProtocol,
+                       Subtitle: StringProtocol,
+                       Footer: StringProtocol>: View {
 
 	// MARK: Initializer
 
@@ -25,7 +25,7 @@ public struct ListItem<LeadingView: View,
 	///   - footer: Text that appears as the third line of text
 	///   - leadingView: The view that appears on the leading edge of the view
 	///   - trailingView: The view that appears on the trailing edge of the view, next to the accessory type if provided
-	public init(title: Title,
+    public init(title: Title,
                 subtitle: Subtitle = String(""),
                 footer: Footer = String(""),
                 leadingView: @escaping () -> LeadingView = { EmptyView() },
@@ -116,10 +116,10 @@ public struct ListItem<LeadingView: View,
 					.tint(Color(fluentTheme.color(.brandForeground1)))
 				accessoryView
 			}
-			.padding(EdgeInsets(top: ListItemTokenSet.paddingVertical + 1,
-								leading: ListItemTokenSet.paddingLeading,
-								bottom: ListItemTokenSet.paddingVertical + 1,
-								trailing: ListItemTokenSet.paddingTrailing))
+            .padding(EdgeInsets(top: ListItemTokenSet.paddingVertical + 1,
+                                leading: ListItemTokenSet.paddingLeading,
+                                bottom: ListItemTokenSet.paddingVertical + 1,
+                                trailing: ListItemTokenSet.paddingTrailing))
 			.frame(minHeight: minHeight)
 			.background(backgroundView)
 			.listRowInsets(EdgeInsets())
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -11,8 +11,8 @@ public typealias ListItemCustomViewSize = MSFTableViewCellCustomViewSize
 public typealias ListItemTokenSet = TableViewCellTokenSet
 
 /// View that represents an item in a List.
-public struct ListItem<LeadingView: View,
-                       TrailingView: View,
+public struct ListItem<LeadingContent: View,
+                       TrailingContent: View,
                        Title: StringProtocol,
                        Subtitle: StringProtocol,
                        Footer: StringProtocol>: View {
@@ -24,18 +24,18 @@ public struct ListItem<LeadingView: View,
 	///   - title: Text that appears as the first line of text
 	///   - subtitle: Text that appears as the second line of text
 	///   - footer: Text that appears as the third line of text
-	///   - leadingView: The view that appears on the leading edge of the view
-	///   - trailingView: The view that appears on the trailing edge of the view, next to the accessory type if provided
+	///   - leadingContent: The content that appears on the leading edge of the view
+	///   - trailingContent: The content that appears on the trailing edge of the view, next to the accessory type if provided
     public init(title: Title,
                 subtitle: Subtitle = String(),
                 footer: Footer = String(),
-                @ViewBuilder leadingView: @escaping () -> LeadingView,
-                @ViewBuilder trailingView: @escaping () -> TrailingView) {
+                @ViewBuilder leadingContent: @escaping () -> LeadingContent,
+                @ViewBuilder trailingContent: @escaping () -> TrailingContent) {
         self.title = title
         self.subtitle = subtitle
         self.footer = footer
-        self.leadingView = leadingView()
-        self.trailingView = trailingView()
+        self.leadingContent = leadingContent
+        self.trailingContent = trailingContent
     }
 
     public var body: some View {
@@ -110,19 +110,23 @@ public struct ListItem<LeadingView: View,
         @ViewBuilder
         var contentView: some View {
             HStack(alignment: .center) {
-                leadingView
-                    .padding(.trailing, ListItemTokenSet.horizontalSpacing)
+                if let leadingContent {
+                    leadingContent()
+                        .padding(.trailing, ListItemTokenSet.horizontalSpacing)
+                }
                 labelStack
                     .padding(.trailing, ListItemTokenSet.horizontalSpacing)
                 Spacer()
-                trailingView
-                    .tint(Color(fluentTheme.color(.brandForeground1)))
+                if let trailingContent {
+                    trailingContent()
+                        .tint(Color(fluentTheme.color(.brandForeground1)))
+                }
                 accessoryView
                     .padding(.leading, ListItemTokenSet.horizontalSpacing)
             }
-            .padding(EdgeInsets(top: ListItemTokenSet.paddingVertical + 1,
+            .padding(EdgeInsets(top: ListItemTokenSet.paddingVertical,
                                 leading: ListItemTokenSet.paddingLeading,
-                                bottom: ListItemTokenSet.paddingVertical + 1,
+                                bottom: ListItemTokenSet.paddingVertical,
                                 trailing: ListItemTokenSet.paddingTrailing))
             .frame(minHeight: layoutType.minHeight)
             .background(backgroundView)
@@ -132,7 +136,7 @@ public struct ListItem<LeadingView: View,
         return contentView
     }
 
-    var state: ListItemState = ListItemState()
+    @StateObject var state: ListItemState = ListItemState()
 
     private var layoutType: LayoutType {
         if !subtitle.isEmpty {
@@ -163,8 +167,8 @@ public struct ListItem<LeadingView: View,
 
     @Environment(\.fluentTheme) private var fluentTheme: FluentTheme
 
-    private var leadingView: LeadingView?
-    private var trailingView: TrailingView?
+    private var leadingContent: (() -> LeadingContent)?
+    private var trailingContent: (() -> TrailingContent)?
 
     private let footer: Footer
     private let subtitle: Subtitle
@@ -174,7 +178,7 @@ public struct ListItem<LeadingView: View,
 
 // MARK: Additional Initializers
 
-public extension ListItem where LeadingView == EmptyView, TrailingView == EmptyView {
+public extension ListItem where LeadingContent == EmptyView, TrailingContent == EmptyView {
     init(title: Title,
          subtitle: Subtitle = String(),
          footer: Footer = String()) {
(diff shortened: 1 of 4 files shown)


ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,154 @@
...
+                    .padding(.trailing, ListItemTokenSet.horizontalSpacing)
+                labelsView
+                    .padding(.trailing, ListItemTokenSet.horizontalSpacing)
+                Spacer()
+                trailingView()
+                    .tint(Color(fluentTheme.color(.brandForeground1)))
+                accessoryView
+                    .padding(.leading, ListItemTokenSet.horizontalSpacing)
+            }
+            .padding(EdgeInsets(top: ListItemTokenSet.paddingVertical + 1,
```
alexanderboswell, line 119: @huwilkes and I were discussing this. I am unsure of what to do for this scenario as the padding in the SwiftUI control should be 12 according to the design. But the token for the TableViewCell is 11 and it isnt super clear why yet.

So for now I did this not so ideal work around.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,154 @@
...
+    
+    @Environment(\.fluentTheme) private var fluentTheme: FluentTheme
+    
+    @ViewBuilder private let leadingView: () -> LeadingView
+    @ViewBuilder private let trailingView: () -> TrailingView
+    
+    private let footer: Footer
+    private let subtitle: Subtitle
+    private let title: Title
+    private let tokenSet: ListItemTokenSet = ListItemTokenSet(customViewSize: { MSFTableViewCellCustomViewSize.zero })
```
alexanderboswell, line 153: Suggestions from anyone on if I should also do a typealias for MSFTableViewCellCustomViewSize to be `ListItemCustomViewSize`? 

 Maybe something to worry about in another iteration when more of the custom view properties are added.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,154 @@
...
+        self.footer = footer
+        self.leadingView = leadingView
+        self.trailingView = trailingView
+    }
+
+    public var body: some View {
+        tokenSet.update(fluentTheme)
+
+        @ViewBuilder
+        var labelsView: some View {
```
huwilkes, line 44: Maybe `labelStack`?

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,154 @@
...
+        
+        return contentView
+    }
+
+    var state: ListItemState = ListItemState()
+    
+    private var minHeight: CGFloat {
+        if !subtitle.isEmpty {
+            if !footer.isEmpty {
+                return ListItemTokenSet.threeLineMinHeight
```
huwilkes, line 136: Do we need/want any guards against setting the footer but not the subtitle? It looks like it'd be possible to set the footer and the title, and then the min height would be for one line.

alexanderboswell: So I have a similar check in the labels stack, so a footer cant be shown if a subtitle is empty.

I was thinking about introducing an enum like in MSFTableViewCell for layoutType:
``` .swift
    fileprivate enum LayoutType {
        case oneLine
        case twoLines
        case threeLines
    }
```

So I could have a computed variable to determine the layoutType based title, subtitle, and footer values.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,154 @@
...
+	/// - Parameters:
+	///   - title: Text that appears as the first line of text
+	///   - subtitle: Text that appears as the second line of text
+	///   - footer: Text that appears as the third line of text
+	///   - leadingView: The view that appears on the leading edge of the view
+	///   - trailingView: The view that appears on the trailing edge of the view, next to the accessory type if provided
+    public init(title: Title,
+                subtitle: Subtitle = String(""),
+                footer: Footer = String(""),
+                leadingView: @escaping () -> LeadingView = { EmptyView() },
```
segamalamine, line 31: I'm not well versed in SwiftUI so apologies if this is a stupid question but why not just use optionals here instead of `EmptyView()`? (Especially considering SwiftUI documentation states "You will rarely, if ever, need to create an EmptyView directly" which sounds like this should be steered away from unless there's no other way).

laminesm, line 31: I'm not well versed in SwiftUI so apologies if this is a stupid question but why not just use optionals here instead of EmptyView()? (Especially considering SwiftUI documentation states "You will rarely, if ever, need to create an EmptyView directly" which sounds like this should be steered away from unless there's no other way).

ios/FluentUI/List/ListItemState.swift
```diff
@@ -0,0 +1,36 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// Properties available to customize the appearance of a `ListItem`
```
laminesm, line 8: nit: missing periods on end of documentations

ios/FluentUI/List/ListItem.swift
laminesm: I think you should  as it'd make it more consistent.

ios/FluentUI/List/ListItem.swift
alexanderboswell: hmm I was struggling to get it to compile as having an optional generic view isnt allowed. Basically, when the view is created it needs to know what the generic type `LeadingView` is, and having a optional puts it in a state in which it wouldnt know (I dont know if this is making any sense ha).


But I think I found a way around it. Basically will need to provide some connivence initializers, but that should be possible.

Here is a good post on stack overflow that answers the question on the updated changes: https://stackoverflow.com/a/62975146

alexanderboswell: at the bottom of the file under `MARK: Additional Initializers`, you can see that it will allow optional for the leading and trailing views.

huwilkes (approved):

ios/FluentUI/List/ListItem.swift
lcapkovic: IMO this has higher probability of causing a hard to find future bug once we forget about this +1. Probably better to just follow what the token has.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,209 @@
...
+
+import SwiftUI
+
+public typealias ListItemAccessoryType = TableViewCellAccessoryType
+public typealias ListItemBackgroundStyleType = TableViewCellBackgroundStyleType
+public typealias ListItemCustomViewSize = MSFTableViewCellCustomViewSize
+public typealias ListItemTokenSet = TableViewCellTokenSet
+
+/// View that represents an item in a List.
+public struct ListItem<LeadingView: View,
```
lcapkovic, line 14: LeadingContent / TrailingContent (also in the public facing param name) seems to be more platform native. See HStack header definition.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,209 @@
...
+	///   - trailingView: The view that appears on the trailing edge of the view, next to the accessory type if provided
+    public init(title: Title,
+                subtitle: Subtitle = String(),
+                footer: Footer = String(),
+                @ViewBuilder leadingView: @escaping () -> LeadingView,
+                @ViewBuilder trailingView: @escaping () -> TrailingView) {
+        self.title = title
+        self.subtitle = subtitle
+        self.footer = footer
+        self.leadingView = leadingView()
```
lcapkovic, line 37: Do we know if there are perf implications of calling this on init, instead of storing the ViewBuilder and only calling it from body? I'd gravitate towards the latter to be safe. Init can get called *much* more than view body.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,209 @@
...
+
+        @ViewBuilder
+        var accessoryView: some View {
+            if let accessoryType = state.accessoryType,
+               let icon = accessoryType.icon,
+               let iconColor = accessoryType.iconColor(tokenSet: tokenSet, fluentTheme: fluentTheme) {
+                let image = Image(uiImage: icon)
+                    .foregroundColor(Color(uiColor: iconColor))
+                if accessoryType == .detailButton {
+                    SwiftUI.Button {
```
lcapkovic, line 90: Why is the SwiftUI prefix needed?

ios/FluentUI/List/ListItem.swift
alexanderboswell: so the trade off was having default values in the init like `leadingView = { EmptyView()}` or not. @ViewBuilder cant have an optional value, so this was trying to address how we could have leadingView and trailingView as nil in the private variable instead of creating `EmptyView()`


This conversation from the first iteration: https://github.com/microsoft/fluentui-apple/pull/1845#discussion_r1274263135

ios/FluentUI/List/ListItem.swift
huwilkes: Fluent's UIKit button is also just named `Button`.

ios/FluentUI/List/ListItem.swift
```diff
@@ -0,0 +1,209 @@
...
+                                trailing: ListItemTokenSet.paddingTrailing))
+            .frame(minHeight: layoutType.minHeight)
+            .background(backgroundView)
+            .listRowInsets(EdgeInsets())
+        }
+
+        return contentView
+    }
+
+    var state: ListItemState = ListItemState()
```
lcapkovic, line 135: This will get recreated on every view re-init, which can happen at any time. SwiftUI destroys and rebuilds views whenever it wants. Should it be a @StateObject ?

ios/FluentUI/List/ListItem.swift
lcapkovic: ohh I see. I think you could store the private vars as `var content: (() -> Content)?` - that would allow you to set them to nil from those other inits, and still defer the actual calling of the closure to view body. Up to you tho, I'm not even sure about the perf impact.

alexanderboswell: ah gotcha. I think that will work, and makes sense

PR #1850, opened 2023-07-27
Tokenize PeoplePicker

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Tokenized `PeoplePicker`. Since `PeoplePicker` inherits from `BadgeField`, `PeoplePicker`'s `updateAppearance` will also call `BadgeField`'s `updateAppearance`.

### Binary change

Total increase: 52,824 bytes
Total decrease: 0 bytes
| Total | 47,031,800 bytes | 47,084,624 bytes | 🛑 52,824 bytes |
<details>

| PeoplePickerTokenSet.o | 0 bytes | 35,056 bytes | ⚠️ 35,056 bytes |
| PeoplePicker.o | 466,040 bytes | 480,664 bytes | ⚠️ 14,624 bytes |
| __.SYMDEF | 2,926,688 bytes | 2,929,408 bytes | ⚠️ 2,720 bytes |
| BadgeField.o | 961,208 bytes | 961,632 bytes | ⚠️ 424 bytes |
</details>

### Verification

<details>

| Before                                       | After                                      | Theme Overrides                                      | Control Overrides                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|--------------------------------------------|
| light before | light after | theme light | per light |
| dark before | dark after | theme dark | per dark |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1850&drop=dogfoodAlpha


ios/FluentUI/People Picker/PeoplePicker.swift
```diff
@@ -146,6 +142,9 @@ open class PeoplePicker: BadgeField {
 
     private let separator = Separator()
 
+    public typealias TokenSetKeyType = PeoplePickerTokenSet.Tokens
+    public var peoplePickerTokenSet: PeoplePickerTokenSet = .init()
```
joannaquu, line 146: Since `PeoplePicker` inherits from `BadgeField`, `tokenSet` already exists. Open to suggestions on how we want to approach two token sets (one for BadgeField and one for PersonaListView)

ios/FluentUI/People Picker/PeoplePicker.swift
```diff
@@ -164,11 +163,14 @@ open class PeoplePicker: BadgeField {
...
         }
         personaListView.searchDirectoryDelegate = self
-        personaListView.backgroundColor = UIColor(light: fluentTheme.color(.background2).light,
-                                                  dark: fluentTheme.color(.background2).dark)
 
         NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardWillShow(_:)), name: UIResponder.keyboardWillShowNotification, object: nil)
         NotificationCenter.default.addObserver(self, selector: #selector(handleKeyboardWillHide(_:)), name: UIResponder.keyboardWillHideNotification, object: nil)
+
+        // Update appearance whenever `peoplePickerTokenSet` changes.
+        peoplePickerTokenSet.registerOnUpdate(for: self) { [weak self] in
```
alexanderboswell, line 171: just trying to understand how the token sets work, so will this be called on init for the initial setting of background color, correct?

alexanderboswell (approved):

ios/FluentUI/People Picker/PeoplePicker.swift
huwilkes: For `PopupMenu`, I made it assert on trying to access the drawer tokens. It's probably better to keep all the values coming from one tokenSet if possible

```swift
public typealias TokenSetKeyType = PopupMenuTokenSet.Tokens
public typealias TokenSetType = PopupMenuTokenSet
public override var tokenSet: DrawerTokenSet {
  get {
    return popupTokenSet
  }
  set {
    assertionFailure("PopupMenuController tokens must be set through popupTokenSet")
  }
}
public var popupTokenSet: PopupMenuTokenSet = .init()
```

huwilkes: Though, I guess that made more sense when overrides were done by setting a custom token class. This should still work, though, since `popupMenu.tokenSet` will return the `popupTokenSet`.

ios/FluentUI/People Picker/PeoplePicker.swift
huwilkes: Sort of. This will get called anytime the token set gets an update, which is usually an override is set, or the theme changes. Usually the theme will change when you move to a new window, so moving to a window will usually trigger this, but there are some cases where the control moves to a window using the default theme, and the update doesn't get called. Our demo app is bad for testing that, since the "default" theme isn't actually the shared default theme. So we should probably be making sure we set the background color in init, also.

ios/FluentUI/People Picker/PeoplePicker.swift
```diff
@@ -195,6 +198,11 @@ open class PeoplePicker: BadgeField {
 
     open override func willMove(toWindow newWindow: UIWindow?) {
         super.willMove(toWindow: newWindow)
+        guard let newWindow else {
```
laminesm, line 201: Seems like the old logic in this function (on line 206) doesn't rely on the window not being nil. Are we sure we want to add the guard here or would it be safer to add it after the if block?

laminesm: Or maybe change the guard into an if let that updates the tokenSet if non nil?

ios/FluentUI/People Picker/PeoplePickerTokenSet.swift
```diff
@@ -0,0 +1,28 @@
...
+/// Design token set for the `PeoplePicker` control.
+public class PeoplePickerTokenSet: ControlTokenSet<PeoplePickerTokenSet.Tokens> {
+    public enum Tokens: TokenSetKey {
+        /// The background color of the PeoplePicker.
+        case backgroundColor
+    }
+
+    init() {
+        super.init { token, theme in
+            switch token {
```
laminesm, line 17: nit: Feels weird to have a switch with a single element to iterate over. Could this be rewritten differently?

joannaquu: It seems like some other controls also do it this way (`Separator`, `PersonaButtonCarousel`). I'm assuming it also makes it easier to add new tokens but would you still prefer it to be rewritten?

laminesm: I guess then it's fine but maybe something that's worth bringing up for potential cleanup in the future.

huwilkes (approved):

laminesm (approved):

PR #1857, opened 2023-08-03
MVP of List Action Item

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

In conjunction with creating the `ListItem`, adding support for the `ActionsCell` with `ListActionItem`. Following a similar pattern done for tableviewcells, the design for having an action in a list is so different than the `ListItem` designs, it makes sense to create another control as it can be small and simple, reducing the complexity needed to have it all in one control.

A couple scenarios in which `ListItem` is needed also needs the `ListActionItem`, so the reason for creating it now. 

### Binary change

`libFluentUI.a`

Total increase: 302,968 bytes
Total decrease: -1,400 bytes
| Total | 30,415,152 bytes | 30,716,720 bytes | ⛔️ 301,568 bytes |
<details>

| ListActionItem.o | 0 bytes | 219,968 bytes | 🛑 219,968 bytes |
| __.SYMDEF | 4,605,704 bytes | 4,664,056 bytes | 🛑 58,352 bytes |
| ListActionItem+Extensions.o | 0 bytes | 20,336 bytes | ⚠️ 20,336 bytes |
| FocusRingView.o | 800,752 bytes | 804,432 bytes | ⚠️ 3,680 bytes |
| IndeterminateProgressBar.o | 231,216 bytes | 231,848 bytes | ⚠️ 632 bytes |
| SwiftUI+ViewPresentation.o | 58,968 bytes | 58,960 bytes | 🎉 -8 bytes |
| PersonaListView.o | 175,640 bytes | 175,624 bytes | 🎉 -16 bytes |
| PersonaButton.o | 243,968 bytes | 243,344 bytes | 🎉 -624 bytes |
| ActionsCell.o | 166,368 bytes | 165,616 bytes | 🎉 -752 bytes |
</details>

`FluentUI.Demo`

| Total | 34,039,846 bytes | 34,198,038 bytes | ⛔️ 158,192 bytes |

### Verification

Tested it in the FluentUI demo app. XCUI tests are out of scope for this PR and will be in follow up interactions.

<details>

| scenario                                       | screenshot                                      |
| light | Simulator Screenshot - iPhone 14 Pro - 2023-08-02 at 22 06 36 |
| dark | Simulator Screenshot - iPhone 14 Pro - 2023-08-02 at 22 06 44 |
| large accessibility class | Simulator Screenshot - iPhone 14 Pro - 2023-08-02 at 22 07 14 |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1857&drop=dogfoodAlpha


ios/FluentUI/List/ListActionItem.swift
```diff
@@ -0,0 +1,199 @@
...
+
+        return content
+            .listRowInsets(EdgeInsets())
+            .listRowSeparator(.hidden)
+            .frame(minHeight: ListItemTokenSet.oneLineMinHeight)
+            .background(backgroundView)
+    }
+
+    /// The  type of separator on the top edge
+    var topSeparatorType: ListActionItemSeparatorType?
```
laminesm, line 147: Seems odd to have top and bottom separator types as optional. Shouldn't there be a default type instead just like `backgroundStyleType`?

ios/FluentUI/List/ListItem.swift
```diff
@@ -106,6 +106,8 @@ public struct ListItem<LeadingContent: View,
         var backgroundView: some View {
             if let backgroundColor = backgroundStyleType.defaultColor(tokenSet: tokenSet) {
```
laminesm, line 107: `defaultColor` sounds like a function that should always return a non-optional color. Is there a scenario in which we would want it to return nil?

ios/FluentUI/List/ListActionItem.swift
alexanderboswell: hmm yeah, I am not sure. It can kinda go either way depending on if you are using grouped sections in list (which has separators) or a plain style which doesnt.

not that it is the correct way, but I followed the pattern of `ActionsCell`, which had then set to none with the table style as plain:
``` swift    
/// Style describing whether or not the cell's custom top separator should be visible and how wide it should extend
    @objc open var topSeparatorType: TableViewCell.SeparatorType = .none {
        didSet {
            if topSeparatorType != oldValue {
                updateHorizontalSeparator(topSeparator, with: topSeparatorType)
            }
        }
    }
    /// Style describing whether or not the cell's custom bottom separator should be visible and how wide it should extend
    @objc open var bottomSeparatorType: TableViewCell.SeparatorType = .none {
        didSet {
            if bottomSeparatorType != oldValue {
                updateHorizontalSeparator(bottomSeparator, with: bottomSeparatorType)
            }
        }
    }
```

ios/FluentUI/List/ListItem.swift
alexanderboswell: It looks like the decision was made that if a client wanted to override the background color **not** using the token system, the color would be nil.

I would assume that having the default for `.custom` background style to be `.clear` could cause issues with someone attempting to override it outside the token system?


``` swift
// Different background color is used for `TableViewCell` by getting the appropriate tokens and integrate with the cell's `UIBackgroundConfiguration`
@objc(MSFTableViewCellBackgroundStyleType)
public enum TableViewCellBackgroundStyleType: Int {
    // use for flat list of cells
    case plain
    // use for grouped list of cells
    case grouped
    // clear background so that TableView's background can be shown
    case clear
    // in case clients want to override the background on their own without using token system
    case custom

    func defaultColor(tokenSet: TableViewCellTokenSet) -> UIColor? {
        switch self {
        case .plain:
            return tokenSet[.cellBackgroundColor].uiColor
        case .grouped:
            return tokenSet[.cellBackgroundGroupedColor].uiColor
        case .clear:
            return .clear
        case .custom:
            return nil
        }
    }
}
```

ios/FluentUI/List/ListActionItem.swift
laminesm: I think to make it more consistent with the pattern of ActionsCell we'd need to add the default none style and make it non-optional. It's not a big deal though so it's your call.

alexanderboswell: I guess I dont agree with having an enum for `.none`.  Is that not the goal of having optional values in swift? (`Optional.none` which is the same as nil) As for this scenario we dont show something in the view at all, its not a modification of a view that will be shown.

That was the reasoning on why I made a new `enum` for this which only has `.inset` and `.full`, instead of following the pattern of having a "none" case for each enum.

Thoughts?

laminesm: I would agree if this was an optional field for the separator itself. However, an _optional type_ reads as if sometimes we have a separator type and sometimes we don't which semantically sounds wrong to me. Also this is the practice we usually adopt in fluent where we routinely use none in enum cases to show a missing element (i.e. DrawerResizingBehavior, AvatarActivityStyle, AvatarPresence, TwoLineTitleViewAccessoryType etc.). It also seems to be common practice within UIKit as well.

alexanderboswell: updated to reflect this.

ios/FluentUI/List/ListActionItem.swift
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        return content
+            .listRowInsets(EdgeInsets())
+            .listRowSeparator(.hidden)
+            .frame(minHeight: ListItemTokenSet.oneLineMinHeight)
+            .background(backgroundView)
+    }
+
+    /// The  type of separator on the top edge
```
joannaquu, line 141: Nit: extra spaces

joannaquu:
I just realized that PRs into feature branches do not run all ci checks. Could you run `ci.sh` locally to ensure that all PR checks pass?

huanluu (approved):

alexanderboswell:
> I just realized that PRs into feature branches do not run all ci checks. Could you run `ci.sh` locally to ensure that all PR checks pass?

Ran and it passed!

joannaquu (approved):

PR #1862, opened 2023-08-09
Add isHidden property to CommandingItem

nodes11:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Add an `isHidden` property to the `CommandingItem` class and update the `BottomCommandingController` implementation to update its list and tab views when that property changes. The updating in the `BottomCommandingController` was achieved by creating private filtered versions of the `heroItems` and `expandedListSections` arrays. When an item's `isHidden` property is changed, these arrays are re-filtered and the tab view and table view are refreshed to show only not hidden items.

This change is being made for an use case in an internal application.

### Binary change

Total increase: 57,120 bytes
Total decrease: -8 bytes
| Total | 30,167,152 bytes | 30,224,264 bytes | 🛑 57,112 bytes |
<details>

| BottomCommandingController.o | 889,880 bytes | 925,520 bytes | ⚠️ 35,640 bytes |
| __.SYMDEF | 4,573,832 bytes | 4,584,728 bytes | ⚠️ 10,896 bytes |
| CommandingItem.o | 87,096 bytes | 92,424 bytes | ⚠️ 5,328 bytes |
| FocusRingView.o | 791,528 bytes | 794,992 bytes | ⚠️ 3,464 bytes |
| HUD.o | 262,896 bytes | 263,832 bytes | ⚠️ 936 bytes |
| BadgeField.o | 578,376 bytes | 579,232 bytes | ⚠️ 856 bytes |
| CommandingSection.o | 27,472 bytes | 27,464 bytes | 🎉 -8 bytes |
</details>

### Verification

- Verified with internal application use case
- Verified behavior with the `BottomCommandingDemoController` when items are set to be hidden
  - Added 2 new toggles to the `BottomCommandingDemoController` to toggle the hidden state of both tab and list items 

<details>

### Before
IMG_0044
__Above image shows the existing set of toggles before changes were made.__

IMG_0043
__Above image shows the existing bottom commanding sheet before changes were made.__

### After

IMG_0037 
_Above image shows the new toggles to hide hero and list items in untoggled states._

Screenshot 2023-08-15 at 4 23 27 PM
_Above image shows the sheet with the hero and list items displayed. None of the items are hidden since the toggles are off._

IMG_0041
_Above image shows the new toggles to hide hero and list items in toggled states._

Screenshot 2023-08-15 at 4 27 21 PM
_Above image shows the sheet with the hero and list items displayed. Some of the items are hidden since the toggles are on._

https://github.com/microsoft/fluentui-apple/assets/10938746/09da82b9-5431-4ee3-a0a2-371241a1b466
_Above video demos all of the controls touched in this PR. New behavior is the hiding and showing of `heroItems` and `expandedListSection` items._

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: codeflow:open?pullrequest=https://github.com/microsoft/fluentui-apple/pull/1862&drop=dogfoodAlpha

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -48,8 +48,8 @@ class BottomCommandingDemoController: DemoController {
         return Array(1...25).map {
             let item = CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
             item.selectedImage = homeSelectedImage
-            item.isOn = $0 % 3 == 1 ? true : false
-            item.isEnabled = $0 % 2 == 1 ? true : false
+            item.isOn = modifiedCommandIndices.contains($0) ? true : false
+            item.isEnabled = modifiedCommandIndices.contains($0) ? true : false
             return item
         }
     }()
@@ -103,17 +103,17 @@ class BottomCommandingDemoController: DemoController {
                 DemoItem(title: "Expanded list items", type: .boolean, action: #selector(toggleExpandedItems), isOn: expandedItemsVisible),
                 DemoItem(title: "Additional expanded list items", type: .boolean, action: #selector(toggleAdditionalExpandedItems(_:)), isOn: additionalExpandedItemsVisible),
                 DemoItem(title: "Popover on hero command tap", type: .boolean, action: #selector(toggleHeroPopover)),
-                DemoItem(title: "Hero command isOn", type: .boolean, action: #selector(toggleHeroCommandOnOff), isOn: true),
-                DemoItem(title: "Hero command isEnabled", type: .boolean, action: #selector(toggleHeroCommandEnabled), isOn: false),
-                DemoItem(title: "List command isEnabled", type: .boolean, action: #selector(toggleListCommandEnabled), isOn: true),
+                DemoItem(title: "Hero command isOn", type: .boolean, action: #selector(toggleModifiableHeroCommandsOnOff), isOn: true),
+                DemoItem(title: "Hero command isEnabled", type: .boolean, action: #selector(toggleModifiableHeroCommandsEnabled), isOn: false),
+                DemoItem(title: "List command isEnabled", type: .boolean, action: #selector(toggleModifiableListCommandsEnabled), isOn: true),
                 DemoItem(title: "Long title hero items", type: .boolean, action: #selector(toggleLongTitleHeroItems), isOn: false),
-                DemoItem(title: "Hero command isHidden", type: .boolean, action: #selector(toggleHeroCommandHidden), isOn: false),
-                DemoItem(title: "List command isHidden", type: .boolean, action: #selector(toggleListCommandHidden), isOn: false),
+                DemoItem(title: "Hero command isHidden", type: .boolean, action: #selector(toggleModifiableHeroCommandsHidden), isOn: false),
+                DemoItem(title: "List command isHidden", type: .boolean, action: #selector(toggleModifiableListCommandsHidden), isOn: false),
                 DemoItem(title: "Toggle boolean cells", type: .action, action: #selector(toggleBooleanCells)),
-                DemoItem(title: "Change hero command titles", type: .action, action: #selector(changeHeroCommandTitle)),
-                DemoItem(title: "Change hero command images", type: .action, action: #selector(changeHeroCommandIcon)),
-                DemoItem(title: "Change list command titles", type: .action, action: #selector(changeListCommandTitle)),
-                DemoItem(title: "Change list command images", type: .action, action: #selector(changeListCommandIcon))
+                DemoItem(title: "Change hero command titles", type: .action, action: #selector(changeModifiableHeroCommandsTitle)),
+                DemoItem(title: "Change hero command images", type: .action, action: #selector(changeModifiableHeroCommandsIcon)),
+                DemoItem(title: "Change list command titles", type: .action, action: #selector(changeModifiableListCommandsTitle)),
+                DemoItem(title: "Change list command images", type: .action, action: #selector(changeModifiableListCommandsIcon))
             ]
         ]
     }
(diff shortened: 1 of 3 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -175,6 +177,20 @@ class BottomCommandingDemoController: DemoController {
         }
     }
 
+    @objc private func toggleHeroCommandHidden(_ sender: BooleanCell) {
+        for (index, heroItem) in heroItems.enumerated() {
+            if index % 2 == 1 {
```
harrieshin, line 182: nit: kind of abstract logic that will be hidden every other one? the function name doesn't say it. the label of the toggle doesn't say it.

nodes11: I agree that abstraction does create for a bit unknown & confusion on first sight. I removed the use of the modulo operator across the file and replaced it with the `modifiedCommandIndices` array. I also updated the selector function names to make it clear that calling them would only act on items inside the `modifiedCommandIndices`.

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -48,8 +48,8 @@ class BottomCommandingDemoController: DemoController {
         return Array(1...25).map {
             let item = CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
             item.selectedImage = homeSelectedImage
-            item.isOn = $0 % 3 == 1 ? true : false
-            item.isEnabled = $0 % 2 == 1 ? true : false
+            item.isOn = modifiedCommandIndices.contains($0) ? true : false
+            item.isEnabled = modifiedCommandIndices.contains($0) ? true : false
```
joannaquu, line 52: Nit: can remove true/false

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -1133,6 +1135,13 @@ extension BottomCommandingController: UITableViewDelegate {
         return configuredHeader
     }
 
+    public func tableView(_ tableView: UITableView, heightForRowAt indexPath: IndexPath) -> CGFloat {
+        let section = expandedListSections[indexPath.section]
+        let item = section.items[indexPath.row]
+
+        return item.isHidden ? 0 : UITableView.automaticDimension
```
anandrajeswaran, line 1142: This feels suspicious. Shouldn't we just not have a cell if an item is hidden? Setting the height to 0 sounds like it will just lead to accessibility bugs with accessible but zero-sized elements

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -1210,6 +1219,26 @@ extension BottomCommandingController: CommandingItemDelegate {
...
     }
+
+    func commandingItem(_ item: CommandingItem, didChangeHiddenTo value: Bool) {
+        guard let view = itemToBindingMap[item]?.view else {
+            return
+        }
+
+        switch view {
+        case let tabBarItemView as TabBarItemView:
+            if tabBarItemView.isHidden != value {
```
anandrajeswaran, line 1230: is this check needed? Shouldn't the no-op if value isn't changing be an implementation detail?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -1210,6 +1219,26 @@ extension BottomCommandingController: CommandingItemDelegate {
...
+                tabBarItemView.isHidden = value
+            }
+        case let cell as TableViewCell:
+            if cell.isHidden != value {
+                cell.isHidden = value
+                tableView.reloadData()
+            }
+        default:
```
anandrajeswaran, line 1238: Is this ok to hit the default case? Should be an error of some sort?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
```diff
@@ -48,8 +48,8 @@ class BottomCommandingDemoController: DemoController {
         return Array(1...25).map {
             let item = CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
             item.selectedImage = homeSelectedImage
-            item.isOn = $0 % 3 == 1 ? true : false
```
huwilkes, line 51: Did you update your after screenshots after making this change? I changed it away from just changing `modifiedCommandIndices` so that we could more easily see all the combination of on/off enabled/disabled, and also so that it would change the commands in the overflow section.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -777,6 +777,7 @@ open class BottomCommandingController: UIViewController, TokenizedControlInterna
         itemView.numberOfTitleLines = Constants.heroButtonMaxTitleLines
         itemView.isSelected = item.isOn
         itemView.isEnabled = item.isEnabled
+        itemView.isHidden = item.isHidden
```
huwilkes, line 780: Seems like if the item is hidden, we just shouldn't be making the view. We should probably just filter out all of the hidden items in `reloadHeroCommandStack` and still show 5 items per row, rather than allow setting `isHidden` to change the number of items per row indirectly.

joannaquu:
Could you add the size comparison data?

ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
nodes11: I did not update the screenshot. I think the use of the modulo makes sense to achieve that behavior so I've added the code back.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
nodes11: This code isn't needed with the most recent changes.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
nodes11: Updated the `heroItems` and `expandedListSections` arrays (i.e. the datasources) to filter out hidden items so that this approach isn't needed. The table view will now only try to render filtered items.

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
nodes11: Added filtering to the `heroItems` array to filter out hidden items so that this change isn't needed.

huwilkes (approved):

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -672,6 +662,58 @@ open class BottomCommandingController: UIViewController, TokenizedControlInterna
...
+            return !item.isHidden
+        }
+
+        visibleHeroItems = updatedVisibleHeroItems
+    }
+
+    /// Filters the `expandedListSections` 2D array for items with that are not hidden
+    private func updateVisibleExpandedListSections() {
+        /// Filter for all `CommandingSection`  that have at least 1 visible `CommandingItem`
+        let updatedVisibleExpandedListSections = expandedListSections.filter {$0.items.filter {!$0.isHidden}.count > 0}
```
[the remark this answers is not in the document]
anandrajeswaran: But maybe can split this line into separate lines anyway. I am likely misunderstanding, but feels like the outer filter is using implicit boolean conversion of whether the inner filter is not nil, but wouldn't that never be nil because it would just return an empty array instead?

So maybe should be 
```
expandedListSections.filter { expandedListSection in 
    return expandedListSection.items.contains { item in
        return !item.isHidden
    }
}
```

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
```diff
@@ -672,6 +662,58 @@ open class BottomCommandingController: UIViewController, TokenizedControlInterna
...
+
+        /// Filter all `CommandingItem` that are not hidden and add to a new `CommandingSection` to holds the filtered items
+        visibleExpandedListSections = updatedVisibleExpandedListSections.map({ (section: CommandingSection) -> CommandingSection in
```
anandrajeswaran, line 682: minor: can `(section: CommandingSection) -> CommandingSection in` just trim down to `section in`?

ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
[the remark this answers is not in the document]
nodes11: Using `contains` has the potential for slightly better perf too!

anandrajeswaran (approved):

PR #1864, opened 2023-08-16
[ListItem] Support VoiceOver

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

The changes made are to match the VoiceOver functionality that is in `MSFTableViewCell`.

To achieve this, the views needed to be combined into one accessibility element so that they would be read together. The edge cases are for the accessory type of `detailButton` and `trailingContent`. For the `detailButton`, it will be the next item in the accessibility tree after reading out the title/subtitle/footer. It will also get its own focus.

For `trailingContent`, it has two scenarios:
1. one focusable element. For example, a toggle on the trailing edge. In this scenario we would want the toggle to be apart of the same accessibility element as the text so that it is all read at once and doesnt require another swipe to navigate to it.
2. more than one focusable element. If an array of buttons were to be supplied, we would want to be able to navigate to each individually. 

To support these variations, it requires a bit of a change of the view hierarchy. The default will be `#1`. This will put the `trailingContent` in the same `HStack` as the text so that it can be combined for one accessibility element. For `#2`, a modifier is provided so a client can explicitly tell if the `trailingContent` should not be combined. This will result in the `trailingContent` being outside of the `HStack` of the rest of the content. Because `trailingContent` is generic content we dont have a way to determine what type of view it is, so we cant just check if it is `Toggle` like we do in `MSFTableViewCell` for `UISwitch`.

### Binary change

N/A

### Verification

https://github.com/microsoft/fluentui-apple/assets/21343215/ef063ef4-af40-47a2-aed7-9f4de4459c0d

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1864)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
@@ -182,51 +182,6 @@ struct ListItemDemoView: View {
                         .alert("Detail button tapped", isPresented: $showingAlert) {
                             Button("OK", role: .cancel) { }
                         }
-                        ListItem(title: title,
-                                 subtitle: showSubtitle ? subtitle : "",
-                                 footer: showFooter ? footer : "",
-                                 leadingContent: {
-                            if showLeadingContent {
-                                leadingContent
-                            }
-                        },
-                                 trailingContent: {
-                            if showTrailingContent {
-                                switch trailingContentFocusableElementCount {
-                                case 0:
-                                    Text("Spreadsheet")
-                                case 1:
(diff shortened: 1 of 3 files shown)


huwilkes:
When you had two buttons and the disclosure icon, it didn't read out the disclosure icon at all, so I think something is missing. And there are probably times where we'll need to add the button trait.

huwilkes:
> When you had two buttons and the disclosure icon, it didn't read out the disclosure icon at all, so I think something is missing. And there are probably times where we'll need to add the button trait.

Looks like there's not an action for the entire ListItem yet, so we probably don't need the button trait until then.

huwilkes (approved):

PR #1731, opened 2023-05-05
Merge NavBar feature branch into main

amgleitman:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This brings all the navigation bar enhancements into the main branch, including the ability to host two-line title views and tokenization.

This change also includes two new features that ended up being requirements: an autolayout-based `TwoLineTitleView` and a decoupling of `TextColorStyle` from `Label`.

This change involved a temporary revert of #1699 (with 5af0153) because it didn't play nicely with `TwoLineTitleView`'s manual layout mechanism. Naturally, this is a moot point now.

### Binary change

As of f85784a:

Total increase: 859,384 bytes
Total decrease: -184,072 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 30,729,400 bytes | 31,404,712 bytes | ⛔️ 675,312 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| NavigationBar.o | 544,856 bytes | 666,840 bytes | 🛑 121,984 bytes |
| __.SYMDEF | 4,638,352 bytes | 4,748,240 bytes | 🛑 109,888 bytes |
| TwoLineTitleView.o | 208,176 bytes | 300,080 bytes | 🛑 91,904 bytes |
| NavigationBarTokenSet.o | 0 bytes | 71,232 bytes | 🛑 71,232 bytes |
| TwoLineTitleViewTokenSet.o | 0 bytes | 60,968 bytes | 🛑 60,968 bytes |
| BadgeView.o | 584,672 bytes | 638,800 bytes | 🛑 54,128 bytes |
| AvatarTitleViewTokenSet.o | 0 bytes | 53,568 bytes | 🛑 53,568 bytes |
| FocusRingView.o | 758,544 bytes | 790,424 bytes | ⚠️ 31,880 bytes |
| AvatarTitleView.o → LargeTitleView.o | 159,488 bytes | 188,288 bytes | ⚠️ 28,800 bytes |
| UINavigationItem+Navigation.o | 48,776 bytes | 77,192 bytes | ⚠️ 28,416 bytes |
| BadgeLabelButton.o | 146,008 bytes | 162,640 bytes | ⚠️ 16,632 bytes |
| BadgeViewTokenSet.o | 89,720 bytes | 98,944 bytes | ⚠️ 9,224 bytes |
[...]

PR #1869, opened 2023-08-21
Add accessibility label to navigation bar's back button

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

There is a voiceover regression where the navigation bar's back button reads "button" instead of "back button". It's a regression because old nav bar used to have the native back button which was already set up with the correct accessibility label. In the new nav bar, we use a [custom back button item](https://github.com/microsoft/fluentui-apple/pull/1731/files#diff-525ee9b5c92f443b1a34cc5182f719cab1854bfdc605164eab9033ed6cabbfc5R330) so we need to set up the accessibility label manually. 
This PR adds "back" to localizable resources and uses it as the new accessibility label for the back button.

### Verification

<details>

| VO: "button" | VO: "back button" |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1869)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -349,6 +349,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
                                              target: nil,
                                              action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
         backButtonItem.accessibilityIdentifier = "Back"
+        backButtonItem.accessibilityLabel = "Accessibility.NavigationBar.BackLabel".localized
         return backButtonItem
     }()
 
--- a/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
+++ b/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
@@ -103,6 +103,9 @@
 /* Accessibility label for the avatar (or profile) view */
 "Accessibility.LargeTitle.ProfileView" = "Account Profile";
 
+/* Accessibility label for the navigation bar's back button */
+"Accessibility.NavigationBar.BackLabel" = "Back";
+
 /* Accessibility label format string for avatar view. Format: "<Name>, <Presence>". Example: "Kat, Available" */
 "Accessibility.AvatarView.LabelFormat" = "%@, %@";
 
--- a/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
+++ b/ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
@@ -9,25 +9,24 @@
 
 /* Accessibility alert for common use */
 "Accessibility.Alert" = "Alert";
-/* Accessibility dismiss label for common use */
-"Accessibility.Dismiss.Label" = "Dismiss";
-/* Accessibility dismiss hint for common use */
-"Accessibility.Dismiss.Hint" = "Double tap to dismiss";
-/* Accessibility done label for common use */
-"Accessibility.Done.Label" = "Done";
-/* Accessibility multi select hint for common use */
-"Accessibility.MultiSelect.Hint" = "Double tap to toggle selection";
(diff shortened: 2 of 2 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -349,6 +349,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
                                              target: nil,
```
harrieshin, line 349: have we tried to create a regular system barbuttonitem and use the backbuttonimage api  to update the image, and see if that would give us the right accessibility hints without doing our own localziation?

ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
```diff
@@ -103,6 +103,9 @@
 /* Accessibility label for the avatar (or profile) view */
 "Accessibility.LargeTitle.ProfileView" = "Account Profile";
 
+/* Accessibility label for the navigation bar's back button */
+"Accessibility.NavigationBar.BackLabel" = "Back";
+
```
joannaquu, line 108: Is this supposed to be in alphabetical order? Seems weird that `Accessibility.AvatarView.LabelFormat` is at the bottom.

ios/FluentUI/Navigation/NavigationBar.swift
laminesm: This is a regular bar button item we are using to replace the native back button. Unfortunately, we need to specify its accessibility label if we want VO to say "back button". I guess we could revert the nav bar to using the native back button in order to avoid the localization but @amgleitman might have more context as to why we're using a custom back button item here.

ios/FluentUI/Resources/Localization/en.lproj/Localizable.strings
laminesm: The file is mostly alphabetical but there are some strings that are out of place but I can fix em up.

joannaquu: Awesome, thanks Lamine!

joannaquu (approved):

ios/FluentUI/Navigation/NavigationBar.swift
amgleitman: [Here](https://github.com/microsoft/fluentui-apple/commit/028b81bc768005933fabc56083149f0c1a57b446) is this commit that added the custom back button in the first place.

I don't remember the exact context, but looking at this, I think it had something to do with the fact that the system back button gets hidden by the overlay we use for leading-aligned title views. Another added benefit if I recall correctly is that the system back button won't be aligned vertically in the compact-versus-regular case (the whole 32px vs 44px nav bar struggle), so that's another thing to look out for.

harrieshin: but in this bug case we aren't overlaying the leading edge title view right? if it is meant to be vanillar system navigation bar why do we need to do it?

laminesm: Right, we aren't overlaying the leading title view but the back button showing up is still the custom one and not the native back button. Looking at the view hierarchy in our demo, it seems the native back button is always hidden on system, leading and landscape.

anandrajeswaran: Would that be the ideal fix though - go back to using the native back button?

laminesm: According to Adam the native back button was causing problems on the nav bar. If we can reliably fix those without relying on a custom back button then going back would be the ideal fix. I'll have to investigate more to know for sure.

anandrajeswaran (approved):

PR #1878, opened 2023-08-28
Fix no such module 'FluentUI_ios' on MacCatalyst

xnth97:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In Swift Package, use FluentUI_ios target for `MacCatalyst` platform.

### Binary change

N/A - Change to Package.swift

### Verification

Before:

Build target on MacCatalyst will fail as no module FluentUI_ios configured for platform.

截屏2023-08-27 下午9 33 06

After:

MacCatalyst target with FluentUI dependency builds successfully.

截屏2023-08-27 下午9 40 08

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1878)

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,7 +22,7 @@ let package = Package(
         .target(
             name: "FluentUI",
             dependencies: [
-                .target(name: "FluentUI_ios", condition: .when(platforms: [.iOS])),
+                .target(name: "FluentUI_ios", condition: .when(platforms: [.iOS, .macCatalyst])),
                 .target(name: "FluentUI_macos", condition: .when(platforms: [.macOS]))
             ],
             path: "public"


anandrajeswaran:
Apologies for the delay on getting to this one and thanks for contributing! The concern here is that we don't *actually* support catalyst. We have never actually tested it and have no idea if the controls work or come close to fulfilling the macOS project of Fluent design, though the existence of this PR suggests something remotely usable shows up, so that's cool if it does :).

If you are able to continue to prototype using the library with Catalyst without merging the change to main, we'd love to hear your feedback/results and can use that super helpful data to decide how to move forward with support and where we need touch-ups/banned controls/etc. If you need the change in main, we're happy to do so, but then we'll probably want to make corresponding updates to the README and the demo app to very clearly indicate that the macOS Catalyst controls are experimental and not actually supported.

xnth97:
Hi @anandrajeswaran,

Thank you for your detailed explanation. I've been using FluentUI in my personal MacCatalyst project for a while. From my own experience it works fine with MacCatalyst + Scaled to Match iPad, as in this mode it's basically still using iOS idioms and linked against UIKit. 

Screenshot 2023-09-05 at 10 36 52 AM

Attaching some screenshots of FluentUI demo under MacCatalyst.

| Cell | Nav Bar | Side Tab | Bottom Sheet |
| --- | --- | --- | --- |
| Screenshot 2023-09-05 at 10 35 47 AM | Screenshot 2023-09-05 at 10 32 29 AM | Screenshot 2023-09-05 at 10 35 17 AM | Screenshot 2023-09-05 at 10 36 19 AM | 

However if switching to Optimize for Mac idiom, it does throw exceptions as some code calling unavailable UIKit APIs under Mac idiom.

Screenshot 2023-09-05 at 10 38 38 AM


This change only adds `macCatalyst` platform to the target dependency so that users who use FluentUI with SPM on MacCatalyst can still build the project. I found [firebase team has a similar issue](https://github.com/firebase/firebase-ios-sdk/issues/11119) indicating that this seems to be a new requirement for MacCatalyst projects introduced by Swift 5.7 toolchain. I think if we want to declare that FluentUI fully supports MacCatalyst, we should also change it in the package platforms as

```swift
platforms: [
    .iOS(.v15),
    .macCatalyst(.v15),
    .macOS(.v10_15),
],
```

as a library user if the platform availability not specifically declared here, I'd assume that this library doesn't officially support MacCatalyst and I'm just leveraging the compatibility with my own risk.

I think this decision should be left to FluentUI team as I understand this is adding compatibility to another unofficially supported platform. Personally I would prefer to have this in master or even investigate support of MacCatalyst so that users like me won't need to maintain a fork just to build our MacCatalyst projects.

Thanks

anandrajeswaran:
> Hi @anandrajeswaran,
> 
> Thank you for your detailed explanation. I've been using FluentUI in my personal MacCatalyst project for a while. From my own experience it works fine with MacCatalyst + Scaled to Match iPad, as in this mode it's basically still using iOS idioms and linked against UIKit.
> 
> Screenshot 2023-09-05 at 10 36 52 AM
> Attaching some screenshots of FluentUI demo under MacCatalyst.
> 
> Cell	Nav Bar	Side Tab	Bottom Sheet
> Screenshot 2023-09-05 at 10 35 47 AM	Screenshot 2023-09-05 at 10 32 29 AM	Screenshot 2023-09-05 at 10 36 19 AM	Screenshot 2023-09-05 at 10 35 17 AM
> However if switching to Optimize for Mac idiom, it does throw exceptions as some code calling unavailable UIKit APIs under Mac idiom.
> 
> Screenshot 2023-09-05 at 10 38 38 AM
> This change only adds `macCatalyst` platform to the target dependency so that users who use FluentUI with SPM on MacCatalyst can still build the project. I found [firebase team has a similar issue](https://github.com/firebase/firebase-ios-sdk/issues/11119) indicating that this seems to be a new requirement for MacCatalyst projects introduced by Swift 5.7 toolchain. I think if we want to declare that FluentUI fully supports MacCatalyst, we should also change it in the package platforms as
> 
> ```swift
> platforms: [
>     .iOS(.v15),
>     .macCatalyst(.v15),
>     .macOS(.v10_15),
> ],
> ```
> 
> as a library user if the platform availability not specifically declared here, I'd assume that this library doesn't officially support MacCatalyst and I'm just leveraging the compatibility with my own risk.
> 
> I think this decision should be left to FluentUI team as I understand this is adding compatibility to another unofficially supported platform. Personally I would prefer to have this in master or even investigate support of MacCatalyst so that users like me won't need to maintain a fork just to build our MacCatalyst projects.
> 
> Thanks

Thanks for the reply and additional detail. For the timebeing we'd like to hold off on having this in the repo because it's something of a partial state that suggests we have thought about macCatalyst but have an impartial implementation. Since we don't officially support it yet, we'd like to keep that clearly apparent in our project definition.

We will keep the investigation of the catalyst implementation on our backlog though and hopefully have some time in the future to do some investigation on how close that gets us to the intended macOS representation of the fluent design language.

PR #1887, opened 2023-09-07
Change alpha of ListItem when disabled

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

Adds support for changing the alpha of a ListItem when it is disabled.

### Binary change

N/A

### Verification

Added a toggle in the test app to visually test this change.

<details>

| Disabled                                       | Enabled                                      |
| Simulator Screenshot - iPad Air (5th generation) - 2023-09-07 at 11 45 02 | Simulator Screenshot - iPad Air (5th generation) - 2023-09-07 at 11 45 07 |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1887)

### DIFF
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -189,6 +189,7 @@ public struct ListItem<LeadingContent: View,
                 accessoryView
             }
             .frame(minHeight: layoutType.minHeight)
+            .opacity(isEnabled ? ListItemTokenSet.enabledAlpha : ListItemTokenSet.disabledAlpha)
             .background(backgroundView)
             .listRowInsets(EdgeInsets())
         }
@@ -274,6 +275,7 @@ public struct ListItem<LeadingContent: View,
     // MARK: Private variables
 
     @Environment(\.fluentTheme) private var fluentTheme: FluentTheme
+    @Environment(\.isEnabled) private var isEnabled: Bool
 
     private var leadingContent: (() -> LeadingContent)?
     private var trailingContent: (() -> TrailingContent)?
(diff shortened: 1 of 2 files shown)


ios/FluentUI/List/ListItem.swift
```diff
@@ -189,6 +189,7 @@ public struct ListItem<LeadingContent: View,
                 accessoryView
             }
             .frame(minHeight: layoutType.minHeight)
+            .opacity(isEnabled ? ListItemTokenSet.enabledAlpha : ListItemTokenSet.disabledAlpha)
```
huwilkes, line 192: Generally we prefer using a specific color rather than changing the opacity of a control. Changing the opacity can lead to unintentional accessibility issues depending on the background color.

alexanderboswell: hmm, I agree. But I guess how do we go about doing that? The token set only has an alpha specified and doesnt have disabled colors. 

I believe that would be needed to support doing colors, instead of alpha like what is being done for MSFTableViewCell.

huwilkes: I guess this is how the TVC handles it, though.

huwilkes (approved):

ios/FluentUI/List/ListItem.swift
```diff
@@ -274,6 +275,7 @@ public struct ListItem<LeadingContent: View,
     // MARK: Private variables
 
     @Environment(\.fluentTheme) private var fluentTheme: FluentTheme
+    @Environment(\.isEnabled) private var isEnabled: Bool
```
anandrajeswaran, line 278: Hopefully dumb question - why does this need to be an Environment variable? Feels like it should be specific to the view?

alexanderboswell: it is set based on the `.disabled()` view modifier. This allows for ListItem to handle changing on disabled state without having to have a custom modifier to pass in a boolean value. 

It is an environment variable as that allows it to be set without having each view have to pass the value down. So for example, if a whole list is disabled, it will allow for this value set and used without having to pass in a boolean to each listItem view.

PR #1889, opened 2023-09-08
[ListItem] Fix static method for getting background color

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [ ] macOS

### Description of changes

A static method was provided on ListItem to get the background color of the list. This allows for implementations of ListItem to ensure the list has the correct background color. 

The issue is that calling a static method requires that we know the types of the generic content for ListItem. This would require the call to the static method to be something like:
`ListItem<EmptyView, EmptyView, String, String, String>.listBackgroundColor(for: .grouped)`

To handle this scenario better, we can provide default values for the generic content so we can call it like:
`ListItem.listBackgroundColor(for: .grouped)`

### Binary change

N/A

### Verification

Added to the Fluent UI demo app and tested that the background color is shown.

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1889)

### DIFF
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -40,21 +40,6 @@ public struct ListItem<LeadingContent: View,
         self.tokenSet = ListItemTokenSet(customViewSize: { layoutType.leadingContentSize })
     }
 
-    /// The background color of `List` based on the style.
-    /// - Parameter backgroundStyle: The background style of the `List`.
-    /// - Returns: The color to use for the background of `List`.
-    public static func listBackgroundColor(for backgroundStyle: ListItemBackgroundStyleType) -> Color {
-        let tokenSet = ListItemTokenSet(customViewSize: { .default })
-        switch backgroundStyle {
-        case .grouped:
-            return Color(uiColor: tokenSet[.backgroundGroupedColor].uiColor)
-        case .plain:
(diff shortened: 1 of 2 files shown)


ios/FluentUI/List/ListItem.swift
```diff
@@ -338,3 +323,22 @@ public extension ListItem where LeadingContent == EmptyView {
         self.tokenSet = ListItemTokenSet(customViewSize: { layoutType.leadingContentSize })
     }
 }
+
+/// Provide defaults for generic types so static methods can be called without needing to specify them.
+public extension ListItem where LeadingContent == EmptyView, TrailingContent == EmptyView, Title == String, Subtitle == String, Footer == String {
```
huwilkes, line 328: I think in a `guard` or `if` statement, the `,` works sort of like an `&&`. Is that also true for `where`? Or, rather, does this still work when any one of the properties is specified?

alexanderboswell: yup! it works like a && in this scenario also.

alexanderboswell: If you see the other extensions defined in this file, you can see I have to do a lot of different variations so that you only need to specify a subset. So if there is another generic content added, it makes the matrix of extensions larger ha.

huwilkes (approved):

PR #1899, opened 2023-09-22
Fix TableViewHeaderFooterView being cropped on large text 

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There is a bug where the `TableViewHeaderFooterView` gets cropped on large text mode with the inset grouped style. This is caused by the height function miscalculating the width of the title view since it takes into account the header footer view's frame instead of the content view's frame. 
To fix the bug, we explicitly pass in the `contentView`'s width in `sizeThatFits(_ size: CGSize)` for the height calculations. In the height func, I also removed the extra`TableViewHeaderFooterViewTokenSet.horizontalMargin` from the width calculations since `TableViewHeaderFooterView.titleLeadingOffset()` and `TableViewHeaderFooterView.titleTrailingOffset(accessoryView: accessoryView)` already add the necessary leading and trailing paddings. With these adjustments, `layoutSubviews` and `height()` now use the exact same width values to calculate the view's height.

### Binary change

Total increase: 632 bytes
Total decrease: 0 bytes
| Total | 31,610,192 bytes | 31,610,824 bytes | ⚠️ 632 bytes |
<details>

| TableViewHeaderFooterView.o | 282,656 bytes | 283,208 bytes | ⚠️ 552 bytes |
| __.SYMDEF | 4,956,736 bytes | 4,956,816 bytes | ⚠️ 80 bytes |
</details>

### Verification

Tested the footer on the demo app.

<details>

| before_large | after_large |
| before_small | after_small |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1899)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -39,20 +39,14 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
                                    containerWidth: CGFloat = .greatestFiniteMagnitude,
                                    accessoryView: UIView? = nil) -> CGFloat {
         let tokenSet: TableViewHeaderFooterViewTokenSet = .init(style: { style }, accessoryButtonStyle: { AccessoryButtonStyle.regular })
-        let verticalMargin: CGFloat
         let font = tokenSet[.textFont].uiFont
-        switch style {
-        case .header, .footer:
-            verticalMargin = TableViewHeaderFooterViewTokenSet.titleDefaultTopMargin + TableViewHeaderFooterViewTokenSet.titleDefaultBottomMargin
-        case .headerPrimary:
-            verticalMargin = TableViewHeaderFooterViewTokenSet.titleDefaultTopMargin + TableViewHeaderFooterViewTokenSet.titleDefaultBottomMargin
-        }
+        let verticalMargin = TableViewHeaderFooterViewTokenSet.titleDefaultTopMargin + TableViewHeaderFooterViewTokenSet.titleDefaultBottomMargin
 
         if let accessoryView = accessoryView {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -39,20 +39,14 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
...
-            verticalMargin = TableViewHeaderFooterViewTokenSet.titleDefaultTopMargin + TableViewHeaderFooterViewTokenSet.titleDefaultBottomMargin
-        }
+        let verticalMargin = TableViewHeaderFooterViewTokenSet.titleDefaultTopMargin + TableViewHeaderFooterViewTokenSet.titleDefaultBottomMargin
 
         if let accessoryView = accessoryView {
             accessoryView.frame.size = accessoryView.systemLayoutSizeFitting(CGSize(width: containerWidth, height: .infinity))
         }
 
-        let titleWidth = containerWidth - (TableViewHeaderFooterViewTokenSet.horizontalMargin + TableViewHeaderFooterView.titleTrailingOffset(accessoryView: accessoryView) + TableViewHeaderFooterView.titleLeadingOffset())
+        let titleWidth = containerWidth - (TableViewHeaderFooterView.titleLeadingOffset() + TableViewHeaderFooterView.titleTrailingOffset(accessoryView: accessoryView))
```
huwilkes, line 49: `titleLeadingOffset` being a private static func seems odd to me, but not really an issue

huwilkes (approved):

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
laminesm: I think that's because the height func is a class func and cannot take instance funcs.

PR #1901, opened 2023-09-26
Add optional customBackgroundColor to TableViewHeaderFooterView

osvaldoalv:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Adds two new setup methods for `TableViewHeaderFooterView` to allow adding a custom background color. 

### Verification

Tested both methods.; here is a screenshot of a `TableViewHeaderFooterView` with a custom background color.

<details>

| Screenshot 2023-09-25 at 6 55 06 PM | Screenshot 2023-09-25 at 6 57 28 PM |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1901)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -176,6 +176,7 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
     }
 
     private let titleView = TableViewHeaderFooterTitleView()
+    private var customBackgroundColor: UIColor?
 
     private var accessoryView: UIView? {
         didSet {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -297,6 +298,24 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
         self.leadingView = leadingView
     }
 
+    /// - Parameters:
+    ///   - style: The `TableViewHeaderFooterView.Style` used to set up the view.
+    ///   - title: The title string.
+    ///   - backgroundColor: The background color of the `backgroundView`
```
anandrajeswaran, line 304: Would be good to get input @sophialee0416 on this parameter's naming and documentation. i.e. should it be `customBackgroundColor`? And what do we want to put in the comment to help inform the consumer when it makes sense to diverge from the default background color and how to do so without sacrificing on "fluent-ness" of design

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -297,6 +298,24 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
         self.leadingView = leadingView
     }
 
+    /// - Parameters:
```
anandrajeswaran, line 301: Thoughts on tagging on to two of the existing `setup` functions and just adding an extra optional parameter (optional to make it not a breaking change)? I'd guess that would reduce binary size impact of this component and it definitely would make our interface simpler?

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -297,6 +298,24 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
...
+    @objc open func setup(style: Style, title: String, backgroundColor: UIColor) {
+        setup(style: style, title: title, accessoryButtonTitle: "")
+        customBackgroundColor = backgroundColor
+    }
+
+    /// - Parameters:
+    ///   - style: The `TableViewHeaderFooterView.Style` used to set up the view.
+    ///   - attributedTitle: Title as an NSAttributedString for additional attributes.
+    ///   - backgroundColor: The background color of the `backgroundView`
+    @objc open func setup(style: Style, attributedTitle: NSAttributedString, backgroundColor: UIColor) {
```
laminesm, line 314: When it comes to using custom colors, the pattern we typically use in fluent is to add a public UIColor variable in the control instead of injecting the custom color in the setup functions (DrawerController, CardView for reference).

osvaldoalv: Got it. I can also see that in **0.17.0** the `backgroundView` background color was changed to use `TableViewHeaderFooterViewTokenSet` values istead of `ColorToken`s. We need to use custom colors for some of our header views but didn't realize we were still using **0.14.1**.

Given that we can override the colors from our side by _registering_ a custom set of `ControlTokenValue`, it seems the change is no longer needed, we can just upgrade to > 0.17.0 and do that.

I'll go ahead and close this PR, sorry for the confusion.

PR #1908, opened 2023-10-06
Adding titleNumberOfLines to PopupMenuItem

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In the PopupMenuController, setting a long string in the headerItem causes the header area to bloat and take up a big chunk of the screen. Let's allow users to configure how many lines they'd like in the header area.

### Binary change

### Verification
Before API vs. Using new API to set `.titleNumberOfLines` to 1

| image | image |
| image | image |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1908)

### DIFF
--- a/ios/FluentUI/Popup Menu/PopupMenuController.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuController.swift
@@ -93,6 +93,7 @@ open class PopupMenuController: DrawerController {
                 } else {
                     headerView.isHidden = false
                     headerView.setup(item: headerItem)
+                    headerView.titleNumberOfLines = headerItem.titleNumberOfLines
                 }
             }
         }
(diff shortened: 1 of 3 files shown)


ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -88,6 +88,7 @@ open class PopupMenuController: DrawerController {
                 if headerItem.subtitle == nil {
```
huwilkes, line 88: What happens when the subtitle isn't nil? It looks like it should default to zero in that case, and I'm not sure there's a concise way to say "titleNumberOfLinesAsLongAsNoSubtitle", but we should probably at least add a documentation comment that warns the user that it doesn't apply when there is a subtitle, or that it wouldn't reset to 0 if they later add one.

sophialee0416: Oh good point. So it looks like when the subtitle isn't nil, then it uses a PopupMenuItemCell, subclassed from TableViewCell, instead of a UIView. In that case we can apply the titleNumberOfLines to the subtitle use case too

huwilkes (approved):

PR #1913, opened 2023-10-11
Deprecate outline button style

huwilkes:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

To align with Fluent Web, we're renaming the neutral outline button style to just `.outline`. In order to do that, we need to get clients off of the current `.outline` style, and get them on to `.outlineAccent`. This is part 1 of that effort. In future PRs, we'll need to deprecate `.outlineNeutral`, then eventually remove `.outlineNeutral` entirely and have `.outline` use the neutral tokens.

### Binary change

Total increase: 3,448 bytes
Total decrease: 0 bytes
| Total | 31,396,368 bytes | 31,399,816 bytes | ⚠️ 3,448 bytes |
<details>

| ButtonTokenSet.o | 121,992 bytes | 125,136 bytes | ⚠️ 3,144 bytes |
| __.SYMDEF | 4,861,208 bytes | 4,861,448 bytes | ⚠️ 240 bytes |
| FocusRingView.o | 822,600 bytes | 822,656 bytes | ⚠️ 56 bytes |
| Button.o | 213,320 bytes | 213,328 bytes | ⚠️ 8 bytes |
</details>

### Verification

Brand colored outline buttons are still brand colored, and there's no deprecation warnings.

<details>

| Button_Outline_Before | Button_Outline_After |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1913)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
@@ -135,7 +135,7 @@ extension ButtonStyle {
         switch self {
         case .accent:
             return "Accent"
-        case .outline:
+        case .outlineAccent:
             return "Outline accent"
         case .outlineNeutral:
             return "Outline neutral"
@@ -151,6 +151,8 @@ extension ButtonStyle {
             return "Floating accent"
         case .floatingSubtle:
             return "Floating subtle"
+        case .outline:
+            return "Outline accent"
         }
     }
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/CommandBarDemoController.swift
@@ -194,37 +194,37 @@ class CommandBarDemoController: DemoController {
 
         itemCustomizationContainer.addArrangedSubview(UIView()) //Spacer
 
-        let refreshButton = Button(style: .outline)
+        let refreshButton = Button(style: .outlineAccent)
         refreshButton.sizeCategory = .small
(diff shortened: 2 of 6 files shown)


harrieshin:
before after screenshots please and size analysis

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
```diff
@@ -151,6 +151,8 @@ extension ButtonStyle {
             return "Floating accent"
         case .floatingSubtle:
             return "Floating subtle"
+        case .outline:
+            return "Outline accent"
```
alexanderboswell, line 155: should this just be `return "Outline"`?

alexanderboswell (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/ButtonDemoController.swift
huwilkes: I kept it as `"Outline accent"` because that's what the `.outline` case said before, but it shouldn't be showing either way right now. And they styling is still the same as `.outlineAccent`.

harrieshin:
may be in the demo app also denote that outline is depcreated.

huwilkes:
> may be in the demo app also denote that outline is depcreated.

`outlineAccent` was added to replace `outline`. So the branded outline style itself isn't deprecated, just the name. And even then the name is only being temporarily deprecated so that clients currently using it will stop and we can replace `outlineNeutral` with just `outline`. The demo says "Outline accent", which is a supported style.

PR #1914, opened 2023-10-16
Deprecating `AliasTokens`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

This deprecates `AliasTokens` in favor of the equivalent APIs on `FluentTheme`. Some particular changes of note:
* `AliasTokens` and all its methods and enums are marked as deprecated.
* `FluentTheme.aliasTokens` is deprecated.
* Any public-facing method that interacts with `AliasTokens` or its types is deprecated.
* Legacy `GlobalTokens` lookup methods are deprecated.

All alias token values have been moved to new helper methods in `FluentTheme+Tokens.swift`. They are value-for-value identical to the old values in `AliasTokens.swift`, and are still directly referenced by the deprecated class.

### Binary change

Total increase: 76,192 bytes
Total decrease: -246,352 bytes
| Total | 31,082,744 bytes | 30,912,584 bytes | 🎉 -170,160 bytes |
<details>

| FluentTheme+Tokens.o | 72,784 bytes | 148,648 bytes | 🛑 75,864 bytes |
| UIColor+Extensions.o | 81,248 bytes | 81,576 bytes | ⚠️ 328 bytes |
| ColorProviding.o | 41,416 bytes | 41,408 bytes | 🎉 -8 bytes |
| BadgeViewTokenSet.o | 95,296 bytes | 95,288 bytes | 🎉 -8 bytes |
| FontInfo.o | 70,576 bytes | 70,568 bytes | 🎉 -8 bytes |
| DateTimePickerController.o | 138,744 bytes | 138,720 bytes | 🎉 -24 bytes |
| CalendarViewWeekdayHeadingView.o | 86,032 bytes | 86,000 bytes | 🎉 -32 bytes |
| DatePickerController.o | 323,568 bytes | 323,536 bytes | 🎉 -32 bytes |
| FluentUIFramework.o | 104,952 bytes | 104,912 bytes | 🎉 -40 bytes |
| DateTimePickerView.o | 207,896 bytes | 207,848 bytes | 🎉 -48 bytes |
| CalendarViewDayTodayCell.o | 36,512 bytes | 36,464 bytes | 🎉 -48 bytes |
| BadgeView.o | 626,472 bytes | 626,416 bytes | 🎉 -56 bytes |
| PillButtonTokenSet.o | 88,320 bytes | 88,248 bytes | 🎉 -72 bytes |
| SearchBarTokenSet.o | 79,504 bytes | 79,432 bytes | 🎉 -72 bytes |
| BottomSheetTokenSet.o | 48,272 bytes | 48,200 bytes | 🎉 -72 bytes |
| GlobalTokens.o | 539,368 bytes | 539,264 bytes | 🎉 -104 bytes |
| DynamicColor.o | 49,720 bytes | 49,280 bytes | 🎉 -440 bytes |
| PageCardPresenterController.o | 147,464 bytes | 146,952 bytes | 🎉 -512 bytes |
| BadgeLabelButton.o | 111,032 bytes | 110,128 bytes | 🎉 -904 bytes |
| CommandBarCommandGroupsView.o | 197,040 bytes | 191,088 bytes | 🎉 -5,952 bytes |
| __.SYMDEF | 4,830,504 bytes | 4,823,000 bytes | 🎉 -7,504 bytes |
| BadgeField.o | 591,264 bytes | 583,424 bytes | 🎉 -7,840 bytes |
| FluentTheme.o | 184,944 bytes | 146,552 bytes | 🎉 -38,392 bytes |
| AliasTokens.o | 281,248 bytes | 97,064 bytes | 🎉 -184,184 bytes |
</details>

### Verification

<details>

| Before-AC | After-AC |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1914)


alexanderboswell (approved):

ios/FluentUI/Core/Theme/FluentTheme.swift
```diff
@@ -36,33 +36,24 @@ public class FluentTheme: NSObject, ObservableObject {
...
-
-        // Pass overrides to AliasTokens
-        aliasTokens = .init(colorOverrides: Dictionary(uniqueKeysWithValues: fixedColorOverrides),
-                            shadowOverrides: Dictionary(uniqueKeysWithValues: fixedShadowOverrides),
-                            typographyOverrides: Dictionary(uniqueKeysWithValues: fixedTypographyOverrides),
-                            gradientOverrides: Dictionary(uniqueKeysWithValues: fixedGradientOverrides))
+
+        // Need to massage UIFonts into FontInfo objects
+        let mappedTypographyOverrides = typographyOverrides?.compactMapValues({ font in
+            return FontInfo(name: font.fontName, size: font.pointSize)
```
huwilkes, line 42: Are we not able to just use UIFont for the typography overrides?

huwilkes: I thought one of the issues we had with `FontInfo` was there wasn't a good way to guarantee the correct conversion of `UIFont` -> `FontInfo` -> `UIFont`, so seems like we shouldn't be going through `FontInfo` here and `defaultTypography` should return `UIFont`. Keeping `FontInfo` around for our internal inits makes sense, but should we also deprecate the public `FontInfo`? Likewise with `ColorValue` and `DynamicColor` if we're trying to move to just `UIColor`.

mischreiber: So I do want to deprecate `ColorValue` and `DynamicColor`, and I even have a separate commit ready for that. It's a little wide ranging, so I wanted to keep it as a follow-up instead of the same PR, but if you prefer then I can fold it in here.

As for `FontInfo`, my intent was to avoid creating UIFont objects until they're needed -- my understanding previously was that they can be a bit heavy to instantiate, but that info might be a little outdated now. Do you have any recollection/feeling there @huwilkes?

huwilkes: Separate PR for colors sounds good, as long as we can also get it in for this version bump.

For `FontInfo`, I don't recall that coming up in the previous `UIFont` vs `FontInfo` conversations, but I could see it being the case. As long as we can guarantee that we return the correct fonts that the clients pass in for their theme overrides I don't see an issue with continuing to use `FontInfo` here. We could also do some profiling tests to see how much instantiating `UIFont`s will cost us here.

mischreiber: Discussed offline -- we're gonna dig deeper into the Font question at a later date, but as far as we can tell this doesn't regress our current situation, so let's move forward with it.

PR #1929, opened 2023-11-13
TableViewHeaderFooterView should not override any font or color provided as part of the AttributedString

owenconnolly:
### Platforms Impacted
- [ x] iOS
- [ ] macOS

### Description of changes

There is an issue with usage of this control when an NSAttributedString is provided for the label. The existing code updates the font and foreground color of the label in a number of places, but this will override any font or foreground color attributes which the consumer may have set in the attributed string. 

We can extract the UIFont and UIColor and store these if they have been provided as part of the attributed string, and wherever we update the font or color, we can apply the provided values where they exist, otherwise fall back to the default values we were previously using.

### Verification

Verified locally when integrated into an app that the default font and color are used where not provided, and if they are provided as part of the attributed string then those values will be used instead.

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1929)


laminesm (commented):
Curious if this couldn't be dealt with externally by overriding the font and title color tokens.

owenconnolly:
> Curious if this couldn't be dealt with externally by overriding the font and title color tokens.

Not sure on that, though it is an interesting question. I would assume that if I override font/color tokens that _could_ apply to other elements of this/other FluentUI components, whereas it should only apply to the attributed string. In addition to that I am not sure that the caller should assume knowledge of the internals of this component (i.e. what font/color tokens are used and where, which could change).

Even considering that, I do think that it would still be correct behaviour to not override the consumer provided string attributes, and that this change would make sense regardless of what the consumer does with the tokens...

laminesm:
> > Curious if this couldn't be dealt with externally by overriding the font and title color tokens.
> 
> Not sure on that, though it is an interesting question. I would assume that if I override font/color tokens that _could_ apply to other elements of this/other FluentUI components, whereas it should only apply to the attributed string. In addition to that I am not sure that the caller should assume knowledge of the internals of this component (i.e. what font/color tokens are used and where, which could change).
> 
> Even considering that, I do think that it would still be correct behaviour to not override the consumer provided string attributes, and that this change would make sense regardless of what the consumer does with the tokens...

Overriding the tokens on a control instance doesn’t propagate to other fluent controls. Also fluent control tokens are publicly exposed and architected to allow partners to customize them as they see fit, so I don’t think there are any particular assumptions about what’s going on internally that need to be considered. 

But yeah I do agree this change makes sense regardless, especially considering TableViewCell seems to have similar custom behavior for attributed titles.

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -430,18 +439,20 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
 
     private func updateTitleViewFont() {
         if let window = window {
-            let titleFont = tokenSet[.textFont].uiFont
+            let titleFont = attributedTitleFont != nil ? attributedTitleFont : tokenSet[.textFont].uiFont
```
laminesm, line 442: To avoid having to do these nil checks, can we just initialize 
```
attributedTitleFont = attributes[NSAttributedString.Key.font] as? UIFont ?? tokenSet[.textFont].uiFont
attributedTitleColor = attributes[NSAttributedString.Key.foregroundColor] as? UIColor ?? tokenSet[.textColor].uiColor
```
Then we can just pass in `attributedTitleFont` and `attributedTitleColor`?

joannaquu: Since `attributedTitleFont` and `attributedTitleColor` are initialized in `setup`, wouldn't that be a little confusing here? That `titleFont` is set to `attributedTitleFont`?

laminesm: `attributedTitleFont` and `attributedTitleColor` would need to be renamed to something like `resolvedFont` and `resolvedTitleColor`.

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -206,6 +206,9 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
         }
     }
 
+    private var attributedTitleFont: UIFont?
```
mischreiber, line 209: Nit: we try to keep our file organized as per the Microsoft Swift docs here: [File Organization
](https://microsoft.github.io/swift-guide/FileOrganization.html).

Specifically, `private` vars are down lower in the file, while `open` and `public` is higher up.

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
mischreiber: Agreed -- I'd much rather we have a single source of truth, at least internally, to avoid having to sprinkle `nil` checks around. That's gonna be super error-prone going forward, I fear.

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
owenconnolly: Thanks, I would usually have these at the end of the file, but wanted to group them with existing private vars. I can move them to the end of the class though, thanks!

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
owenconnolly: Thanks for the suggestion, that makes sense alright so I will update it!

owenconnolly: At present leaving the vars as optional so I don't need to assign values in as many places, but if we want them non-optional I can make sure they are set on init

huwilkes: Another option could be to make the `lazy`, and init them with the `tokenSet`, then overwrite that with the values from the attributed string if needed.

huwilkes (approved):

laminesm (approved):

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
owenconnolly: Thanks, I think that would end up with similar additional code, so I might just go with the current iteration for now...

PR #1935, opened 2023-11-21
Make FluentUI be iOS 17 safe

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Changes to the library to make it build against iOS 17, in case any clients are trying to do so:
- `traitCollectionDidChange(_:)` is deprecated, so add `@available(iOS, deprecated: 17.0)` to all overrides.
- `UITraitCollection(traitsFrom:)` is deprecated, so adding a conditional check and using the new `UITraitCollection(mutations:)` API when iOS 17 is available.
- `onChange(of:_:)` has changed signatures, so I added a new `onChange_iOS17(of:_:)` to abstract that difference away. Feel free to suggest a better name if you have one 🤠

Note: this change doesn't do anything for the demo app itself. There are some significant refactors I'd like to do there before we worry too much about API deprecations. Don't worry, this won't affect consumers who pull us via CocoaPods or SPM.

### Binary change

Total increase: 39,104 bytes
Total decrease: 0 bytes
| Total | 30,982,416 bytes | 31,021,520 bytes | ⚠️ 39,104 bytes |
<details>

| FluentNotification.o | 676,088 bytes | 698,680 bytes | ⚠️ 22,592 bytes |
| __.SYMDEF | 4,821,272 bytes | 4,829,384 bytes | ⚠️ 8,112 bytes |
| UIColor+Extensions.o | 73,328 bytes | 78,896 bytes | ⚠️ 5,568 bytes |
| SwiftUI+ViewModifiers.o | 133,128 bytes | 135,960 bytes | ⚠️ 2,832 bytes |
</details>

### Verification

App builds and runs as expected -- no functional changes here.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1935)

### DIFF
--- a/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
+++ b/ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
@@ -59,22 +59,22 @@ extension View {
         modifier(ShadowModifier(shadowInfo: shadowInfo))
     }
 
-	/// Abstracts away differences in pre-iOS 17 `onChange(of:perform:)` versus post-iOS 17 `onChange(of:_:)`.
-	///
-	/// This function will be removed once FluentUI moves to iOS 17 as a minimum target.
-	/// - Parameters:
-	///   - value: The value to check against when determining whether to run the closure.
-	///   - action: A closure to run when the value changes.
-	/// - Returns: A view that fires an action when the specified value changes.
-	func onChange_iOS17<V>(of value: V, _ action: @escaping (V) -> Void) -> some View where V: Equatable {
-		if #available(iOS 17, *) {
-			return self.onChange(of: value) { _, newValue in
-				return action(newValue)
-			}
-		} else {
-			return self.onChange(of: value, perform: action)
-		}
-	}
+    /// Abstracts away differences in pre-iOS 17 `onChange(of:perform:)` versus post-iOS 17 `onChange(of:_:)`.
+    ///
+    /// This function will be removed once FluentUI moves to iOS 17 as a minimum target.
+    /// - Parameters:
+    ///   - value: The value to check against when determining whether to run the closure.
+    ///   - action: A closure to run when the value changes.
+    /// - Returns: A view that fires an action when the specified value changes.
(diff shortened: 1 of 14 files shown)


laminesm (commented):
Don't we need to migrate the logic in `traitCollectionDidChange(_:)` to the new APIs Apple wants us to use for iOS 17?

anandrajeswaran (approved):

mischreiber:
> Don't we need to migrate the logic in `traitCollectionDidChange(_:)` to the new APIs Apple wants us to use for iOS 17?

Eventually, but for now this is good. Unfortunately, the new APIs don't exist before iOS 17, so as long as our code *actually* targets iOS 15 or 16, we can't actually rely on that new logic.

laminesm (approved):

ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -52,12 +52,29 @@ extension View {
...
+    /// - Parameters:
     ///  - shadowInfo: The values of the two shadows to be applied
     /// - Returns: The modified view.
     func applyShadow(shadowInfo: ShadowInfo) -> some View {
         modifier(ShadowModifier(shadowInfo: shadowInfo))
     }
+
+    /// Abstracts away differences in pre-iOS 17 `onChange(of:perform:)` versus post-iOS 17 `onChange(of:_:)`.
+    ///
+    /// This function will be removed once FluentUI moves to iOS 17 as a minimum target.
```
joannaquu, line 64: Do we have a bug tracking this?

mischreiber: I think that any time we bumped minimum OS versions in the past, the bumper would also search the codebase for any relevant availability checks -- the `if #available(iOS 17, *)` should trip that search query when we get there. But let's chat more offline to see if there's something we can do to better formalize this process.

PR #1936, opened 2023-11-30
Adding a search bar option to the PopupMenuController

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

- Added Search bar to PopupMenuController
- Fixed bug where Popup view would not move with keyboard
- Fixed Popup colors (all default background should match the drawer background)

### Binary change

### Verification

<details>

</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Popup Menu/PopupMenuController.swift
+++ b/ios/FluentUI/Popup Menu/PopupMenuController.swift
@@ -64,6 +64,9 @@ open class PopupMenuController: DrawerController {
                 height += headerItem.cellClass.preferredHeight(for: headerItem)
             }
         }
+        if !searchView.isHidden {
+            height += searchView.frame.height
+        }
         for section in sections {
             height += PopupMenuSectionHeaderView.preferredHeight(for: section)
             for item in section.items {
@@ -98,6 +101,17 @@ open class PopupMenuController: DrawerController {
             }
         }
     }
+
+    @objc open func searchBar(isVisible: Bool, placeholderText: String? = "") {
+        searchBar.isHidden = !isVisible
+        searchBar.placeholderText = placeholderText
+        adjustsHeightForKeyboard = true
+    }
+
+    @objc open var searchText: String? {
+        return searchBar.isHidden ? nil : searchBar.searchText
+    }
+
     /// Use `selectedItemIndexPath` to get or set the selected menu item instead of doing this via `PopupMenuItem` directly
     @objc open var selectedItemIndexPath: IndexPath? {
         get {
@@ -144,6 +158,7 @@ open class PopupMenuController: DrawerController {
         view.axis = .vertical
         view.addArrangedSubview(descriptionView)
         view.addArrangedSubview(headerView)
+        view.addArrangedSubview(searchView)
         view.addArrangedSubview(tableView)
         return view
     }()
@@ -193,6 +208,21 @@ open class PopupMenuController: DrawerController {
         view.isHeader = true
         view.isHidden = true
         return view
+    }()
+    private lazy var searchView: UIView = {
+        let view = UIStackView()
+        view.axis = .vertical
+        view.addArrangedSubview(searchBar)
+        view.isHidden = searchBar.isHidden
+        view.isLayoutMarginsRelativeArrangement = true
+        view.layoutMargins = UIEdgeInsets(top: 16, left: 16, bottom: 12, right: 16)
+        return view
+    }()
+    private let searchBar: SearchBar = {
+        let searchBar = SearchBar()
+        searchBar.style = .onSystemNavigationBar
+        searchBar.isHidden = true
+        return searchBar
     }()
     private let tableView = UITableView(frame: .zero, style: .grouped)
 
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/PopupMenuDemoController.swift
@@ -17,6 +17,7 @@ class PopupMenuDemoController: DemoController {
 
     private let navButtonTitleSwitch = BrandedSwitch()
     private let navButtonSubtitleSwitch = BrandedSwitch()
+    private let toolbarSearchSwitch = BrandedSwitch()
     private let switchTextWidth: CGFloat = 150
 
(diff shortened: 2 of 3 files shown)


ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -194,6 +209,21 @@ open class PopupMenuController: DrawerController {
         view.isHidden = true
         return view
     }()
+    private lazy var searchView: UIView = {
+        let view = UIStackView()
+        view.axis = .vertical
+        view.addArrangedSubview(searchBar)
+        view.isHidden = searchBar.isHidden
+        view.isLayoutMarginsRelativeArrangement = true
+        view.layoutMargins = UIEdgeInsets(top: 16, left: 16, bottom: 12, right: 16)
```
harrieshin, line 218: should it be using our global tokens?

sophialee0416: Oh good catch forgot to update that. Thanks!

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -194,6 +209,21 @@ open class PopupMenuController: DrawerController {
...
+        return view
+    }()
+    private let searchBar: SearchBar = {
```
harrieshin, line 221: does your pod spec need to be updated for new dependencies?

sophialee0416: Yep added that in the new update but I'm wondering if we need to add the entirety of the Nav subspec or if we should create a new one just for Search, thoughts?

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -98,6 +101,17 @@ open class PopupMenuController: DrawerController {
             }
         }
     }
+
+    @objc open func searchBar(isVisible: Bool, placeholderText: String? = "") {
```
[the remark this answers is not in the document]
mischreiber: Normally I'd say yes, these should be datasource/delegate methods. But the whole class is currently structured this way -- I see a TON of methods/vars that I would move into a delegate/datasource. Is that level of refactor something we want to commit to? I'm leaning towards leaving it alone for now, just based on how gnarly this class already is.

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -98,6 +101,17 @@ open class PopupMenuController: DrawerController {
...
+        searchBar.isHidden = !isVisible
+        searchBar.placeholderText = placeholderText
+        adjustsHeightForKeyboard = true
+    }
+
+    @objc open var searchText: String? {
```
laminesm, line 111: Will this be overridden at this level or should it be a readonly property?

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -194,6 +209,26 @@ open class PopupMenuController: DrawerController {
         view.isHidden = true
         return view
     }()
+    private lazy var searchView: UIView = {
+        let view = UIStackView()
+        let marginsSpacing = GlobalTokens.spacing(.size160)
```
laminesm, line 214: nit: I'd suggest renaming these variables since margin and spacing mean the same thing in this context.

ios/FluentUI/Popup Menu/PopupMenuController.swift
sophialee0416: I think this should be a read only, but I couldn't think of any scenarios where we'd want to overwrite the search text content

ios/FluentUI/Popup Menu/PopupMenuController.swift
sophialee0416: Oh interesting! Hm would it be any better to add the dependency as a file path (ie. SearchBar folder path) inside the pod spec for Popup, or add the Navigation subspace entirely? Not sure if I want to factor Searcher into its own subspace in this PR.

huwilkes: Probably also not a great solution, but it might make sense to make this just a different control so that we don't force a dependency on anything navigation or search bar on someone that just wants a plain popup menu.

ios/FluentUI/Popup Menu/PopupMenuController.swift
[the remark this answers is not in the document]
sophialee0416: The urgency for the Search Bar APIs + Pill Button Bar (coming soon) is pretty strong so I am happy to try my hand at refactoring at a later time, if possible?

ios/FluentUI/Popup Menu/PopupMenuController.swift
```diff
@@ -257,7 +292,7 @@ open class PopupMenuController: DrawerController {
     }
 
     private func initTableView() {
-        tableView.backgroundColor = backgroundColor
+        tableView.backgroundColor = .clear
```
huwilkes, line 295: Did you test overrides?

sophialee0416: Yep those work!
image
image

ios/FluentUI/Popup Menu/PopupMenuController.swift
huwilkes: Different control as in subclass of PopupMenuController with a search bar instead of adding the search bar to the PopupMenuController.

sophialee0416: Hmmm... I don't know how I feel to be honest especially because we have another addition to the Popup coming up (pill button bar). Would that mean a subclass for every popupmenu variant? Or maybe a "search/filter" version of popup that contains these add-ons? And then there's also binary size to consider I suppose?

huwilkes: I think splitting out to a search/filter version makes sense. I imagine the search bar and pill button bar for filtering will be used together, but there's lots of times where someone would want just a popup menu without any of the extra bits and dependencies adding those to the base class would add.

sophialee0416:
Closing PR for now to pivot some eng decisions

PR #1938, opened 2023-12-04
Fixing keyboard support in Drawers

sophialee0416:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

In #1936 , there was a bug fix for drawers not being able to support expand with keyboard if the drawer was presented in a custom-set origin.  This is because we automatically calculate the frame origin on `contentFrame.maxY - contentSize.height`, which is not necessarily always the correct origin especially if a keyboard is in use. The keyboard height code is also simplified so that it is more accurate in the presentation calculations.

### Binary change

### Verification

Manual tests below, no change should be observed. This should also only affect presentations that go `.up`.

<details>

| image | image |
| image | !image |
| image | image |
</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Drawer/DrawerPresentationController.swift
+++ b/ios/FluentUI/Drawer/DrawerPresentationController.swift
@@ -225,7 +225,7 @@ class DrawerPresentationController: UIPresentationController {
             }
         case .up:
             if actualPresentationOrigin == containerView.bounds.maxY {
-                return containerView.safeAreaInsets.bottom + keyboardHeight
+                return keyboardHeight == 0 ? containerView.safeAreaInsets.bottom : keyboardHeight
             }
         case .fromLeading:
             if actualPresentationOrigin == containerView.bounds.minX {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Drawer/DrawerPresentationController.swift
```diff
@@ -225,7 +225,7 @@ class DrawerPresentationController: UIPresentationController {
             }
         case .up:
             if actualPresentationOrigin == containerView.bounds.maxY {
-                return containerView.safeAreaInsets.bottom + keyboardHeight
+                return keyboardHeight == 0 ? containerView.safeAreaInsets.bottom : keyboardHeight
```
mischreiber, line 228: Can you verify this using the "Some editable content" keyboard test, but simulating a hardware keyboard?

sophialee0416: Like this? 

https://github.com/microsoft/fluentui-apple/assets/22566866/013b90a6-d942-4317-8e44-8924e6d4bd9f

mischreiber (approved):

PR #1939, opened 2023-12-05
[iOS] Add podspec for ListItem

alexanderboswell:
### Platforms Impacted
- [ X] iOS
- [ ] macOS

### Description of changes
Add ListItem in the podspec file so it can be consumed.


huwilkes:
> @huwilkes When I created the ListItem it seemed ideal to use the same tokens as TableViewCell. I do wonder if that doesnt make as much sense anymore because it is required for ListItem - which makes it so you have more needed (which is easy to see now when we define the podspec)
> 
> Thoughts?

Probably makes sense to move the tokens into their own podspec that both list item and tvc depend on rather than adding a new identical token set.

PR #1946, opened 2023-12-18
Update all macOS targets to 12.0

mischreiber:
### Platforms Impacted
- [ ] iOS
- [x] macOS

### Description of changes

Some of our macOS targets were providing their own deployment overrides in their xcscheme files, which has been triggering GitHub failure checks. This removes those overrides and brings us to 12.0 across the board.

### Binary change

n/a

### Verification

n/a

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1946)


alexanderboswell (approved):

anandrajeswaran:
Not caused by this review so feel free to split to a separate change - I totally missed that the settings for the xctest target were added directly into the .pbxproj. Should be quick enough to move to an xcconfig to prevent the duplication and over specification?

anandrajeswaran (approved):

PR #1951, opened 2024-01-24
Cbowdoin/bottom sheet supports x axis constraints

cbowdoin:
### Platforms Impacted
- [x] iOS
- [ ] macOS

#### BottomSheetController.swift
Adds 2 new properties to bottom sheets: preferredWidth and anchoredEdge.
PreferredWidth allows for custom width specification for bottom sheets. The sheet will attempt to set the width to the size specified as long as the value is between the minimum and maximum width of the sheet. If the value is not within that range, we will revert back to the original behavior of using the max width.

~~To avoid conflicts with should always fill width these properties unset each other when one is set.~~
 If should always fill width is set we will take this as the primary source of truth and defer behavior to this value being set.

AnchoredEdge allows the user to specify which edge we will constrain the sheet to. By default this remains keeping the sheet centered but when set users can move the sheet to be attached to the leading or trailing edge with a padding of 8.

#### BottomSheetDemoController.swift
- Adds 2 new toggles that allow the user to test the preferredWidth and anchoredEdge.
    - preferredWidth will set the width to 400
    - the anchoredEdge toggle allows users to test attaching the sheet to the trailing edge. Disabling the toggle will 
       recenter the sheet

### Binary change

Total increase: 39,000 bytes
Total decrease: -3,816 bytes
| Total | 30,927,296 bytes | 30,962,480 bytes | ⚠️ 35,184 bytes |
<details>

| BottomSheetController.o | 490,440 bytes | 521,072 bytes | ⚠️ 30,632 bytes |
| __.SYMDEF | 4,790,704 bytes | 4,797,328 bytes | ⚠️ 6,624 bytes |
| FocusRingView.o | 815,024 bytes | 816,768 bytes | ⚠️ 1,744 bytes |
| BottomCommandingController.o | 846,984 bytes | 843,168 bytes | 🎉 -3,816 bytes |
</details>

### Verification

Tested using the bottom sheet demo. Initialized the sheet with properties for the various scenarios.
Preferred Width
- width declared to a value less than the minimum sheet size
- width declared to a value greater than max sheet size
- width declared to a value between min and maximum
- width declared after the sheet is set to always fill width
- width declared before the sheet is to always fill width

AnchoredEdge
- centered, leading, and trailing all position the sheet in the correct spot in both LTR and RTL languages

<details>

| Before                                 | After                                      |
| Screenshot 2024-01-24 at 2 52 08 PM | Screenshot 2024-01-24 at 2 54 55 PM |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1951)


ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -722,7 +759,23 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     // The output is only meaningful once view.bounds is non-zero i.e. a layout pass has occured.
     private func sheetFrame(offset: CGFloat) -> CGRect {
         let availableWidth: CGFloat = view.bounds.width
-        let sheetWidth = shouldAlwaysFillWidth ? availableWidth : min(Constants.maxSheetWidth, availableWidth)
+        let maxWidth = min(Constants.maxSheetWidth, availableWidth)
+
+        /// Width will the full screen if should always fill width
```
nodes11, line 764: spelling: fill

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -732,8 +785,28 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
             sheetHeight = expandedSheetHeight
         }
 
-        return CGRect(origin: CGPoint(x: (view.bounds.width - sheetWidth) / 2, y: offset),
-                      size: CGSize(width: sheetWidth, height: sheetHeight))
+        let padding: CGFloat = 8
```
nodes11, line 788: Should this be a constant defined in the `Constants` struct?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -732,8 +785,28 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
...
+            }
+        } else if anchoredEdge == .leading {
+            if UIView.userInterfaceLayoutDirection(for: view.semanticContentAttribute) == .leftToRight {
+                xPosition = padding
+            } else {
+                xPosition = view.bounds.width - sheetWidth - padding
+            }
+        } else {
+            xPosition = (view.bounds.width - sheetWidth) / 2
+        }
```
nodes11, line 805: Not sure if there is a style preference for Fluent, but you may consider using a switch statement here since all of the `BottomSheetAnchorEdge` values need to be covered. Also adding a short comment describing how the `xPosition` is determined may be helpful.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -722,7 +759,23 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     // The output is only meaningful once view.bounds is non-zero i.e. a layout pass has occured.
     private func sheetFrame(offset: CGFloat) -> CGRect {
         let availableWidth: CGFloat = view.bounds.width
-        let sheetWidth = shouldAlwaysFillWidth ? availableWidth : min(Constants.maxSheetWidth, availableWidth)
+        let maxWidth = min(Constants.maxSheetWidth, availableWidth)
+
+        /// Width will the fill the screen if should always fill width
```
laminesm, line 764: These should be // comments since they are inside a function they probably won't be showing as Xcode documentation.

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -732,8 +785,32 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
             sheetHeight = expandedSheetHeight
         }
 
-        return CGRect(origin: CGPoint(x: (view.bounds.width - sheetWidth) / 2, y: offset),
-                      size: CGSize(width: sheetWidth, height: sheetHeight))
+        /// Calculates the location to put the left edge of the sheet relative to the view
+        /// For right aligned we get the width of the view offset by the sheets width and the padding
+        /// For left aligned we only need to add in the padding
+        /// For center aligned we need the position of the center offset by half the sheets width
```
laminesm, line 791: Same here

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -1062,6 +1139,9 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         // Minimum padding from top when the sheet is fully expanded
         static let minimumTopExpandedPadding: CGFloat = 25.0
 
+        // The padding allocated to the space between the sheet and the edge when attached to the leading or trailing edge
+        static let horizontalSheetPadding: CGFloat = 8
```
laminesm, line 1143: Since fluent 2 we try to use our tokens whenever possible so this should be:
`static let horizontalSheetPadding: CGFloat = GlobalTokens.spacing(.size80)`

laminesm (commented):
Is this change meant to be iPad only?

cbowdoin:
> Is this change meant to be iPad only?

No. The most likely scenarios where this is useful is on iPad but I see no reason to limit consumers to use it on iPad. For example, I could see a scenario where you would want a bottom sheet that is presented on iPhone in landscape in a non centered view.
Screenshot 2024-01-25 at 9 57 35 AM

nodes11 (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -722,7 +752,23 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
...
+            let determinedWidth: CGFloat
+            if shouldAlwaysFillWidth {
+                determinedWidth = availableWidth
+            } else if Constants.minSheetWidth...maxWidth ~= preferredWidth {
+                determinedWidth = preferredWidth
+            } else {
+                determinedWidth = maxWidth
+            }
+            return determinedWidth
+        }()
```
anandrajeswaran, line 770: Should we just split this out to a function? I thought in some cases we've seen binary size impact to using closures and nested functions as opposed to more standard functions, especially when exposed to objC. But also good to just simplify and be consistent with the codebase

Or doesn't need to be a function, can just keep lines 761-769 and rename `determinedWidth` to `sheetWidth`

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -39,6 +39,13 @@ public protocol BottomSheetControllerDelegate: AnyObject {
     case transitioning // Sheet is between states, only used during user interaction / animation
 }
 
+/// Defines where the sheet should be postionioned relative to the screen space
+@objc public enum BottomSheetAnchorEdge: Int {
```
anandrajeswaran, line 43: Sorry I missed this before, but does it need to be @objc(MSFBottomSheetAnchorEdge) to get the proper class prefix for objC usage?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -223,6 +230,29 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         }
     }
 
+    /// Setting this property  will result in the sheet trying to be as close to this width as possible.
+    /// If the declared width is too large it will roll back to the maximum width
+    @objc open var preferredWidth: CGFloat = 0 {
+        didSet {
+            guard preferredWidth != oldValue && isViewLoaded else {
```
anandrajeswaran, line 237: Random small thing - `guard` statements are essentially a negation, so "complex" guard statements with && or || become kinda difficult for simpletons like me to read and parse because I have to do boolean logic in my head. Would recommend either using `,` syntax to join conditions or just splitting out the complexity to a separate bool. In this case, possible recommendation which avoids the need for early returns as well:

```
let shouldInvalidateLayout = (preferredWidth != oldValue) && isViewLoaded
if shouldInvalidateLayout {
    view.setNeedsLayout()
}
 ```

File: ios/FluentUI/Bottom Sheet/BottomSheetController.swift [hunk 10]
```diff
@@ -223,6 +230,29 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         }
     }
 
+    /// Setting this property  will result in the sheet trying to be as close to this width as possible.
+    /// If the declared width is too large it will roll back to the maximum width
+    @objc open var preferredWidth: CGFloat = 0 {
```
anandrajeswaran, line 235: Another small thing - are these meant to be subclassed? If they are meant to be set and not subclassed, could remove `open` I think and see what that does to generated code size?

anandrajeswaran (approved):

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
cbowdoin: Public and open both provide the same binary size. Going to leave it as open remain consistent with the other public properties that support subclassing.

nodes11 (approved):

PR #1955, opened 2024-01-26
[Localization] Localized Resource Files

huwilkes:
Latest localized resource files from Touchdown Build
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1955)


joannaquu (approved):

ios/FluentUI/Resources/Localization/ar.lproj/Localizable.strings
```diff
@@ -1,205 +0,0 @@
-//
```
huwilkes, line 1: Not sure what triggered this PR. I'll investigate shortly, but may end up just abandoning this PR since I don't imagine we actually wanted to delete all our localized files.

mischreiber (changes requested):
Yeah something is wrong with this. Blocking so we don't accidentally auto-merge while we sort it out.

huwilkes:
Going to abandon for now, we'll see if we get some repeats over the weekend.

PR #1958, opened 2024-01-26
Activity Indicator Demo: Dynamic start/stop activity button title

imthath-m:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

In the iOS demo app, open ActivityIndicator. The button in the second section (Settings) will now read "Stop Activity". If you tap that, the activity indicators below will now stop spinning (and also hide if that is enabled) and the button title will change to "Start Activity". 

Previously the button was named "Start/Stop Activity" which did not exactly represent the state of the activity indicators below.

### Binary change

No change in binary size as the change is only in the iOS demo app.

### Verification

- The button title is "Stop Activity" when activity indicators are running
- The button title is "Start Activity" when activity indicators are stopped/hidden

<details>

### Before
https://github.com/microsoft/fluentui-apple/assets/46041492/8e0185f1-a1f6-4d29-a4d6-21b9e33a5a5f
### After
https://github.com/microsoft/fluentui-apple/assets/46041492/6b964386-bc99-4462-a494-3462e9b02f7e
</details>

### Pull request checklist

This PR has considered:


ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift
```diff
@@ -233,8 +235,9 @@ class ActivityIndicatorDemoController: DemoTableViewController {
 
     private let xLargeSize: CGFloat = 36
 
-    @objc private func startStopActivity() {
+    @objc private func startStopActivity(_ sender: UIButton) {
         isAnimating.toggle()
+        sender.setTitle(activityActionButtonTitle, for: .normal)
```
mischreiber, line 240: Instead of coupling the UI type of the sender to this logic, it might be safer to reload the contents of the relevant table view cell and let the existing logic do the work.

imthath-m: Yeah, sounds good. 

Code updated to reload row on button action.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ActivityIndicatorDemoController.swift
```diff
@@ -235,6 +235,13 @@ class ActivityIndicatorDemoController: DemoTableViewController {
 
     @objc private func startStopActivity() {
         isAnimating.toggle()
+
+        guard let section: Int = ActivityIndicatorDemoSection.allCases.firstIndex(of: .settings),
+              let row: Int = ActivityIndicatorDemoSection.settings.rows.firstIndex(of: .startStopActivity) else {
```
mischreiber, line 240: Nit: we avoid declaring types for inline variables unless strictly needed. See the specific reference in our style guide at [Type Inference | swift-guide (microsoft.github.io)](https://microsoft.github.io/swift-guide/TypeInference.html)

imthath-m: Based on the linked guide, it would still make sense to declare types in such cases. Here `Int` is not directly used as literal. It is the result of a function `firstIndex` on another type. So for quick understanding, I think it is helpful to have the type `Int` specified here.

mischreiber: The guide is consistent on this point:

> Use type annotation as little as possible when declaring a constant or a variable; instead let the compiler infer the variable type.

> Let the compiler infer the type of a constant or variable whenever possible.

The primary exception to this policy is for stored properties; local variables should almost never need a type declared, particularly when they are storing the result of a well-understood method like `firstIndex`.

PR #1963, opened 2024-02-01
Set segment pill button corner radius

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Currently, `SegmentedControl` sets the `cornerRadius`. On iOS, this trickles down to `SegmentPillButton` as well. This is not true on other platforms. To fix this, set the `cornerRadius` explicitly.

To avoid magic numbers, move `SegmentedControl.Constants` to `SegmentedControlTokenSet` and access the corner radius from there. While I'm here, remove `selectionBarHeight` since it's not being used anywhere.

### Binary change

Total increase: 2,224 bytes
Total decrease: -1,600 bytes
| Total | 30,917,640 bytes | 30,918,264 bytes | ⚠️ 624 bytes |
<details>

| SegmentedControlTokenSet.o | 90,128 bytes | 91,448 bytes | ⚠️ 1,320 bytes |
| SegmentPillButton.o | 90,648 bytes | 91,552 bytes | ⚠️ 904 bytes |
| FocusRingView.o | 814,848 bytes | 814,688 bytes | 🎉 -160 bytes |
| __.SYMDEF | 4,770,960 bytes | 4,770,720 bytes | 🎉 -240 bytes |
| SegmentedControl.o | 333,384 bytes | 332,184 bytes | 🎉 -1,200 bytes |
</details>

### Verification

<details>

| before | after |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1963)

### DIFF
--- a/ios/FluentUI/SegmentedControl/SegmentPillButton.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentPillButton.swift
@@ -41,6 +41,7 @@ class SegmentPillButton: UIButton {
                                                               trailing: horizontalInset)
         configuration.background.backgroundColor = .clear
         configuration.baseForegroundColor = tokenSet[.restLabelColor].uiColor
+        configuration.background.cornerRadius = 16
         let titleTransformer = UIConfigurationTextAttributesTransformer { [weak self] incoming in
             var outgoing = incoming
             outgoing.font = self?.tokenSet[.font].uiFont
--- a/ios/FluentUI/SegmentedControl/SegmentPillButton.swift
+++ b/ios/FluentUI/SegmentedControl/SegmentPillButton.swift
@@ -40,8 +40,8 @@ class SegmentPillButton: UIButton {
                                                               bottom: verticalInset,
                                                               trailing: horizontalInset)
         configuration.background.backgroundColor = .clear
+        configuration.background.cornerRadius = SegmentedControlTokenSet.pillButtonCornerRadius
         configuration.baseForegroundColor = tokenSet[.restLabelColor].uiColor
-        configuration.background.cornerRadius = 16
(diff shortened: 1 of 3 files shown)


laminesm (approved):
Should we consider making this a constant token instead of using a magic number?

ios/FluentUI/SegmentedControl/SegmentPillButton.swift
```diff
@@ -41,6 +41,7 @@ class SegmentPillButton: UIButton {
                                                               trailing: horizontalInset)
         configuration.background.backgroundColor = .clear
         configuration.baseForegroundColor = tokenSet[.restLabelColor].uiColor
+        configuration.background.cornerRadius = 16
```
mischreiber, line 44: As mentioned by @laminesm, I'd like to see this stored as a named constant somewhere--either as a token constant, or potentially added as a `CornerRadiusToken.radius160`, though that would require updating the token design library.

mischreiber: Alternatively, can we just use `CornerRadiusToken.radiusCircular` for the same effect?

joannaquu: `CornerRadiusToken.radiusCircular` would result in a corner radius that's too large. It's already stored in `SegmentedControl.Constants` so I'll move that to `SegmentedControlTokenSet` and access it there.

mischreiber (approved):

PR #1966, opened 2024-02-08
Bump CI to macOS 14 and build visionOS debug simulator

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] macOS

### Description of changes

* Updating our workflows to use macOS 14 runners, which are ARM64 devices by default
* Using these new runners, verify visionOS as part of the CI pipeline

### Binary change

n/a -- only pipeline changes

### Verification

Hopefully CI passes :)

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1966)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,7 +12,7 @@ on:
 
 jobs:
   validation:
-    runs-on: macos-13
+    runs-on: macos-14
     strategy:
       fail-fast: true
     steps:
@@ -24,7 +24,7 @@ jobs:
     - name: validation
       run: scripts/validation.sh
   xcodebuild:
-    runs-on: macos-13
+    runs-on: macos-14
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


harrieshin (approved):

joannaquu (approved):
do we want device too or is simulator enough

mischreiber:
> do we want device too or is simulator enough

I was wondering about that too. In your experience, have there been any interesting differences between device and simulator that we should consider? I assumed that for most cases (catching deprecated API usage, for example), one should be sufficient.

joannaquu:
> > do we want device too or is simulator enough
> 
> I was wondering about that too. In your experience, have there been any interesting differences between device and simulator that we should consider? I assumed that for most cases (catching deprecated API usage, for example), one should be sufficient.

nope!

PR #1971, opened 2024-02-13
Fix layout for custom views in `BadgeView`

mischreiber:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

Existing frame calculations for `BadgeView` when it has a custom view set (e.g. an `Avatar`) were not quite accurate. This resulted in badges that were a bit too wide, and that would vary in their internal layout by a bit depending on the exact text used.

Basically, we had two problems:
* Internal sizing doubled down on padding. When determining the width of the badge, we included both the padding to the right of the custom view, *and* the padding to the left of the label, resulting in 16pt (by default) instead of 8pt between the view and the label.
* The actual layout did not account for padding at all, letting the label's text center itself in the overly-large space. This could result in padding between the custom view and the label that was too wide, and too variable.

The new logic is more explicit about every bit of padding and sizing being used.

### Binary change

Total increase: 0 bytes
Total decrease: -80 bytes
| Total | 31,064,208 bytes | 31,064,128 bytes | 🎉 -80 bytes |
<details>

| BadgeView.o | 625,296 bytes | 625,216 bytes | 🎉 -80 bytes |
</details>

### Verification

Ensured that the labels and custom views behaved as expected at all text sizes, including extra large sizing.

<details>

| before | after |
| before-l | after-l |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1971)

### DIFF
--- a/ios/FluentUI/Badge Field/BadgeView.swift
+++ b/ios/FluentUI/Badge Field/BadgeView.swift
@@ -227,13 +227,27 @@ open class BadgeView: UIView, TokenizedControlInternal {
         backgroundView.frame = bounds
 
         if let customViewSize = customViewSize(for: frame.size), customViewSize != .zero {
-            let customViewOrigin = CGPoint(x: customViewPadding.left, y: (frame.height - customViewSize.height) / 2)
-            dataSource?.customView?.frame = CGRect(origin: customViewOrigin, size: customViewSize)
-            let labelOrigin = CGPoint(x: customViewPadding.left + customViewPadding.right + customViewSize.width, y: (frame.height - labelSize.height) / 2)
-            let labelSizeThatFits = CGSize(width: frame.size.width - labelOrigin.x, height: labelSize.height)
+            // H:|-(cvp.left)-[customView]-(cvp.right)-[label]-(lhp)-|
+            let customViewPadding = customViewPadding
+            let customViewOrigin = CGPoint(x: customViewPadding.left,
+                                           y: floor((frame.height - customViewSize.height) / 2))
+            let customViewFrame = CGRect(origin: customViewOrigin, size: customViewSize)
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Badge Field/BadgeView.swift
```diff
@@ -227,13 +227,27 @@ open class BadgeView: UIView, TokenizedControlInternal {
         backgroundView.frame = bounds
 
         if let customViewSize = customViewSize(for: frame.size), customViewSize != .zero {
-            let customViewOrigin = CGPoint(x: customViewPadding.left, y: (frame.height - customViewSize.height) / 2)
-            dataSource?.customView?.frame = CGRect(origin: customViewOrigin, size: customViewSize)
-            let labelOrigin = CGPoint(x: customViewPadding.left + customViewPadding.right + customViewSize.width, y: (frame.height - labelSize.height) / 2)
-            let labelSizeThatFits = CGSize(width: frame.size.width - labelOrigin.x, height: labelSize.height)
+            // H:|-(cvp.left)-[customView]-(cvp.right)-[label]-(lhp)-|
+            let customViewPadding = customViewPadding
```
joannaquu, line 231: This is just for perf right? Fetch customViewPadding once?

mischreiber: Exactly. The custom getter for `customViewPadding` does a bit of work, so no need to run it more than once here.

mischreiber: Same with `let labelSize = labelSize` below.

PR #1818, opened 2023-07-05
Expose badge style from TabBar and SideTabBar

laminesm:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

There's a request to change the badge color for tab bar items. Since the colors needed are satisfied by one of the existing badge styles, we can leverage that and add a `badgeStyle` property that will allow the client to override the badge style of all items within tab bar and side tab bar.

Since `BadgeLabel.Style` was internal, I created a new `BadgeStyle` in `TabBarItem` that will be used to set the badge colors.

### Binary change

Total increase: 31,032 bytes
Total decrease: -18,624 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 29,946,488 bytes | 29,958,896 bytes | ⚠️ 12,408 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| SideTabBar.o | 193,608 bytes | 202,696 bytes | ⚠️ 9,088 bytes |
| TabBarItem.o | 68,864 bytes | 77,944 bytes | ⚠️ 9,080 bytes |
| TabBarView.o | 127,360 bytes | 135,488 bytes | ⚠️ 8,128 bytes |
| __.SYMDEF | 4,653,392 bytes | 4,656,104 bytes | ⚠️ 2,712 bytes |
| FocusRingView.o | 789,952 bytes | 791,416 bytes | ⚠️ 1,464 bytes |
| BadgeLabel.o | 56,296 bytes | 56,656 bytes | ⚠️ 360 bytes |
| NavigationBar.o | 542,392 bytes | 542,520 bytes | ⚠️ 128 bytes |
| BadgeLabelButton.o | 110,360 bytes | 110,432 bytes | ⚠️ 72 bytes |
| PopupMenuSectionHeaderView.o | 28,240 bytes | 28,088 bytes | 🎉 -152 bytes |
| FluentTheme.o | 181,880 bytes | 181,176 bytes | 🎉 -704 bytes |
| TableViewHeaderFooterViewTokenSet.o | 83,920 bytes | 80,856 bytes | 🎉 -3,064 bytes |
| TableViewHeaderFooterView.o | 282,256 bytes | 275,312 bytes | 🎉 -6,944 bytes |
| BadgeLabelTokenSet.o | 55,488 bytes | 47,728 bytes | 🎉 -7,760 bytes |
</details>

### Verification

<details>
<summary>Visual Verification</summary>

[...]

PR #1974, opened 2024-02-15
implement red dot support for UIBarButtonItem

kumarmohit5189:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

As part of this change, I have added red dot support for `UIBarButtonItem`. Currently `UIBarButtonItem` supports only Badge label, to make it similar to `TabBarItem` in Fluent UI I introduced a property called `shouldShowRedDot` in similar way as we are using for badge. It will work in similar approach to `TabBarItem` where:

- If user set `shouldShowRedDot` and `badgeValue` both, it will show badge value and logic will run as it was before change.
- If user just set `shouldShowRedDot`, it will display red dot
- If user just set `badgeValue`, it will show badge label as it was previously.

### Binary change

| Total | 3,10,97,936 bytes | 3,11,54,096 bytes | 🛑 56,160 bytes |
<details>

| UIBarButtonItem+RedDot.o | 0 bytes | 25,832 bytes | ⚠️ 25,832 bytes |
| BadgeLabelButton.o | 1,09,936 bytes | 1,30,520 bytes | ⚠️ 20,584 bytes |
| __.SYMDEF | 48,49,960 bytes | 48,59,232 bytes | ⚠️ 9,272 bytes |
| FocusRingView.o | 8,21,464 bytes | 8,21,928 bytes | ⚠️ 464 bytes |
| NavigationBar.o | 5,49,544 bytes | 5,49,552 bytes | ⚠️ 8 bytes |

### Verification

<details>

|-------
Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 11 44 40
Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 11 44 42

---------------------------------------|--------------------------------------------|
| | 
Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 11 25 19
Simulator Screenshot - iPhone 15 Pro - 2024-02-15 at 11 25 22
|
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1974)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -473,6 +473,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
     var showSearchProgressSpinner: Bool = true
     var showRainbowRingForAvatar: Bool = false
     var showBadgeOnBarButtonItem: Bool = false
+	var showRedDotOnBarButtonItem: Bool = false
 
     var allowsCellSelection: Bool = false {
         didSet {
@@ -623,7 +624,21 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
             return cell
         }
 
-        if indexPath.row == 3 {
+		if indexPath.row == 3 {
+			guard let cell = tableView.dequeueReusableCell(withIdentifier: BooleanCell.identifier, for: indexPath) as? BooleanCell else {
+				return UITableViewCell()
+			}
+			cell.setup(title: "Show Red Dot on right bar button items",
+					   isOn: showRedDotOnBarButtonItem,
+					   isSwitchEnabled: navigationItem.titleStyle == .largeLeading)
+			cell.titleNumberOfLines = 0
+			cell.onValueChanged = { [weak self, weak cell] in
+				self?.shouldShowRedDot(isOn: cell?.isOn ?? false)
+			}
+			return cell
+		}
+
+        if indexPath.row == 4 {
             guard let cell = tableView.dequeueReusableCell(withIdentifier: ActionsCell.identifier, for: indexPath) as? ActionsCell else {
                 return UITableViewCell()
             }
@@ -747,6 +762,16 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
         }
         showBadgeOnBarButtonItem = isOn
     }
+
+	@objc private func shouldShowRedDot(isOn: Bool) {
+		guard let items = navigationItem.rightBarButtonItems, !items.isEmpty else {
+			return
+		}
+		for item in items {
+			item.showRedDot(isOn)
+		}
+		showRedDotOnBarButtonItem = isOn
+	}
 
     @objc private func showTooltipButtonPressed() {
         let navigationBar = msfNavigationController?.msfNavigationBar
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
@@ -473,7 +473,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
     var showSearchProgressSpinner: Bool = true
     var showRainbowRingForAvatar: Bool = false
     var showBadgeOnBarButtonItem: Bool = false
-	var showRedDotOnBarButtonItem: Bool = false
+    var showRedDotOnBarButtonItem: Bool = false
 
     var allowsCellSelection: Bool = false {
         didSet {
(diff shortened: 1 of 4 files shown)


ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
```diff
@@ -473,6 +473,7 @@ class RootViewController: UIViewController, UITableViewDataSource, UITableViewDe
     var showSearchProgressSpinner: Bool = true
     var showRainbowRingForAvatar: Bool = false
     var showBadgeOnBarButtonItem: Bool = false
+	var showRedDotOnBarButtonItem: Bool = false
```
laminesm, line 476: Many of your changes are not properly aligned. This might be caused by using tabs instead of spaces on Xcode.

laminesm (commented):
Also, `BadgeLabelButton` already supports the red dot. All that is needed is to set `BadgeLabelStyle` to `.system` and have no text as `badgeValue`. Though the style and the control are internal, could we just leverage that?
Screenshot 2024-02-15 at 11 47 04 AM

laminesm (commented):
Also, how do these changes interact with the existing badge styles? Does the red dot override the badge label style?

ios/FluentUI/Navigation/UIBarButtonItem+RedDot.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import UIKit
+
+@objc public extension UIBarButtonItem {
```
mischreiber, line 8: We generally do not want to create public extensions to system classes in Fluent, as it forces all downstream consumers to deal with our extension in their namespace. Instead, either use a custom `UIBarButtonItem` subclass, or provide some form of helper class that can perform the mutation for you.

ios/FluentUI/Navigation/BadgeLabelButton.swift
```diff
@@ -212,10 +266,14 @@ class BadgeLabelButton: UIButton {
     }
 
     @objc private func badgeValueDidChange() {
-        updateBadgeLabel()
+        updateBadge()
         updateAccessibilityLabel()
     }
 
+    @objc private func redDotVisibilitDidChanged() {
```
mischreiber, line 273: spelling

mischreiber (changes requested):

harrieshin:
the concern i have is that this isn't a design paradigm in fluent design for navigation header. the small dot was only used for controls like pillbuttonbar or segmented control so far. why can't you use "1" to represent the unread notification? it follows most apple navigation standards ..

kumarmohit5189:
@harrie

> Also, `BadgeLabelButton` already supports the red dot. All that is needed is to set `BadgeLabelStyle` to `.system` and have no text as `badgeValue`. Though the style and the control are internal, could we just leverage that? Screenshot 2024-02-15 at 11 47 04 AM

Yes, we can do, but from API prospect it looks weird when we set blank title. So, i though to keep clean API so its clearer on what API is exacty does.

kumarmohit5189:
> the concern i have is that this isn't a design paradigm in fluent design for navigation header. the small dot was only used for controls like pillbuttonbar or segmented control so far. why can't you use "1" to represent the unread notification? it follows most apple navigation standards ..

showing 1 looks weird, 

> 

Another option to show red dot is passing blank value, and expose Button style property to App. But passing blank value seems weird here, So I try to make separate path similar to tab bar. Let me know your input for this.

kumarmohit5189:
> Also, how do these changes interact with the existing badge styles? Does the red dot override the badge label style?

its similar to Tab, If we set Badge and Red dot both, preference will be given to Badge and it will work as it was.

ios/FluentUI/Navigation/UIBarButtonItem+RedDot.swift
kumarmohit5189: But we are following same styling for Badge value as well. I am just adding one extra functionality on top of badge button by using same mechanism.

ios/FluentUI.Demo/FluentUI.Demo/Demos/NavigationControllerDemoController.swift
kumarmohit5189: yes updated it

ios/FluentUI/Navigation/UIBarButtonItem+RedDot.swift
mischreiber: The existence of technical debt does not automatically justify the addition of more. We should move the badge styling away from this pattern as well; adding more that needs to be changed will only increase future work, *especially* when it comes to public APIs that downstream partners expect to remain supported.

mischreiber (changes requested):

laminesm:
> @Harrie
> 
> > Also, `BadgeLabelButton` already supports the red dot. All that is needed is to set `BadgeLabelStyle` to `.system` and have no text as `badgeValue`. Though the style and the control are internal, could we just leverage that? Screenshot 2024-02-15 at 11 47 04 AM
> 
> Yes, we can do, but from API prospect it looks weird when we set blank title. So, i though to keep clean API so its clearer on what API is exactly does.
> 
> Also, about style as System, this is being set in private method itself, so may be confusing while exposing these things.
> 
> If still in case we wanted to use BadgeLabel, what is your suggestion for style thing. We can simply add a property similar to badge value? And pick this if set while creating Badge label button.
> 
> Please let me know your input here.

You can expose the style as a public property through the control that uses the BadgeLabel. For reference, see how we did it here for the tabbar https://github.com/microsoft/fluentui-apple/pull/1818/files

kumarmohit5189:
Open another PR as suggested above, please help to review @laminesm @mischreiber @harrieshin 

PR link https://github.com/microsoft/fluentui-apple/pull/1977

Closing it.

PR #1977, opened 2024-02-20
add property to override badge style in navigation bar

kumarmohit5189:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

For `NavigationBar` currently we don't have a way in case we want to change background color for `BadgeLabel`. It is picking using `NavigationBar` theme defined. As part of this change, added capability to override `BadgeLableStyle` for a particular screen, If this override happen then `BadgeLabel` will be using updated style for its UIBarButtonItems else it will follow `NavigationBar` style to set background. 

### Binary change

Total increase: 5,360 bytes
Total decrease: 0 bytes
| Total | 3,10,97,936 bytes | 3,11,03,296 bytes | ⚠️ 5,360 bytes |
<details>

| NavigationBar.o | 5,49,544 bytes | 5,53,480 bytes | ⚠️ 3,936 bytes |
| __.SYMDEF | 48,49,960 bytes | 48,50,824 bytes | ⚠️ 864 bytes |
| FocusRingView.o | 8,21,464 bytes | 8,22,024 bytes | ⚠️ 560 bytes |
</details>
### Verification

Validated this change by manual testing with different mode like dark mode / light mode etc. Since this change is using existing UI and flows, so is safe fix.

<details>

|  
Simulator Screenshot - iPhone 15 Pro - 2024-02-20 at 15 24 25 |
Simulator Screenshot - iPhone 15 Pro - 2024-02-20 at 15 27 13
  |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1977)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -724,9 +732,10 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
     private func createBarButtonItemButton(with item: UIBarButtonItem, isLeftItem: Bool) -> UIButton {
         let button = BadgeLabelButton(type: .system)
         button.item = item
-        if style == .system {
+        let finalStyle = badgeLabelStyle != nil ? badgeLabelStyle : style
```
mischreiber, line 735: Nit: can simplify as `let finalStyle = badgeLabelStyle ?? style`

kumarmohit5189: updated accordingly.

anandrajeswaran:
Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -589,6 +591,12 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         titleView.avatarOverrideStyle = style
     }
 
+    /// Override BadgeLabelStyle for navigation bar
```
anandrajeswaran, line 594: Can just make the `badgeLabelStyle` property public instead of manually implementing a setter with a different name? Or maybe rename that property to `overrideBadgeLabelStyle`?

kumarmohit5189:
> Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience

We need this to make Badgelabel style to designed format, without making any change in navigationBar.
Example if we keep navigationBar ar eprimary but want badge to system design (red color).

With primary nav style text and badge is of white color. Please refer attached screenshot.


Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 40 36
Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 46 15

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: if we do so and mark that as @objc, it doesn't allow optional values. Also, in swift we need to mark optional if want to assign nil.  So have created this approach. Please let me know your input.

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -589,6 +591,12 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
...
+    /// - Parameter badgeLabelStyle: updated style to be used
+    @objc public func overrideBadgeLabelStyle(_ badgeLabelStyle: Style) {
```
mischreiber, line 596: This API is unclear and relies on very specific behaviors about `Style`. That class describes the entire notification bar, not just the label style, so we are making two assumptions:
1) Callers of this API will understand exactly what will happen to the badge label if they specify a different nav bar style.
2) We will never change what label styles are associated with each navigation bar style.

If we are going to provide a way to change the color of the label, we should expose the final outcome of this: the button's `BadgeLabelStyle`, and have people override that.

mischreiber (changes requested):

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: only issue with exposing `BadgeLabelStyle` is that we can't keep as null by default since @Objc don't allow this with optional in swift. So, we need to assign some default value which means even when user not setting this explicitly it will be using that value.

Purpose of this API is that user explicitly call this to set, else it will be null.

Please let me know if we can achieve null handling in some better way? Basically we should have it null by default.

kumarmohit5189: making that property optional direct not allowed as it seems optional Enum in swift with property marked as @Objc not allowed. My thought process here is to use this property only when not nil. Assigning default value will create some confusion.
Screenshot 2024-02-29 at 11 56 23 AM

kumarmohit5189:
> Additional thought: the UI changes in your "before" and "after" images, but there are no new changes in the Demo app. If the default behavior of the button has changed, that is a major design change and can't be submitted. Otherwise, we should expose a toggle in the navigation controller demo to replicate the change you illustrated.

I wanted to close on https://github.com/microsoft/fluentui-apple/pull/1977#discussion_r1501177320, I will update UI accordingly. Can you please let me know your feedback for my comment? How we should handle this case?

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: @anandrajeswaran I have created one wrapper class to achieve this, with Enum optional its not allowing to expose, Give error as attached. To enable app to set null when not needed, We need to ensure this is optional property.

Screenshot 2024-02-29 at 11 56 23 AM

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: @mischreiber Have used wrapper class to fix this, please have a look.

kumarmohit5189:
> Additional thought: the UI changes in your "before" and "after" images, but there are no new changes in the Demo app. If the default behavior of the button has changed, that is a major design change and can't be submitted. Otherwise, we should expose a toggle in the navigation controller demo to replicate the change you illustrated.

@mischreiber here is UI changes demo, please have a look.

https://github.com/microsoft/fluentui-apple/assets/102029000/4d1bf0b1-43e8-4e99-b411-0ebeef3b9cec

anandrajeswaran:
> > Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience
> 
> We need this to make Badgelabel style to designed format, without making any change in navigationBar. Example if we keep navigationBar ar eprimary but want badge to system design (red color).
> 
> With primary nav style text and badge is of white color. Please refer attached screenshot.
> 
> Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 40 36 Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 46 15

Do we have design input on this - is it really intended that both are possible? Or is one of those the actual design and the other is just a bug?

ios/FluentUI/Navigation/NavigationBar.swift
anandrajeswaran: We definitely don't want to have to create a wrapper class just to expose to objC. One question first - do we actually need to consume this from objC? Assuming we do, still need to go towards the suggestion in this thread of exposing a `BadgeLabelStyle` rather than a `Style`

kumarmohit5189:
> > > Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience
> > 
> > 
> > We need this to make Badgelabel style to designed format, without making any change in navigationBar. Example if we keep navigationBar ar eprimary but want badge to system design (red color).
> > With primary nav style text and badge is of white color. Please refer attached screenshot.
> > Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 40 36 Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 46 15
> 
> Do we have design input on this - is it really intended that both are possible? Or is one of those the actual design and the other is just a bug?

@anandrajeswaran This is an addition to support unread dot like tab bar item. Generally unread dot should be Red. I raised another PR https://github.com/microsoft/fluentui-apple/pull/1974 for this to keep red dot as separate. But came with suggestion to leverage https://github.com/microsoft/fluentui-apple/pull/1974#pullrequestreview-1883723233  badgeLabel itself. Its just we need to expose property.
Please let me know your input for this. Any other better approach we can take here?

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: Yes, we need to access this in Objective C class, making it optional (since override happening, user can reset too) we need to add wrapper as Enum optional is not compliable.

Not clear with point - `Assuming we do, still need to go towards the suggestion in this thread of exposing a BadgeLabelStyle rather than a Style`, can you please explain bit more.

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: @anandrajeswaran have updated it, removed setter method now,

ios/FluentUI/Navigation/NavigationBar.swift
anandrajeswaran: `Style` (exposed to objC with MSFNavigationBarTitleAccessoryStyle) and `BadgeLabelStyle` (exposed to objC with MSFBadgeLabelStyle) are 2 different types, right?

anandrajeswaran:
> > > > Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience
> > > 
> > > 
> > > We need this to make Badgelabel style to designed format, without making any change in navigationBar. Example if we keep navigationBar ar eprimary but want badge to system design (red color).
> > > With primary nav style text and badge is of white color. Please refer attached screenshot.
> > > Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 40 36 Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 46 15
> > 
> > 
> > Do we have design input on this - is it really intended that both are possible? Or is one of those the actual design and the other is just a bug?
> 
> @anandrajeswaran This is an addition to support unread dot like tab bar item. Generally unread dot should be Red. I raised another PR #1974 for this to keep red dot as separate. But came with suggestion to leverage [#1974 (review)](https://github.com/microsoft/fluentui-apple/pull/1974#pullrequestreview-1883723233) badgeLabel itself. Its just we need to expose property. Please let me know your input for this. Any other better approach we can take here?

Let's chat offline with designers as well - presumably there was a reason for this existing design to not be red in this case.

ios/FluentUI/Navigation/NavigationBar.swift
kumarmohit5189: BadgeLabelButton is created internally in this class. So we won't have access to that object, ideally main app not aware about this thing.
Screenshot 2024-03-05 at 10 29 40 AM

kumarmohit5189: Updated code to override `badgeLabelStyle` instead of `Style`. But still we require wrapper to make sure we can set to nil if we don't want override.

mohit5189: marking resolve as changes are made

kumarmohit5189:
> > > Do we really want to expose this? Good to add an example to the Demo app and that should help with Before/After screenshots to justify why this is a desirable experience
> > 
> > 
> > We need this to make Badgelabel style to designed format, without making any change in navigationBar. Example if we keep navigationBar ar eprimary but want badge to system design (red color).
> > With primary nav style text and badge is of white color. Please refer attached screenshot.
> > Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 40 36 Simulator Screenshot - iPhone 15 Pro - 2024-02-22 at 10 46 15
> 
> Do we have design input on this - is it really intended that both are possible? Or is one of those the actual design and the other is just a bug?

We have it now in another discussion.

PR #1985, opened 2024-03-08
DrawerController escape accessibility

miolsako:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Users navigating with the Full Keyboard Access accessibility feature can get stranded in a DrawerController with no way to escape. We should not assume that every implementation of the DrawerController will include an explicit dismiss button but support the Escape key for dismissal, similar to other Apple controls.

### Binary change

N/A

### Verification

Validated using full keyboard access

<details>

| Escape key does not work | Here I am pressing the escape key and you can see we dismiss

EscapeKeyDismiss |

</details>

### Pull request checklist

This PR has considered:
- N/A Light and Dark appearances
- N/A Internationalization and Right to Left layouts
- N/A Different resolutions (1x, 2x, 3x)
- N/A Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- N/A iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- N/A [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- N/A Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1985)

### DIFF
--- a/ios/FluentUI/Drawer/DrawerController.swift
+++ b/ios/FluentUI/Drawer/DrawerController.swift
@@ -588,6 +588,16 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         }
     }
 
+	@objc func handleEscapeKey() {
+		dismissPresentingViewController(animated: true)
+	}
+
+	open override var keyCommands: [UIKeyCommand]? {
+		return [
+			UIKeyCommand(input: UIKeyCommand.inputEscape, modifierFlags: [], action: #selector(self.handleEscapeKey))
+		]
+	}
+
     open override func accessibilityPerformEscape() -> Bool {
         return dismissPresentingViewController(animated: true)
     }


ios/FluentUI/Drawer/DrawerController.swift
```diff
@@ -588,6 +588,16 @@ open class DrawerController: UIViewController, TokenizedControlInternal {
         }
     }
 
+	@objc func handleEscapeKey() {
+		dismissPresentingViewController(animated: true)
+	}
+
+	open override var keyCommands: [UIKeyCommand]? {
+		return [
+			UIKeyCommand(input: UIKeyCommand.inputEscape, modifierFlags: [], action: #selector(self.handleEscapeKey))
```
edjamesmsft, line 597: Looks like none of our other Fluent controls explicitly handle UIKeyCommand.inputEscape (or any other UIKeCommands). I will do a little digging around, but I wonder if this is the right fix. It works, but is it a bandaid? Do our other controls have the same problem? If just the DrawerController, it'd be good to understand why it differs from the other controls

joannaquu:
Isn't not an explicit dismiss button not an accessibility issue?

ios/FluentUI/Drawer/DrawerController.swift
miolsako: This is in response to an accessibility bug we go in Loop iOS. Instead of a patch fix just on our end I figured it would be worth it to fix it for Fluent all up as I was able to repro the issue in the test app. Other native Apple controls that pop out a menu do support dismissal with the Escape key (you can see an example in the Notes all), but I suspect since a ViewController has various methods of presentation possible there was no default behaviour.

I did not look into other Fluent controls as I felt that was out of scope for the change.

miolsako:
> Isn't not having an explicit dismiss button an accessibility issue?

It does not seem to be a requirement. Apple itself has menus without explicit close buttons in its native apps. Example from the Notes app:

ios/FluentUI/Drawer/DrawerController.swift
anandrajeswaran: From side conversation, looks like Tab B is the default method of dismissing popovers on iOS with Full Keyboard Access (ESC doesn't actually work for me with Full Keyboard Access in other apps). Are you seeing that doesn't work in the Fluent Demo app?
image

miolsako: Closing this PR based on conversation above. Tab B can be used as well as overridden by a user to use Esc if that is their preference

PR #2000, opened 2024-04-22
Add support for ListItem on visionOS

joannaquu:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

Adding support for `ListItem` on visionOS.

Currently, `ListItem` internally is a `View`, which means on visionOS, there is no hover/gaze effect. When there are no tap gestures added to the view, this is ok. However, when the `ListItem` is tappable, this presents a usability issue (users can't tell where they are and may not be aware the view is tappable). To fix this, update `ListItem` to be a `Button` when necessary.

List of changes
- New parameter: `action` is an optional closure that executes when the `ListItem` is tapped.
- `ListItemButtonStyle`: a custom SwiftUI button style that is applied when `ListItem` is a `Button`
  - The default cornerRadius makes the button a capsule, update it to match `TableViewCell`
  - Use `cellBackgroundSelectedColor` as the background color when selected
  - The resting color is already set on the `ListItem` itself
- Setting a `buttonStyle` on the `ListItem` will trickle down to the detail button. Explicitly set the detail button's style. On iOS, it should be `.plain` and on visionOS it should be `.bordered`
  - When `ListItem` was first implemented, it was done so to mimic `TableViewCell` as closely as possible. This meant a larger padded detail button. To achieve this, the accessory view's padding was applied to the image and an extra background was added to address VoiceOver. Both of these changes do not work on visionOS (bloated button and darker background).
    - To fix the padding, apply it to the accessory view/`innerContent` instead. Changes made:
      - Move the padding values around so `paddingLeading/paddingTrailing/paddingVertical` are always be on `innerContent`. No matter what's inside the list item, these padding values always apply (this allows us to remove all the conditional code depending on if there's a `trailingContentView`/accessory view.
      - `horizontalSpacing` should then apply to the leading edge of the `trailingContentView` and accessory view. Since it's on the leading edge, we don't run into the issue of double padding.
    - Remove the VoiceOver change, this is actually a bug in `TableViewCell`, not `ListItem` (see screenshot of `UITableViewCell`)
- To preserve existing behavior, when `action` is nil, `ListItem` should still be a `View`
- Update `ListItemDemoController_SwiftUI` with an option to add/remove an `action`
- Add option to display `ListItem` outside `List`

### Binary change

Total increase: 56,264 bytes
Total decrease: -1,568 bytes
| Total | 31,359,184 bytes | 31,413,880 bytes | 🛑 54,696 bytes |
<details>

| ListItem.o | 268,680 bytes | 315,992 bytes | ⚠️ 47,312 bytes |
| __.SYMDEF | 4,830,512 bytes | 4,839,192 bytes | ⚠️ 8,680 bytes |
| FocusRingView.o | 837,952 bytes | 838,224 bytes | ⚠️ 272 bytes |
| SwiftUI+ViewModifiers.o | 128,464 bytes | 126,896 bytes | 🎉 -1,568 bytes |
</details>

### Verification

<details>

| Before                                       | After iOS                                     | After visionOS                                     |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
| none before | none after | none |
| detail before | detail after | detail |

Standalone `ListItem`
| iOS                                       | visionOS                                      |
| standalone ipad | standalone vision |

VoiceOver change
| UITableViewCell                                       | Before                                      | After                                      |
|----------------------------------------------|--------------------------------------------|--------------------------------------------|
| vo | ios before | ios after |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2000)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
@@ -21,7 +21,8 @@ class ListItemDemoControllerSwiftUI: UIHostingController<ListItemDemoView> {
 }
 
 struct ListItemDemoView: View {
-    @State var showingAlert: Bool = false
+    @State var showingPrimaryAlert: Bool = false
+    @State var showingSecondaryAlert: Bool = false
     @ObservedObject var fluentTheme: FluentTheme = .shared
     let accessoryTypes: [ListItemAccessoryType] = [.none, .checkmark, .detailButton, .disclosureIndicator]
 
@@ -32,6 +33,7 @@ struct ListItemDemoView: View {
     @State var showFooter: Bool = false
     @State var showLeadingContent: Bool = true
     @State var showTrailingContent: Bool = true
+    @State var isTappable: Bool = true
     @State var isDisabled: Bool = false
     @State var accessoryType: ListItemAccessoryType = .none
     @State var leadingContentSize: ListItemLeadingContentSize = .default
@@ -41,6 +43,7 @@ struct ListItemDemoView: View {
     @State var footerLineLimit: Int = 1
     @State var trailingContentFocusableElementCount: Int = 0
     @State var trailingContentToggleEnabled: Bool = true
+    @State var renderStandalone: Bool = false
 
     public var body: some View {
 
@@ -73,7 +76,9 @@ struct ListItemDemoView: View {
                 .accessibilityIdentifier("leadingContentSwitch")
             FluentUIDemoToggle(titleKey: "Show trailing content", isOn: $showTrailingContent)
                 .accessibilityIdentifier("trailingContentSwitch")
+            FluentUIDemoToggle(titleKey: "Tappable", isOn: $isTappable)
             FluentUIDemoToggle(titleKey: "Disabled", isOn: $isDisabled)
+            FluentUIDemoToggle(titleKey: "Render standalone", isOn: $renderStandalone)
         }
 
         @ViewBuilder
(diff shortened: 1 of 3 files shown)


alexanderboswell:
What happens if the `ListItem` is part of a navigationLink? I believe that is going to have it as a button already. Is that going to prevent that from working?

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
```diff
@@ -73,7 +76,9 @@ struct ListItemDemoView: View {
                 .accessibilityIdentifier("leadingContentSwitch")
             FluentUIDemoToggle(titleKey: "Show trailing content", isOn: $showTrailingContent)
                 .accessibilityIdentifier("trailingContentSwitch")
+            FluentUIDemoToggle(titleKey: "Tappable", isOn: $isTappable)
             FluentUIDemoToggle(titleKey: "Disabled", isOn: $isDisabled)
+            FluentUIDemoToggle(titleKey: "Render standalone", isOn: $renderStandalone)
```
mischreiber, line 81: Do we just want to always render standalone? One of each?

joannaquu: I left it as the non-standalone one (in a List) since that's what it was before but did we want both?

alexanderboswell:
Can you add a screenshot of what it looks like in VoiceOver when the whole listItem is selected?

joannaquu:
> Can you add a screenshot of what it looks like in VoiceOver when the whole listItem is selected?

yup! they all look good except when the listitem is standalone and in a vstack (we run into the same issue where voiceover does not wrap the padded view). looking into if we can shift our padding values around
| | List | Standalone|
| - | - | - |
| View | list view | vstack view |
| Button | list button | vstack button |

nodes11 (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
mischreiber: Probably fine for now, but a future revision might be nice to feature both.

mischreiber (approved):

PR #2005, opened 2024-04-24
Add localized strings to expose accessibility string

alexanderboswell:
### Platforms Impacted
- iOS
- visionOS

### Description of changes

Today we do not have a control for a navigation controller in swiftUI. Part of the navigation controller is providing a custom back button which will need an accessibility identifier. Until we support a NavigationStack for swiftUI, we need the same accessibility identifier. Exposing it so it can be used.
### Binary change

Total increase: 3,176 bytes
Total decrease: 0 bytes
| Total | 31,260,080 bytes | 31,263,256 bytes | ⚠️ 3,176 bytes |
<details>

| NavigationBar.o | 553,064 bytes | 555,896 bytes | ⚠️ 2,832 bytes |
| __.SYMDEF | 4,769,144 bytes | 4,769,488 bytes | ⚠️ 344 bytes |
</details>

### Verification

Ran VoiceOver and the accessibility identifier is still read properly on the navigation bar.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2005)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
+++ b/ios/FluentUI.Demo/FluentUI.Demo.xcodeproj/project.pbxproj
@@ -85,6 +85,7 @@
 		8F0B81122670200300463726 /* AppCenterDistribute in Frameworks */ = {isa = PBXBuildFile; platformFilter = ios; productRef = 8F0B81112670200300463726 /* AppCenterDistribute */; };
 		8F0B8114267021A700463726 /* AppCenterAnalytics in Frameworks */ = {isa = PBXBuildFile; platformFilter = ios; productRef = 8F0B8113267021A700463726 /* AppCenterAnalytics */; };
 		8F0B8116267021A700463726 /* AppCenterCrashes in Frameworks */ = {isa = PBXBuildFile; platformFilter = ios; productRef = 8F0B8115267021A700463726 /* AppCenterCrashes */; };
+		9211BC612BD97B7900B95200 /* DemoHostingController.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9211BC602BD97B7900B95200 /* DemoHostingController.swift */; };
 		92279B352B97F5DA00994D88 /* ButtonDemoController_SwiftUI.swift in Sources */ = {isa = PBXBuildFile; fileRef = 92279B342B97F5D900994D88 /* ButtonDemoController_SwiftUI.swift */; };
 		923DF2DB271158C900637646 /* libFluentUI.a in Frameworks */ = {isa = PBXBuildFile; fileRef = 923DF2DA271158C900637646 /* libFluentUI.a */; };
 		923DF2DF27115B4700637646 /* FluentUIResources-ios.bundle in Resources */ = {isa = PBXBuildFile; fileRef = 923DF2DC271158CD00637646 /* FluentUIResources-ios.bundle */; };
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Resources/Localization/StringKeys.swift
```diff
@@ -0,0 +1,13 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import Foundation
+
+/// Keys for getting localized strings.
+@objc(MSFStringKeys)
+public class StringKeys: NSObject {
```
mischreiber, line 10: I disagree on the `N/A` for binary size: adding a new public class will definitely have a measurable impact. Let's get it recorded just for reference.

mischreiber (commented):
Could we also get a demo of how this would be consumed by an outside client in the Demo app?

alexanderboswell:
> Could we also get an example of how this would be consumed by an outside client in the Demo app?

just in the PR you mean, or in the fluent demo app?

ios/FluentUI/Resources/Localization/StringKeys.swift
alexanderboswell: updated

alexanderboswell:
> Could we also get an example of how this would be consumed by an outside client in the Demo app?

sorry read this too fast the first time. Do we really care to do that? Just because we are under the shared understanding that this shouldnt live on forever. Also, we would have to create a whole new entry in the demo app, as we dont have a navigationstack in the demo app yet, and that seems a bit overkill?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -349,7 +349,7 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
                                              target: nil,
                                              action: #selector(NavigationBarBackButtonDelegate.backButtonWasPressed))
         backButtonItem.accessibilityIdentifier = "Back"
-        backButtonItem.accessibilityLabel = "Accessibility.NavigationBar.BackLabel".localized
+        backButtonItem.accessibilityLabel = StringKeys.accessibilityNavigationBarBackLabel.localized
```
laminesm, line 352: I don't think this change is necessary especially if the new class is created temporarily. We should limit its usage to be strictly external and not internal so we don't have to touch this again when StringKeys changes later on.

alexanderboswell: for code maintainability though, someone could change internally and break the external usage without knowing it?

laminesm: I doubt this would change but even if a change were to happen it would probably be the value of the string and not the key itself that would change and that shouldn't result in any breaks as far as I can tell.

laminesm (commented):
Would this need to be exposed too https://github.com/microsoft/fluentui-apple/blob/0e7c12e038b47125c6dba7ad065ce7e5ef41c3bb/ios/FluentUI/Extensions/String%2BExtension.swift#L9?

ios/FluentUI/Navigation/NavigationBar.swift
alexanderboswell: yeah oops, ill make a change to expose the localized string

ios/FluentUI/Resources/Localization/LocalizedStrings.swift
```diff
@@ -0,0 +1,15 @@
+//
```
anandrajeswaran, line 1: Probably don't want a generic API for exposing localized strings per offline conversation so let's not add this file. Instead just expose this specific string from NavigationBar.swift?

anandrajeswaran (changes requested):

ios/FluentUI/Resources/Localization/LocalizedStrings.swift
alexanderboswell: Done

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -165,6 +165,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         }
     }
 
+    /// The accessibility label that should be applied for the back button.
+    @objc public static var backButtonAccessibilityLabel: String {
```
laminesm, line 169: Maybe this should be a let?

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -165,6 +165,9 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
         }
     }
 
+    /// The accessibility label that should be applied for the back button.
```
anandrajeswaran, line 168: In the spirit of not wanting this to live on - probably should add in a comment that consumers should ideally not use this - this is just temporary to avoid duplicating resources until we bring better SwiftUI support into the repo such that SwiftUI clients don't need to reconstruct fluent elements

anandrajeswaran (approved):

PR #2011, opened 2024-04-29
expose accessibility string for done action in navigation bar

alexanderboswell:
### Platforms Impacted
- iOS
- visionOS

### Description of changes

Today we do not have a control for a navigation controller in swiftUI. Part of the navigation controller is providing a dismiss action which will need an accessibility identifier. Until we support a NavigationStack for swiftUI, we need the same accessibility identifier. Exposing it so it can be used.
### Binary change

Total increase: 3,101 bytes
Total decrease: 0 bytes
| Total | 31,260,080 bytes | 31,263,181 bytes | ⚠️ 3,101 bytes |

### Verification

Ran VoiceOver and the accessibility identifier is still read properly on the navigation bar.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2005)

### DIFF
--- a/ios/FluentUI/Navigation/NavigationBar.swift
+++ b/ios/FluentUI/Navigation/NavigationBar.swift
@@ -169,6 +169,10 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
     /// A temporary change so that consumers who use SwiftUI for navigation can avoid duplicated resources until support of a swiftUI control is available.
     @objc public static let backButtonAccessibilityLabel: String = "Accessibility.NavigationBar.BackLabel".localized
 
+	/// The accessibility label that should be applied for the done button for when navigation bar is shown in a modal view.
+	/// A temporary change so that consumers who use SwiftUI for navigation can avoid duplicated resources until support of a swiftUI control is available.
+	@objc public static let doneButtonAccessibilityLabel: String = "Accessibility.Done.Label".localized
+
     /// An element size to describe the behavior of large title's avatar. If `.automatic`, avatar will resize when `expand(animated:)` and `contract(animated:)` are called.
     @objc open var avatarSize: ElementSize = .automatic {
         didSet {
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -169,6 +169,10 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
     /// A temporary change so that consumers who use SwiftUI for navigation can avoid duplicated resources until support of a swiftUI control is available.
     @objc public static let backButtonAccessibilityLabel: String = "Accessibility.NavigationBar.BackLabel".localized
 
+	/// The accessibility label that should be applied for the done button for when navigation bar is shown in a modal view.
+	/// A temporary change so that consumers who use SwiftUI for navigation can avoid duplicated resources until support of a swiftUI control is available.
+	@objc public static let doneButtonAccessibilityLabel: String = "Accessibility.Done.Label".localized
```
joannaquu, line 174: nit: whitespace, fluent uses spaces instead of tabs

joannaquu (approved):

ios/FluentUI/Navigation/NavigationBar.swift
joannaquu, line 174: since the nav bar doesn't actually use the accessibility label, do we want to export it from a control that does? I'm seeing BarButtonItems?

alexanderboswell: yeah I wasn't sure what the best way to do this was, as the string isn't specific to barbuttonItems per say, its just the only place we use it. 

thinking about when we move to SwiftUI for list, it seems like we will be making our own navigation bar implementation, so it seemed like this would be a good spot given that is where we would have to add it. 

I dont have a strong opinion one way or the other, but grouping it with the other label that had to be exposed seemed also useful

PR #2015, opened 2024-05-03
Update button minimum height on visionOS

joannaquu:
### Platforms Impacted
- [ ] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

As recommended by apple hig, buttons on visionOS should be at least 44 px tall. Until we have an official updated size ramp for visionOS, both `.small` and `.medium` buttons will have a minimum of 44.

### Binary change

Total increase: 1,048 bytes
Total decrease: 0 bytes
| Total | 31,515,176 bytes | 31,516,224 bytes | ⚠️ 1,048 bytes |
<details>

| ButtonTokenSet.o | 125,216 bytes | 126,264 bytes | ⚠️ 1,048 bytes |
</details>

### Verification

<details>

| before | after |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2015)


ios/FluentUI/Button/ButtonTokenSet.swift
```diff
@@ -344,9 +344,9 @@ extension ButtonTokenSet {
             case .large:
                 return 52
             case .medium:
-                return 40
+                return Compatibility.isDeviceIdiomVision() ? 44 : 40
```
mischreiber, line 347: How is this different from the iOS HIG, which also recommends 44pt high buttons on iOS?

mischreiber: Actually, [per the HIG](https://developer.apple.com/design/human-interface-guidelines/buttons#Best-practices) even that's not quite right:
> As a general rule, a button needs a hit region of at least 44x44 pt — in visionOS, 60x60 pt — to ensure that people can select it easily, whether they use a fingertip, a pointer, their eyes, or a remote.

joannaquu: that's hit area right? this is for the actual size of the button?

mischreiber: Technically true, but our existing buttons all use the same hit and visual bounding boxes. And I'm not sure I'd want to change that except in specific one-off cases.

PR #2019, opened 2024-05-15
Initial Implementation of FluentList and FluentListSection

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

This is an initial version of `FluentList` and `FluentListSection`, wrappers around `SwiftUI.List` and `SwiftUI.Section` that apply the appropriate fluent style. The intention is that that these views can be used with `FluentUI.ListItem` to create a completely fluentized list or menu in SwiftUI.

### Binary change

Total increase: 76,760 bytes
Total decrease: 0 bytes
| Total | 30,668,792 bytes | 30,745,552 bytes | 🛑 76,760 bytes |
<details>

| FluentListSection.o | 0 bytes | 38,448 bytes | ⚠️ 38,448 bytes |
| FluentList.o | 0 bytes | 27,368 bytes | ⚠️ 27,368 bytes |
| __.SYMDEF | 4,724,544 bytes | 4,734,424 bytes | ⚠️ 9,880 bytes |
| FocusRingView.o | 829,400 bytes | 829,992 bytes | ⚠️ 592 bytes |
| ListItem.o | 319,984 bytes | 320,456 bytes | ⚠️ 472 bytes |
</details>

### Verification

The changes have been validated in the 2 `ListItem` demo controllers to ensure there are no visual or functional regressions between the usage base SwiftUI components and the new FluentUI wrappers.

<details>

| NA | Screenshot 2024-05-15 at 2 43 00 PM |
| NA | Screenshot 2024-05-15 at 2 43 13 PM |

Demo of changing the ListStyle:

https://github.com/microsoft/fluentui-apple/assets/10938746/812c9712-ae57-4ad6-a048-e5cad8077a11

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2019)

### DIFF
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
@@ -84,8 +84,8 @@ struct ListActionItemDemoView: View {
 
         @ViewBuilder
         var content: some View {
-            List {
-                Section {
+            FluentList {
+                FluentListSection {
                     if showSecondaryAction {
                         ListActionItem(primaryActionTitle: primaryActionTitle,
                                        onPrimaryActionTapped: {
@@ -110,7 +110,6 @@ struct ListActionItemDemoView: View {
                         .bottomSeparatorType(bottomSeparatorType)
                         .backgroundStyleType(backgroundStyleType)
                     }
-
                 } header: {
                     Text("ListActionItem")
                 }
@@ -119,7 +118,7 @@ struct ListActionItemDemoView: View {
                         .accessibilityIdentifier("DismissAlertButton")
                 }
 
-                Section {
+                FluentListSection {
                     FluentUIDemoToggle(titleKey: "Show secondary action", isOn: $showSecondaryAction)
                         .accessibilityIdentifier("showSecondaryActionSwitch")
                     textFields
@@ -129,7 +128,6 @@ struct ListActionItemDemoView: View {
                 }
             }
             .fluentTheme(fluentTheme)
-            .listStyle(.insetGrouped)
         }
 
         return content
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
@@ -138,27 +138,11 @@ struct ListActionItemDemoView: View {
             }
         }
 
-        @ViewBuilder
-        var list: some View {
-            if listStyle == .grouped {
-                FluentList(listStyle: .grouped) {
-                    content
-                }
-            } else if listStyle == .inset {
-                FluentList(listStyle: .inset) {
-                    content
-                }
-            } else if listStyle == .insetGrouped {
-                FluentList(listStyle: .insetGrouped) {
-                    content
-                }
-            } else {
-                FluentList {
-                    content
-                }
-            }
+        return FluentList {
+            content
         }
-
-        return list.fluentTheme(fluentTheme)
+        .fluentListStyle(listStyle)
+        .background(ListItem.listBackgroundColor(for: .grouped))
+        .fluentTheme(fluentTheme)
     }
 }
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
@@ -83,7 +83,6 @@ struct ListActionItemDemoView: View {
 
             let listStylePickerOptions = Group {
                 Text(".plain").tag(FluentListStyle.plain)
-                Text(".grouped").tag(FluentListStyle.grouped)
                 Text(".insetGrouped").tag(FluentListStyle.insetGrouped)
                 Text(".inset").tag(FluentListStyle.inset)
             }
@@ -138,11 +137,15 @@ struct ListActionItemDemoView: View {
             }
         }
 
-        return FluentList {
-            content
+        @ViewBuilder
+        var list: some View {
+            FluentList {
+                content
+            }
+            .fluentListStyle(listStyle)
+            .fluentTheme(fluentTheme)
         }
-        .fluentListStyle(listStyle)
-        .background(ListItem.listBackgroundColor(for: .grouped))
-        .fluentTheme(fluentTheme)
+
+        return list
     }
 }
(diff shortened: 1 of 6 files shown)


ios/FluentUI/List/FluentList.swift
```diff
@@ -0,0 +1,33 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// This a wrapper around `SwiftUI.List` that has fluent style applied. It is intended to be used in conjunction with `FluentUI.FluentListSection` and `FluentUI.ListItem`
```
alexanderboswell, line 8: should we provide any comments that this is a work-in-progress?

Or maybe not needed because we dont expose any customization yet, so just get the default styling?

ios/FluentUI/List/FluentListSection.swift
```diff
@@ -0,0 +1,52 @@
...
+
+    // MARK: Initializer
+
+    /// Creates a `FluentListSection`
+    /// - Parameters:
+    ///   - content: content to show inside of the section.
+    ///   - header: content to show inside of the header. Defaults to an `EmptyView`.
+    ///   - footer: content to show inside of the footer. Defaults to an `EmptyView`.
+    public init(@ViewBuilder content: @escaping () -> SectionContent,
+                @ViewBuilder header: @escaping () -> SectionHeaderContent = { EmptyView() },
```
alexanderboswell, line 20: how is spacing between sections affected if we provide a default "EmptyView()" for header and footer? I think it might add additional unwanted spacing.

alexanderboswell, line 20: I guess we can't have nullable, but should we have other inits that only specify a header or a footer?  I did this in listItem as extensions

joannaquu:
neither have anything Fluent specific attached to them yet right (fluent theme/color is still added externally)? we're just creating the wrappers here

alexanderboswell (approved):

ios/FluentUI/List/FluentListSection.swift
nodes11: Great idea. I will update this.

nodes11: The spacing of the sections is identical if the `EmptyView()` is used as a default. I compared a `SwiftUI.Section` with a `FluentListSection` and both had the same dimensions/spacing.

ios/FluentUI/List/FluentList.swift
nodes11: I think leaving a brief comment is a good idea just so it's clear.

ios/FluentUI/List/FluentList.swift
```diff
@@ -0,0 +1,35 @@
...
+    ///   - content: SwiftUI content to show inside of the list.
+    public init(@ViewBuilder content: @escaping () -> ListContent) {
+        self.content = content
+    }
+
+    public var body: some View {
+        List {
+            content()
+        }
+        .listStyle(.insetGrouped)
```
joannaquu, line 27: similar to tvc and list item, do we also want to support plain, grouped style?

joannaquu (approved):

ios/FluentUI/List/FluentList.swift
nodes11: Good idea. Updated to take an optional list style and updated the demo controllers. See demo video in validation section.

alexanderboswell: I think this would be better as a modifier? So we have some default value, but then could set the value in a modifier. Makes it not an init and just add if we dont want the default style. (Like what is done in `ListItemModifiers.swift` as one example)

joannaquu: wait I just realized. this is just a system modifier so we don’t even need to set this right? leave it up to the client to set the style. unless we wanted to enforce fluent background colors

alexanderboswell: yeah we will be enforcing a fluent color for the background

nodes11: Okay, more good ideas. I realized the awkwardness and misalignment with Apple of the style param in the init shortly after posting the revision. I've taken Alex's suggestion and moved the list style to a view modifier for `FluentList`. It and the custom style enum should make supporting the fluent background color easier.

ios/FluentUI/List/FluentList.swift
```diff
@@ -0,0 +1,64 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+import SwiftUI
+
+/// Fluent specific list style enum
+public enum FluentListStyle {
+    case plain
```
mischreiber, line 10: Do we actually plan to support all four of these styles?

mischreiber (approved):

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListActionItemDemoController_SwiftUI.swift
```diff
@@ -80,58 +80,69 @@ struct ListActionItemDemoView: View {
...
-            .listStyle(.insetGrouped)
         }
 
-        return content
+        return FluentList {
+            content
+        }
+        .fluentListStyle(listStyle)
+        .background(ListItem.listBackgroundColor(for: .grouped))
+        .fluentTheme(fluentTheme)
```
joannaquu, line 146: nit: can we move this to a viewbuilder and just have return contentView or something like that

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
```diff
@@ -202,23 +209,22 @@ struct ListItemDemoView: View {
...
                 }
-                .background(ListItem.listBackgroundColor(for: .grouped))
+                .fluentListStyle(listStyle)
                 .fluentTheme(fluentTheme)
-                .listStyle(.insetGrouped)
             }
         }
 
-        return content
+        return content.background(ListItem.listBackgroundColor(for: .grouped))
```
joannaquu, line 227: why'd we have to move the background here?

ios/FluentUI/List/FluentList.swift
nodes11: Went ahead and removed `.grouped`.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
nodes11: This was a part of validation and should not be moved. Reverted.

alexanderboswell (approved):

PR #2021, opened 2024-05-16
Avatar custom default image

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

Add a new API to override the default image in `Avatar`. This is the little "person" image that is displayed when there is no text or image set.

New APIs: 
```swift
// MARK: - UIKit

@objc public protocol MSFAvatarState {
    /// An override for the template icon to use when there is no set image or name.
    var customDefaultImage: UIImage? { get set }
}

// MARK: - SwiftUI

extension Avatar {
    /// An override for the template icon to use when there is no set image or name.
    /// - Parameter customDefaultImage: Image to be used as the Avatar's default image.
    /// - Returns: The modified Avatar with the property set.
    func customDefaultImage(_ customDefaultImage: UIImage?) -> Avatar
}
```

### Binary change

Total increase: 8,424 bytes
Total decrease: 0 bytes
| Total | 31,086,904 bytes | 31,095,328 bytes | ⚠️ 8,424 bytes |
<details>

| Avatar.o | 604,704 bytes | 609,424 bytes | ⚠️ 4,720 bytes |
| AvatarModifiers.o | 44,312 bytes | 46,104 bytes | ⚠️ 1,792 bytes |
| __.SYMDEF | 4,849,552 bytes | 4,850,912 bytes | ⚠️ 1,360 bytes |
| FocusRingView.o | 839,832 bytes | 840,192 bytes | ⚠️ 360 bytes |
| AvatarGroup.o | 417,456 bytes | 417,648 bytes | ⚠️ 192 bytes |
</details>

### Verification

<details>

| Avatar UIKit | Avatar SwiftUI |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2021)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -166,8 +166,7 @@ public struct Avatar: View, TokenizedControlView {
             if shouldUseDefaultImage {
                 var defaultImage = state.defaultImage
                 if defaultImage == nil {
-                    let isOutlinedStyle = style == .outlined || style == .outlinedPrimary
-                    defaultImage = UIImage.staticImageNamed(isOutlinedStyle ? "person_48_regular" : "person_48_filled")
+                    defaultImage = UIImage.staticImageNamed(AvatarTokenSet.defaultImageName(style))
                 }
                 return (defaultImage, .template)
             } else {
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -34,6 +34,9 @@ import SwiftUI
     /// The image used to fill the ring as a custom color.
     var imageBasedRingColor: UIImage? { get set }
 
+    /// An override for the template icon to use when there is no set image or name.
+    var customDefaultImage: UIImage? { get set }
+
     /// Defines whether the avatar state transitions are animated or not. Animations are enabled by default.
     var isAnimated: Bool { get set }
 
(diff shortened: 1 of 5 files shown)


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -34,6 +34,9 @@ import SwiftUI
     /// The image used to fill the ring as a custom color.
     var imageBasedRingColor: UIImage? { get set }
 
+    /// An override for the template icon to use when there is no set image or name.
+    var customDefaultImage: UIImage? { get set }
```
alexanderboswell, line 38: Does `Custom` really mean much here? Could it be simplified to just `defaultImage`? By setting it you are implying it is custom already?

alexanderboswell: or at least for the modifier API, here it would still make sense to distinguish between variables.

mischreiber: Yeah I went back and forth there. Due to the design of `MSFAvatar`, this name will have to match the public API for at least UIKit consumers.

I don't love having a variable named `defaultImage` that might be `nil` even when we clearly have a value, but it does match the rest of the API, so I guess we should be consistent.

ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -161,11 +164,15 @@ public struct Avatar: View, TokenizedControlView {
 
         let avatarImageInfo: (image: UIImage?, renderingMode: Image.TemplateRenderingMode) = {
             if shouldUseDefaultImage {
-                let isOutlinedStyle = style == .outlined || style == .outlinedPrimary
-                return (UIImage.staticImageNamed(isOutlinedStyle ? "person_48_regular" : "person_48_filled"), .template)
+                var defaultImage = state.defaultImage
+                if defaultImage == nil {
+                    let isOutlinedStyle = style == .outlined || style == .outlinedPrimary
+                    defaultImage = UIImage.staticImageNamed(isOutlinedStyle ? "person_48_regular" : "person_48_filled")
```
nodes11, line 170: minor: Should these image names be move to a constants struct?

nodes11 (approved):

alexanderboswell (approved):

ios/FluentUI/Avatar/Avatar.swift
mischreiber: That would be nice! Not a huge fan of having them inline either.

nodes11 (approved):

PR #2023, opened 2024-05-17
[FluentList] Set environment variable for styling so child views can utilize it

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes

The `FluentList` and `FluentItem` are designed to be used together but there will need to be some coordination between them for styling. This is so that the item can be styled to match the type of list. 

To achieve this, it seemed to make sense to use an environment variable so the list can set the list style and pass it to any child views that want to use it, like a `ListItem` view. This way implementors don't need to set a matching style on every `ListItem` to be the same as the style of the wrapping `FluentList`. We will be able to also use this value for `FluentListSection`, `FluentListSectionHeader`, and `FluentListSectionFooter`.

If an implementor for some reason still wanted to specify a different style on `ListItem`, we respect that value over the environment variable.

Updated both the background color and the spacing for the `insetGrouped` list style.

### Binary change

Total increase: 36,016 bytes
Total decrease: 0 bytes
| Total | 30,968,016 bytes | 31,004,032 bytes | ⚠️ 36,016 bytes |
<details>

| FluentList.o | 49,288 bytes | 70,896 bytes | ⚠️ 21,608 bytes |
| ListItem.o | 319,352 bytes | 325,840 bytes | ⚠️ 6,488 bytes |
| __.SYMDEF | 4,750,168 bytes | 4,756,024 bytes | ⚠️ 5,856 bytes |
| IndeterminateProgressBar.o | 231,256 bytes | 232,400 bytes | ⚠️ 1,144 bytes |
| FocusRingView.o | 831,480 bytes | 832,400 bytes | ⚠️ 920 bytes |
</details>

### Verification

<details>

| | Before                                       | After                                      |
|--|----------------------------------------------|--------------------------------------------|
| header height and spacing (with a header string, with an empty string, and with no header provided) | image| image |

With `.grouped` style specified for `ListItem`

With `.plain` style specified for `ListItem`

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2023)

### DIFF
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -249,7 +249,7 @@ public struct ListItem<LeadingContent: View,
     /// The `ListItemAccessoryType` that the `ListItem` should display.
     var accessoryType: ListItemAccessoryType = .none
 
-    /// The custom background styling of the `ListItem`, which is preferred over the `ListStyle` environment value.
+    /// The custom background styling of the `ListItem`, which is preferred over the `FLuentListStyle` environment value.
     var customBackgroundStyleType: ListItemBackgroundStyleType?
 
     /// The handler for when the `detailButton` accessory view is tapped.
--- a/ios/FluentUI/List/ListItem.swift
+++ b/ios/FluentUI/List/ListItem.swift
@@ -249,8 +249,8 @@ public struct ListItem<LeadingContent: View,
     /// The `ListItemAccessoryType` that the `ListItem` should display.
     var accessoryType: ListItemAccessoryType = .none
 
-    /// The background styling of the `ListItem` to match the type of `List` it is displayed in.
-    var backgroundStyleType: ListItemBackgroundStyleType = .plain
+    /// The custom background styling of the `ListItem`, which is preferred over the `ListStyle` environment value.
+    var customBackgroundStyleType: ListItemBackgroundStyleType?
 
     /// The handler for when the `detailButton` accessory view is tapped.
     var onAccessoryTapped: (() -> Void)?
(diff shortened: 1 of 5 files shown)


ios/FluentUI/List/ListItem.swift
```diff
@@ -281,8 +281,24 @@ public struct ListItem<LeadingContent: View,
 
     // MARK: Private variables
 
+    /// The background styling of the `ListItem`.
+    private var backgroundStyleType: ListItemBackgroundStyleType {
```
nodes11, line 285: I think we generally try to avoid early returns unless it's a `guard...else` to validate parameters. Maybe this could just be:

```
let styleType: ListItemBackgroundStyleType
if let customBackgroundStyleType {
    styleType = customBackgroundStyleType
} else {
    switch listStyle {
    case .plain, .inset:
        styleType = .plain
    case .insetGrouped:
        styleType = .grouped
    }
}
return styleType
```

https://microsoft.github.io/swift-guide/EarlyReturn.html

ios/FluentUI/List/ListItem.swift
```diff
@@ -249,8 +249,8 @@ public struct ListItem<LeadingContent: View,
     /// The `ListItemAccessoryType` that the `ListItem` should display.
     var accessoryType: ListItemAccessoryType = .none
 
-    /// The background styling of the `ListItem` to match the type of `List` it is displayed in.
-    var backgroundStyleType: ListItemBackgroundStyleType = .plain
+    /// The custom background styling of the `ListItem`, which is preferred over the `FLuentListStyle` environment value.
```
nodes11, line 252: minor: `FLuentListStyle` --> `FluentListStyle`

nodes11 (approved):

PR #2031, opened 2024-05-22
[TableViewCell] fix height calculation based on font size

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes

When calculating the height of a table view cell, we are incorrectly accounting for which subtitle font to use. This means we are calculating using a font that is bigger than what is expected, meaning our table view cell is rendered larger than needed. 

The fix is to ensure we only use the 3 line font if we have 3 lines, otherwise default back to the two line font, which is also the font for one line.

### Binary change

N/A

### Verification

| Before  rendered height: 66, calculated height: 64                                     | After    rendered height: 64, calculated height: 64                                        |
|image|image|

saw the height of each cell in the view debugger and compared the value. All of these scenarios it was still as expected

regular text size

|scenario | Before                                       | After                                      |
| --- |----------------------------------------------|--------------------------------------------|
| single line cell  48 height |Screenshot 2024-05-22 at 11 45 53 AM|Screenshot 2024-05-22 at 11 39 22 AM|
| double line cell  64 height| Screenshot 2024-05-22 at 11 46 01 AM|Screenshot 2024-05-22 at 11 39 31 AM|
| triple line cell 84 height| Screenshot 2024-05-22 at 11 46 19 AM | Screenshot 2024-05-22 at 11 40 02 AM |
| cell without custom view 64 height | Screenshot 2024-05-22 at 11 46 45 AM |Screenshot 2024-05-22 at 11 40 23 AM|
| cell with custom accessory view 48 height | Screenshot 2024-05-22 at 11 46 51 AM |Screenshot 2024-05-22 at 11 40 32 AM|
| cell with text truncation 84 height |Screenshot 2024-05-22 at 11 47 00 AM|Screenshot 2024-05-22 at 11 40 40 AM|
| cell with text wrapping  84 height|Screenshot 2024-05-22 at 11 47 08 AM|Screenshot 2024-05-22 at 11 41 11 AM|

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2031)

### DIFF
--- a/ios/FluentUI/Table View/TableViewCell.swift
+++ b/ios/FluentUI/Table View/TableViewCell.swift
@@ -351,7 +351,7 @@ open class TableViewCell: UITableViewCell, TokenizedControlInternal {
             subtitleHeight: labelSize(text: subtitle,
                                       attributedText: attributedSubtitle,
                                       isAttributedTextSet: isAttributedSubtitleSet,
-                                      font: subtitleFont ?? (layoutType == .twoLines ? tokenSet[.subtitleTwoLinesFont].uiFont : tokenSet[.subtitleThreeLinesFont].uiFont),
+                                      font: subtitleFont ?? (layoutType == .threeLines ? tokenSet[.subtitleThreeLinesFont].uiFont : tokenSet[.subtitleTwoLinesFont].uiFont),
                                       numberOfLines: subtitleNumberOfLines,
                                       textAreaWidth: textAreaWidth,
                                       leadingAccessoryView: subtitleLeadingAccessoryView,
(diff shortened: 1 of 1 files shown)


lyzhan7 (approved):

joannaquu:
should be one more case in line 1570

alexanderboswell:
> should be one more case in line 1570

Actually, that scenario is wrapped in 
```swift
if layoutType == .twoLines || layoutType == .threeLines {
}
```

so there is no scenario in which we could hit that line and it be the `.oneLine` scenario

PR #2037, opened 2024-05-29
Make our onChange backcompat extension public

lcapkovic:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

Anand recently made a local change to make `onChange_iOS17` public because we hit yet another new case where this was needed, and didn't want to create another duplicate implementation. Given the non-obvious dangers of an incorrect implementation (Apple's availability check bug), making this public seems to be the lesser evil.

### Verification

Build + run test app.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2037)


ios/FluentUI/Core/SwiftUI+ViewModifiers.swift
```diff
@@ -66,7 +66,7 @@ extension View {
     ///   - value: The value to check against when determining whether to run the closure.
     ///   - action: A closure to run when the value changes.
     /// - Returns: A view that fires an action when the specified value changes.
-    func onChange_iOS17<V>(of value: V, _ action: @escaping (V) -> Void) -> some View where V: Equatable {
+    public func onChange_iOS17<V>(of value: V, _ action: @escaping (V) -> Void) -> some View where V: Equatable {
```
mischreiber, line 69: Sorry, but this should not be made public. This is not a feature of Fluent, and we should not be polluting `View` downstream with this change, regardless of convenience to any one client.

mischreiber (changes requested):

mischreiber (approved):

PR #2038, opened 2024-05-29
Fix nav bar system title not hiding for some transparent backgrounds

lcapkovic:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

We've had repeated issues with the system navbar title showing up together with our fluent title.
This came up again recently on visionOS due to the recent UIColor -> Color token change, because the `.clear` comparison check fails now ((`Color.clear` converted to `UIColor`) != `UIColor.clear`).

I changed the check to not compare against `.clear` and instead check if the color alpha is less than 1.0. This should cover all the needed cases.

### Binary change

Change too tiny to matter?

### Verification

Build + testing in iOS and visionOS demo app, running through all `NavigationController` demo cases.
No diff on iOS, fixed doubled title on visionOS.

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2038)


mischreiber (approved):

harrieshin:
why wouldn't https://developer.apple.com/documentation/uikit/uibarappearance/3197999-configurewithtransparentbackgrou?language=objc work?

laminesm (commented):
These changes were made to address some strange bugs with the gradient nav bar iirc. I'd suggest validating that gradient nav bar works properly on portrait and landscape. Depending on how fast you want to merge this, it would be worth validating the client app that makes use of the gradient nav bar to make sure we aren't causing any regressions for them.

mischreiber:
> why wouldn't https://developer.apple.com/documentation/uikit/uibarappearance/3197999-configurewithtransparentbackgrou?language=objc work?

I'm guessing some of this code predates that API being available. Probably worth re-evaluating and seeing if that can get us what we want with less overhead.

lcapkovic:
> why wouldn't https://developer.apple.com/documentation/uikit/uibarappearance/3197999-configurewithtransparentbackgrou?language=objc work?

Nav bar transparency is not the issue, system title visibility is the issue. This is the only way to hide it AFAIK.

lcapkovic:
> These changes were made to address some strange bugs with the gradient nav bar iirc. I'd suggest validating that gradient nav bar works properly on portrait and landscape. Depending on how fast you want to merge this, it would be worth validating the client app that makes use of the gradient nav bar to make sure we aren't causing any regressions for them.

This doesn't take away the gradient bar fix, it instead brings the fix to all styles (including gradient bar). But I see your point.

lcapkovic:
We weren't able to prove that we *never* want to show a system title, so I decided to go with a less risky fix that only hides the system titles for non-opaque backgrounds.

joannaquu (approved):

ios/FluentUI/Navigation/NavigationBar.swift
```diff
@@ -639,11 +640,11 @@ open class NavigationBar: UINavigationBar, TokenizedControlInternal, TwoLineTitl
...
 
-        backgroundView.backgroundColor = style == .gradient ? .clear : color
-        standardAppearance.backgroundColor = color
+        backgroundView.backgroundColor = style == .gradient ? .clear : backgroundColor
+        standardAppearance.backgroundColor = backgroundColor
         tintColor = tokenSet[.buttonTintColor].uiColor
-        standardAppearance.titleTextAttributes[NSAttributedString.Key.foregroundColor] = shouldHideRegularTitle ? UIColor.clear : tokenSet[.titleColor].uiColor
-        standardAppearance.largeTitleTextAttributes[NSAttributedString.Key.foregroundColor] = tokenSet[.titleColor].uiColor
+        standardAppearance.titleTextAttributes[NSAttributedString.Key.foregroundColor] = systemTitleColor
+        standardAppearance.largeTitleTextAttributes[NSAttributedString.Key.foregroundColor] = systemTitleColor
```
joannaquu, line 647: oo nice! bug fix we never caught?

lcapkovic: ya this would come up if the nav bar was set to have large titles. I didn't find a case of it but technically a bug.

PR #1945, opened 2023-12-15
Revert #1929

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] macOS

### Description of changes

#1929 seems to be causing some very weird behavior with the `TableViewHeaderFooterView`. Headers that should be `headerPrimary` are showing up as `header` and vice versa. And some headers are changing on scroll. Reverting this change until it can be investigated.

### Verification

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| primary before | primary after |
| header before | header after |
| scroll before | scroll after |
</details>

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/1945)

PR #2040, opened 2024-05-31
Reapply #1929 and fix issue where header title uses incorrect font

nodes11:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes

**Issue:**
#1929 ensured that the `TableViewHeaderFooterView` does not override any font or color provided as part of the AttributedString. However, an issue was observed where headers that should be `headerPrimary` are showing up as `header` and vice versa. The change was reverted in #1945.

We were also seeing an issue where the title font would change after scroll where the title goes off screen.

This change reapplies #1929 and fixes the issue where headers that should be `headerPrimary` are showing up as `header` and vice versa, as well as the issue where the title font changes after it goes off screen.

**Root Cause:**
1. In #1929, if no attributed string was provided, the resolved `resolvedTitleFont` and `resolvedTitleColor` were only set once at inside `setup(...)`. This call occurs before the `tokenSet` is updated with overrides and the `resolvedTitleFont` would be set to the wrong font as a result. 
2. The setup method wasn't correctly "recycling" the views when `dequeueReusableHeaderFooterView` was called, so `resolvedTitleFont` and `resolvedTitleColor` properties were still set in some cases and the values would be used when rendering the header for a different header with a different font.

**Fix:**
- Change `resolvedTitleFont` and `resolvedTitleColor` to `attributedTitleFont` and `attributedTitleColor` as they'll only be used to hold values for the attributed string.
- Replace usages of `resolvedTitleFont` and `resolvedTitleColor` as an r-value with `attributedTitleFont ?? tokenSet[.textFont].uiFont` and `attributedTitleColor ?? tokenSet[.textColor].uiColor`.
- Cleanup the setup methods so they call `setupBase(...)` with their parameters. `setupBase` handles the cleanup and ensures all the properties are set in the same order for call cases.

### Binary change

Total increase: 19,968 bytes
Total decrease: -808 bytes
| Total | 30,874,424 bytes | 30,893,584 bytes | ⚠️ 19,160 bytes |
<details>

| TableViewHeaderFooterView.o | 277,496 bytes | 291,664 bytes | ⚠️ 14,168 bytes |
| __.SYMDEF | 4,751,744 bytes | 4,755,896 bytes | ⚠️ 4,152 bytes |
| FocusRingView.o | 832,280 bytes | 833,752 bytes | ⚠️ 1,472 bytes |
| PopupMenuSectionHeaderView.o | 24,504 bytes | 24,680 bytes | ⚠️ 176 bytes |
| PopupMenuController.o | 376,616 bytes | 375,808 bytes | 🎉 -808 bytes |
</details>

### Verification

- Validated that the header titles show in the correct font.
  - Verified that the font displayed matches the style listed in `TableViewHeaderFooterSampleData`
  - Verified that the font displayed matches the screenshots in the `After` category of this #1945

<details>

**Before**

https://github.com/microsoft/fluentui-apple/assets/10938746/1668c82a-9199-45fe-8703-698221ae8dea

**After**

https://github.com/microsoft/fluentui-apple/assets/10938746/9f301eb3-c114-4215-aea5-748510ce2e84

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2040)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -320,17 +320,12 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
                            accessoryButtonTitle: String = "",
                            accessoryView: UIView? = nil,
                            leadingView: UIView? = nil) {
-        attributedTitleFont = nil
-        attributedTitleColor = nil
-
-        if let attributedTitle = attributedTitle {
+        if let attributedTitle {
+            self.attributedTitle = attributedTitle
             titleView.attributedText = attributedTitle
             titleView.isSelectable = true
-
-            let attributes = attributedTitle.attributes(at: 0, effectiveRange: nil)
-            attributedTitleFont = attributes[NSAttributedString.Key.font] as? UIFont ?? tokenSet[.textFont].uiFont
-            attributedTitleColor = attributes[NSAttributedString.Key.foregroundColor] as? UIColor ?? tokenSet[.textColor].uiColor
         } else {
+            self.attributedTitle = nil
             titleView.attributedText = NSAttributedString(string: " ") // to clear attributes
             titleView.text = title
             titleView.isSelectable = false
@@ -458,8 +453,10 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
 
     private func updateTitleViewFont() {
         if let window = window {
-            let titleFont = attributedTitleFont ?? tokenSet[.textFont].uiFont
-            titleView.font = titleFont
+            let titleFont = tokenSet[.textFont].uiFont
+            if !isUsingAttributedTitle {
+                titleView.font = titleFont
+            }
 
             // offset text container to center its content
 #if os(iOS)
@@ -474,8 +471,6 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
     }
 
     private func updateTitleAndBackgroundColors() {
-        titleView.textColor = attributedTitleColor ?? tokenSet[.textColor].uiColor
-
         if tableViewCellStyle == .grouped {
             backgroundView?.backgroundColor = tokenSet[.backgroundColorGrouped].uiColor
         } else if tableViewCellStyle == .plain {
@@ -484,7 +479,10 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
             backgroundView?.backgroundColor = .clear
         }
 
-        titleView.font = attributedTitleFont ?? tokenSet[.textFont].uiFont
+        if !isUsingAttributedTitle {
+            titleView.textColor = tokenSet[.textColor].uiColor
+            titleView.font = tokenSet[.textFont].uiFont
+        }
         titleView.linkColor = tokenSet[.linkTextColor].uiColor
     }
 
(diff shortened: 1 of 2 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -296,30 +289,53 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
...
+    private func setupBase(style: Style,
+                           title: String? = nil,
+                           attributedTitle: NSAttributedString? = nil,
+                           accessoryButtonTitle: String = "",
+                           accessoryView: UIView? = nil,
+                           leadingView: UIView? = nil) {
+        attributedTitleFont = nil
+        attributedTitleColor = nil
+
+        if let attributedTitle = attributedTitle {
```
mischreiber, line 326: Nit: we've been using the more concise version of this going forward:

```swift
if let attributedTitle {
    // ...
}
```

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -329,6 +345,12 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
         }
 
         accessoryButton = !accessoryButtonTitle.isEmpty ? createAccessoryButton(withTitle: accessoryButtonTitle) : nil
+
+        /// `accessoryButton` and `accessoryView` occupy the same space at the trailing end of the view. If both are provided, the `accessoryView` will be given priority
+        if accessoryView != nil {
```
mischreiber, line 350: Why do we check before setting this? What happens if we set `nil` here?

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -296,30 +289,53 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
...
+        attributedTitleFont = nil
+        attributedTitleColor = nil
+
+        if let attributedTitle = attributedTitle {
+            titleView.attributedText = attributedTitle
+            titleView.isSelectable = true
+
+            let attributes = attributedTitle.attributes(at: 0, effectiveRange: nil)
+            attributedTitleFont = attributes[NSAttributedString.Key.font] as? UIFont ?? tokenSet[.textFont].uiFont
+            attributedTitleColor = attributes[NSAttributedString.Key.foregroundColor] as? UIColor ?? tokenSet[.textColor].uiColor
```
mischreiber, line 332: This will only work if the attributed string has only a single color and font associated. I expect we would want to support any arbitrary `NSAttributedString`, right?

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
nodes11: The `accessoryButton` above uses the `accessoryView` to display itself. The `accessoryButton` won't be shown if we set `nil` here after setting a non-nil `accessoryButton`. `accessoryButton` setter for reference: 

```
    private var accessoryButton: UIButton? {
        didSet {
            accessoryView = accessoryButton
            if accessoryButton != nil {
                updateAccessoryButtonTitleStyle()
            }
        }
    }
  ```

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
nodes11: Yeah, I think allowing that make sense. I took a similar approach to the `Label` component and its `isUsingCustomAttributedText` property. The one thing to note is that if an attributed string is provided, no additional token overrides can/will be applied. The string will solely rely on the attributes it was assigned when it was created.

mischreiber (approved):

PR #2047, opened 2024-06-17
Extend ListItem to support overrideTokens

joannaquu:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

Extending `ListItem` to support `.overrideTokens` so that consumers can customize tokens.
- Update `ListItem` so that every time the `body` runs, the `tokenSet` will be recreated with the current `tokenOverrides`
- Update `ListItemDemoController_SwiftUI` to use the `.overrideTokens` modifier
  - `.overrideTokens` requires that `listItem` be mutable so move this to a separate var.
- Remove the extension in `ListItemDemoController`
  - This was currently updating the fluent theme for the `TableViewCellTokenSet`, which reflected in the list item demo controller. Since this is not the "right" way to override list item tokens, let's remove it so consumers don't try this as well.
 - Add `leadingContentSize` as a var in `ListItem` so that when `leadingContentSize` changes, we update our tokens that rely on `customViewSize`.
   - Previously the `.leadingContentSize` modifier was creating a brand new token set. This meant that any tokens that were overridden would be lost. Now we can ensure that when `leadingContentSize`, only the tokens dependent on that update.

### Binary change

Total increase: 1,344 bytes
Total decrease: -16,168 bytes
| Total | 31,095,248 bytes | 31,080,424 bytes | 🎉 -14,824 bytes |
<details>

| FocusRingView.o | 838,936 bytes | 840,280 bytes | ⚠️ 1,344 bytes |
| __.SYMDEF | 4,808,816 bytes | 4,806,904 bytes | 🎉 -1,912 bytes |
| ListItemModifiers.o | 25,856 bytes | 20,544 bytes | 🎉 -5,312 bytes |
| ListItem.o | 357,984 bytes | 349,040 bytes | 🎉 -8,944 bytes |
</details>

### Verification

<details>

| iOS                                       | visionOS                                      |
| Simulator Screenshot - iPhone 15 Pro - 2024-06-17 at 11 48 53 | Simulator Screenshot - Apple Vision Pro - 2024-06-17 at 13 08 54 |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2047)


ios/FluentUI/List/ListItemModifiers.swift
```diff
@@ -63,8 +65,25 @@ public extension ListItem {
...
+            customViewTrailingMargin = GlobalTokens.spacing(.sizeNone)
+        case .small:
+            customViewDimensions = GlobalTokens.icon(.size240)
+            customViewTrailingMargin = GlobalTokens.spacing(.size160)
+        case .medium, .default:
+            customViewDimensions = GlobalTokens.icon(.size400)
+            customViewTrailingMargin = GlobalTokens.spacing(.size120)
+        }
+
+        listItem.tokenSet[.customViewDimensions] = .float { customViewDimensions }
```
joannaquu, line 84: this may cause a bug where if a new token is added to the tvc or list item token set that is dependent on the `customViewSize`, it's not updated here but I wasn't sure how else to ensure that other overridden tokens don't get mangled here.

nodes11: Could you describe a scenario where this might happen? Trying to understand it better. If this change is merged as-is with this potential bug, I think a comment should be added to call out the behavior.

joannaquu: so this would only happen if we added a new token to the `TableViewCellTokenSet` that depended on `customViewSize`. essentially what would happen is that token would never update even though `customViewSize` changed. the current tokens that depend on `customViewSize` work because I'm manually updating them here.

joannaquu: however, I talked to Mike and we're going to do the "proper" fix which is ensuring that whenever `customViewSize` updates, all the tokens that rely on it update. so fix coming soon

ios/FluentUI/List/ListItem.swift
```diff
@@ -15,7 +15,9 @@ public struct ListItem<LeadingContent: View,
                        TrailingContent: View,
                        Title: StringProtocol,
                        Subtitle: StringProtocol,
-                       Footer: StringProtocol>: View {
+                       Footer: StringProtocol>: View, TokenizedControlView {
+    public typealias TokenSetKeyType = ListItemTokenSet.Tokens
+    @ObservedObject public var tokenSet: ListItemTokenSet
```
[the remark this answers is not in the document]
joannaquu: these are just a part of the `TokenizedControl` protocol so not specific to `ListItem`

ios/FluentUI/List/ListItem.swift
```diff
@@ -39,8 +46,12 @@ public struct ListItem<LeadingContent: View,
         self.leadingContent = leadingContent
         self.trailingContent = trailingContent
         self.action = action
+
         let layoutType = ListItem.layoutType(subtitle: subtitle, footer: footer)
-        self.tokenSet = ListItemTokenSet(customViewSize: { layoutType.leadingContentSize })
+        let leadingContentSize = layoutType.leadingContentSize
+        let state = ListItemStateImpl(leadingContentSize: leadingContentSize)
+        self.state = state
```
mischreiber, line 53: This does match the existing pattern in other components, but it's one we'd like to move away from due to its complexity. Could we simply store `leadingContentSize` as a var on the `ListItem` itself and do away with `ListItemState` altogether?

mischreiber: You won't be able to do that either, because a SwiftUI `View` has to be a `struct`. What if you just don't make it `@Published`? A mutating var on the struct should force a re-render automatically, if I'm remembering my SwiftUI correctly.

joannaquu: ooh I didn't know that about views, turns out I don't know my swiftui very well hehe

joannaquu: but I believe that doesn't work because the compiler is complaining `Escaping closure captures mutating 'self' parameter`

joannaquu: spoke offline, going a different approach and following what `FluentButtonStyle` does instead

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
```diff
@@ -146,60 +148,66 @@ struct ListItemDemoView: View {
...
-            }
-            .disabled(isDisabled)
-            .alert("List Item tapped", isPresented: $showingPrimaryAlert) {
-                Button("OK", role: .cancel) { }
-            }
-            .alert("Detail button tapped", isPresented: $showingSecondaryAlert) {
-                Button("OK", role: .cancel) { }
-            }
+            overridenListItem()
+                .disabled(isDisabled)
```
joannaquu, line 204: these modifiers are kinda tricky. since they're not in `ListItemModifiers`, applying them on `listItem` before we apply `overrideTokens` changes `listItem` from type `ListItem` to type `some View` so we need to apply them after `overrideTokens`. however we can't directly apply them after `overrideTokens` because the type that returns has no member `.disabled`/`.alert` so I had to move it to a separate view builder. this really is not that clean so if I'm missing something obvious let me know.

ios/FluentUI/List/ListItem.swift
```diff
@@ -436,3 +433,10 @@ public extension ListItem where LeadingContent == EmptyView, TrailingContent ==
         }
     }
 }
+
+public extension ListItem {
+    /// Provide override values for various `ListItem` values.
+    mutating func overrideTokens(_ overrides: [ListItemToken: ControlTokenValue]) {
```
mischreiber, line 439: This function needs to return a type of `Self` and look like this:
```swift
mutating func overrideTokens(_ overrides: [ListItemToken: ControlTokenValue]) {
    tokenOverrides = overrides
    return self
}
```
That way, it can be chained with other modifiers (like your other comment mentions).

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
mischreiber: Mentioned the issue in your other comment: the `overrideTokens` function should return `Self`.

The modifying function on `FluentButton[Toggle]Style` is a little different because it's not meant to be chained--that might be something we investigate and change later as well.

ios/FluentUI/List/ListItem.swift
joannaquu: ah yes this! I knew I was missing something obvious. thanks Mike!

ios/FluentUI/List/ListItem.swift
```diff
@@ -436,3 +433,11 @@ public extension ListItem where LeadingContent == EmptyView, TrailingContent ==
         }
     }
 }
+
+public extension ListItem {
+    /// Provide override values for various `ListItem` values.
+    mutating func overrideTokens(_ overrides: [ListItemToken: ControlTokenValue]) -> some View {
```
mischreiber, line 439: Instead of returning `some View`, can you return either `Self` or `ListItem`? That'll retain the typing a little longer.

ios/FluentUI.Demo/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
```diff
@@ -146,72 +148,72 @@ struct ListItemDemoView: View {
...
-                showingSecondaryAlert = true
-            }
-            .disabled(isDisabled)
-            .alert("List Item tapped", isPresented: $showingPrimaryAlert) {
-                Button("OK", role: .cancel) { }
-            }
-            .alert("Detail button tapped", isPresented: $showingSecondaryAlert) {
-                Button("OK", role: .cancel) { }
-            }
+        func listItem() -> some View {
```
mischreiber, line 151: Should still be possible to tag this as a `@ViewBuilder` if you remove the `return` later in the function, I think?

mischreiber (approved):

ios/FluentUI/List/ListItem.swift
```diff
@@ -39,11 +40,11 @@ public struct ListItem<LeadingContent: View,
         self.leadingContent = leadingContent
         self.trailingContent = trailingContent
         self.action = action
-        let layoutType = ListItem.layoutType(subtitle: subtitle, footer: footer)
-        self.tokenSet = ListItemTokenSet(customViewSize: { layoutType.leadingContentSize })
     }
 
     public var body: some View {
+        let tokenSet = ListItemTokenSet(customViewSize: { leadingContentSize })
+        tokenSet.replaceAllOverrides(with: tokenOverrides)
```
anandrajeswaran, line 47: Out of curiosity, what's the perf impact observed or expected here? We can't guarantee when body runs so hopefully can guarantee this is dirt cheap?

mischreiber: We talked about this offline quite a bit, and yep, token set creation is dirt cheap. We'll work in the future to make it even cheaper, but performance should be totally acceptable right now.

PR #2048, opened 2024-06-20
Add ListItem to podspec

lyzhan7:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Continuation of https://github.com/microsoft/fluentui-apple/pull/1939
- There was a suggestion in that PR to create a separate subspec for both TVC/ListItem to depend on rather than having ListItem depend on all of TVC - followed up on that, and also thought it would be cleaner for the shared tokens/enums that both TVC/ListItem use to be in its own folder, but happy to hear thoughts/go back to only the podspec changes
- ListItem depended on TVC's tokens, as well as several enums. There were a couple enums used that weren't defined in the TableViewCellTokenSet file -  I saw there were a couple other enums defined in TableViewCellTokenSet, so I just moved all the shared ones into that file - but maybe would make more sense to put all the enums in its own file

### Binary change
Not adding/removing any source code, so I don't think this PR should impact binary size - but if there's a reason to run this let me know

### Verification

`pod lib lint` passes. Unfortunately each run of this seemed like it took about 20 minutes. If there's a faster way to validate podspec changes in fluentui-apple let me know!

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2048)


MicrosoftFluentUI.podspec
```diff
@@ -183,6 +183,14 @@ fi', :execution_position => :before_compile }
     label_ios.source_files = ["ios/FluentUI/Label/**/*.{swift,h}"]
   end
 
+  s.subspec 'ListItem_ios' do |listitem_ios|
+    listitem_ios.platform = :ios
+    listitem_ios.dependency 'MicrosoftFluentUI/OtherCells_ios'
+    listitem_ios.dependency 'MicrosoftFluentUI/Separator_ios'
+    listitem_ios.dependency 'MicrosoftFluentUI/TableViewListShared_ios'
```
alexanderboswell, line 190: I see the benefit of `TableViewListShared` in the project structure, but would it be simpler just to have a file path directly to the tokenset? We are establishing a pattern, so just have to be aware if this would make sense for other controls that have a swiftui implementation and a UIKit implementation>

The trade off of the one time work around versus the setting up a new pattern with only the insight into how one control needs to share things between swiftui vs UIKit.

mischreiber (approved):
Looks good overall! I'm debating about the name `TableViewListShared`, but I also don't have a better idea.

alexanderboswell (approved):

ios/FluentUI.xcodeproj/project.pbxproj
```diff
@@ -957,6 +957,7 @@
 		A5CEC15320D980B20016922A /* FluentUI */ = {
 			isa = PBXGroup;
 			children = (
+				F5854ED02C220AC100938291 /* TableViewListShared */,
```
joannaquu, line 960: nit: this unalphabetizes the folders in Xcode right?

MicrosoftFluentUI.podspec
lyzhan7: Hmm so to clarify - are you suggesting instead of having a separate subspec for it and adding a dependency to it, to instead update listitem_ios.source_files to include a path to the tokens?

lyzhan7: Talked offline - will keep this pattern as is, I think its also more extensible if other pairs of swiftui/UIKit implementation want to share more than just the tokens

ios/FluentUI.xcodeproj/project.pbxproj
lyzhan7: Yes you're right thanks for catching that! Going to merge as is since the PR checks have passed but will follow up in a separate PR

MicrosoftFluentUI.podspec
```diff
@@ -183,6 +183,14 @@ fi', :execution_position => :before_compile }
     label_ios.source_files = ["ios/FluentUI/Label/**/*.{swift,h}"]
   end
 
+  s.subspec 'ListItem_ios' do |listitem_ios|
+    listitem_ios.platform = :ios
```
Saadnajmi, line 187: Should this include visionOS?

ios/FluentUI.xcodeproj/project.pbxproj
lyzhan7: Never mind will fix, need to rerun the checks anyway

MicrosoftFluentUI.podspec
lyzhan7: Hmm I'm not sure actually, we don't specify visionos in any of the other subspecs, I assume some of the other pods are being used in vision and they didn't need to specify visionos. @joannaquu @mischreiber do either of you know whether we need to specify it?

lyzhan7: Will leave as is for now, and follow up if needed

Saadnajmi: Sounds like we may not be allowing cocoapods users of FUA to use visionOS at all, which is outside the scope of this PR. Fair for followup :)

MicrosoftFluentUI.podspec
mischreiber: I think this is interesting because in most cases, our SwiftUI and UIKit components are intentionally packaged together. But List and TableViewCell are odd enough that I do think it makes sense to fork them like this.

MicrosoftFluentUI.podspec
mischreiber: That's correct, we haven't dealt with visionOS and Cocoapods yet.

PR #2041, opened 2024-05-31
Calculate Bottom Commanding header height instead of using constant

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Currently, `BottomCommandingController`'s `headerContentHeight` is hard-coded in as a constant of 66. When the hero item's titles are one line, this is ok. However when it is multi-lined, there is not enough bottom padding. Update it so that `headerContentHeight` is calculated using `systemLayoutSizeFitting`.

### Binary change

Total increase: 6,072 bytes
Total decrease: 0 bytes
| File | Before | After | Delta |
|------|-------:|------:|------:|
| Total | 31,087,872 bytes | 31,093,944 bytes | ⚠️ 6,072 bytes |
<details>
<summary> Full breakdown </summary>

| File | Before | After | Delta |
|------|-------:|------:|------:|
| BottomCommandingController.o | 837,944 bytes | 840,832 bytes | ⚠️ 2,888 bytes |
| __.SYMDEF | 4,813,352 bytes | 4,815,168 bytes | ⚠️ 1,816 bytes |
| FocusRingView.o | 837,864 bytes | 839,232 bytes | ⚠️ 1,368 bytes |
</details>

### Verification

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| before | after |
</details>

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2041)

PR #2072, opened 2024-07-17
Fix bottom commanding bug where dimming view is still accessible when collapsed 

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

We've gotten a bug from clients that even when collapsed, `BottomCommandingController`'s dimming view (which dismisses the sheet when tapped) is still accessible. 

There are some dependent changes in this pr (otherwise I would have opened separate prs) so I've broken it down by commit
- #2041 attempted to fix an issue where two line hero items in `BottomCommandingController` did not have enough padding. This ended up subtly hiding a pre-exisiting bug (the one this pr is solving). Revert this change.
- Actually solve the padding issue.
  - Given our fixed `headerHeight` of 66, the padding between the resizing handle and the hero items should be small enough to allow for two-line titles. This was the original design and somewhere along the shuffle with tokenization, this got lost. See screenshots for the original design.
  - Update the padding to `.size60`. Revert the name back to `headerTopMargin` since it gets confusing with `resizingHandleHeight`
- Fix the accessible dimming view bug. This requires a long explanation:
  - In `BottomCommandingController`, we calculate `collapsedContentHeight` by starting from the height of the `heroCommandStack` and working our way up to the header height. In `BottomSheetController`. we then use `collapsedSheetHeight` to calculate the offset, which we use to calculate the alpha of the dimming view, which we use to determine whether or not the dimming view should be accessible.
  - When `collapsedContentHeight` is a whole number, this is fine. However when it's a decimal, specifically one that can not be precisely represented in binary, we run into issues.
  - Take for example the one line case in bottom commanding.
    - In `updateDimmingViewAlpha`, `currentOffset = 750.6666666666665` and `lowestUndimmedOffset = 750.6666666666666`
    - The `0.0000000000001` difference causes `currentOffset >= lowestUndimmedOffset` to fail
    - `dimmingView.alpha` is set to be a number very close to 0
    - `dimmingView.alpha == 0` fails and our dimming view is left as accessible when it should not
  - The fix?
    - In `updateDimmingViewAlpha`, when `targetAlpha` is expected to be 0 or 1 (all states except `.transitioning`), set it explicitly if that is not the case.
    - When `handleCompletedStateChange` is hit, it was previously assumed that `dimmingView.alpha` would already be 0 or 1. Call `updateDimmingViewAlpha/updateDimmingViewAccessibility` to ensure that is the case.

### Binary change

Total increase: 0 bytes
Total decrease: -1,848 bytes
| Total | 31,087,520 bytes | 31,085,672 bytes | 🎉 -1,848 bytes |
<details>

| BottomCommandingTokenSet.o | 71,944 bytes | 71,928 bytes | 🎉 -16 bytes |
| BottomSheetController.o | 523,880 bytes | 523,848 bytes | 🎉 -32 bytes |
| __.SYMDEF | 4,805,640 bytes | 4,805,360 bytes | 🎉 -280 bytes |
| BottomCommandingController.o | 837,912 bytes | 836,392 bytes | 🎉 -1,520 bytes |
</details>

### Verification

<details>

| before | after |
| two line before | two line after |
| collapsed before | collapsed after |
| expanded before | expanded after |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2072)

### DIFF
--- a/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
+++ b/ios/FluentUI/Bottom Sheet/BottomSheetController.swift
@@ -740,6 +740,15 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
                 targetAlpha = abs(currentOffset - lowestUndimmedOffset) / (lowestUndimmedOffset - highestDimmedOffset)
             }
         }
+
+        if currentExpansionState != .transitioning {
+            // In the case that there has been a floating point precision error and
+            // targetAlpha is a value very close to 0 or 1, set it explicitly
+            if targetAlpha != 0 || targetAlpha != 1 {
+                targetAlpha = currentExpansionState == .expanded ? 1 : 0
+            }
+        }
+
         dimmingView.alpha = targetAlpha
     }
 
@@ -1050,6 +1059,9 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         }
 
         bottomSheetView.isHidden = targetExpansionState == .hidden
+
+        updateDimmingViewAlpha()
+        updateDimmingViewAccessibility()
 
         // UIKit doesn't properly handle interrupted constraint animations, so we need to
         // detect and fix a possible desync here
--- a/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
+++ b/ios/FluentUI.Demo/FluentUI.Demo/Demos/BottomCommandingDemoController.swift
@@ -46,8 +46,7 @@ class BottomCommandingDemoController: DemoController {
 
     private lazy var heroItems: [CommandingItem] = {
         return Array(1...25).map {
-            let title = ($0 == 4) ? "Two line item" : "Item"
-            let item = CommandingItem(title: title + String($0), image: homeImage, action: commandAction)
+            let item = CommandingItem(title: "Item " + String($0), image: homeImage, action: commandAction)
             item.selectedImage = homeSelectedImage
             item.isOn = ($0 % 3 == 1)
             item.isEnabled = ($0 % 2 == 1)
--- a/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
+++ b/ios/FluentUI/Bottom Commanding/BottomCommandingController.swift
@@ -920,7 +920,7 @@ open class BottomCommandingController: UIViewController, TokenizedControlInterna
         bottomSheetController.isExpandable = isExpandable
 
         let maxHeroItemHeight = heroCommandStack.arrangedSubviews.map { $0.intrinsicContentSize.height }.max() ?? Constants.defaultHeroButtonHeight
-        let headerHeightWithoutBottomWhitespace = BottomCommandingTokenSet.handleHeaderHeight + maxHeroItemHeight
+        let headerHeightWithoutBottomWhitespace = BottomCommandingTokenSet.headerTopMargin + maxHeroItemHeight
 
         // How much more whitespace is required at the bottom of the sheet header
         let requiredBottomWhitespace = max(0, Constants.BottomSheet.headerHeight - headerHeightWithoutBottomWhitespace)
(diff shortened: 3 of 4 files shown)


anandrajeswaran:
CGFloat should automatically translate to Double under the hood, but it's possible somewhere we're initializing one with a Float implicitly or explicitly and that's why it's getting different floating point precision issues. Is it possible this problem goes away if we be explicit about using Double and just make everything in this particular file a Double?

If not that, hopefully we still can get to a more durable solution - ostensibly some point when an animation/interaction completes, we should set the alpha explicitly to 0 or 1?

cbowdoin:
"Turns out values smaller than 0.99999998 evaluate to 1, which is large enough of an epsilon that we won't run into the same issue."

Is this suppose be larger than `0.99999998` or does a value of `0.0000001` also evaluate to 1?

joannaquu:
> "Turns out values smaller than 0.99999998 evaluate to 1, which is large enough of an epsilon that we won't run into the same issue."
> 
> Is this suppose be larger than `0.99999998` or does a value of `0.0000001` also evaluate to 1?

oh haha you're right. updated!

joannaquu:
> CGFloat should automatically translate to Double under the hood, but it's possible somewhere we're initializing one with a Float implicitly or explicitly and that's why it's getting different floating point precision issues. Is it possible this problem goes away if we be explicit about using Double and just make everything in this particular file a Double?
> 
> If not that, hopefully we still can get to a more durable solution - ostensibly some point when an animation/interaction completes, we should set the alpha explicitly to 0 or 1?

Changing everything to Double doesn't change anything because we're still fetching `currentOffset` and `lowestUndimmedOffset` from two separate paths. One comes from our math and one comes from the bottom sheet's frame

mischreiber:
Would it be easiest to simply check if the value is `< 0.001` or `> 0.999`, to account for minor floating point error? Or is there a risk that that would not bring us close enough?

ios/FluentUI/Bottom Sheet/BottomSheetController.swift
```diff
@@ -743,15 +743,15 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
             return
         }
 
-        if dimmingView.alpha == 0 {
-            dimmingView.isAccessibilityElement = false
-            view.accessibilityViewIsModal = false
-        } else {
+        if dimmingView.alpha == 1 {
```
anandrajeswaran, line 746: With the change to make alpha explicitly 1 or 0, is this part needed? In fact, 0 may technically be the more correct value since it appears we're checking alpha as an indirect proxy for whether the view is hidden. From that lens, hidden=false usually means alpha between 0.00001 and 1 and hidden=true usually means alpha == 0.

Natural follow-up question - should we be explicitly hiding this view when it's at 0 alpha and can we then key off the `hidden` property instead?

lcapkovic: I think I agree with this. Lots of potential for simplification here I think, because if we actually set `hidden` to the correct value on completed state change, we could get away with keeping `isAccessibilityElement` set to true, right? A hidden view should not show up in a11y hierarchy automatically.

I'd also recommend revisiting the need for `accessibilityFrame` updates - it seems a little sus we need to do this. Overlapping a11y elements should naturally appear across the system, so I don't understand why we need to handle that case here specifically.

Btw I realize this is scope creep but I see potential for tons of code deletion here :D

joannaquu: this is so so so close to working with one caveat: we lose the alpha animation when tapping on the resizing handle and it snaps from hidden to not hidden. since changing from 0 alpha to hidden is not a 1 to 1 swap, i'm leaving as is for now to get this fix to customers before the end of the month

joannaquu:
> Would it be easiest to simply check if the value is `< 0.001` or `> 0.999`, to account for minor floating point error? Or is there a risk that that would not bring us close enough?

yeah my fear with this would be that there is a chance it's not close enough

nodes11 (approved):

anandrajeswaran (approved):

PR #2076, opened 2024-07-24
Ensure TableViewHeaderFooterView has a themed default text color and font style

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Issue: Attributed strings in the `TableViewHeaderFooterView` were showing as black if a foreground color attribute was not provided.

Root Cause: The attributed text color will default to the `textColor` of the `UITextView`. [According to the Apple documentation, the default text color for a `UITextView` is black](https://developer.apple.com/documentation/uikit/uitextview/1618601-textcolor), explaining why we were seeing the text show as black in both light and dark modes

Fix: Check if an attributed foreground color is provided with string. If not, apply the fluent theme text color. Also do the same thing for font.

### Binary change

Total increase: 19,560 bytes
Total decrease: 0 bytes
| Total | 30,900,560 bytes | 30,920,120 bytes | ⚠️ 19,560 bytes |
<details>

| TableViewHeaderFooterView.o | 282,296 bytes | 297,712 bytes | ⚠️ 15,416 bytes |
| __.SYMDEF | 4,744,576 bytes | 4,746,832 bytes | ⚠️ 2,256 bytes |
| FocusRingView.o | 834,848 bytes | 836,576 bytes | ⚠️ 1,728 bytes |
| PopupMenuSectionHeaderView.o | 24,496 bytes | 24,656 bytes | ⚠️ 160 bytes |
</details>

### Verification

- In the demo controller, I validated that an attributed string with both foreground color and font attributes uses the attributes.
- In the demo controller, I validated that an attributed string without a foreground color and font attributes uses the fluent theme colors.
- Validated that the change fixes our internal use case.

<details>

Before change, without attributes applied to string:
Screenshot 2024-07-24 at 11 04 00 AM
Screenshot 2024-07-24 at 11 03 54 AM

After change, without attributes applied to string:
Screenshot 2024-07-24 at 11 05 38 AM
Screenshot 2024-07-24 at 11 05 42 AM

After change, with attributes applied to string:
Screenshot 2024-07-24 at 11 01 04 AM
Screenshot 2024-07-24 at 11 00 57 AM

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2076)

### DIFF
--- a/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
+++ b/ios/FluentUI/Table View/TableViewHeaderFooterView.swift
@@ -454,7 +454,7 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
     private func updateTitleViewFont() {
         if let window = window {
             let titleFont = tokenSet[.textFont].uiFont
-            if !isUsingAttributedTitle {
+            if !hasOverriddenAttributedFont {
                 titleView.font = titleFont
             }
 
@@ -479,8 +479,10 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
             backgroundView?.backgroundColor = .clear
         }
 
-        if !isUsingAttributedTitle {
+        if !hasOverriddenAttributedForegroundColor {
             titleView.textColor = tokenSet[.textColor].uiColor
+        }
+        if !hasOverriddenAttributedFont {
             titleView.font = tokenSet[.textFont].uiFont
         }
         titleView.linkColor = tokenSet[.linkTextColor].uiColor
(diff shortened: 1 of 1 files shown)


ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -522,13 +524,25 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
         onHeaderViewTapped?()
     }
 
-    private var attributedTitle: NSAttributedString? {
-        didSet {
-            isUsingAttributedTitle = attributedTitle != nil
+    private var attributedTitle: NSAttributedString?
+
+    private var hasOverriddenAttributedForegroundColor: Bool {
+        if let attributes = attributedTitle?.attributes(at: 0, effectiveRange: nil) {
```
joannaquu, line 530: this is a niche case but in the case that clients only override one letter's foreground color, we preserve that they are responsible for overriding all the other letters as well right?

nodes11: Yep! A screenshot to verify. I applied a green foreground color attribute to just the substring "Footer - A description". The rest is left in the default UITextView text color.

Screenshot 2024-07-24 at 1 01 14 PM

joannaquu (approved):
can you run the binary size change

ios/FluentUI/Table View/TableViewHeaderFooterView.swift
```diff
@@ -479,8 +479,10 @@ open class TableViewHeaderFooterView: UITableViewHeaderFooterView, TokenizedCont
             backgroundView?.backgroundColor = .clear
         }
 
-        if !isUsingAttributedTitle {
+        if !hasOverriddenAttributedForegroundColor {
             titleView.textColor = tokenSet[.textColor].uiColor
```
mischreiber, line 483: I'm not sure about this logic still--following up from @joannaquu's earlier comment, if someone applies a color to a substring, the rest will be stuck with default black color, right? Seems like the default should always be to use our Fluent color for any spot that doesn't have custom color applied.

nodes11: I'm not opposed to supporting that. Just unsure how cleanly and efficiently we can support that. I think the operation would be o(n) worst case scenario. Generally speaking, I think we could:

1. Enumerate over attributes on string to create mappings of ranges with foregroundColor attributes
2. Sort the items in the map by range
3. Starting the 0 index, find all ranges that don't have a foregroundColor attribute

nodes11: I think I found a better & easier way to do this.

PR #2080, opened 2024-07-31
Add wide accessory view to ShyHeaderView

huanluu:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Based on my understanding, the current shy header behavior has 2 limitations:
(1) The accessory view is expected to be a SearchBar
(2) The accessory view has an inset on 4 edges so that the accessory view will always looks like it's part of the shy header. This is very obvious when the shy header has a background color.
Now the issue is sometimes we want to make other views to hide and appear when users scroll as well, but we don't want them to have the header background color. 

So to solve this problem, in `ShyHeaderView` I added a `wideAccessoryView`, which sits at the bottom of the `ShyHeaderView`. It can be set with or without an accessory view. As a result, on iPhone where there is an `accessoryView`(`SearchBar`), both the `SearchBar` and the `wideAccessoryView` will shy; and on iPad where there is a `topAccessoryView`(`SearchBar`) then only `wideAccessoryView` will shy. So as to matching the design.

Also I make the `wideAccessoryView` such that it doesn't have leading, trailing, and bottom inset. As a result, the client code can set the background color of it to match the content view controller so that it doesn't look like a part of header(otherwise the header could look too massive).

### Binary change

Small code change, no assets, no SwiftUI added. So I don't think there will be a binary change concern.

### Verification

- Added two buttons in `NavigationControllerDemoController`:
"Show with collapsible search bar and pill segmented control", 
https://github.com/user-attachments/assets/832e4714-c820-474c-b274-23447e809797
Note that now both the search bar and the pill segment control will shy when user scrolls. And when user taps on the search bar, the pill segment control will contract as well.

"Show with top search bar for large screen width with a shy pill segment control"
https://github.com/user-attachments/assets/148ad5f0-a2a5-43b3-a696-57323149c00f
Note that the search bar doesn't shy because it's a `topAccessoryView` not `accessoryView`, and only the pill segment control will shy. And when user taps on the search bar, the pill segment control doesn't appear.

- Also verified that switching between iPad regular and compact mode works well, i.e. handling adding and removing `accessoryView` is working as expected:
https://github.com/user-attachments/assets/202d4cee-36da-40f6-bfed-ac1f5ddf3c2a

### Pull request checklist

This PR has considered:

### DIFF
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -46,7 +46,6 @@ class ShyHeaderController: UIViewController {
 
     private var accessoryViewObservation: NSKeyValueObservation?
     private var wideAccessoryViewObservation: NSKeyValueObservation?
-    private var wideAccessoryViewHeightObservation: NSKeyValueObservation?
 
     private var navigationBarCenterObservation: NSKeyValueObservation?
     private var navigationBarStyleObservation: NSKeyValueObservation?
@@ -106,10 +105,6 @@ class ShyHeaderController: UIViewController {
 
         wideAccessoryViewObservation = contentViewController.navigationItem.observe(\UINavigationItem.wideAccessoryView) { [weak self] item, _ in
             self?.shyHeaderView.wideAccessoryView = item.wideAccessoryView
-        }
-
-        wideAccessoryViewHeightObservation = contentViewController.navigationItem.observe(\UINavigationItem.wideAccessoryViewHeight) { [weak self] item, _ in
-            self?.shyHeaderView.wideAccessoryViewHeight = item.wideAccessoryViewHeight
         }
     }
 
@@ -221,10 +216,10 @@ class ShyHeaderController: UIViewController {
     }
 
     private func setupShyHeaderView() {
-        shyHeaderView.accessoryView = contentViewController.navigationItem.accessoryView
-        shyHeaderView.wideAccessoryView = contentViewController.navigationItem.wideAccessoryView
-        shyHeaderView.wideAccessoryViewHeight = contentViewController.navigationItem.wideAccessoryViewHeight
-        shyHeaderView.navigationBarShadow = contentViewController.navigationItem.navigationBarShadow
+        let navigationItem = contentViewController.navigationItem
+        shyHeaderView.accessoryView = navigationItem.accessoryView
+        shyHeaderView.wideAccessoryView = navigationItem.wideAccessoryView
+        shyHeaderView.navigationBarShadow = navigationItem.navigationBarShadow
         shyHeaderView.paddingView = paddingView
         shyHeaderView.parentController = self
         shyHeaderView.maxHeightChanged = { [weak self] in
--- a/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
+++ b/ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
@@ -45,6 +45,8 @@ class ShyHeaderController: UIViewController {
     }
 
     private var accessoryViewObservation: NSKeyValueObservation?
+    private var wideAccessoryViewObservation: NSKeyValueObservation?
+    private var wideAccessoryViewHeightObservation: NSKeyValueObservation?
 
     private var navigationBarCenterObservation: NSKeyValueObservation?
     private var navigationBarStyleObservation: NSKeyValueObservation?
@@ -100,6 +102,14 @@ class ShyHeaderController: UIViewController {
 
         accessoryViewObservation = contentViewController.navigationItem.observe(\UINavigationItem.accessoryView) { [weak self] item, _ in
             self?.shyHeaderView.accessoryView = item.accessoryView
+        }
+
+        wideAccessoryViewObservation = contentViewController.navigationItem.observe(\UINavigationItem.wideAccessoryView) { [weak self] item, _ in
+            self?.shyHeaderView.wideAccessoryView = item.wideAccessoryView
+        }
+
+        wideAccessoryViewHeightObservation = contentViewController.navigationItem.observe(\UINavigationItem.wideAccessoryViewHeight) { [weak self] item, _ in
+            self?.shyHeaderView.wideAccessoryViewHeight = item.wideAccessoryViewHeight
         }
     }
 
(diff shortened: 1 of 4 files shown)


ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
```diff
@@ -31,6 +33,28 @@ import UIKit
...
+        get {
+            return objc_getAssociatedObject(self, &AssociatedKeys.wideAccessoryView) as? UIView
+        }
+        set {
+            objc_setAssociatedObject(self, &AssociatedKeys.wideAccessoryView, newValue, .OBJC_ASSOCIATION_RETAIN_NONATOMIC)
+        }
+    }
+
+    /// The height of the wideAccessoryView.
+    var wideAccessoryViewHeight: CGFloat {
```
laminesm, line 49: Can't we just use the height of the view instead of asking for it to be explicitly provided in the navitem?

ios/FluentUI/Navigation/Shy Header/ShyHeaderView.swift
```diff
@@ -222,12 +253,43 @@ class ShyHeaderView: UIView, TokenizedControlInternal {
 
     private func initContentStackView() {
         contentStackView.isLayoutMarginsRelativeArrangement = true
+        contentStackView.translatesAutoresizingMaskIntoConstraints = false
         addSubview(contentStackView)
-        contentStackView.fitIntoSuperview(usingConstraints: true)
+
+        if let wideContentStackViewTopAnchorConstraint {
```
laminesm, line 259: nit: Might be worth adding a comment here explaining why we are deactivating this constraints we the shy header's content is being initialized.

ios/FluentUI/Navigation/Shy Header/ShyHeaderView.swift
```diff
@@ -135,13 +141,35 @@ class ShyHeaderView: UIView, TokenizedControlInternal {
...
     }
 
-    var maxHeight: CGFloat {
+    var wideAccessoryView: UIView? {
+        willSet {
+            wideAccessoryView?.removeFromSuperview()
+        }
+        didSet {
+            if let newContentView = wideAccessoryView {
+                wideContentStackView.addArrangedSubview(newContentView)
```
laminesm, line 150: I might be misunderstanding but does this mean every time we set `wideAccessoryView`, it just gets added to the content stack view?

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -212,6 +222,8 @@ class ShyHeaderController: UIViewController {
 
     private func setupShyHeaderView() {
```
laminesm, line 223: nit: maybe we should just cache the nav item and reuse it `let navigationItem = contentViewController.navigationItem`

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
huanluu: My understanding is that we can't. Because not all subviews of `UIView` have `intrinsicContentSize`, so there is no reliable way to get the size from the provided view. And before this change the `maxHeight` is 36.0(the height of the SearchBar) + top and bottom insets, and I guess the reason is also because of that.

huanluu: Curious what other people think. I would love to know if we can do that.

ios/FluentUI/Navigation/Shy Header/ShyHeaderView.swift
huanluu: Yes, every time we set `wideAccessoryView `, the old one will first get removed from `wideContentStackView`, and then the new one gets added to it, similar to how a`ccessoryView` is handled.

ios/FluentUI/Navigation/Shy Header/ShyHeaderView.swift
huanluu: Maybe the name `contentStackView` is a little confusing? This stack view only contains the accessoryView, and the new `wideAccessoryView` is hosted in another stack view `wideContentStackView`. When we don't have an `accessoryView` we set the top anchor constraint of the `wideContentStackView` to be equal to the current view's top anchor. But when `accessoryView` is added here(whenever a `accessoryView` is added we call `initContentStackView()`) we need to deactivate that constraints and instead constraint the top anchor of the `wideContentStackView` to be equal to the bottom anchor of the `contentStackView`. Does that make sense?

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
joannaquu: the height of searchbar is fixed so I think it just worked out conveniently that we could calculate `maxHeight` based off the height of the search bar and insets. however, I'm also curious if we can just get the height from the view. does using `systemLayoutSizeFitting` work?

huanluu: `systemLayoutSizeFitting ` does work! I didn't know we can do that! Thank you so much to you both!

ios/FluentUI/Navigation/Shy Header/ShyHeaderView.swift
huanluu: Added a comment.

ios/FluentUI/Navigation/Shy Header/ShyHeaderController.swift
```diff
@@ -45,6 +45,7 @@ class ShyHeaderController: UIViewController {
     }
 
     private var accessoryViewObservation: NSKeyValueObservation?
+    private var wideAccessoryViewObservation: NSKeyValueObservation?
```
mischreiber, line 48: I think I would prefer a name that better describes that the view will live under the main accessory view. Maybe `secondaryAccessoryView`?

ios/FluentUI/Navigation/UINavigationItem+Navigation.swift
```diff
@@ -20,6 +20,7 @@ import UIKit
         static var customNavigationBarColor: UInt8 = 0
         static var customSubtitleTrailingImage: UInt8 = 0
         static var isTitleImageLeadingForTitleAndSubtitle: UInt8 = 0
+        static var wideAccessoryView: UInt8 = 0
```
mischreiber, line 23: Nit: could you put this up next to `accessoryView`?

mischreiber (approved):

PR #2081, opened 2024-08-02
New Avatar API: `hasBackgroundOutline`

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

Adding a new API for `Avatar` and `AvatarGroup` that optionally renders an outline around the `Avatar` instances with the Fluent background color.

This is the same color (and the same logic) that is used when drawing a ring around the `Avatar`. Useful for when the control is being rendered atop a dark or complicated background, such as a photo.

As an addendum, this change also brings formal deprecation to the `isTransparent` property of `Avatar`. This is from a very old version of the Fluent spec, and should not be used anymore. The default value has been changed to `false`, and will be fully removed in a future version of Fluent.

### Binary change

Total increase: 31,000 bytes
Total decrease: 0 bytes
| Total | 31,248,472 bytes | 31,279,472 bytes | ⚠️ 31,000 bytes |
<details>

| AvatarGroup.o | 383,488 bytes | 399,688 bytes | ⚠️ 16,200 bytes |
| Avatar.o | 601,344 bytes | 609,088 bytes | ⚠️ 7,744 bytes |
| __.SYMDEF | 4,860,560 bytes | 4,865,256 bytes | ⚠️ 4,696 bytes |
| AvatarModifiers.o | 45,952 bytes | 47,760 bytes | ⚠️ 1,808 bytes |
| FocusRingView.o | 846,992 bytes | 847,520 bytes | ⚠️ 528 bytes |
| MSFAvatarGroup.o | 30,352 bytes | 30,376 bytes | ⚠️ 24 bytes |
</details>

### Verification

Verified enabling and disabling the value in both UIKit and SwiftUI.

<details>

| UIKit | SwiftUI |
| UIKit_Dark | SwiftUI_Dark |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2081)

### DIFF
--- a/ios/FluentUI/Avatar/Avatar.swift
+++ b/ios/FluentUI/Avatar/Avatar.swift
@@ -49,7 +49,12 @@ import SwiftUI
 
     /// Sets the transparency of the avatar elements (inner and outer ring gaps, presence, and activity icon outline).
     /// Uses the solid default background color if set to false.
+    @available(iOS, deprecated, message: "Transparency is deprecated, and will default to `false`. This property will be removed in a future version of Fluent.")
     var isTransparent: Bool { get set }
+
+    /// Whether a background-colored outline should be drawn behind the avatar.
+    /// Uses the solid default background color if set to true.
+    var hasBackgroundOutline: Bool { get set }
 
     /// Defines the presence displayed by the Avatar.
     /// Image displayed depends on the value of the isOutOfOffice property.
@@ -134,6 +139,7 @@ public struct Avatar: View, TokenizedControlView, Equatable {
         let cornerRadius = shouldDisplayActivity ? (activityStyle == .square ? AvatarTokenSet.activityIconRadius(size) : .infinity) : .infinity
         let isRingVisible = state.isRingVisible
         let hasRingInnerGap = state.hasRingInnerGap
+        let hasBackgroundOutline = state.hasBackgroundOutline
         let ringThicknessToken: CGFloat = tokenSet[.ringThickness].float
         let accessoryBorderThicknessToken: CGFloat = tokenSet[.borderThickness].float
         let accessoryBorderColorToken: UIColor = tokenSet[.borderColor].uiColor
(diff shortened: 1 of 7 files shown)


ios/FluentUI/Avatar/Avatar.swift
```diff
@@ -143,7 +149,7 @@ public struct Avatar: View, TokenizedControlView, Equatable {
         // Adding ringInnerGapOffset to ringInnerGap & ringThickness to accommodate for a small space between
         // the ring and avatar when the ring is visible and there is no inner ring gap
         let ringInnerGapOffset = 0.5
-        let ringInnerGap: CGFloat = isRingVisible ? (hasRingInnerGap ? tokenSet[.ringInnerGap].float : -ringInnerGapOffset) : 0
+        let ringInnerGap: CGFloat = (isRingVisible || hasBackgroundOutline) ? ((hasRingInnerGap || hasBackgroundOutline) ? tokenSet[.ringInnerGap].float : -ringInnerGapOffset) : 0
```
joannaquu, line 152: nit: is there a way to make this easier to read? maybe multiple lines like we have below?

joannaquu (approved):

ios/FluentUI/Avatar/Avatar.swift
mischreiber: Yeah I have another, bigger refactor in the wings that completely refactored this logic, but I agree that we should still try to keep it legible in the interim.

mischreiber: Update: I've switch to using the outer gap for this API, which I think fits better with the intent anyway (and the values are identical). Plus, it makes for cleaner code!

joannaquu: looks good, ship it!

PR #2094, opened 2024-10-01
Create `FluentUI_common` module, part 1

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

This is part 1 of creating a new `FluentUI_common` module to share `FluentTheme` (and others) between iOS and macOS. The goal of this PR is to rework several files and interfaces in ways that will directly contribute towards the goal of moving this important logic out of the module.

Some highlights of the changes:
* Making many previously internal classes/structures `public` so the `FluentUI_ios` module can consume them out of `FluentUI_common`. For example, much of `ControlTokenSet`.
* No more `TokenizedControlInternal` interface. It's been fully folded into `TokenizedControl`.
* The iOS `DynamicColor` has been reduced further in importance, and now largely exists only to have something to conform to `ShapeStyle` for iOS 17.
* Several extensions have been broken up, separating UIKit and SwiftUI logic. The latter of these will often be shared with macOS.
* Move `isApplicableThemeChange()` onto a `UIView` extension. This will make it possible to share theme changing logic in the future by creating an equivalent `NSView` extension.

### Binary change

Still working on new system, but when comparing Archive output of the release iOS app:
| Before | After |
|-----|-----|
| 16,755,710 | 16,792,142 |

### Verification

Key scenario validation pass across iOS demo app.

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2094)


Sources/FluentUI_iOS/Core/Theme/Tokens/ControlTokenSet.swift
```diff
@@ -204,9 +206,11 @@ public class ControlTokenSet<T: TokenSetKey>: ObservableObject {
 /// Union-type enumeration of all possible token values to be stored by a `ControlTokenSet`.
 public enum ControlTokenValue {
     case float(() -> CGFloat)
-    case uiColor(() -> UIColor)
     case color(() -> Color)
+#if os(iOS) || os(visionOS)
+    case uiColor(() -> UIColor)
```
joannaquu, line 211: I'm assuming macOS will exclusively use swift color then? (no plans to add a nsColor control token). if so any reason to keep uiColor around (besides the fact that deprecating it will be a pain)

Sources/FluentUI_iOS/Core/Theme/FluentTheme+UIKit.swift
```diff
@@ -23,6 +87,61 @@ public extension FluentTheme {
...
+
+        // If there is no object, or it is not a UIView, we must assume that we need to update.
+        guard let themeView = notification.object as? UIView else {
+            return true
+        }
+
+        // If the object is a UIView, we only update if `view` is a descendant thereof.
+        return self.isDescendant(of: themeView)
+    }
+
```
joannaquu, line 145: nit: whitespace

Sources/FluentUI_iOS/Core/Theme/Tokens/DynamicColor.swift
```diff
@@ -0,0 +1,44 @@
...
+/// A container that stores a dynamic set of `Color` values.
+struct DynamicColor: Hashable {
+
+    /// Creates a custom `ShapeStyle` that stores a dynamic set of `Color` values.
+    ///
+    /// - Parameter light: The default `Color` for a light context. Required.
+    /// - Parameter dark: The override `Color` for a dark context. Optional.
+    /// - Parameter darkElevated: The override `Color` for a dark elevated context. Optional.
+    init(light: Color,
+                dark: Color? = nil,
```
joannaquu, line 17: nit: whitespace

joannaquu (approved):

Sources/FluentUI_iOS/Core/Theme/Tokens/ControlTokenSet.swift
mischreiber: > I'm assuming macOS will exclusively use swift color then? (no plans to add a nsColor control token). 
For the time being, that's the goal. We'll see if there's customer feedback asking for NSColor, in which case we can consider it.

> I'm assuming macOS will exclusively use swift color then? (no plans to add a nsColor control token). if so any reason to keep uiColor around (besides the fact that deprecating it will be a pain)
100%, it's just a deprecation thing we'd have to stage and plan and roll out. Maybe someday!

PR #2101, opened 2024-10-31
[A11y][VoiceOver] Set accessibility value for the sheet expansion state

nodes11:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Currently, the resizing handle of the bottom sheet does not read its current state (i.e. "Collapsed" or "Expanded"). This change adds that functionality to align with Apple system sheets and improve the VoiceOver experience.

I `updateResizingHandleViewAccessibility` to set the `accessibilityValue` on the `resizingHandle` based on the provided expansion state. Updated where `updateResizingHandleViewAccessibility` is called to ensure that the accessibility values are updated before the voiceover reads the new state.

### Binary change

### Verification

On iPhone, in the Fluent Test app, I validated that the "Expanded" value is read when the sheet is in an expanded state and that the "Collapsed" value is read when the sheet is in a collapsed state.

<details>
After:

https://github.com/user-attachments/assets/812cb6cc-fecd-4a74-a7ca-f99b353d2183

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2101)

### DIFF
--- a/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
+++ b/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
@@ -174,7 +174,7 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     /// The message should convey the "Expand" action and will be used when the bottom sheet is collapsed.
     @objc public var handleExpandCustomAccessibilityLabel: String? {
         didSet {
-            updateResizingHandleViewAccessibility()
+            updateResizingHandleViewAccessibility(for: currentExpansionState)
         }
     }
 
@@ -182,7 +182,7 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     /// The message should convey the "Collapse" action and will be used when the bottom sheet is expanded.
     @objc public var handleCollapseCustomAccessibilityLabel: String? {
         didSet {
-            updateResizingHandleViewAccessibility()
+            updateResizingHandleViewAccessibility(for: currentExpansionState)
         }
     }
 
@@ -496,6 +496,8 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         super.viewDidAppear(animated)
 
         tokenSet.update(fluentTheme)
+
+        updateResizingHandleViewAccessibility(for: currentExpansionState)
     }
 
     public override func viewDidLayoutSubviews() {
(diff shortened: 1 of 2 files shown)


joannaquu:
can you just call `updateResizingHandleViewAccessibility` in `move(...)`? or is the issue still that the value is wrong

Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
```diff
@@ -677,6 +677,20 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         }
     }
 
+    private func updateResizingHandleViewAccessibilityValue(for state: BottomSheetExpansionState) {
+        switch state {
+        case .collapsed:
+            resizingHandleView.accessibilityValue = "Accessibility.Drawer.ResizingHandle.Value.Collapsed".localized
+            break
```
anandrajeswaran, line 684: No need for `break`? (I <3 swift)

Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
```diff
@@ -677,6 +677,20 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
...
+        case .expanded:
+            resizingHandleView.accessibilityValue = "Accessibility.Drawer.ResizingHandle.Value.Expanded".localized
+            break
+        default:
```
anandrajeswaran, line 688: General best practice to avoid default cases (especially for enums we define in our module) so any added cases must explicitly consider whether there should be an accessibility value

Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
```diff
@@ -667,13 +670,15 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
         move(to: isExpanded ? .collapsed : .expanded, interaction: .resizingHandleTap)
     }
 
-    private func updateResizingHandleViewAccessibility() {
-        if currentExpansionState == .expanded {
+    private func updateResizingHandleViewAccessibility(for state: BottomSheetExpansionState) {
```
nodes11, line 673: This method was being called from the `didSet` method for `currentExpansionState` when the `currentExpansionState` is `.transitioning`. The accessibility properties would be set for the collapsed state via the `else` block and the sheet would read "Expand, Collapsed, Douple to tap expand" after the sheet was in the expanded state. The `currentExpansionState` is set again to the `.expanded` state but the updated accessibility properties are not set in time for VoiceOver to read them correctly.

 There is no real Accessibility/VoiceOver use case for the transitioning state. Let's only update the accessibility properties if the state is `expanded` or `collapsed` and no-op for other states.

mischreiber (approved):

PR #2113, opened 2025-01-07
Fixing floating button style horizontal padding

daceball:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Fixing UI bugs in `FluentButtonStyle(style: .floatingSubtle)` style from the fluent library.

###Actual behavior:
The icon in the Floating fluent button style is not centered because of some padding added to the icon. Icon must be centered. The fluent icon shape is an oval, but it must be a circle.

###Expected behavior:
Icon must be centered, and button frame should be a circle.

Screenshot 2024-12-18 at 4 55 43 PM

### Verification

The floating button horizontal margin from icon to frame must be symmetric.

<details>

</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/Sources/FluentUI_iOS/Components/Button/FluentButtonStyle.swift
+++ b/Sources/FluentUI_iOS/Components/Button/FluentButtonStyle.swift
@@ -97,12 +97,11 @@ public struct FluentButtonStyle: SwiftUI.ButtonStyle {
     private var edgeInsets: EdgeInsets {
         let size = size
         let horizontalPadding = ButtonTokenSet.horizontalPadding(style: style, size: size)
-        let fabAlternativePadding = ButtonTokenSet.fabAlternativePadding(size)
         return EdgeInsets(
             top: .zero,
             leading: horizontalPadding,
             bottom: .zero,
-            trailing: style.isFloating ? fabAlternativePadding : horizontalPadding
+            trailing: horizontalPadding
         )
     }
 


Sources/FluentUI_iOS/Components/Button/FluentButtonStyle.swift
```diff
@@ -97,12 +97,11 @@ public struct FluentButtonStyle: SwiftUI.ButtonStyle {
     private var edgeInsets: EdgeInsets {
         let size = size
         let horizontalPadding = ButtonTokenSet.horizontalPadding(style: style, size: size)
-        let fabAlternativePadding = ButtonTokenSet.fabAlternativePadding(size)
```
mischreiber, line 100: Can you check to see how this affects FAB controls with text? IIRC this alternate padding was used to ensure that buttons with text had the right amount of padding, which was slightly higher on the trailing edge.

daceball: Sure! In the Figma the trailing and leading padding have same values.

image

image

This is how the FAQ with text looks after the update I made. Please let me know if it looks good and if there are other scenarios I should test. Thanks! (:

Screenshot 2025-01-07 at 12 25 18 PM

daceball: I could also add one more condition to only use the extra padding if there is text in the label, just to make sure I won't affect other scenarios. Working on it!

daceball: @mischreiber Could you confirm if the FAB with text example I added above looks correct? thanks!

mischreiber: There's something odd going on -- look at how much space there is between the right edge of the word `Label` and the actual bounding box that's being calculated. I'm investigating what's going on there now.

mischreiber: Confirmed: there appears to be a 4pt spacing between the actual label's bounding box, and the box being measured against.

image

daceball: oh got it! Thanks Mike! Gonna add this extra padding only when there is some text in the label and 0 extra padding for icon only. Working on it!

daceball: Updated to only use the alternative padding if the button is a FAB with text, so that clients can call it as:
```
FluentButtonStyle(style: .floatingSubtle, isIconOnly: true)
```

Screenshot 2025-01-09 at 4 33 44 PM

anandrajeswaran: Is there a way we can figure out where those extra 4pts of spacing are coming from? Feels like a slightly painful API if users have to duplicate information of whether it has a label to us?

mischreiber: It's definitely an intentional design decision. The component looks unbalanced without the extra padding.

daceball: @mischreiber Thank you for confirming that this is by design. Could you please let me know if you agree with the fix I made by adding the `isIconOnly` parameter to the style init: `FluentButtonStyle(style: .floatingSubtle, isIconOnly: true)`? Thanks

joannaquu: @daceball is there a way to internally determine if it's an iconOnly/textOnly button, I agree with Anand that this is a slightly painful api then

daceball: @mischreiber Looks like `ButtonStyle` doesn’t have a built-in way to figure out if a button is icon-only or icon+text. In `makeBody(configuration:)`, the `configuration.label` is just a `View`, so the style doesn’t get any extra info about whether there’s text or just an icon. We’d have to pass this info as a parameter (like in my last commit). I can think of the following approaches:

- Pass a parameter indicating the type of floating button: icon-only or icon+text.
- Create a new style type (e.g., `style: .floatingSubtleIconOnly`) and pick the right padding for each type.
- Update the design and use the same padding for both icon-only and icon+text.

Let me know what you think about these options or if you have another idea to solve this bug. I'd really appreciate your input!

daceball: @mischreiber friendly ping. Thanks :)

PR #2114, opened 2025-01-13
Added large custom view size to MSFTableViewCell

nabil-k:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
- Added a large custom view size option MSFTableViewCell to accommodate larger custom views. 
- TableViewCell.swift
  - Forcing two line layout when using the new large size if the layout type was trying to be one line.
- Added the large custom view size example to the demo app
### Binary change

N/A

### Verification
When using a fluent avatar with size 40 and then setting an activity, for example:
```
let avatar = MSFAvatar(style: .default, size: .size40)
avatar.state.activityStyle = .circle
avatar.state.activityImage = UIImage(systemName: "star")
avatar.state.image = UIImage(systemName: "person")
```
The avatar view will actually go from having 40x40 dimensions, to having 46x46 dimensions. When using the fluent table view cell which can only accommodate 40x40, this causes the unread status to make contact with the custom view. However, when using the new large custom view, that is no longer the case.

| Before (Using medium custom view size) | After (using new large custom view size)     |
| image | image |
</details>

| Before (Not forcing two line layout for large custom view size) | After (forcing two line layout for large custom view size) |
| image  | image |
</details>

Large custom view size added to testapp:
image

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2114)

### DIFF
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/TableViewSampleData.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/TableViewSampleData.swift
@@ -25,6 +25,7 @@ class TableViewSampleData {
         let hasCustomAccessoryView: Bool
         let hasCustomLeadingView: Bool
         let hasHandler: Bool
+        let customViewSize: MSFTableViewCellCustomViewSize?
 
         init(title: String,
              items: [Item] = [],
@@ -41,7 +42,8 @@ class TableViewSampleData {
              hasCustomLinkHandler: Bool = false,
              hasCustomAccessoryView: Bool = false,
              hasCustomLeadingView: Bool = false,
-             hasHandler: Bool = false) {
+             hasHandler: Bool = false,
+             customViewSize: MSFTableViewCellCustomViewSize? = nil) {
             self.title = title
             self.items = items
             self.isUnreadDotVisible = isUnreadDotVisible
@@ -58,6 +60,7 @@ class TableViewSampleData {
             self.hasCustomAccessoryView = hasCustomAccessoryView
             self.hasCustomLeadingView = hasCustomLeadingView
             self.hasHandler = hasHandler
+            self.customViewSize = customViewSize
         }
     }
 
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/TableViewSampleData.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/TableViewSampleData.swift
@@ -24,8 +24,8 @@ class TableViewSampleData {
         let hasCustomLinkHandler: Bool
         let hasCustomAccessoryView: Bool
         let hasCustomLeadingView: Bool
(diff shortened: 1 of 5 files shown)


joannaquu:
have we run this pass design yet?

nabil-k:
Haven't followed up with any designers about this regarding fluent specifically.

anandrajeswaran:
Is the issue really that we need a new size? Or is it just a layout (implementation) issue at an existing size?

nabil-k:
Followed up with @joannaquu offline and reaching out to designers. Will provide an update when I hear back.

@anandrajeswaran I think the issue is that we'd need a new size. From that I've seen in a design the avatars with activities are supposed to be 46x46. But I'll pass this question along with design and see what they say if the adjustment that needs to be made is related to the table view cell or the avatar.

nabil-k:
Spoke with design and they were okay with this change. They didn't have any opinions though on if the change should be made to the avatar v.s. the table cell.

anandrajeswaran:
> Followed up with @joannaquu offline and reaching out to designers. Will provide an update when I hear back.
> 
> @anandrajeswaran I think the issue is that we'd need a new size. From that I've seen in a design the avatars with activities are supposed to be 46x46. But I'll pass this question along with design and see what they say about if the adjustment that needs to be made is related to the table view cell or the avatar.

Can you add an example in the Demo app? May be interesting to  show how this looks relative to other similarly sized cells and what impact that has on consuming code

nabil-k:

Demos/FluentUIDemo_iOS/FluentUI.Demo/TableViewSampleData.swift
```diff
@@ -41,7 +42,8 @@ class TableViewSampleData {
              hasCustomLinkHandler: Bool = false,
              hasCustomAccessoryView: Bool = false,
              hasCustomLeadingView: Bool = false,
-             hasHandler: Bool = false) {
+             hasHandler: Bool = false,
+             customViewSize: MSFTableViewCellCustomViewSize? = nil) {
```
joannaquu, line 46: what's a nil size end up being/should we make it an explicit default size

joannaquu: also nit: can we move this under `hasCustomLeadingView`

joannaquu:
sorry, just thought of one more thing to check. `MSFTableViewCellCustomViewSize` should be used in `ListItem`, as well `PersonaCell`, `PopupMenuCell`. can you double check that this change does not affect them?

nabil-k:
Checked with those controls and didn't see they'd be affected. But I'm going to close this PR since I was able to get the same results by just overriding the tokens for the custom view.

PR #2118, opened 2025-01-17
Bottom Sheet Animation Completes Regardless of Safe Area Changes

cbowdoin:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
When a safe area change happens, we get callback to `viewSafeAreaInsetsDidChange` which will call `completeAnimationsIfNeeded(skipToEnd: true)`. When we get this call we currently pass in skipToEnd as true, this results in us not properly animating the sheet in scenarios where the sheets safe area changes during the animation such as when the keyboard is being presented. By passing in false to skipToEnd we can avoid skipping an animation that the user is currently watching.

### Binary change
TBD

### Verification
- [x] Safe area changes while the sheet is animating are not interrupted
- [x] Changing the safe area while sheet is not animating puts the sheet in the correct location 
- [x] Safe Area changes while sheet is not animating do not animate the sheet

<details>

| Safe area change that occurred during a sheet animation skips it to the completed state | Animations continue on their current path while animating during a safe area change |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2118)


nodes11 (approved):

anandrajeswaran:
For future reference, can you be more specific in the "Verification" section? Which customer scenarios in the demo app did you verify that exercised those code paths where safe areas change?

PR #2119, opened 2025-01-17
Tab Bar View can hide separator and disable blur effect

cbowdoin:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
Consumers have no way currently to control the separator and blur effect created by the tab bar view. This PR adds the ability to specify in the init that you want to disable either/both the separator or the blur effect.  We do not apply any effect to the background view if the blur effect is disabled. If the separator is disabled we will set the hidden property on the separator view to true.

Tab Bar View demo app has the incrementor on the right and decrementor on the left to match LTR design patterns.
Additionally, it now has toggles for the control of the separator and blur effect.

### Binary change

### Verification
- [x] Built objective c projects that use Tab Bar View
- [x] Tab Bar View can be seen with no separator and blur effect in the demo controller.

<details>

| TabBarViewDefault | TabBarViewNoBlurNoDivider |
| image | image | 
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2119)

### DIFF
--- a/Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
+++ b/Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
@@ -92,24 +92,11 @@ open class TabBarView: UIView, TokenizedControl {
             stackView.setCustomSpacing(spacing, after: view)
         }
     }
-    /// Initializes MSTabBarView
-    /// - Parameter showsItemTitles: Determines whether or not to show the titles of the tab bar items.
-    @objc public convenience init(showsItemTitles: Bool = false) {
-        self.init(showsItemTitles: showsItemTitles, hideSeparator: false, disableBlur: false)
-    }
 
     /// Initializes MSTabBarView
     /// - Parameter showsItemTitles: Determines whether or not to show the titles of the tab bar items.
-    /// - Parameter hideSeparator: Removes the top divider displayed above the tab bar.
-    /// - Parameter disableBlur: Disables the blur effect applied to the tab bar.
-    @objc public init(showsItemTitles: Bool = false, hideSeparator: Bool = false, disableBlur: Bool = false) {
+    @objc public init(showsItemTitles: Bool = false) {
         self.showsItemTitles = showsItemTitles
-
-        if disableBlur {
-            self.backgroundView = UIVisualEffectView()
-        } else {
-          self.backgroundView = UIVisualEffectView(effect: UIBlurEffect(style: UIBlurEffect.Style.systemChromeMaterial))
-        }
 
         super.init(frame: .zero)
 
@@ -121,7 +108,6 @@ open class TabBarView: UIView, TokenizedControl {
         let stackViewInsets = Compatibility.isDeviceIdiomVision() ? UIEdgeInsets(top: 12.0, left: 0.0, bottom: 12.0, right: 0.0) : .zero
         contain(view: stackView, withInsets: stackViewInsets, respectingSafeAreaInsets: true)
 
-        topBorderLine.isHidden = hideSeparator
         topBorderLine.translatesAutoresizingMaskIntoConstraints = false
         addSubview(topBorderLine)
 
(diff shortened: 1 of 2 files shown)


Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
```diff
@@ -92,11 +92,25 @@ open class TabBarView: UIView, TokenizedControl {
...
+    /// - Parameter hideSeparator: Removes the top divider displayed above the tab bar.
+    /// - Parameter disableBlur: Disables the blur effect applied to the tab bar.
+    @objc public init(showsItemTitles: Bool = false, hideSeparator: Bool = false, disableBlur: Bool = false) {
         self.showsItemTitles = showsItemTitles
+
+        if disableBlur {
+            self.backgroundView = UIVisualEffectView()
+        } else {
+          self.backgroundView = UIVisualEffectView(effect: UIBlurEffect(style: UIBlurEffect.Style.systemChromeMaterial))
+        }
```
nodes11, line 112: minor: I think this could be reduced to

```
var blurEffect: UIVisualEffect? = disableBlur ? nil : UIBlurEffect(style: UIBlurEffect.Style.systemChromeMaterial)
self.backgroundView = UIVisualEffectView(effect: blurEffect) 
```

Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
```diff
@@ -107,6 +121,7 @@ open class TabBarView: UIView, TokenizedControl {
         let stackViewInsets = Compatibility.isDeviceIdiomVision() ? UIEdgeInsets(top: 12.0, left: 0.0, bottom: 12.0, right: 0.0) : .zero
         contain(view: stackView, withInsets: stackViewInsets, respectingSafeAreaInsets: true)
 
+        topBorderLine.isHidden = hideSeparator
```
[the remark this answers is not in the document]
nodes11: Alternatively, should `hideSeparator` be changed to a public property so that the view shows/hides when set?

Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
```diff
@@ -92,11 +92,25 @@ open class TabBarView: UIView, TokenizedControl {
...
 
     /// Initializes MSTabBarView
     /// - Parameter showsItemTitles: Determines whether or not to show the titles of the tab bar items.
-    @objc public init(showsItemTitles: Bool = false) {
+    /// - Parameter hideSeparator: Removes the top divider displayed above the tab bar.
+    /// - Parameter disableBlur: Disables the blur effect applied to the tab bar.
+    @objc public init(showsItemTitles: Bool = false, hideSeparator: Bool = false, disableBlur: Bool = false) {
         self.showsItemTitles = showsItemTitles
+
+        if disableBlur {
```
joannaquu, line 108: nit: can we move this to where the other `backgroundView` code is in init

Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
```diff
@@ -92,11 +92,25 @@ open class TabBarView: UIView, TokenizedControl {
...
+    @objc public convenience init(showsItemTitles: Bool = false) {
+        self.init(showsItemTitles: showsItemTitles, hideSeparator: false, disableBlur: false)
+    }
 
     /// Initializes MSTabBarView
     /// - Parameter showsItemTitles: Determines whether or not to show the titles of the tab bar items.
-    @objc public init(showsItemTitles: Bool = false) {
+    /// - Parameter hideSeparator: Removes the top divider displayed above the tab bar.
+    /// - Parameter disableBlur: Disables the blur effect applied to the tab bar.
+    @objc public init(showsItemTitles: Bool = false, hideSeparator: Bool = false, disableBlur: Bool = false) {
```
joannaquu, line 105: would it be cleaner for these to be public properties instead of passing them in on init?

Sources/FluentUI_iOS/Components/Tab Bar/TabBarView.swift
cbowdoin: Unfortunately no since we need to initialize `backgroundView` before we call `super.init()`

joannaquu: oooh I see, since we moved this out into init, ok ignore me

nodes11:
A video demonstrating the updates to the demo controller should be added to the visual verification section.

nodes11 (approved):

anandrajeswaran:
Random thought - do we ever expect someone to toggle one of these but not both of these? If not, could encapsulate into single declarative property that effectively saws "are you detaching a tab bar from its child view controller"?

PR #2121, opened 2025-01-27
[macOS] Multiline Pill Picker

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [x] macOS

### Description of changes

Created a new component, `MultilinePillPicker`, for macOS. Built in SwiftUI and wrapped (via `[MSF]MultilinePillPickerView`) for AppKit. Supports two rows of controls, with optional actions on selection.

### Binary change

n/a - no measurement system for macOS as of yet.

### Verification

Verified light and dark modes, enabled and disabled modes. See video and screenshots below.

<details>
<summary>Visual Verification</summary>

https://github.com/user-attachments/assets/038582f1-cbfc-49c8-8d6b-2e9e1a1c00c2

| Enabled                                    | Disabled                                  |
|--------------------------------------------|-------------------------------------------|
| light-enabled | light-disabled |
| dark-enabled | dark-disabled |

</details>

### Pull request checklist

This PR has considered:
- [x] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [x] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [x] Objective-C exposure (provide it only if needed)
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2121)

PR #2122, opened 2025-01-28
[macOS] Fix ObjC export of MultilinePillPickerView

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [x] macOS

### Description of changes

My previous PR #2121 didn't work in ObjC as expected. Since we don't have a real ObjC demo in macOS yet, this was manually validated to quickly unblock the issue.

(The problem was inheriting from a generic class -- that silently fails to export to Objective-C, even though there's an `@objc` in the title.)

### Binary change

n/a, still can't measure macOS

### Verification

Verified MPP component still renders as expected.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2122)

### DIFF
--- a/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
+++ b/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
@@ -42,7 +42,7 @@ public final class MultilinePillPickerView: ControlHostingView, ObservableObject
 	@MainActor @Published public var labels: [String]
 	@MainActor @Published public var action: (@MainActor (Int) -> Void)?
 
-	@MainActor private var viewModel: MultilinePillPickerViewModel = .init()
+	@MainActor private let viewModel: MultilinePillPickerViewModel = .init()
 }
 
 /// Maps properties from `MultilinePillPickerView` to `MultilinePillPickerViewWrapper`.
--- a/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
+++ b/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
@@ -56,8 +56,10 @@ fileprivate struct MultilinePillPickerWrapper: View {
 	@ObservedObject var viewModel: MultilinePillPickerViewModel
 
 	var body: some View {
-		MultilinePillPicker(labels: viewModel.labels,
-							action: viewModel.action)
+		MultilinePillPicker(
+			labels: viewModel.labels,
+			action: viewModel.action
+		)
 		.disabled(!viewModel.isEnabled)
 	}
 }
--- a/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
+++ b/Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
@@ -21,8 +21,18 @@ public final class MultilinePillPickerView: ControlHostingView, ObservableObject
 		self.action = action
 		super.init(AnyView(EmptyView()))
 
-		let wrapper = MultilinePillPickerWrapper(viewModel: self)
+		let wrapper = MultilinePillPickerWrapper(viewModel: viewModel)
 		self.hostingView.rootView = AnyView(wrapper)
+
+		// Set up observation to keep the view model in sync.
+		viewModel.loadProperties(from: self)
+		cancellable = self.objectWillChange.sink { [weak self] in
+			DispatchQueue.main.async {
+				if let self {
+					self.viewModel.loadProperties(from: self)
+				}
+			}
+		}
 	}
 
 	@MainActor required dynamic init?(coder aDecoder: NSCoder) {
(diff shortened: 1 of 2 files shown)


Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
```diff
@@ -49,6 +49,6 @@ public final class MultilinePillPickerView: ControlHostingView {
 	private func updatePicker() {
 		let picker = MultilinePillPicker(labels: labels, action: action)
```
lukas-cap, line 50: Does this approach block SwiftUI from nicely handling the state updates with transitions? Specifically `isEnabled` and `labels` feel like they should be Published props in an `ObservableObject`.

lukas-cap (approved):

Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
mischreiber: It's a good point! That's actually how most of our Fluent iOS components work. I'm not sure how I'd want to arrange that though -- the approach we used there is a bit cumbersome and doesn't play well with established things like pulling `isEnabled` from the environment.

nodes11 (approved):

Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
mischreiber: Talked offline, and came up with a very clever solution. Thanks @lukas-cap!

Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
```diff
@@ -30,25 +33,18 @@ public final class MultilinePillPickerView: ControlHostingView {
...
+fileprivate struct MultilinePillPickerWrapper: View {
+	@ObservedObject var viewModel: MultilinePillPickerView
 
-	private func updatePicker() {
-		let picker = MultilinePillPicker(labels: labels, action: action)
-			.disabled(!isEnabled)
-		rootView = AnyView(picker)
+	var body: some View {
+		MultilinePillPicker(labels: viewModel.labels,
+							action: viewModel.action)
```
lukas-cap, line 47: Intended whitespace? 

Screenshot 2025-01-28 at 3 49 20 PM

mischreiber: Ah it's GitHub not playing nicely with our existing indentation in FluentUI macOS. Looks right when tabs are set to four spaces instead of eight.

Saadnajmi (approved):

Sources/FluentUI_macOS/Core/ControlHostingView.swift
```diff
@@ -4,34 +4,71 @@
...
+	}
+
 	/// Initializes and returns an instance of `ControlHostingContainer` that wraps `controlView`.
 	///
 	/// Unfortunately this class can't use Swift generics, which are incompatible with Objective-C interop. Instead we have to wrap
 	/// the control view in an `AnyView.`
 	///
 	/// - Parameter controlView: An `AnyView`-wrapped component to host.
-	/// - Parameter safeAreaRegions: Passthrough to the respective property on NSHostingView.
+	/// - Parameter safeAreaRegions: Passthrough to the respective property on UIHostingController.
```
lukas-cap, line 25: Assuming this should still say NSHostingView

Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
```diff
@@ -18,37 +19,48 @@ public final class MultilinePillPickerView: ControlHostingView {
...
-	@MainActor public var action: (@MainActor (Int) -> Void)? {
-		didSet {
-			updatePicker()
-		}
-	}
+	@MainActor @Published public var isEnabled: Bool = true
+	@MainActor @Published public var labels: [String]
+	@MainActor @Published public var action: (@MainActor (Int) -> Void)?
+
+	@MainActor private var viewModel: MultilinePillPickerViewModel = .init()
```
lukas-cap, line 45: I think this can be let bc the VM is not a value type?

lukas-cap (approved):

Sources/FluentUI_macOS/Core/ControlHostingView.swift
mischreiber: Whoops, yeah copied a bit from the iOS one

Sources/FluentUI_macOS/Components/MultilinePillPicker/MultilinePillPickerView.swift
mischreiber: Yep good catch

PR #2134, opened 2025-02-20
[macOS] Announce `DatePickerController` changes in VoiceOver

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [x] macOS

### Description of changes

* Create a new `NSAccessibilityStepper` for the header, making it possible to read and change the value of the header as one entire element.
* Add an `NSAccessibility` announcement when the visible month in the `DatePickerController` changes.
* Clean up `TestDatePickerController` so popover-hosted picker works as expected.

### Binary change

n/a, no Mac size testing

### Verification

* Verified VoiceOver announcements happen for all three date pickers in the test screen.
* Ensured announcements happen exactly once, and only when the picker changes (not on initial appearance).

https://github.com/user-attachments/assets/28af074e-68d2-4a06-8f20-df513797e47a

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2134)


Demos/FluentUIDemo_macOS/FluentUITestViewControllers/TestDatePickerController.swift
```diff
@@ -14,14 +14,12 @@ class TestDatePickerController: NSViewController {
...
 
-		datePickerController = DatePickerController(date: date, calendar: calendar, style: .dateTime)
-		menuDatePickerController = DatePickerController(date: date, calendar: calendar, style: .dateTime)
-
-		datePickerController?.delegate = self
-		menuDatePickerController?.delegate = self
-
-		datePickerController?.hasEdgePadding = true
-		menuDatePickerController?.hasEdgePadding = true
+		DatePickerLocation.allCases.forEach { location in
```
anandrajeswaran, line 17: Totally ignorable - I've mostly stopped using `.forEach` in favor of old-fashioned `for...in`. I personally pretty strongly agree with https://forums.swift.org/t/do-we-want-foreach/56929/2m, plus it makes debugging easier to just be able to step through a for loop and not to have to set a separate breakpoint inside because it's fired as a closure

Sources/FluentUI_macOS/Components/DatePicker/DatePickerView.swift
```diff
@@ -252,8 +252,18 @@ class DatePickerView: NSView {
...
 		headerView.weekdayStrings = Array(zip(dataSource.shortWeekdays, dataSource.longWeekdays))
-		headerView.monthYearLabel.stringValue = dateFormatter.string(from: dataSource.visibleRange.first)
+		headerView.monthYearLabel.stringValue = newMonthYearValue
+
+		if newMonthYearValue != previousValue && !previousValue.isEmpty {
+			NSAccessibility.post(element: NSApp.mainWindow as Any, notification: .announcementRequested, userInfo: [
+				NSAccessibility.NotificationUserInfoKey.announcement: newMonthYearValue,
+				NSAccessibility.NotificationUserInfoKey.priority: NSAccessibilityPriorityLevel.medium.rawValue
+			])
+		}
```
anandrajeswaran, line 266: Doesn't this usually happen because of direct user interaction? In which case we shouldn't need an announcement? Or is there a case where it's happening indirectly?

Demos/FluentUIDemo_macOS/FluentUITestViewControllers/TestDatePickerController.swift
anandrajeswaran: I think it's still a matter of preference and not a proven guideline so I'm good either way

Sources/FluentUI_macOS/Components/DatePicker/DatePickerView.swift
anandrajeswaran: What about just making the header title accessible so it's announced it a user moves focus to it? Or is that not trivial?

mischreiber: It already is, but that's not sufficiently accessible. We need an actual announcement when the value changes.

Demos/FluentUIDemo_macOS/FluentUITestViewControllers/TestDatePickerController.swift
mischreiber: Changed all the complex multi-line loops to use `for x in y`, but left the one-liners as-is

Sources/FluentUI_macOS/Components/DatePicker/DatePickerView.swift
anandrajeswaran: Still a bit concerned - that feels like it would solve the letter of the law of accessibility compliance but maybe not actually be better for the user? I think Apple has a general assumption that the user knows what they are focused on when they activate a button so announcing additional info beyond their activation being successful seems like it may add noise rather than resolve confusion.

Another way to put it - if the user is in a situation where they are focused on the "next month" button and don't know or can't easily infer which month that will take them to, we have already failed them?

mischreiber: The scenario that was brought up was less about a single press, and more about multiple taps. So if a user is on, say, March 2025, but needs to back up to September 2024, how do they accomplish this? Using visuals, a user would click the back button until they see that the UI shows September 2024, but what about using VoiceOver? Since there's currently no feedback on tap, a user has to (A) do mental math to figure out how many times to select a button, an/or (B) regularly move the focus back to the title to see where they are.

Add that to the fact that we do really need to be announcing major visual changes in the UI, and this feels like a very correct change.

For a comparison point, see Apple's `NSStepper`. The same behavior occurs: incrementing or decrementing the stepper via VoiceOver always announces the new value of the text field tied to the stepper.

mischreiber: Went ahead and just made it an `NSAccessibilityStepper` 😎

Sources/FluentUI_macOS/Components/DatePicker/CalendarHeaderView.swift
```diff
@@ -28,6 +28,66 @@ private struct Constants {
...
+
+	private func performStep(_ button: NSButton) -> Bool {
+		let isEnabled = button.isEnabled
+		if (isEnabled) {
+			button.performClick(button)
+			NSAccessibility.post(element: self, notification: .announcementRequested, userInfo: [
+				NSAccessibility.NotificationUserInfoKey.announcement: self.accessibilityValue() ?? "",
+				NSAccessibility.NotificationUserInfoKey.priority: NSAccessibilityPriorityLevel.medium.rawValue
+			])
+		}
```
brentpe, line 82: Is this still necessary or is there a way to get this for free since we're now a NSAccessibilityStepper?

mischreiber: Unfortunately it does still seem necessary :(

brentpe (approved):

Sources/FluentUI_macOS/Components/DatePicker/CalendarHeaderView.swift
```diff
@@ -28,6 +28,66 @@ private struct Constants {
...
+		self.setAccessibilityElement(true)
+		self.setAccessibilityRole(.incrementor)
+
+		self.addView(monthYearLabel, in: .center)
+		self.addView(leadingButton, in: .leading)
+		self.addView(trailingButton, in: .trailing)
+	}
+
+	required init?(coder: NSCoder) {
+		fatalError("init(coder:) has not been implemented")
```
anandrajeswaran, line 51: minor: Should be `preconditionFailure` per https://lapcatsoftware.com/articles/fatalError.html. And can also mark func as unavailable?

Sources/FluentUI_macOS/Components/DatePicker/CalendarHeaderView.swift
```diff
@@ -28,6 +28,66 @@ private struct Constants {
...
+		self.leadingButton = leadingButton
+		self.trailingButton = trailingButton
+
+		super.init(frame: .zero)
+
+		self.setAccessibilityElement(true)
+		self.setAccessibilityRole(.incrementor)
+
+		self.addView(monthYearLabel, in: .center)
+		self.addView(leadingButton, in: .leading)
```
anandrajeswaran, line 46: Should we mark these nested controls as not accessibility elements to match OS pattern of stepper elements not having children?

PR #2138, opened 2025-03-07
[macOS] Add dynamic color support directly to NSColor

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [x] macOS

### Description of changes

Creates a new extension on `NSColor` that adds dynamic color support, similar to the equivalent methods in `UIColor+Extensions.swift`.

These changes will be important for `FluentTheme` integration, which is almost ready :)

New:
- Create an `NSColor` with explicit `light` and `dark` variants.
- Fetch these values back on demand.
- Add test UI to the demo app to validate this.

### Verification

Verified that the new dynamic colors work as expected:
- The `light` method always returns the light value.
- The `dark` method returns either dark if it exists, or light if not.
- Otherwise, the color will draw according to the current `NSAppearance` as expected. This includes redrawing when the system appearance changes.

<details>

| dark | light |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2138)

### DIFF
--- a/Sources/FluentUI_macOS/Core/Extensions/NSColor+Extensions.swift
+++ b/Sources/FluentUI_macOS/Core/Extensions/NSColor+Extensions.swift
@@ -1,0 +1,86 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+#if canImport(AppKit)
+import AppKit
+import SwiftUI
+
+extension NSColor {
+
+    /// Creates a dynamic color object that returns the appropriate color value based on the current
+    /// rendering context.
+    ///
+    /// The decision order for choosing between the colors is based on the following questions, in order:
+    /// - Is the current `userInterfaceStyle` `.dark` or `.light`?
+    ///
+    /// - Parameter light: The default color for a light context. Required.
+    /// - Parameter dark: The override color for a dark context. Optional.
+    @objc public convenience init(light: NSColor,
+                                  dark: NSColor? = nil) {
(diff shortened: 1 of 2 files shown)


Sources/FluentUI_macOS/Core/Extensions/NSColor+Extensions.swift
```diff
@@ -0,0 +1,86 @@
...
+    /// rendering context.
+    ///
+    /// The decision order for choosing between the colors is based on the following questions, in order:
+    /// - Is the current `userInterfaceStyle` `.dark` or `.light`?
+    ///
+    /// - Parameter light: The default color for a light context. Required.
+    /// - Parameter dark: The override color for a dark context. Optional.
+    @objc public convenience init(light: NSColor,
+                                  dark: NSColor? = nil) {
+        self.init(name: nil) { appearance -> NSColor in
```
Saadnajmi, line 22: Could possibly use https://developer.apple.com/documentation/appkit/nscolor/init(name:dynamicprovider:) ?

mischreiber: Yep, that's what this method is! Just using the implicit closure attachment thing in Swift.

Saadnajmi: Dang. I can't read apparently 😂

Saadnajmi (approved):

PR #2139, opened 2025-03-07
Move `FluentTheme` code to new `FluentUI_common` module

mischreiber:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [x] macOS

### Description of changes

This PR lights up `FluentTheme` and various other low-level Fluent utilities on macOS by moving them to a new cross-platform module, `FluentUI_common`. To walk through the changes, I've split the diff into four commits.

#### 320e3748278da6354ed3e612a85f28cd57723b55: Create FluentUI_common target
* Nice and easy: updating `Package.swift` to specify a new build target, `FluentUI_common`.
* Updating both `FluentUI_macos` and `FluentUI_ios` to depend on this new target.

#### e6b23707616e6c43f5ba780dc5d20a456720b905: Move common files to FluentUI_common directory
* This moves many `FluentTheme`- and token-related files to the new `FluentUI_common` directory, with almost no changes.
* First exception to this is `Color+Extensions.swift`, which hooks into the new macOS-specific `DynamicColor` integration found in `NSColor+Extensions.swift`.
* Second exception is `FluentUI.swift`, which now also exports `FluentUI_common` in addition to the platform-specific module. This means users can continue to simply `import FluentUI` and get the code they expect.

#### 069ec24b4cdb0484f2e319af02ed7a5b308e1b2f: Add import FluentUI_common to iOS files
* This is a big commit, but nothing interesting is happening.
* I've just added `import FluentUI_common` to every remaining file in `FluentUI_ios`, now that so much of the support structure of Fluent iOS has been moved to that library.

#### 1e07c89bf11ed13317c30eb7281d5f0a8d07688b: Add import FluentUI_common to tests
* Same for tests :)

### Binary change

TBA, will be testing the built app to see impact.

### Verification

Ran all automation on both iOS and macOS, and got the same level of success as before my change (there are some failing Notification tests that we should fix later).

Also added a direct `FluentTheme` reference in the new FluentUI demo app, demonstrating the ability to access Fluent colors from macOS.

| Before | After |
|-----|-----|
| before | after |
| before_dark | after_dark |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2139)


Saadnajmi (dismissed):

Sources/FluentUI/FluentUI.swift
```diff
@@ -3,7 +3,8 @@
 //  Licensed under the MIT License.
 //
 
-#if os(iOS) || os(visionOS)
+@_exported import FluentUI_common
+#if os(iOS) || os(visionOS) || targetEnvironment(macCatalyst)
 @_exported import FluentUI_ios
 #elseif os(macOS)
 @_exported import FluentUI_macos
```
Saadnajmi, line 10: Do they need different names still?

Sources/FluentUI_common/Core/Extensions/NSColor+Extensions.swift
```diff
@@ -51,6 +51,12 @@ extension NSColor {
         return resolvedColorValue(appearance: NSAppearance(named: .darkAqua))
     }
 
+    convenience init(dynamicColor: DynamicColor) {
+        let colorResolver = { $0 != nil ? NSColor($0!) : nil }
+        self.init(light: NSColor(dynamicColor.light),
+                  dark: colorResolver(dynamicColor.dark))
+    }
```
Saadnajmi, line 58: Maybe add a comment that the purpose of color resolver is to ensure proper fallback behavior? IDK, your choice

Sources/FluentUI_common/Core/Theme/Tokens/DynamicColor.swift
```diff
@@ -6,30 +6,31 @@
...
     /// - Parameter light: The default `Color` for a light context. Required.
     /// - Parameter dark: The override `Color` for a dark context. Optional.
     /// - Parameter darkElevated: The override `Color` for a dark elevated context. Optional.
-    init(light: Color,
-         dark: Color? = nil,
-         darkElevated: Color? = nil) {
-        self.light = light
+	public init(light: Color,
+				dark: Color? = nil,
+				darkElevated: Color? = nil) {
```
Saadnajmi, line 19: Alignment looks off? Or is that just Github?

Sources/FluentUI_common/Core/Extensions/UIColor+Extensions.swift
```diff
@@ -3,8 +3,9 @@
 //  Licensed under the MIT License.
 //
 
-import UIKit
+#if canImport(UIKit)
 import SwiftUI
```
joannaquu, line 7: does anything in this file need swiftui?

Sources/FluentUI_common/Core/Theme/Tokens/DynamicColor.swift
```diff
@@ -6,30 +6,31 @@
...
     /// Creates a custom `ShapeStyle` that stores a dynamic set of `Color` values.
     ///
     /// - Parameter light: The default `Color` for a light context. Required.
     /// - Parameter dark: The override `Color` for a dark context. Optional.
     /// - Parameter darkElevated: The override `Color` for a dark elevated context. Optional.
-    init(light: Color,
-         dark: Color? = nil,
-         darkElevated: Color? = nil) {
-        self.light = light
+	public init(light: Color,
```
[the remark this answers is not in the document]
mischreiber: This can't be `@objc` because `Color` is a SwiftUI struct, which can't be expressed in Objective-C.

Sources/FluentUI_common/Core/Extensions/UIColor+Extensions.swift
mischreiber: Ah I guess not anymore! Will remove.

Sources/FluentUI_common/Core/Extensions/NSColor+Extensions.swift
mischreiber: Sure -- it's copied from the UIColor equivalent of this file, so I can comment both.

Sources/FluentUI/FluentUI.swift
mischreiber: As far as I know, yeah. The name is derived from the target name, which I *think* needs to be unique within the `Package.swift` definition. I can play with it a little, but I'm fairly certain this is required.

Sources/FluentUI_common/Core/Theme/Tokens/DynamicColor.swift
mischreiber: Good eye, it's because the file was a mix of tabs and spaces, and github uses 8-space tab characters by default for some reason. Converting to all spaces.

Sources/FluentUI/FluentUI.swift
thomasjross: FYI, but the `targetEnvironment()` check here is redundant. When building for Mac Catalyst, the `os(iOS)` check will be true.

So this is probably not needed, but there may be some value in being explicit about which module should be exported.

Sources/FluentUI_common/Core/Extensions/NSColor+Extensions.swift
```diff
@@ -51,6 +50,14 @@ extension NSColor {
         return resolvedColorValue(appearance: NSAppearance(named: .darkAqua))
     }
 
+    convenience init(dynamicColor: DynamicColor) {
+        // Simple closure to return a nil NSColor if the passed-in Color is
+        // also nil, since the argument to `NSColor.init(_:)` is not optional.
+        let colorResolver = { $0 != nil ? NSColor($0!) : nil }
```
thomasjross, line 56: Since `Optional` includes support for `.map()` and `.flatMap()`, I wonder if you could remove this closure and instead do something like:

```
self.init(light: NSColor(dynamicColor.light),
          dark: dynamicColor.dark.map { NSColor($0) }
```

Sources/FluentUI_common/Core/Extensions/Color+Extensions.swift
```diff
@@ -40,15 +40,19 @@ extension Color {
     ///
     /// - Parameter dynamicColor: A dynmic color structure that describes the `Color` to be created.
     init(dynamicColor: DynamicColor) {
-        if #available(iOS 17, *) {
+        if #available(iOS 17, macOS 14, *) {
             self.init(dynamicColor)
         } else {
+#if os(macOS)
```
thomasjross, line 46: This could be a good excuse to rename `Color+Extensions.swift` to something more descriptive and then introduce something like `DynamicColor+UIKit` and `DynamicColor+AppKit` (or similar names)

Sources/FluentUI_common/Core/Extensions/Color+Extensions.swift
```diff
@@ -40,15 +40,19 @@ extension Color {
     ///
     /// - Parameter dynamicColor: A dynmic color structure that describes the `Color` to be created.
```
thomasjross, line 41: Nitpick: Not related to this PR, but there's a typo here. `dynmic` -> `dynamic`

Sources/FluentUI_iOS/Core/UIKit+SwiftUI_interoperability.swift
```diff
@@ -3,8 +3,9 @@
 //  Licensed under the MIT License.
 //
 
-import UIKit
+import FluentUI_common
 import SwiftUI
+import UIKit
```
thomasjross, line 8: A little surprised this didn't generate a lint warning. There's the `sorted_imports` rule for this, which you could opt into.

Sources/FluentUI_common/Core/Extensions/NSColor+Extensions.swift
mischreiber: Ah I hadn't thought of that, good to keep in the back of my mind for the future too!

Sources/FluentUI_common/Core/Extensions/Color+Extensions.swift
mischreiber: I see the appeal, but two reasons I'm inclined to keep the current names for now:
1. `Color+Extensions.swift` isn't purely based around `DynamicColor`; it also adds `Color.init(hexValue:)`. In fact, the public API is half and half, and doesn't even mention `DynamicColor` at all.
2. Our convention in Fluent is to name extension files after the class they're extending. Since the files in question extend `Color`, `NSColor`, and `UIColor`, they should all be named `[Color|NSColor|UIColor]+Something.swift`. We could change the `+Something` part, but we run into the same problem that there are multiple kinds of additions to each class, not just `DynamicColor`.

Sources/FluentUI_common/Core/Extensions/Color+Extensions.swift
mischreiber: Nice catch, might as well grab it now!

Sources/FluentUI_iOS/Core/UIKit+SwiftUI_interoperability.swift
mischreiber: It looks like our linting is having some problems right now -- I'll pencil that in as a good rule to add once I have some time to restore the check.

Sources/FluentUI_common/Core/Theme/Tokens/DynamicColor.swift
```diff
@@ -6,30 +6,31 @@
 import SwiftUI
 
 /// A container that stores a dynamic set of `Color` values.
-struct DynamicColor: Hashable {
+@objc(MSFDynamicColor)
+public final class DynamicColor: NSObject {
```
anandrajeswaran, line 10: That a bunch of stuff (well actually not that much, but not zero) now has to be exposed publicly when it wasn't before makes me second guess myself. There's no way to share code selectively across platforms with use of path includes/excludes in Package.swift/MicrosoftFluentUI.podspec?

anandrajeswaran: Basically, is it actually gross to do this without introducing a new module?

Sources/FluentUI_common/Core/Extensions/UIColor+Extensions.swift
```diff
@@ -3,8 +3,8 @@
 //  Licensed under the MIT License.
 //
 
+#if canImport(UIKit)
```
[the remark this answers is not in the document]
mischreiber: I'm not sure if we can exclude individual files from every possible iteration of build systems that would touch this (SPM, Cocoapods, etc).

We could do an OS check instead of a `canImport`, but it can be a little more finicky considering the wide array of OSes we need to consider (iOS, visionOS, macCatalyst, mac), and will be less future-proof if we ever wanted to support more (e.g. tvOS, watchOS).

Sources/FluentUI_common/Core/Theme/Tokens/DynamicColor.swift
mischreiber: I've been beating my head against this for months now, and it's really just a limitation of SPM. If we went back to Cocoapods and Xcode projects alone, this would be very easy to support!

The problem is that Swift Packages only allow gating per-platform on a target (read: module) level. There is no way to conditionally include/exclude specific paths or files as part of a build.

The only options I saw that would work for our needs:
* The one you see here: two modules, more public than we hoped, etc.
* A single module/target, but every platform-specific file needs an `#if` gate around itself, *a la* `UIColor+Extensions.swift` from your earlier comment.
* Drop Swift Package Manager support 🧐

mischreiber: Got a solution in place! Cocoapods will continue to work as it has, while SPM will treat `common` as a separate module.

joannaquu (approved):

amoggo (approved):

PR #2141, opened 2025-03-20
[List][SwiftUI control] Add color and font from token set for List header & footer

alexanderboswell:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes

For the swiftUI control for List / ListItem, which is the TableViewCell swiftUI version, we don't support tokens yet. This change is to add in color and font.

- Added in a token set on the view
- Added a new enum for header style. We have `TableViewHeaderFooterView.Style` which maps to the token set, but the new swiftUI control, header and footer are distinctly different and will be in the future. To break that trend, from a consumer standpoint we can separate out the styles to just be an option for a header as we only have 1 footer style.

### Binary change

Not measured

### Verification

Tested with the demo app. We dont have a demo page for this yet, but could be added in a future change.

<details>

| header regular style                                       | header primary style                                      |
| image | image |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2141)

### DIFF
--- a/Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
+++ b/Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
@@ -5,6 +5,8 @@
 
 import SwiftUI
 
+public typealias ListSectionHeaderFooterTokenSet = TableViewHeaderFooterViewTokenSet
+
 /// Defines the visual style of the FluentListSectionHeader
 public enum ListSectionHeaderStyle {
     case regular
@@ -72,7 +74,7 @@ public struct FluentListSectionHeader<Title: StringProtocol, TrailingContent: Vi
 
     private var trailingContent: (() -> TrailingContent)?
     private let title: Title
-    private let tokenSet: TableViewHeaderFooterViewTokenSet
+    private let tokenSet: ListSectionHeaderFooterTokenSet
     @Environment(\.fluentTheme) private var fluentTheme: FluentTheme
 }
 
--- a/Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
+++ b/Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
@@ -5,6 +5,12 @@
 
 import SwiftUI
 
+/// Defines the visual style of the FluentListSectionHeader
+public enum ListSectionHeaderStyle {
+	case regular
+	case primary
+}
+
 /// This component is a work in progress. Expect changes to be made to it on a somewhat regular basis.
 ///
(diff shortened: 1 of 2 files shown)


joannaquu:
```
but the new swiftUI control, header and footer are distinctly different and will be in the future
```
does it still make sense to reuse the tableview token set then?

Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
```diff
@@ -44,16 +57,37 @@ public struct FluentListSectionHeader<Title: StringProtocol, TrailingContent: Vi
...
+        case .primary:
+            return .headerPrimary
+        }
+    }
+
     // MARK: Private variables
 
     private var trailingContent: (() -> TrailingContent)?
     private let title: Title
+    private let tokenSet: TableViewHeaderFooterViewTokenSet
```
joannaquu, line 75: similar to `ListItem`, should we have a `FluentListSectionHeaderTokenSet` alias

alexanderboswell:
> ```
> but the new swiftUI control, header and footer are distinctly different and will be in the future
> ```
> 
> does it still make sense to reuse the tableview token set then?

Well I guess if I understand, the token system is already well established and having two two token sets for the same thing is not preferred (like we want the same font and color for the section header no matter if it is in swiftui or UIKit).


I guess what I meant by this is that it's not a "HeaderFooterView" as one component anymore, its HeaderView and a FooterView which dont overlap. 

Thoughts? This was my motivation on this approach.

Sources/FluentUI_iOS/Components/List/FluentListSectionHeader.swift
alexanderboswell: Yeah makes sense, I can add!

joannaquu:
> > ```
> > but the new swiftUI control, header and footer are distinctly different and will be in the future
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > does it still make sense to reuse the tableview token set then?
> 
> @joannaquu Well I guess if I understand, the token system is already well established and having two two token sets for the same thing is not preferred (like we want the same font and color for the section header no matter if it is in swiftui or UIKit).
> 
> I guess what I meant by this is that it's not a "HeaderFooterView" as one component anymore, its HeaderView and a FooterView which dont overlap.
> 
> Thoughts? This was my motivation on this approach.

ahh ok gotcha, I totally read that wrong and interpreted it as our header and footer styles would change in the future

joannaquu (approved):

PR #2143, opened 2025-04-14
[WIP] Add CommandBarItemGroup labels

joannaquu:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Adding support for `CommandBarItemGroup` labels
- Update `CommandBarItemGroup`, which is currently a typealias (`[CommandBarItem]`) to be a class that takes in a list of items and a label
- `CommandBarButton`: prevent buttons from stretching for long labels
- `CommandBarButtonGroupView`: add a stack view that contains buttons + label
  - In this stack view, the label should only take the vertical space it needs. The buttons should fill the remaining space

Note: this work is experimental

### Verification

<details>

| before | after |

(This is just an example of labels. `CommandBarDemoController` has not been updated)

</details>

### Pull request checklist

This PR has considered:


Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/CommandBarDemoController.swift
```diff
@@ -339,17 +343,15 @@ class CommandBarDemoController: DemoController {
...
-            commandGroup.map { command in
-                newItem(for: command)
-            }
+        let itemGroups: [CommandBarItemGroup] = commandGroups.map {
+            CommandBarItemGroup(items: $0.map { newItem(for: $0) })
         }
 
-        itemGroups[0][1].isEnabled = false
-        itemGroups[2][0].isEnabled = false
+        itemGroups[0].items[1].isEnabled = false
```
mischreiber, line 350: Idea: were you interested in implementing the [Subscript](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/subscripts/) operator on `CommandBarItemGroup` so you can access values directly via square brackets?

Sources/FluentUI_iOS/Components/Command Bar/CommandBarItem.swift
```diff
@@ -5,7 +5,22 @@
 
 import UIKit
 
-public typealias CommandBarItemGroup = [CommandBarItem]
+@objc(MSFCommandBarItemGroup)
+public class CommandBarItemGroup: NSObject {
+    public let items: [CommandBarItem]
+    public let label: String?
+
+    @objc public init(items: [CommandBarItem]) {
```
mischreiber, line 13: I don't have a *strong* preference, but how do you feel about making this a suppressed argument name with an underscore?
```swift
    @objc public init(_ items: [CommandBarItem]) {
        // ...
    }

    @objc public init(_ items: [CommandBarItem], label: String? = nil) {
        // ...
    }
```

Sources/FluentUI_iOS/Components/Command Bar/CommandBarItem.swift
```diff
@@ -5,7 +5,22 @@
...
+        self.items = items
+        self.label = nil
+    }
+
+    @available(*, deprecated, message: "CommandBarItemGroup labels are experimental.")
```
mischreiber, line 18: If this is deprecated, downstream consumers won't be able to use this API at all without disabling *all* deprecation errors. We can probably make do with a simple warning in its documentation comment (which, by the way, we should probably do anyway).

Sources/FluentUI_iOS/Components/Command Bar/CommandBar.swift
```diff
@@ -58,15 +58,15 @@ public class CommandBar: UIView, TokenizedControl {
                 return nil
             }
 
-            return [[leadingItem]]
+            return [CommandBarItemGroup(items: [leadingItem])]
```
mischreiber, line 61: TIL about [ExpressibleByArrayLiteral](https://developer.apple.com/documentation/swift/expressiblebyarrayliteral), which I *think* would mean we could make this change without breaking most uses of the API?

joannaquu: unfortunately only works with swift usages but still cleans up the changes we need in our demo controllers!

mischreiber (approved):

PR #2144, opened 2025-04-29
MSFNotification can set a closure to run regardless of the source of the disimissal

cbowdoin:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes
Added a new onDismiss closure that will be called whenever the Notification View disappears. This allows for consumers to ensure that some action will always be taken whenever the notification is dismissed regardless of what the source of the notification being dismissed is. 

### Binary change

### Verification
Tested with closures in demo app.
Created notificatiosn from both swift and objective c code

<details>

| Consumers of MSFNotifications could only have a dimiss block if they have a custom dismiss button or if they explicity call hide on the notification  | Consumers can define a closure that will be called everytime the notification view will disappear |
</details>

### Pull request checklist

This PR has considered:

### DIFF
--- a/Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
+++ b/Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
@@ -51,6 +51,9 @@ import SwiftUI
     /// Action to be dispatched by tapping on the toast/bar notification.
     var messageButtonAction: (() -> Void)? { get set }
 
+    /// The callback to execute when the notification is dismissed.
+    var onDismiss: (() -> Void)? { get set }
+
     /// Defines whether the notification shows from the bottom of the presenting view or the top.
     var showFromBottom: Bool { get set }
 
@@ -118,6 +121,7 @@ public struct FluentNotification: View, TokenizedControlView {
         self.state = state
         self.shouldSelfPresent = shouldSelfPresent
         self.isFlexibleWidthToast = isFlexibleWidthToast && style.isToast
+        self.onDismiss = nil
 
         self.tokenSet = NotificationTokenSet(style: { state.style })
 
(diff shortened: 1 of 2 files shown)


Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -118,6 +121,7 @@ public struct FluentNotification: View, TokenizedControlView {
         self.state = state
         self.shouldSelfPresent = shouldSelfPresent
         self.isFlexibleWidthToast = isFlexibleWidthToast && style.isToast
+        self.onDismiss = nil
```
mischreiber, line 124: Defaults to `nil`, so no need to explicitly call it out

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -335,7 +339,21 @@ public struct FluentNotification: View, TokenizedControlView {
             }
         }
 
-        return presentableNotification
+        @ViewBuilder
```
mischreiber, line 342: Any reason we can't skip the `@ViewBuilder` entirely by always returning
```swift
presentableNotification
    .onDisappear {
        state.onDismiss?()
    }
```

cbowdoin: We can absolutely do that. I assumed that was discouraged with the ViewBuilder structure of SwiftUI

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -393,6 +399,9 @@ public struct FluentNotification: View, TokenizedControlView {
     // When true, the notification will fit the size of its contents.
     // When false, the notification will be fixed based on the size of the screen.
     private let isFlexibleWidthToast: Bool
+
+    // The callback to execute when the notification is dismissed.
+    private let onDismiss: (() -> Void)? = nil
```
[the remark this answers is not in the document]
cbowdoin: Ya thats correct. This should be a var and not a let. Will make that change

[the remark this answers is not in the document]
cbowdoin: Oh yes my bad this onDismiss is extraneous and can be removed. We only need the one on the state and MSFNotification

mischreiber (approved):

PR #2147, opened 2025-05-05
Add custom height functionality to Bottomsheet VC

sharath273kumar:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Supporting custom height functionality for fluent Bottom Sheet Controller in Opensource project 

The custom height functionality will be an opt-in feature, which means any client consuming the bottomsheet can configure this as an option to enjoy using the functionality. 

Exposing customHeightState and customVelocity properties as objc variables, which should be set by caller for using custom height functionality 

CustomHeightState  

Type: Array of CGFloat, Optional 

Description: The array defines custom height intervals for the bottom sheet. These intervals allow the bottom sheet to snap to specific heights during user interactions, such as dragging or swiping. The array is expected to be sorted in ascending order. 

Default Value: If the array is empty ([]), the bottom sheet will fall back to its default behavior, which includes snapping to predefined states like .collapsed or .expanded. 

CustomVelocity 

Type: CGFloat Optional 

Description:  

It determines the minimum velocity required for the bottom sheet to snap to a specific state (e.g., collapsed, expanded, or custom) when the user releases the drag gesture. 

If the velocity of the gesture is below this threshold, the bottom sheet will snap to the nearest state based on its current position. 

If the velocity exceeds this threshold, the bottom sheet will snap to the state in the direction of the swipe. 

Default Value: The default value is 0, meaning the bottom sheet will always snap to the nearest state regardless of the swipe velocity. 

To Handle the Panning gesture for custom height expansion states we check of the bottomsheet is in custom height mode in HandlePane complete and handle in completeCustomPane() 

CustomCompletePan- The completeCustomPan function is responsible for determining the final state of the bottom sheet when a pan gesture (dragging motion) ends, specifically when custom expansion states are defined. It calculates whether the bottom sheet should snap to a specific custom height or transition to a default state (e.g., collapsed, expanded, or hidden) based on the gesture's velocity and the current position of the sheet. 

### Binary change

Negligible

### Verification

Tested with existing bottomsheet VCs

<details>

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2147)

### DIFF
--- a/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
+++ b/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
@@ -58,6 +58,7 @@ public protocol BottomSheetControllerDelegate: AnyObject {
     case collapsed // Sheet is collapsed
     case hidden // Sheet is hidden (fully off-screen)
     case transitioning // Sheet is between states, only used during user interaction / animation
+    case custom // Sheet is at a custom height
 }
 
 /// Defines where the sheet should be postionioned relative to the screen space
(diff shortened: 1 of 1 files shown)


sharath273kumar:
@microsoft-github-policy-service agree [company="Microsoft"]

sharath273kumar:

lukas-cap (changes requested):
Hi! Limiting the number of detents was a conscious design decision for the bottom sheet. With collapsed and expanded - we already have two detents, and I can't think of an example across iOS that requires more than that (and I'm concerned that wouldn't be good UX anyways).

Do you actually need another detent between collapsed and expanded, or is this more about customizability of the existing detents?

If this is more about customizability - We already have API for controlling the collapsed height - `collapsedContentHeight` and `collapsedHeightResolver` (the resolver being the more robust option). And for expanded height as well - `preferredExpandedContentHeight`.

Please let me know if those work for you. If not, let's work together to understand what the requirements are, and what we can do to accommodate.

PR #2149, opened 2025-05-07
Partial sheet expansion support

lukas-cap:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Adds support for a new partial expansion state to bottom sheet. Most of the diff is fairly simple, the main behavioral change is that on pan end, we don't just determine the target state based on direction + velocity, but also position to see if we go towards `.partial`, `.expanded`, or `.collapsed`.

The other bigger change is the addition of the `partialHeightResolver`, which is needed to pull fresh height from the client. It follows design of the existing `collapsedHeightResolver`. I chose to cache them together, because we usually pull them together as well.

Some semantic choices had to be made, like `isExpanded` still maps to `.expanded`, which also means tapping the resizing handle while in `.partial` state takes the sheet to `.expanded`. That choice is fairly arbitrary, given that `.partial` is in-between by definition.

To enable the partial state, we don't have a distinct bool - it's just inferred from the existence of the `partialHeightResolver`.

### Verification

Various manual testing in the demo app + sanity check with client team to early integrate and test E2E. Restricted height scenarios, landscape / portrait changes - verifying sheet frame recalculates to correct new calue.

<details>

https://github.com/user-attachments/assets/4c27c10f-371f-4919-ad50-8d429349b2f0

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2149)

### DIFF
--- a/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
+++ b/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
@@ -1165,7 +1165,7 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     // Height of the sheet in collapsed state
     private var collapsedSheetHeight: CGFloat {
         let safeAreaSheetHeight: CGFloat
-        if let dynamicHeight = resolvedDynamicSheetHeights?.collapsedHeight, dynamicHeight > 0  {
+        if let dynamicHeight = resolvedDynamicSheetHeights?.collapsedHeight, dynamicHeight > 0 {
             safeAreaSheetHeight = dynamicHeight
         } else if collapsedContentHeight > 0 {
             safeAreaSheetHeight = collapsedContentHeight + currentResizingHandleHeight
--- a/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
+++ b/Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
@@ -195,7 +195,7 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     /// Changes to this property are animated. A new value is reflected in the getter only after the animation completes.
     @objc open var isExpanded: Bool {
         get {
-            return currentExpansionState == .expanded || currentExpansionState == .partial
+            return currentExpansionState == .expanded
         }
         set {
             setIsExpanded(newValue)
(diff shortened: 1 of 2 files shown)


lukas-cap:

sharath273kumar (approved):

Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
```diff
@@ -1130,8 +1165,8 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     // Height of the sheet in collapsed state
     private var collapsedSheetHeight: CGFloat {
         let safeAreaSheetHeight: CGFloat
-        if resolvedCollapsedSheetHeight > 0 {
-            safeAreaSheetHeight = resolvedCollapsedSheetHeight
+        if let dynamicHeight = resolvedDynamicSheetHeights?.collapsedHeight, dynamicHeight > 0  {
```
nodes11, line 1168: super minor: extra space between `0` and `{`

nodes11 (approved):

Sources/FluentUI_iOS/Components/Bottom Sheet/BottomSheetController.swift
lukas-cap: Don't mind me, just making a suggestion so github gives me the nice button to make the change directly.

```suggestion
        if let dynamicHeight = resolvedDynamicSheetHeights?.collapsedHeight, dynamicHeight > 0 {
```

lukas-cap: it worked! Nice.

amoggo:
Don't we care to add some default value for the partial height if the client doesn't provide it in the partialHeightResolver? In which case, we'd probably need a bool opt in to supportsPartialHeight

lukas-cap:
> Don't we care to add some default value for the partial height if the client doesn't provide it in the partialHeightResolver? In which case, we'd probably need a bool opt in to supportsPartialHeight

The way it's designed now, it's a fully opt-in feature, so we shouldn't have cases where we're trying to resolve a height and the resolver isn't there. And if it's there but doesn't return a valid value - that's a client side bug.

We could have a separate bool and a default, but I'm leaning towards not adding that unless this becomes a heavily used feature. My expectation is that this will be used rarely, and often it's straight up not good UX to have that many detents.

nodes11 (approved):

amoggo (approved):

PR #2154, opened 2025-05-12
[FluentNotification] Enable showing an action button and a dismiss button

nodes11:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes

There is a use case where we need to be able to show an action button and a dismiss button on a single notification. Both show iff the `MSFNotificationState.showDefaultDismissActionButton ` is true, `actionButtonAction` is not `nil`,and `actionButtonTitle` is not `nil` and not empty.

To implement this:
- Add a `defaultDimissButtonAction` block property on the `MSFNotificationState`. If the value is not set by the consumer, we will give it a default value in the `MSFNotification` initializer.
- Create a `dismissButton` viewbuilder in the view implementation. This button will only show when the conditions in the first paragraph above are met.
- Add a call to add the dimiss button if the `dismissButtonAction` is non-nil. The code to determine the `dismissButtonAction` was broken out into its own private function so that it could be more easily called from two places in the view.

### Binary change

### Verification

- Validated the behavior in the Fluent Test App (see demo)

<details>

https://github.com/user-attachments/assets/77b750b2-6c64-4875-967b-769fd93053c7

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2154)

### DIFF
--- a/Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
+++ b/Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
@@ -48,7 +48,8 @@ import SwiftUI
     /// Bool to control if the Notification has a dismiss action by default.
     var showDefaultDismissActionButton: Bool { get set }
 
-	var defaultDimissButtonAction: (() -> Void)? { get set }
+    /// Action to be dispatched by the dismiss button on the trailing edge of the control.
+    var defaultDimissButtonAction: (() -> Void)? { get set }
 
     /// Action to be dispatched by tapping on the toast/bar notification.
     var messageButtonAction: (() -> Void)? { get set }
@@ -191,7 +192,7 @@ public struct FluentNotification: View, TokenizedControlView {
         @ViewBuilder
         var button: some View {
             let shouldHaveDefaultAction = state.showDefaultDismissActionButton && shouldSelfPresent
-            if let buttonAction = state.actionButtonAction ?? (shouldHaveDefaultAction ? dismissAnimated : nil) {
+            if let buttonAction = state.actionButtonAction ?? (shouldHaveDefaultAction ? (state.defaultDimissButtonAction ?? dismissAnimated) : nil) {
                 if let actionTitle = state.actionButtonTitle, !actionTitle.isEmpty {
                     SwiftUI.Button(actionTitle) {
                         isPresented = false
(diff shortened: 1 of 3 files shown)


Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -189,7 +192,7 @@ public struct FluentNotification: View, TokenizedControlView {
         @ViewBuilder
         var button: some View {
             let shouldHaveDefaultAction = state.showDefaultDismissActionButton && shouldSelfPresent
-            if let buttonAction = state.actionButtonAction ?? (shouldHaveDefaultAction ? dismissAnimated : nil) {
+            if let buttonAction = state.actionButtonAction ?? (shouldHaveDefaultAction ? (state.defaultDimissButtonAction ?? dismissAnimated) : nil) {
```
cbowdoin, line 195: Do we still need the shouldHaveDefaultAction check for the button if we have a dedicated builder for the dimiss button?

nodes11: I've refactored the view a bit so it is no longer required here. In the first revision, this `button` would become a dismiss button in some cases so the check was still required. I've updated it so that this `button` (now `actionButton`) will only be shown if there is an `actionButtonAction` set and there is a custom title or image to show. This button will never appear as a dismiss button with an `X`. That is now reserved for the `dismissButton`.

joannaquu:
@huwilkes do you remember the history behind why there's this complicated logic of if we show dismiss button, action button, which one gets linked to which action, etc? why was it not just declarative with action button, dismiss button

huwilkes:
> @huwilkes do you remember the history behind why there's this complicated logic of if we show dismiss button, action button, which one gets linked to which action, etc? why was it not just declarative with action button, dismiss button

The design doesn't have an option with two buttons.

cbowdoin (approved):

mischreiber:
> > @huwilkes do you remember the history behind why there's this complicated logic of if we show dismiss button, action button, which one gets linked to which action, etc? why was it not just declarative with action button, dismiss button
> 
> The design doesn't have an option with two buttons.

IIRC most of this logic was inherited from the previous maintainers of this codebase. But yeah like @huwilkes said, it was a safe assumption to say "if we have an action then show a button, and if we have a dismissal show an `×`".

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -101,6 +105,7 @@ public struct FluentNotification: View, TokenizedControlView {
                 actionButtonTitle: String? = nil,
                 actionButtonAction: (() -> Void)? = nil,
                 showDefaultDismissActionButton: Bool? = nil,
+                defaultDismissButtonAction: (() -> Void)? = nil,
```
mischreiber, line 108: Do we need to think about ay of the Objective-C/UIKit API for this, or are we good with adding it for SwiftUI only for now?

mischreiber (approved):

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
nodes11: Objective C callers should be able to configure this property through the `MSFNotificationState` property exposed on the `MSFNotification`.

nodes11: Resolving this for now so the change can be merged. Happy to continue this conversation and follow up with another change if we want to improve the API in the future.

nodes11:
The latest commit introduces the property `showActionButtonAndDismissButton`, which determines whether an action button and a dismiss button can be displayed simultaneously on a notification. Without this property, and with the new behavior applied, all toast notifications that do not explicitly set `showDefaultDismissActionButton` to `true` would start showing a dismiss icon. This would be an undesirable and unexpected UX change for clients. To opt into the new behavior—where both an action button and a dismiss button can be displayed at once—clients will need to explicitly set `showActionButtonAndDismissButton` to `true`.

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -101,6 +108,8 @@ public struct FluentNotification: View, TokenizedControlView {
                 actionButtonTitle: String? = nil,
                 actionButtonAction: (() -> Void)? = nil,
                 showDefaultDismissActionButton: Bool? = nil,
+                showActionButtonAndDismissButton: Bool = false,
```
mischreiber, line 111: No need to act on this PR, but noting that this class's `init` is getting a little out of hand in terms of arguments. Probably time to circle back on the idea of using a configuration struct of some sort, though we need to work out the details on that still.

mischreiber (approved):

cbowdoin (approved):

PR #2168, opened 2025-05-29
[macOS] Adding theme tokens

DanielMorM:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [X] macOS

### Description of changes

Adding an initial set of tokens for macOS.

### DIFF
--- a/Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
+++ b/Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
@@ -25,6 +25,7 @@ public extension FluentTheme {
         case background3Pressed
         case background3Selected
         case background4
+        case background4Hover
         case background4Pressed
         case background4Selected
         case background5
@@ -44,6 +45,7 @@ public extension FluentTheme {
         case foreground1
         case foreground2
         case foreground3
+        case foreground4
         case foregroundDisabled1
         case foregroundDisabled2
         case foregroundOnColor
@@ -194,7 +196,7 @@ extension FluentTheme {
         case .foreground2:
             color = .init(light: GlobalTokens.neutralSwiftUIColor(.grey38),
                           dark: GlobalTokens.neutralSwiftUIColor(.grey84))
-        case .foreground3:
+        case .foreground3, .foreground4:
             color = .init(light: GlobalTokens.neutralSwiftUIColor(.grey50),
                           dark: GlobalTokens.neutralSwiftUIColor(.grey68))
         case .foregroundDisabled1:
@@ -280,7 +282,7 @@ extension FluentTheme {
             color = .init(light: GlobalTokens.neutralSwiftUIColor(.grey92),
                           dark: GlobalTokens.neutralSwiftUIColor(.grey30),
                           darkElevated: GlobalTokens.neutralSwiftUIColor(.grey30))
-        case .background4:
+        case .background4, .background4Hover:
             color = .init(light: GlobalTokens.neutralSwiftUIColor(.grey98),
                           dark: GlobalTokens.neutralSwiftUIColor(.grey20),
                           darkElevated: GlobalTokens.neutralSwiftUIColor(.grey24))
@@ -473,8 +475,7 @@ extension FluentTheme {
         }
 
         // Allow overriding if there is a `PlatformThemeProviding` implementation.
-        if let platformThemeType = self as? PlatformThemeProviding.Type,
-           let platformColor = platformThemeType.platformColorValue(token, defaultColor: color) {
+        if let platformColor = (self as PlatformThemeProviding.Type).platformColorValue(token, defaultColor: color) {
             color = platformColor
         }
         return color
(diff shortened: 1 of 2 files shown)


Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
```diff
@@ -194,7 +196,7 @@ extension FluentTheme {
         case .foreground2:
             color = .init(light: GlobalTokens.neutralSwiftUIColor(.grey38),
                           dark: GlobalTokens.neutralSwiftUIColor(.grey84))
-        case .foreground3:
+        case .foreground3, .foreground4:
```
joannaquu, line 199: why introduce a new foreground instead of just using foreground3?

Sources/FluentUI_macOS/Theme/FluentTheme+AppKit.swift
```diff
@@ -27,7 +26,25 @@ public extension FluentTheme {
...
+        case .background2:
+            overrideColor = .init(light: GlobalTokens.neutralSwiftUIColor(.white), dark: GlobalTokens.neutralSwiftUIColor(.grey16))
+        case .background4:
+            overrideColor = .init(light: GlobalTokens.neutralSwiftUIColor(.grey96), dark: GlobalTokens.neutralSwiftUIColor(.grey8))
+        case .background4Hover:
+            overrideColor = .init(light: GlobalTokens.neutralSwiftUIColor(.grey92), dark: GlobalTokens.neutralSwiftUIColor(.grey12))
+        case .background6:
+            overrideColor = .init(light: GlobalTokens.neutralSwiftUIColor(.grey88), dark: GlobalTokens.neutralSwiftUIColor(.black))
+        case .brandStroke1:
+            overrideColor = .init(light: GlobalTokens.neutralSwiftUIColor(.white), dark: GlobalTokens.neutralSwiftUIColor(.grey16))
```
joannaquu, line 40: brand colors should always be app colors

Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
DanielMorM: These tokens are new for macOS and are not included in the FluentUI-iOS specification, but they are part of the macOS specification. Therefore, I am not setting a default value and am delegating to the override logic so that macOS can set the correct values.

joannaquu: ahhh gotcha, this'll be more of an architecture question for @mischreiber but in that case i wonder if we should just have no defaults and have iOS also provide platform values so it's clear which platforms support which values

mischreiber: That’s interesting. I was under the impression that we would have the same token names across these platforms, just with different values. Let me do some exploring too, but @joannaquu’s approach may be best

Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
```diff
@@ -120,6 +120,12 @@ public extension FluentTheme {
         case presenceDnd
         case presenceAvailable
         case presenceOof
+
+        // MacOS Only Tokens
+        @available(iOS, unavailable)
+        @available(visionOS, unavailable)
+        @available(macCatalyst, unavailable)
+        case background4Hover, foreground4
```
mischreiber, line 128: Nit: could you leave this in place with the other `background4` tokens? As we add more tokens that are platform-specific, I think it'll be easier to keep them organized by category.

Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
```diff
@@ -182,6 +188,14 @@ public extension FluentTheme {
     func gradient(_ token: GradientToken) -> [Color] {
         return gradientTokenSet[token].map { Color(dynamicColor: $0) }
     }
+
+    @available(iOS, unavailable)
+    @available(visionOS, unavailable)
+    @available(macCatalyst, unavailable)
+    private enum MacOSOnlyColorToken: Int, TokenSetKey {
```
mischreiber, line 195: What is this enum for?

DanielMorM: My initial idea was to avoid repeating the "@available" checks for each token that required them. However, as you suggested, it would be better to group them. Therefore, I have removed the grouped checks and opted to add each token individually.

Sources/FluentUI_iOS/Core/Theme/FluentTheme+iOS.swift
```diff
@@ -357,4 +356,3 @@ extension FluentTheme: PlatformThemeProviding {
         }
     }
 }
-
```
joannaquu, line 360: nit: can we restore new line

joannaquu (approved):

Sources/FluentUI_common/Core/Theme/FluentTheme+Tokens.swift
```diff
@@ -25,6 +25,12 @@ public extension FluentTheme {
         case background3Pressed
         case background3Selected
         case background4
+
+        @available(iOS, unavailable)
+        @available(visionOS, unavailable)
+        @available(macCatalyst, unavailable)
```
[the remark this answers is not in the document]
DanielMorM: I tried that but compiler complains with the following error:
`Expected 'available' option such as 'unavailable', 'introduced', 'deprecated', 'obsoleted', 'message', or 'renamed'`

mischreiber (approved):

PR #2193, opened 2025-07-22
Introduce an API to the Fluent Notification that draws attention to the notification

cbowdoin:
### Platforms Impacted
- [X] iOS
- [x] visionOS

### Description of changes
Added in a bump API to the notification that preforms a bouncy animation to highlight the notification while it is presented. This is done using 2 spring animations. We animate to some offset and then animated using a spring animation back to the origin point with enouch velocity that it animates past the point.

Implemented a demo of this in the demo controller

### Verification
Validated using the demo controller that the animation runs an looks smooth without causing issues to the dismiss animation. 

<details>

| Did not exist | Simulator Screen Recording - iPad Pro 13-inch (M4) - 2025-07-22 at 15 52 24 |
</details>

### Pull request checklist

This PR has considered:
- [X] Objective-C exposure 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2193)

### DIFF
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -245,15 +245,12 @@ class NotificationViewDemoController: DemoController {
             notification.state.message = "Tap the buttons below to test bump animations"
             notification.state.title = "Bump Demo"
             notification.state.image = UIImage(named: "play-in-circle-24x24")
-            notification.state.actionButtonTitle = "Gentle"
+            notification.state.actionButtonTitle = "Bump"
+
             notification.state.actionButtonAction = { [weak notification] in
-                // Gentle bump: low intensity, longer duration
-                notification?.bumpWithIntensity(0.3, duration: 1.0, useDecreasingCurve: true)
+                notification?.bump()
             }
-            notification.state.messageButtonAction = { [weak notification] in
-                // Strong bump: high intensity, shorter duration
-                notification?.bumpWithIntensity(1.0, duration: 0.8, useDecreasingCurve: false)
-            }
+
             return notification
         }
     }
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -25,7 +25,7 @@ class NotificationViewDemoController: DemoController {
         case neutralToastWithOverriddenTokens
         case neutralToastWithGradientBackground
         case warningToastWithFlexibleWidth
-        case bumpDemo // Add this new case
+        case bumpDemo
 
         var displayText: String {
             switch self {
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
@@ -25,6 +25,7 @@ class NotificationViewDemoController: DemoController {
         case neutralToastWithOverriddenTokens
         case neutralToastWithGradientBackground
         case warningToastWithFlexibleWidth
+        case bumpDemo // Add this new case
 
         var displayText: String {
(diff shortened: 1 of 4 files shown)


Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -67,6 +67,10 @@ import SwiftUI
     ///
     /// If this property is nil, then this notification will use the background color defined by its design tokens.
     var backgroundGradient: LinearGradientInfo? { get set }
+
+    /// Performs an animation emphasizing the notification.
+    /// The animation alternates between upward and downward movements with spring physics.
+    @objc func bump()
```
nodes11, line 73: minor: This protocol is marked as `@objc` so I don't think it's needed here as well

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -323,6 +327,13 @@ public struct FluentNotification: View, TokenizedControlView {
 #if os(visionOS)
                 .glassBackgroundEffect(in: RoundedRectangle(cornerRadius: tokenSet[.cornerRadius].float))
 #endif // os(visionOS)
+                .offset(y: -bumpVerticalOffset)
+                .onChange_iOS17(of: state.shouldPerformBump) { shouldBump in
```
nodes11, line 331: Just to be sure, there is no way we can use the SwiftUI `animation` modifier to achieve this since we need the animation to be continously moving up and down?

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -549,4 +577,10 @@ class MSFNotificationStateImpl: ControlState, MSFNotificationState {
 
         super.init()
     }
+
+    @objc func bump() {
```
mischreiber, line 581: Do we need both this function and the `shouldPerformBump` variable?

Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/NotificationViewDemoController.swift
```diff
@@ -236,6 +239,18 @@ class NotificationViewDemoController: DemoController {
                 self?.showMessage("`Dismiss` tapped")
                 notification.hide()
             }
+            return notification
+        case .bumpDemo:
+            let notification = MSFNotification(style: .primaryToast)
+            notification.state.message = "Tap the buttons below to test bump animations"
```
joannaquu, line 245: nit: no button below, tap the action button?

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -442,6 +453,19 @@ public struct FluentNotification: View, TokenizedControlView {
         }
     }
 
+    private func preformBumpAnimated() {
+        // Use the dedicated bump offset instead of bottomOffset
+        withAnimation(.interpolatingSpring(stiffness: state.style.animationSpringStiffness, damping: state.style.animationDampingForBump)) {
+            bumpVerticalOffset = state.style.offsetDistanceForBumpAnimation
+        }
+
+        DispatchQueue.main.asyncAfter(deadline: .now() + state.style.bumpAnimationDelay) {
```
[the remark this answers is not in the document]
cbowdoin: 1st animation is to move the toast into an offset position. The 2nd animation is to start the osicaltion effect we want to show here.

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -442,6 +453,20 @@ public struct FluentNotification: View, TokenizedControlView {
         }
     }
 
+    private func preformBumpAnimated() {
```
joannaquu, line 456: nit: performBumpAnimated

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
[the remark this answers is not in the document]
joannaquu: ah I see there's no completion handler when using `withAnimation`. is the issue that the oscillation would happen too early if we did it in the same animation block?

PR #2194, opened 2025-07-30
Updating bottom sheet corner radius

JaquelineGomez:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

- Added a new corner radius token
- Changed the corner radius of the BottomSheets to the new corner radius only if they are using the new iOS version that will be released soon
- This new corner radius makes the BottomSheets more consistent with Apple styling

### Verification

Ran it on two iPhone simulators:
1. Running on iOS 18
2. Running on iOS 26

Ran it on two iPad simulators:
1. Running on iOS 18
2. Running on iOS 26

The change was only seen on the iPhone and iPad running on iOS26.

<details>

| image| image |
| image| image|
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2194)

### DIFF
--- a/Sources/FluentUI_iOS/Components/BottomSheet/BottomSheetTokenSet.swift
+++ b/Sources/FluentUI_iOS/Components/BottomSheet/BottomSheetTokenSet.swift
@@ -32,7 +32,7 @@ public class BottomSheetTokenSet: ControlTokenSet<BottomSheetToken> {
                 }
             case .cornerRadius:
                 if #available(iOS 19, *) {
-                    return .float { BottomSheetTokenSet.cornerRadius400 }
+                    return .float { BottomSheetTokenSet.cornerRadius }
                 } else {
                     return .float { GlobalTokens.corner(.radius120) }
                 }
@@ -51,7 +51,7 @@ extension BottomSheetTokenSet {
     static let blurEffectShadowOpacity: Float = 0.25
     static let blurEffectShadowOffset: CGSize = CGSize(width: 0, height: -2)
     static let blurEffectShadowRadius: CGFloat = 8
-    static let cornerRadius400: CGFloat = 40
+    static let cornerRadius: CGFloat = 40
 }
 
 // MARK: - BottomSheetControllerStyle


JaquelineGomez:

Sources/FluentUI_common/Core/Theme/Tokens/GlobalTokens.swift
```diff
@@ -1796,6 +1796,7 @@ public class GlobalTokens: NSObject {
         case radius60
         case radius80
         case radius120
+        case radius400
```
anandrajeswaran, line 1799: Would love to get input from @joannaquu and @mischreiber here - naively I would want to avoid having to add something to GlobalTokens (cost/complexity to every user of any controls) which is only used in the BottomSheet control.

But 2 potential considerations which could counter:
1. Do we expect to use this in more places as we adapt more of the design to iOS 26 and this is just the first step?
2. Is there something about the tokens system where forcing these to be added globally is better than just defining values in each TokenSet?

mischreiber: Agreed that this is not something that should live in the GlobalTokens. It's not currently part of the Fluent design language, and shouldn't be added here unless we decide to update that. Instead, let's store the constant in the `BottomSheetTokenSet` for now and reevaluate as we implement OS26 support.

joannaquu: tbh I never super understood why we had global tokens for ints. maybe just consistency? but I'll defer to Mike

joannaquu: oops he's already responded

Sources/FluentUI_iOS/Components/BottomSheet/BottomSheetTokenSet.swift
```diff
@@ -47,6 +51,7 @@ extension BottomSheetTokenSet {
     static let blurEffectShadowOpacity: Float = 0.25
     static let blurEffectShadowOffset: CGSize = CGSize(width: 0, height: -2)
     static let blurEffectShadowRadius: CGFloat = 8
+    static let cornerRadius400: CGFloat = 40
```
joannaquu, line 54: nit: don't need 400 anymore

joannaquu (approved):

joannaquu (approved):

PR #2196, opened 2025-08-13
Rename common method to not sound iOS-specific

eschwieb:
The "onChange_iOS17" method is in common code and we'd like to use it in macOS components. Rename it to be non-iOS-specific.

### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

Renamed onChange_iOS17 to onChange_Compatibility.

### Binary change

Trivial increase in size based on the method name.

### Verification

Verified builds.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2196)

### DIFF
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/DemoAppearanceMenu.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/DemoAppearanceMenu.swift
@@ -68,22 +68,22 @@ struct DemoAppearanceMenu: View {
         })
 
         // Changes
-        .onChange_iOS17(of: configuration.userInterfaceStyle) { newValue in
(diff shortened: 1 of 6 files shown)


joannaquu:
fix should be a simple find+replace but I'd double check with @mischreiber which one of clients this change breaks

joannaquu (approved):

anandrajeswaran:
> fix should be a simple find+replace but I'd double check with @mischreiber which one of our clients this change breaks

Curious on bundle size impact if we just leave the old one around and have it call into the new one temporarily. If small, that enables everyone to update on their own and then can delete the old version. Makes a one-step process a two-step process, but may be worth it given diamond dependency chain

PR #2188, opened 2025-06-19
[Draft] Add SwiftUI PillButton

laminesm:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

(a summary of the changes made, often organized by file)

### Binary change

(how is our binary size impacted -- see https://github.com/microsoft/fluentui-apple/wiki/Size-Comparison)

### Verification

(how the change was tested, including both manual and automated tests)

<details>
<summary>Visual Verification</summary>

| Before                                       | After                                      |
|----------------------------------------------|--------------------------------------------|
| Screenshot or description before this change | Screenshot or description with this change |
</details>

### Pull request checklist

This PR has considered:
- [ ] Light and Dark appearances
- [ ] iOS supported versions (all major versions greater than or equal current target deployment version)
- [ ] VoiceOver and Keyboard Accessibility
- [ ] Internationalization and Right to Left layouts
- [ ] Different resolutions (1x, 2x, 3x)
- [ ] Size classes and window sizes (iPhone vs iPad, notched devices, multitasking, different window sizes, etc)
- [ ] iPad [Pointer interaction](https://developer.apple.com/documentation/uikit/pointer_interactions)
- [ ] [SwiftUI](https://developer.apple.com/tutorials/swiftui) consumption (validation or new demo scenarios needed)
- [ ] Objective-C exposure (provide it only if needed)

PR #2197, opened 2025-08-20
Add SwiftUI PillButtonBarView

laminesm:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [ ] macOS

### Description of changes

This PR adds a SwiftUI pill button bar.

This implementation uses generic binding to allow clients to set up pill selection with a `Hashable` `Selection` type that they provide to the `PillButtonBarView`. Each individual's pill button will have a matching `PillButtonViewModel` which will provide their respective `Selection` value used by the bar.

Moreover, this implementation supports 2 types of pill button bars:

1 - The traditional pill button bar that is restricted to 1 selected pill button at all times.
2 - A pill button bar that supports no selected pill button and a maximum of 1. In other words, if the user taps on the currently selected pill button it will be deselected (client request).

To best support this,  I opted to have two separate initializers to help distinguish between the two types of pill button bars: one initializer that takes a non-optional `Selection` binding for the traditional pill button bar experience, and the other initializer taking an optional `Selection` binding for pill bars that support no selected pills. Internally, we use an optional `Selection` property and a bool (determining which kind of pill bar this is) to correctly set up and change pill selection.

This PR also reverts changes to the UIKit pill button made in #2188 as it was causing a bug on the UIKit pill button bar where it wouldn't show the unread dot. I've investigated the issue quite a bit and could accurately pinpoint what was happening but it seems related to some obscure logic in the bar where the buttons get created and destroyed during layout phases. Decided to simply revert as that control is very fragile and has been a source of headaches in the past with minor changes.

### Binary change

Between 80 and 100KB binary impact.

### Verification

Visually verified light/dark and all platforms. Disabled state doesn't look obvious on vision dark mode (though I'm not familiar with the platform), so we might need different disabled colors.

<details>

| iPhone                                       | Vision                                      |
| iphone_pill_selection | vision_pill_selection |
| iphone_dark | vision_light |
| iphone_custom_theme | vision_custom_theme |
| iphone_token_overrides | vision_token_overrides |
| iphone_disabled | vision_disabled |

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2197)

### DIFF
--- a/Sources/FluentUI_iOS/Components/PillButtonBar/SwiftUI/PillButtonBarView.swift
+++ b/Sources/FluentUI_iOS/Components/PillButtonBar/SwiftUI/PillButtonBarView.swift
@@ -11,7 +11,7 @@ import SwiftUI
 /// `PillButtonBarView` is a horizontal scrollable list of pill shape text buttons.
 /// The bar can either have 1 selected pill at all times or allow no selected pills.
 public struct PillButtonBarView<Selection: Hashable>: View {
-    /// Initializes a new `PillButtonBarView` that allows a the bar to have no selected pill buttons and a maximum of 1 selected pill button.
+    /// Initializes a new `PillButtonBarView` that allows the bar to have no selected pill buttons and a maximum of 1 selected pill button.
     ///
     /// Use this initializer only if you want to allow the pill button bar to have no selected pills. In other words
     /// selecting an already selected pill will deselect it.


Sources/FluentUI_iOS/Components/Pill Button Bar/SwiftUI/PillButtonBarView.swift
```diff
@@ -0,0 +1,235 @@
...
+        self._selected = selected
+        self.centerAlignIfContentFits = centerAlignIfContentFits
+        self.supportsPillDeselection = true
+        self.tokenOverrides = tokenOverrides
+    }
+
+    /// Initializes a new `PillButtonBarView` that restricts the bar to 1 selected pill button at all times.
+    /// - Parameters:
+    ///   - style: The style of the pill buttons.
+    ///   - viewModels: The view model objects representing the pill buttons data.
```
nabil-k, line 40: Instead of using view models for the pill buttons have we considered doing something similar to apple's tab view: https://developer.apple.com/documentation/swiftui/tabview

Where we pass in PillButton views with their specifici information (title, image, etc.)

Sources/FluentUI_iOS/Components/Pill Button Bar/SwiftUI/PillButtonBarView.swift
```diff
@@ -0,0 +1,235 @@
...
+        } action: { newWidth in
+            contentWidth = newWidth
+        }
+        .padding(.vertical, Constants.verticalPadding)
+        .padding(.horizontal, Constants.horizontalPadding)
+        .frame(minWidth: disableScroll ? max(0, containerWidth - Constants.horizontalPadding * 2) : 0,
+               alignment: shouldCenterAlign ? .center : .leading)
+    }
+
+    private func pillButton(for item: PillButtonViewModel<Selection>) -> some View {
```
nabil-k, line 128: nit: since we're passing a view model should we just rename `item` to `viewModel`

Sources/FluentUI_iOS/Components/Pill Button Bar/SwiftUI/PillButtonBarView.swift
```diff
@@ -0,0 +1,235 @@
...
+            .environment(\.isEnabled, true)
+            .scrollDisabled(disableScroll)
+            .onGeometryChange(for: CGRect.self) { proxy in
+                proxy.frame(in: .global)
+            } action: { newFrame in
+                scrollViewFrame = newFrame
+                containerWidth = newFrame.width
+            }
+            .onAppear {
+                DispatchQueue.main.async {
```
anandrajeswaran, line 79: Wherever we use a magic dispatch async, should add a comment explaining why we can't do something more deterministic in terms of timing. But natural question - can we do something more deterministic? What happens if we just do this directly?

laminesm: Since we're calling this inside onAppear, we defer to the next layout pass here to make sure scrollTo runs when scroll view's layout is fully set up. It's not strictly needed and more of a safety mechanism. I can add a comment for it.

laminesm: Nvm the new APIs are only available on iOS 17+. I can add a comment here, and a TODO to migrate to the newer APIs once iOS support is dropped in fluent.

laminesm: We could, though I don't think we should since the current API (ScrollViewReader) is wrapped around the ScrollView so we might end up with large chunks of code behind the availability checks. This would also make it harder to reason about this code if we decide to add liquid glass support which would also be behind availability checks and result in duplicated chunks of code. I think considering fluent should be dropping iOS 16 supports relatively soon we're better off keeping the current API until that happens for better code clarity. Please let me know if that makes sense, otherwise I can go ahead and make the change.

laminesm: With fluent having dropped iOS 16 support, I've migrated the APIs thanks!

AllieStoyanova (approved):

PR #2202, opened 2025-09-04
[macOS] Improving Objective-C support on macOS

mischreiber:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [x] macOS

### Description of changes

This change is aimed to help AppKit-heavy apps better utilize FluentUI.

Changes:
* Add Objective-C support to as much of `GlobalTokens` as possible.
* Improve `NSColor` support for existing tokens.
* Add built-in `NSShadow` support for our `ShadowInfo` class.

### Verification

Tested locally in an Objective-C app

### Pull request checklist

This PR has considered:

### DIFF
--- a/Sources/FluentUI_macOS/Theme/Tokens/ShadowInfo+AppKit.swift
+++ b/Sources/FluentUI_macOS/Theme/Tokens/ShadowInfo+AppKit.swift
@@ -1,0 +1,22 @@
+//
+//  Copyright (c) Microsoft Corporation. All rights reserved.
+//  Licensed under the MIT License.
+//
+
+#if canImport(FluentUI_common)
+import FluentUI_common
+#endif
+import SwiftUI
+import AppKit
+
+public extension ShadowInfo {
+    /// Creates an NSShadow from the `key` elements of this instance.
+    @objc var keyShadow: NSShadow {
+        let dropShadow = NSShadow()
+        dropShadow.shadowColor = NSColor(keyColor)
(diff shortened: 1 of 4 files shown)


joannaquu (approved):

Sources/FluentUI_macOS/Theme/Tokens/ShadowInfo+AppKit.swift
```diff
@@ -0,0 +1,32 @@
...
+//  Licensed under the MIT License.
+//
+
+#if canImport(FluentUI_common)
+import FluentUI_common
+#endif
+import SwiftUI
+import AppKit
+
+public extension ShadowInfo {
```
anandrajeswaran, line 12: I'll add my obligatory "can we do this without extensions"? It's an extension of something in the same module which is less risky, but still adds some indirection. In this case, what if it was something like `ShadowFactoryAppKit` which creates NSShadows based on passed-in `ShadowInfo` objects?

mischreiber: While that's very valid for Objective-C, and doubly-so when crossing modules, we heavily utilize same-module anonymous extensions in FluentUI's Swift code for organizing our code. This aligns well with Apple's own practices for SwiftUI, which extensively relies on extensions (pun slightly intended). This also reduces our public footprint and binary size by avoiding additional public factory classes.

PR #2215, opened 2025-10-01
Added layout guide option to notification presentation

cbowdoin:
### Platforms Impacted
- [x] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
Added the ability to present notifications relative to a layout guide when there is no anchor view provided. 

### Verification
Verified by creating a layout guide in the test app to present the notification higher than it normally would present.

### Pull request checklist

This PR has considered:

### DIFF
--- a/Sources/FluentUI_iOS/Components/Notification/MSFNotification.swift
+++ b/Sources/FluentUI_iOS/Components/Notification/MSFNotification.swift
@@ -53,6 +53,13 @@ import UIKit
         return notification.tokenSet
     }
 
+    @objc public func show(in view: UIView,
+                           from anchorView: UIView? = nil,
+                           animated: Bool = true,
+                           completion: ((MSFNotification) -> Void)? = nil) {
+        show(in: view, from: anchorView, with:nil, animated: animated, completion: completion);
+    }
+
     // MARK: - Show/Hide Methods
     /// `show` is used to present the view inside a container view:
     /// insert into layout and show with optional animation. Constraints are used for the view positioning.
@@ -66,6 +73,7 @@ import UIKit
     ///   Can be used to call `hide` with a delay.
     @objc public func show(in view: UIView,
                            from anchorView: UIView? = nil,
+                           with layoutGuide: UILayoutGuide? = nil,
                            animated: Bool = true,
                            completion: ((MSFNotification) -> Void)? = nil) {
         guard self.window == nil else {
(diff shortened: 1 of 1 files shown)


nodes11:
Can an example of a notification using the `layoutGuide` parameter be added to the demo controller.

Sources/FluentUI_iOS/Components/Notification/MSFNotification.swift
```diff
@@ -61,11 +79,13 @@ import UIKit
     ///   - anchorView: The view used as the bottom anchor for presentation
     ///   (notification view is always presented up from the anchor). When no anchor view is provided the
     ///   bottom anchor of the container's safe area is used.
+    ///  - layoutGuide: The layout guide used to present the toast inside of when no `anchorView` is provided
     ///   - animated: Indicates whether to use animation during presentation or not.
     ///   - completion: The closure to be called after presentation is completed.
     ///   Can be used to call `hide` with a delay.
     @objc public func show(in view: UIView,
                            from anchorView: UIView? = nil,
+                           with layoutGuide: UILayoutGuide? = nil,
```
anandrajeswaran, line 88: minor: preference - I've tended away from using argument labels (i.e. `with`). For some small cases it helps make things more natural, but especially in functions like this with a lot of parameters, it tends to remove some clarity in the caller in how the parameter is being used

Sources/FluentUI_iOS/Components/Notification/MSFNotification.swift
```diff
@@ -61,11 +79,13 @@ import UIKit
     ///   - anchorView: The view used as the bottom anchor for presentation
     ///   (notification view is always presented up from the anchor). When no anchor view is provided the
     ///   bottom anchor of the container's safe area is used.
+    ///  - layoutGuide: The layout guide used to present the toast inside of when no `anchorView` is provided
```
anandrajeswaran, line 82: Can we describe "how" the layout guide is used? Which part of the notification is aligned with the layout guide? And probably worth explaining what happens if both an `anchorView` and a `layoutGuide` are provided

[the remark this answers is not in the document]
anandrajeswaran: I think objC just ignores default parameters and forces us to add a value. I'm personally mixed as to whether that mismatch has enough practical badness to cause us to avoid it iff we feel there is substantial benefit to swift consumers of the API

Sources/FluentUI_iOS/Components/Notification/MSFNotification.swift
mischreiber: Agreed; I'd like to redo the MSFNotification API entirely one of these days (for lots of reasons, but this is one of them).

PR #2265, opened 2026-04-16
Make CCB Truncate Button Text Before  Trying to Squish Buttons

cbowdoin:
### Platforms Impacted
- [ ] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
This change makes it so titles in the CommandBar Buttons truncate before we start having to layout out buttons off the screen when the CommandBar is in a non scrollable mode.
To do this we set the content hugging priority of button with texts to be low leading to them getting truncated before buttons that do not have the capacity to be squished.

### Binary change

### Verification
Scrolling CommandBar had no behavioral changes.
Tested with a resizable view to see titles with text getting shrinked before other buttons.

<details>
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2265)

### DIFF
--- a/Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
+++ b/Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
@@ -106,6 +106,10 @@ class CommandBarButton: UIButton {
         configuration?.title = item.title
         configuration?.image = item.iconImage
 
+        if item.title != nil {
+            setContentHuggingPriority(.defaultLow, for: .horizontal)
+        }
+
         if let font = item.titleFont {
             let attributeContainer = AttributeContainer([NSAttributedString.Key.font: font])
             configuration?.attributedTitle?.setAttributes(attributeContainer)


joannaquu:
can we revert the max width change now?

joannaquu (approved):

Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
```diff
@@ -106,6 +106,10 @@ class CommandBarButton: UIButton {
         configuration?.title = item.title
         configuration?.image = item.iconImage
 
+        if item.title != nil {
```
nodes11, line 109: minor sanity check: there is not a case where the title changes and we need to restore the `huggingPriority` to the previous or default value right?

nodes11 (approved):

cbowdoin:
> can we revert the max width change now?

We may still need it around for scenario where we want to truncate a scrolling CommandBar

Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
cbowdoin: Clients should never modify this value directly so it should be safe.

PR #2267, opened 2026-04-20
[CommandBar] Add Glass Styling Support

cbowdoin:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes
Introduces a new CommmandBarStyle  .primary and .glass  to style the CommandBar using UIGlassEffect on iOS26 and UIBlurEffect on older OS. The styling also adjust the shadows to match the glass style.

Change also includes a minor bug fix to the corner radius calculation so we get full half circles in command bars that try and layout below the safe area which previously could be seen in the DemoController Accessaory Input View when the soft keyboard was not showing

### Binary change

### Verification

<details>

| image | image  |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2267)

### DIFF
--- a/Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
+++ b/Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
@@ -78,10 +78,21 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
                   trailingItemGroups: trailingItems)
     }
 
+    @objc public convenience init(itemGroups: [CommandBarItemGroup],
+                                  leadingItemGroups: [CommandBarItemGroup]? = nil,
+                                  trailingItemGroups: [CommandBarItemGroup]? = nil) {
+        self.init(itemGroups: itemGroups,
+                  leadingItemGroups: leadingItemGroups,
+                  trailingItemGroups: trailingItemGroups,
+                  style: .primary)
+    }
+
     @objc public init(itemGroups: [CommandBarItemGroup],
                       leadingItemGroups: [CommandBarItemGroup]? = nil,
-                      trailingItemGroups: [CommandBarItemGroup]? = nil) {
-        self.tokenSet = CommandBarTokenSet()
+                      trailingItemGroups: [CommandBarItemGroup]? = nil,
+                      style: CommandBarStyle) {
+        self.style = style
+        self.tokenSet = CommandBarTokenSet(style: { style })
 
         leadingCommandGroupsView = CommandBarCommandGroupsView(itemGroups: leadingItemGroups,
                                                                buttonsPersistSelection: false,
@@ -104,6 +115,9 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
         super.init(frame: .zero)
 
         configureHierarchy()
+        if style == .glass {
+            setupGlassBackground()
+        }
         updateBackgroundColor()
         updateShadow()
 
@@ -159,6 +173,7 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
 
         updateShadow()
         updateScrollViewShadow()
+        updateCornerRadius()
     }
 
 #if DEBUG
@@ -190,6 +205,9 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
 
     public typealias TokenSetKeyType = CommandBarTokenSet.Tokens
     public var tokenSet: CommandBarTokenSet
+
+    /// The visual style of the CommandBar.
+    public let style: CommandBarStyle
 
     /// Items shown in the center of the CommandBar
     @objc public var itemGroups: [CommandBarItemGroup] {
@@ -245,6 +263,8 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
     public weak var delegate: CommandBarDelegate?
 
     // MARK: - Private properties
+
+    private var glassEffectView: UIVisualEffectView?
 
     /// Container UIStackView that holds the leading, main and trailing views
     private var commandBarContainerStackView: UIStackView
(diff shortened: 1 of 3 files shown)


Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
```diff
@@ -424,12 +446,32 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
...
 
     private func updateBackgroundColor() {
-        backgroundColor = tokenSet[.backgroundColor].uiColor
+        switch style {
+        case .primary:
+            backgroundColor = tokenSet[.backgroundColor].uiColor
+        case .glass:
+            backgroundColor = .clear
+#if !os(visionOS)
+            if #available(iOS 26, *), let effectView = glassEffectView {
```
mischreiber: Nit: we've been optimizing the `let effectView = glassEffectView` to the simpler `let glassEffectView`

Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
```diff
@@ -438,6 +480,59 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
...
+            effectView.bottomAnchor.constraint(equalTo: commandBarContainerStackView.bottomAnchor),
+            effectView.trailingAnchor.constraint(equalTo: commandBarContainerStackView.trailingAnchor)
+        ])
+        commandBarContainerStackView.backgroundColor = .clear
+        glassEffectView = effectView
+    }
+
+#if !os(visionOS)
+    @available(iOS 26, *)
+    private func getGlassEffectView() -> UIVisualEffectView {
```
mischreiber: More naming nits: we're maintaining the old Objective-C convention of avoiding the word "get" as a prefix for a method. Especially in these cases, where they aren't fetching a value but are creating something. Maybe start with "make" or something similar?

mischreiber: Actually a second point, we could simplify by merging these functions together, and using the `#if os(visionOS) / #else / #endif` logic from `setupGlassBackground()` inside it to decide whether to use a `UIGlassEffect` or `UIBlurEffect`.

Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
```diff
@@ -152,13 +168,14 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
...
 
-        let cornerRadius = bounds.height / 2
+        let cornerRadius = commandBarContainerStackView.bounds.height / 2
         layer.cornerRadius = cornerRadius
         commandBarContainerStackView.layer.cornerRadius = cornerRadius
         commandBarContainerStackView.layoutIfNeeded()
 
         updateShadow()
         updateScrollViewShadow()
+        updateCornerRadius()
```
mischreiber, line 170: I was a little confused here with this name, given that we also set the corner radius above. I'm thinking we could either move that logic into this function, or rename it to be explicitly about updating the effect view's corner radius.

It also has a duplicate calculation of the corner radius from a few lines above. That should either be part of the method if they're merged, or passed as an argument if not.

Sources/FluentUI_iOS/Components/CommandBar/CommandBarTokenSet.swift
```diff
@@ -107,6 +124,16 @@ public class CommandBarTokenSet: ControlTokenSet<CommandBarToken> {
...
     }
+
+    var style: () -> CommandBarStyle
+}
+
+// MARK: Constants
+extension CommandBarTokenSet {
+    static let glassEffectShadowColor: CGColor = UIColor.black.cgColor
+    static let glassEffectShadowOpacity: Float = 0.25
+    static let glassEffectShadowOffset: CGSize = CGSize(width: 0, height: -2)
```
mischreiber: It feels in the screenshots that the shadow is facing the wrong direction--it's centered higher than the midpoint of the bar, which makes it feel like the light source is coming from below. But we always want light sources coming from above, so maybe this height should be 2 instead of -2?

Sources/FluentUI_iOS/Components/CommandBar/CommandBarTokenSet.swift
```diff
@@ -107,6 +124,16 @@ public class CommandBarTokenSet: ControlTokenSet<CommandBarToken> {
             }
         }
     }
+
+    var style: () -> CommandBarStyle
+}
+
+// MARK: Constants
+extension CommandBarTokenSet {
+    static let glassEffectShadowColor: CGColor = UIColor.black.cgColor
```
joannaquu, line 133: shouldn't we pull shadow values from the shadow token? also does uikit glass have a default shadow? if so probably should use the default shadow

Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
```diff
@@ -438,6 +480,59 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
...
+        effectView = getBlurEffectView()
+#else
+        if #available(iOS 26, *) {
+            effectView = getGlassEffectView()
+        } else {
+            effectView = getBlurEffectView()
+        }
+#endif
+        effectView.translatesAutoresizingMaskIntoConstraints = false
+        commandBarContainerStackView.insertSubview(effectView, at: 0)
```
joannaquu: this is backwards, we should be adding our view to the effectView's content view

Sources/FluentUI_iOS/Components/CommandBar/CommandBarTokenSet.swift
cbowdoin: Updated to not prescribe our shadow when we are using glass. I kept these tokens around through so on older iOS our "fake glass" is consistent between here and the BottomSheet.

mischreiber (approved):

Sources/FluentUI_iOS/Components/CommandBar/CommandBar.swift
```diff
@@ -104,6 +115,11 @@ public class CommandBar: UIView, Shadowable, TokenizedControl {
         super.init(frame: .zero)
 
         configureHierarchy()
+
+        if style == .glass {
+            setupGlassBackground()
```
joannaquu: nit: can we move this into configureHierarchy so all the constraint setup is done in the same place. also allows us to do a switch case on the style so we account for all future potential styles as well

joannaquu: also it'll help you address the case you have right now where commandBarContainerStackView gets added as a subview twice

joannaquu (approved):

PR #2268, opened 2026-04-29
Fix Command Bar Title Compression

cbowdoin:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
Fixed truncation logic in CommandBar by switching from low contentHuggingPriority to low contentCompressionResistancePriority. We want compression resistance because we want titles to shrink before we try and push other buttons off the screen.

### Binary change

### Verification
Validated in test app that titles shrink correctly.

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2268)

### DIFF
--- a/Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
+++ b/Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
@@ -107,7 +107,7 @@ class CommandBarButton: UIButton {
         configuration?.image = item.iconImage
 
         if item.title != nil {
-            setContentHuggingPriority(.defaultLow, for: .horizontal)
+            setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
         }
 
         if let font = item.titleFont {


Sources/FluentUI_iOS/Components/CommandBar/CommandBarButton.swift
```diff
@@ -107,7 +107,7 @@ class CommandBarButton: UIButton {
         configuration?.image = item.iconImage
 
         if item.title != nil {
-            setContentHuggingPriority(.defaultLow, for: .horizontal)
+            setContentCompressionResistancePriority(.defaultLow, for: .horizontal)
```
mischreiber, line 110: Do we still also need the hugging priority, in addition to the compression resistance, to keep the title from stretching?

cbowdoin: Nope. The default content hugigng prioty of defaultHigh is perfect for our needs since our constraints do not need any smarts for how stick to the view.

mischreiber (approved):

PR #2270, opened 2026-06-05
[CommandBar] Removed safe area checking from command bar

cbowdoin:
### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes
Update the command bar to not make any adjustments internally to layout according to safe area. It should not be the responsibility of the command bar to know about the safe area. It internally should just layout its views as instructed by whoever owns it. The owner of the Command Bar should instead layout the Command Bar to stay out of the safe area. This is needed because currently clients have no way to position the Command Bar inside the unsafe area if they wanted to do something like animate the command bar into its position in the safe area.

### Binary change

### Verification
Confirmed that the command bar buttons do not get squished when animating in from the unsafe area into the safe area.

<details>
No change

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2270)


joannaquu (approved):

nodes11:
Does the demo controller still work as expected with these changes?

JaquelineGomez (approved):

PR #2284, opened 2026-08-07
[ListItem] Add titleTrailingAccessory

cbowdoin:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes
Adds a `titleTrailingAccessory` slot to `ListItem` for content that sits immediately trailing the `title` text, rather than against the trailing edge of the view like `trailingContent`. This is useful for status indicators that need to read as a sort of title metadata.

- `ListItem`: Adds the `titleTrailingAccessory` property. When set, the title is wrapped in an `HStack` with the accessory trailing it; when unset, the title renders exactly as before. Adds the `ListItemTitleTrailingAccessory` accessibility identifier.
- `ListItemModifiers`: Adds the public `.titleTrailingAccessory { }` modifier, taking a `@ViewBuilder` closure like the existing content slots.
- `TableViewCellTokenSet`: Adds `titleTrailingAccessorySpacing` (8pt) for the gap between the title and the accessory.
- `ListItemDemoController_SwiftUI`: Adds a "Show title trailing accessory" toggle to the SwiftUI demo.

The slot is exposed as a modifier since adding it to the existing init would have broken existing explicit `ListItem<...>` type annotations.

We use Text instead of Label because Label would make it so we lose VoiceOver support for the icon since out `ListItem` suppresses the accessibility drill in of its children so we only readout the top level and Label marks its icon as decorative.

### Verification
Added a "Show title trailing accessory" toggle to the SwiftUI `ListItem` demo and verified the accessory renders immediately after the title text, that the title's leading edge does not move when the accessory is added, and that the title returns to its original width when the accessory is removed — confirming no spacing is reserved while the slot is unset.

Existing `ListItem` call sites are unaffected: when `titleTrailingAccessory` is unset the title takes the original rendering path unchanged.

We use 

<details>

| image | image |
| image | image |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2284)

### DIFF
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
@@ -153,7 +153,7 @@ struct ListItemDemoView: View {
 
         @ViewBuilder
         var listItem: some View {
-            let listItem = ListItem(title: title,
+            var listItem = ListItem(title: title,
                                     subtitle: showSubtitle ? subtitle : "",
                                     footer: showFooter ? footer : "",
                                     leadingContent: {
@@ -196,10 +196,9 @@ struct ListItemDemoView: View {
                 .subtitleLineLimit(subtitleLineLimit)
                 .footerLineLimit(footerLineLimit)
                 .combineTrailingContentAccessibilityElement(trailingContentFocusableElementCount < 2)
-            var demoListItem = showTitleTrailingAccessory
-                ? listItem.titleTrailingAccessory { Image(systemName: "star") }
-                : listItem
-            demoListItem
+                .titleTrailingAccessory(showTitleTrailingAccessory ? Image(systemName: "star.fill") : nil,
+                                        accessibilityLabel: "Star icon")
+            listItem
                 .overrideTokens($overrideTokens.wrappedValue ? listItemTokenOverrides : [:])
                 .disabled(isDisabled)
                 .alert("List Item tapped", isPresented: $showingPrimaryAlert) {
--- a/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
+++ b/Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
@@ -32,6 +32,7 @@ struct ListItemDemoView: View {
     @State var showSubtitle: Bool = false
     @State var showFooter: Bool = false
     @State var showLeadingContent: Bool = true
+    @State var showTitleTrailingAccessory: Bool = false
     @State var showTrailingContent: Bool = true
     @State var isTappable: Bool = true
     @State var isDisabled: Bool = false
@@ -76,6 +77,8 @@ struct ListItemDemoView: View {
                 .accessibilityIdentifier("footerSwitch")
             FluentUIDemoToggle(titleKey: "Show leading content", isOn: $showLeadingContent)
                 .accessibilityIdentifier("leadingContentSwitch")
+            FluentUIDemoToggle(titleKey: "Show title trailing accessory", isOn: $showTitleTrailingAccessory)
+                .accessibilityIdentifier("titleTrailingAccessorySwitch")
             FluentUIDemoToggle(titleKey: "Show trailing content", isOn: $showTrailingContent)
                 .accessibilityIdentifier("trailingContentSwitch")
             FluentUIDemoToggle(titleKey: "Tappable", isOn: $isTappable)
@@ -110,7 +113,6 @@ struct ListItemDemoView: View {
                 Text(".plain").tag(FluentListStyle.plain)
                 Text(".insetGrouped").tag(FluentListStyle.insetGrouped)
                 Text(".inset").tag(FluentListStyle.inset)
-                Text(".glass").tag(FluentListStyle.glass)
             }
         }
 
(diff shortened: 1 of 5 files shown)


Sources/FluentUI_iOS/Components/List/ListItem.swift
```diff
@@ -50,13 +50,24 @@ public struct ListItem<LeadingContent: View,
 
         @ViewBuilder
         var titleView: some View {
-            Text(title)
+            let titleView = Text(title)
```
joannaquu: can we do all this with a Label? which has built in icon support https://developer.apple.com/documentation/swiftui/label

cbowdoin: Spoke offline. 

In short cannot use Label because it labels its icon as decorative resulting in us not being able to pick up the icon for VoiceOver.

Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
```diff
@@ -110,7 +113,6 @@ struct ListItemDemoView: View {
                 Text(".plain").tag(FluentListStyle.plain)
                 Text(".insetGrouped").tag(FluentListStyle.insetGrouped)
                 Text(".inset").tag(FluentListStyle.inset)
-                Text(".glass").tag(FluentListStyle.glass)
```
joannaquu, line 113: intentional?

Sources/FluentUI_iOS/Components/List/ListItem.swift
```diff
@@ -50,13 +50,30 @@ public struct ListItem<LeadingContent: View,
...
+            let titleFont = Font(tokenSet[.titleFont].uiFont)
+            let titleText = Text(title)
+                .foregroundColor(titleColor)
+                .font(titleFont)
                 .frame(minHeight: ListItemTokenSet.titleHeight)
                 .lineLimit(titleLineLimit)
                 .truncationMode(titleTruncationMode)
                 .accessibilityIdentifier(AccessibilityIdentifiers.title)
+
+            if let titleTrailingAccessory {
```
joannaquu: move the if let inside the hstack work. swiftui should handle the spacing/not rendering accessory when it's nil for you

Sources/FluentUI_iOS/Components/List/ListItem.swift
```diff
@@ -50,13 +50,30 @@ public struct ListItem<LeadingContent: View,
...
+                HStack(spacing: ListItemTokenSet.titleTrailingAccessorySpacing) {
+                    titleText
+                    titleTrailingAccessory
+                        .modifyIf(titleTrailingAccessoryAccessibilityLabel != nil, { accessory in
```
joannaquu: move all this into ternary operator to remove the modifyIf

Sources/FluentUI_iOS/Components/List/ListItem.swift
```diff
@@ -50,13 +50,30 @@ public struct ListItem<LeadingContent: View,
 
         @ViewBuilder
         var titleView: some View {
-            Text(title)
-                .foregroundColor(Color(uiColor: tokenSet[.titleColor].uiColor))
-                .font(Font(tokenSet[.titleFont].uiFont))
+            let titleColor = Color(uiColor: tokenSet[.titleColor].uiColor)
+            let titleFont = Font(tokenSet[.titleFont].uiFont)
+            let titleText = Text(title)
+                .foregroundColor(titleColor)
```
joannaquu: if we apply on hstack, we don't need to apply on text too

Demos/FluentUIDemo_iOS/FluentUI.Demo/Demos/ListItemDemoController_SwiftUI.swift
cbowdoin: Shoot. How did I miss that. Reverting. Thanks

Sources/FluentUI_iOS/Components/List/ListItemModifiers.swift
```diff
@@ -23,6 +25,19 @@ public extension ListItem {
         return listItem
     }
 
+    /// Image that appears on the trailing edge of the `title` text.
+    /// - Parameters:
+    ///   - image: The image to display after the `title`.
+    ///   - accessibilityLabel: A localized description of the image, announced by VoiceOver after the `title`.
+    ///   Leave this `nil` when the image is purely decorative, so that VoiceOver ignores it.
+    /// - Returns: The modified `ListItem` with the property set.
+    func titleTrailingAccessory(_ image: Image?, accessibilityLabel: String? = nil) -> ListItem {
```
joannaquu: what is this for?

joannaquu: oh I see. is it for the .titleTrailingAccessory? if so it feels kinda off that leadingContent/trailingContent are passed on init and titleTrailingAccessory is not and uses the modifier pattern instead

cbowdoin: I agree the problem was that I did not want to break the existing API for clients so I could not add things to the init

nodes11 (approved):

Sources/FluentUI_iOS/Components/List/ListItemModifiers.swift
```diff
@@ -3,6 +3,8 @@
 //  Licensed under the MIT License.
 //
 
+import SwiftUI
```
nodes11: Can be removed

nodes11 (approved):

nodes11 (approved):

PR #2286, opened 2026-08-14
[Avatar] Remove unused background circle tint when image exists 

StefanyFGo:
### Platforms Impacted
- [x] iOS
- [x] visionOS
- [x] macOS

### Description of changes

Avatar.swift: In `avatarBody`, the `backgroundColor` circle drawn behind the avatar's image content was always rendered, even when a real image is present and fully covers it. This causes a redundant offscreen draw and can produce visual artifacts (e.g., color bleeding at the circle's edges from anti-aliasing/opacity of the image on top).

### Binary change

None - one line change and no assets removed or added

### Verification

Manually verified Avatar renders correctly with:
- An avatar with a real image set (background circle no longer drawn behind it)
- An avatar with no image (initials/fallback still shows its background color as before)

<details>

| Scenario | Before                                       | After                                      |
|----------------------------------------------|----------------------------------------------|--------------------------------------------|
| Light mode | IMG_0188 | IMG_0187 |
| Dark mode |  IMG_0190 | IMG_0189 |
| Full test app validation showing no regression in normal opacity or initials |  |  |

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2286)


Sources/FluentUI_iOS/Components/Avatar/Avatar.swift
```diff
@@ -282,7 +282,8 @@ public struct Avatar: View, TokenizedControlView, Equatable {
                     .overlay(avatarRingView
                                 .frame(width: ringSize, height: ringSize, alignment: .center)
                                 .overlay(Circle()
-                                            .foregroundColor(Color(backgroundColor))
+                                            // If image exists, remove background circle that's redundant
+                                            .foregroundColor(state.image != nil ? Color.clear : Color(backgroundColor))
```
nodes11: Could the `modifyIf` modifier be used here so that foreground color is only applied when the image is nil? https://github.com/microsoft/fluentui-apple/blob/1669fa238d0a13af4e8d9b77f7dd427c3a5ecd0d/Sources/FluentUI_iOS/Core/Extensions/SwiftUI%2BViewModifiers.swift#L19

StefanyFGo: Unfortunately, Color.clear  still needs to be explicitly set: "[Shapes without an explicit fill or stroke get a default fill based on the foreground color.](https://developer.apple.com/documentation/swiftui/shape)" . Without it, the circle would render with the default foreground color instead of being hidden when there's an image.
Using modifyIf would be more verbose than the current single-line ternary, since we'd need both branches explicitly set:

```
Circle()
    .modifyIf(state.image != nil, { $0.foregroundColor(Color.clear) })
    .modifyIf(state.image == nil, { $0.foregroundColor(Color(backgroundColor)) })
```

huwilkes: Would it make sense to move the check for the image to where we initially set the background color? Around line 169

huwilkes: A slightly more involved alternative - should we remove this background modifier entirely from here and move it to the Text case of `avatarContent`?

StefanyFGo: Sounds good, decided to go with the slightly more involved approach of applying the background only where it's needed, rather than unconditionally on the outer circle. It ended up touching a bit more than I expected -  avatarContent  is shared with the  .group  style, which already paints its own background externally, so a flag was added to keep that path unaffected. Validated with the existing test suite and visually confirmed all the avatar variants look correct. Let me know if you have any concerns or want to me to validate anything specific

StefanyFGo:

huwilkes (approved):

huwilkes (approved):

Sources/FluentUI_iOS/Components/Avatar/Avatar.swift
StefanyFGo: Hi Des, let me know if you have any other questions or concerns. I'll be merging this PR soon but I can still follow up after, thanks!

StefanyFGo:
No new feedback/suggestions - merging to fix a transulcent avatar bug being tinted with inconsistent background colors

PR #2289, opened 2026-08-25
[FluentNotification] Expose API to enable changing expand chevron direction

nodes11:
…rection

### Platforms Impacted
- [X] iOS
- [ ] visionOS
- [ ] macOS

### Description of changes

The existing Expand button for the `FluentNotification` always points up. This works fine for apps which anchor the notification to the bottom of the screen, but breaks for apps which anchor the notification to the top of the screen.

Expose an `expandButtonDirection` which takes an `MSFNotificationExpandButtonDirection`. This allows the client to easily specify if the arrow should point up or down.

### Verification

Validated in internal use cases and in demo controller. See video in Visual Verification.

<details>

https://github.com/user-attachments/assets/09396621-f771-496e-a688-472fcb56c93d

</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2289)


Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
```diff
@@ -26,6 +26,9 @@ import SwiftUI
     /// If the message text should be expandabled to view the entire mesage if it is truncated due to the messageLineLimit
     var enableExpandableMessageText: Bool { get set }
 
+    /// Direction the expand button's chevron icon should point when collapsed. Defaults to `.up`.
+    var expandButtonDirection: MSFNotificationExpandButtonDirection { get set }
```
nodes11, line 30: wondering if there is a way we could do this without a API breaking change...

joannaquu (approved):

Sources/FluentUI_iOS/Components/Notification/FluentNotification.swift
joannaquu: we have a default on init so it won't be breaking right?

nodes11: Any consumer that has a custom state object that implements the protocol will break. It is a bit of edge case as having an object that implements the protocol is not required since fluent provides a default object, but still just trying to be considerate of those consumers.

PR #2290, opened 2026-08-26
Fix Mac Catalyst build

thomasjross:
### Platforms Impacted
- [x] iOS / Mac Catalyst
- [ ] visionOS
- [ ] macOS

### Description of changes

Fixes compilation of the FluentUI Swift package for Mac Catalyst.

Mac Catalyst compilation regressed after:

- #2249 introduced the `UIUpdateLink`-based custom bottom-sheet animator, which is unavailable on Mac Catalyst.
- #2251 moved the `NSFont` extensions into the common target behind `canImport(AppKit)`. That condition also succeeds for Mac Catalyst even though the required AppKit font APIs are unavailable there.

This change:

- Excludes the AppKit-specific `NSFont` extensions from Mac Catalyst.
- Excludes the `UIUpdateLink`-based `SheetAnimator` implementation from Mac Catalyst.
- Marks the custom spring animator APIs as unavailable on Mac Catalyst.
- Uses the existing `UIViewPropertyAnimator` path for bottom-sheet animations on Mac Catalyst.

Existing iOS and visionOS behavior remains unchanged.

### Binary change

No meaningful binary-size impact is expected. This change only adjusts platform-specific compilation.

### Verification

Verified that the `FluentUI` scheme builds for the generic Mac Catalyst destination in Release configuration.

<details>

Not applicable; this is a build compatibility change with no intended visual impact.
</details>

### Pull request checklist

This PR has considered:

No changes to appearance, accessibility, localization, layout, pointer interaction, SwiftUI consumption, or Objective-C exposure are expected.

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2290)

### DIFF
--- a/Sources/FluentUI_common/Core/Extensions/NSFont+Extensions.swift
+++ b/Sources/FluentUI_common/Core/Extensions/NSFont+Extensions.swift
@@ -3,7 +3,7 @@
 //  Licensed under the MIT License.
 //
 
-#if canImport(AppKit)
+#if canImport(AppKit) && !targetEnvironment(macCatalyst)
 import AppKit
 import SwiftUI
 
@@ -127,4 +127,4 @@ extension NSFont {
         }
     }
 }
-#endif // canImport(AppKit)
+#endif // canImport(AppKit) && !targetEnvironment(macCatalyst)
--- a/Sources/FluentUI_iOS/Components/BottomSheet/BottomSheetController.swift
+++ b/Sources/FluentUI_iOS/Components/BottomSheet/BottomSheetController.swift
@@ -181,6 +181,7 @@ public class BottomSheetController: UIViewController, Shadowable, TokenizedContr
     /// Defaults to `false` to preserve existing behavior. Requires iOS 18+.
     /// Note: This will cause more layout passes during animations, which may impact performance.
     @available(iOS 18.0, visionOS 2.0, *)
+    @available(macCatalyst, unavailable)
     open var usesCustomSpringAnimator: Bool {
         get { _usesCustomSpringAnimator }
         set { _usesCustomSpringAnimator = newValue }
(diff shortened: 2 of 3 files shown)


thomasjross:
One possible follow-up would be to hide the platform conditionals behind a small facade in `BottomSheetController`. The shared call sites could invoke helpers such as `startCustomSpringAnimationIfSupported`, `setHighFrameRateIdleIfSupported`, and `completeCustomSpringAnimationIfNeeded`, with the Mac Catalyst implementation falling back or doing nothing as appropriate.

That would keep the `#if !targetEnvironment(macCatalyst)` checks localized around the unsupported animator rather than distributing them through the controller’s animation flow. I kept this PR focused on the minimal build fix, but I could do that refactor maintainers would prefer to include it here.

joannaquu:
what `UIUpdateLink` api is unavailable? I see it's available from Mac Catalyst 18+ https://developer.apple.com/documentation/uikit/uiupdatelink

joannaquu:
@mischreiber is it worth adding catalyst validation or are we still "unofficially" supporting catalyst

thomasjross:
> what `UIUpdateLink` api is unavailable? I see it's available from Mac Catalyst 18+ https://developer.apple.com/documentation/uikit/uiupdatelink

I think Apple's documentation is wrong. The header in the Catalyst SDK explicitly marks `UIUpdateLink` as `API_UNAVAILABLE(macCatalyst)` and the compiler seems emit an error if it is referenced or constructed.

We _could_ probably rebuild this using `CADisplayLink` for Catalyst while retaining `UIUpdateLink` on iOS/visionOS. That feels a bit unnecessary to me based on the Catalyst usage, but open to feedback.

PR #2292, opened 2026-09-11
[List Item] Add Support for Editable Text in ListItems

cbowdoin:
### Platforms Impacted
- [X] iOS
- [X] visionOS
- [ ] macOS

### Description of changes
Introduces the option to use a `TextField` in a `ListItem` instead of just `Text`. 
- Exposes a new init the lets clients specify the placeholder title string, the text currently in the text field, and the action they want to preform when the user submits the text in the text field. 
- When we have a placeholder text and the editable text we will construct the text field variant of the view instead of the standard `Text` view.

The Demo Controller will now feature an editable text toggle that lets you see what the ListItemCell looks like when the text is editable.

*Note*: This change also includes a refactor to make `ListItemTitleTrailingAccessory` a struct to prevent invalid combinations of titleTrailingAccessory where we are given an acessibilty string but not an image.

#### Breaking Change
The migration to ListItemTitleTrailingAccessory is a breaking change but no consumers should have adopted this new API yet so this should be safe.

### Binary change

### Verification

<details>

| N/A | 
image |
</details>

### Pull request checklist

This PR has considered:
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/fluentui-apple/pull/2292)

### DIFF
--- a/Sources/FluentUI_iOS/Components/List/ListItem.swift
+++ b/Sources/FluentUI_iOS/Components/List/ListItem.swift
@@ -23,9 +23,7 @@ public struct ListItem<LeadingContent: View,
     ///   - title: Text that appears as the first line of text
     ///   - subtitle: Text that appears as the second line of text
     ///   - footer: Text that appears as the third line of text
-    ///   - titleTrailingAccessory: Image that appears immediately trailing the `title` text
-    ///   - titleTrailingAccessoryAccessibilityLabel: A localized description of `titleTrailingAccessory`, announced by VoiceOver after
-    ///   the `title`. Leave this `nil` when the image is purely decorative, so that VoiceOver ignores it.
+    ///   - titleTrailingAccessory: The accessory that appears immediately trailing the `title` text
     ///   - leadingContent: The content that appears on the leading edge of the view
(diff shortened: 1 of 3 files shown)


Sources/FluentUI_iOS/Components/List/ListItem.swift
```diff
@@ -59,22 +55,35 @@ public struct ListItem<LeadingContent: View,
...
+                TextField(title, text: editingText)
+                    .focused(editingIsFocused)
+                    .onSubmit {
+                        onSubmit?()
+                    }
+                    .frame(minHeight: ListItemTokenSet.titleHeight)
+                    .accessibilityIdentifier(AccessibilityIdentifiers.title)
+                    .foregroundColor(titleColor)
+                    .font(titleFont)
+            } else {
```
mischreiber, line 69: I think we can tighten this up a bit -- could we either (A) make it always an `HStack` and then just vary the items in it, or (B) move some of this out to one or two `@ViewBuilder` blocks?

My main goal is to remove the duplicate modifiers as much as possible, since that's an easy place to end up with bugs.

